PR #11, opened 2015-11-07
fixed bug in location command

Nystul-the-Magician:
location command no longer reports incorrectly that location index is invalid when it is correct and the player was teleported


Interkarma:
Lypyl already has a nearly-complete fast travel map that will be integrated at a later date. Please reduce this pull request to just the fixes in location command.

Scripts/Game/DaggerfallUI.cs
```diff
@@ -214,6 +218,13 @@ void ProcessMessages()
                     GameManager.Instance.PauseGame(true);
                     uiManager.PushWindow(dfCharacterSheet);
                     break;
+                case DaggerfallUIMessages.dfuiOpenAutomap:
+                    if (!GameObject.Find("Exterior")) // open automap only if player is in interior or dungeon - TODO: location automap for exterior locations
```
Interkarma: Should use PlayerEnterExit.IsPlayerInside here.

Resources/fallback.ini.txt
```diff
@@ -36,8 +36,8 @@ StartInDungeon=False
 
 [Enhancements]
```
Interkarma: Please do not change fallback.ini.

Interkarma:
Issues to resolve. See line comments.

Nystul-the-Magician:
I never intended to add this to the pull request (how did I mess this up?) - I started automap (interior and dungeons) but I thought it is only in my own work in progress. the pull request was done before. somehow my commit and push afterwards seems to ended up in the previous pull request - but why?

PR #16, opened 2015-11-14
Work in progress

HowardRoeske:
This started out as a way to expose point sampling on videos for a more old-school experience, but along the way I noticed that GUIFilterMode wasn't hooked up to anything. I studied the revision history and played around with the code for a bit, and ultimately the break seemed like an oversight rather than an intentional ("this isn't quite working yet") omission. 

So now, the GUIFilterMode setting gets applied to DaggerfallUI's global filter mode. In turn, DaggerfallUI's filter mode is applied to DaggerfallVideos. Smoothed UIs get smoothed video, and pixely UIs get pixely video.

This does eliminate the current shipping default configuration of point-scaled UI and smooth videos, though. It might be better to expose a separate filter setting for the videos? Let me know if that's what you'd prefer.

### DIFF
--- a/Scripts/Game/UserInterface/DaggerfallVideo.cs
+++ b/Scripts/Game/UserInterface/DaggerfallVideo.cs
@@ -54,6 +54,7 @@ namespace DaggerfallWorkshop.Game.UserInterface
             // Init empty texture
             vidTexture = TextureReader.CreateFromSolidColor(1, 1, Color.black, false, false);
             vidTexture.wrapMode = TextureWrapMode.Clamp;
+            vidTexture.filterMode = DaggerfallUI.Instance.GlobalFilterMode;
         }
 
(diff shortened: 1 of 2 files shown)


Interkarma:
GUIFilterMode hasn't been hooked up as native UI still needs work to fix cracking and other artifacts with bilinear filtering. I'm also not a fan of how the pixel fonts look with anything other than point, as they don't have enough pixel thickness to avoid looking crimped. My plan was to revisit this later and possibly allow for higher-resolution pixel fonts. Until then, I'd rather not have the setting available.

I do agree that videos could use their own setting. Please resubmit pull request with GUIFilterMode left alone, but with new setting called VideoFilterMode which controls the video filtering as you have here. Let me know on forums if any assistance needed. Good work. :)

PR #20, opened 2015-11-15
VideoFilterMode

HowardRoeske:
Reverted GUIFilterMode connection and added VideoFilterMode to the GUI settings. Defaults to 0 (point sampling), to match the rest of the filtering defaults in the project.

The DaggerfallVideo.cs diff looks huge because I made the changes to my fork before realizing there were changes to the file in your repository and then learning how to merge in upstream changes to my own. The actual diff is just two lines, #57 and #68 from my copy. Sorry about that, still learning how to Git.


Interkarma:
No worries mate, thanks for the pull request. Just to keep things neat, and since it’s only a few lines of code, I will make the changes manually and close the pull request. :)

 

One more quick thing. If you can, configure your git client to use Windows-styled line endings (CR LF) so it matches the rest of the project.

 

Are you using Visual Studio for development, or Unity’s built-in MonoDevelop?

 

Gav.

HowardRoeske:
Okay, sounds good.

As far as line endings go, I didn't realize that Git had automatic management built-in. I work in Visual Studio, which pretty much insists on using CRLF unless you set a per-file preference. I had inspected some of the source files in Notepad++ to see what line endings were in use. They all ended with just LF, so I was manually converting my changes over to LF before committing. :-P

It looks like the way to deal with this going forward is for me to set core.autocrlf to true in my git client. I've made the change locally to my working copy, but haven't really noticed a difference. Git didn't update any files on sync or anything. If you want, we can move this part of the discussion to the message board, since it's probably something good for other people to know.

Interkarma:
Manually merged changes. Will be in next push.

PR #21, opened 2015-11-22
Automap implementation

Nystul-the-Magician:
unfortunately git hub tells me it cannot be automatically merged. I think it is because of the changed scene
had to change it by adding a GameObject called "Automap" which has to have script Game/DaggerfallAutomap.cs attached


Interkarma:
Closing as branch was created from wrong starting point. Will create new branch and Nystul can re-issue pull request.

PR #60, opened 2016-03-17
Automap fixes (+small change to IncreasedTerrainDistance mod)

Nystul-the-Magician:
automap geometry should now correctly be destroyed when player is in exterior environment
automap geometry should now correctly be disabled when automap is closed
automap geometry should now be correctly created on game start and when a savegame is loaded

IncreasedTerrainDistance mod exposes the rendertexture reflectionSeaTexture now (so one can easily see in the unity editor if the mod successfully hijacked the reflectionmods texture or not)


Interkarma:
Recommended to use Destroy() over DestroyImmediate() at runtime. DestroyImmediate() should only be used for editor functions.
http://docs.unity3d.com/ScriptReference/Object.DestroyImmediate.html

I will merge and make this minor change at my end.

Nystul-the-Magician:
will do so ;) thanks for pointing this out!

PR #126, opened 2016-08-11
FarTerrain - found a fix for global fog to work with camera stack

Nystul-the-Magician:
actually these are really good news! I found a fix for global fog to work with the far terrain camera stack. It was easier than expected in the end. only had to fix some bugs on my side and uncheck the checkbox "Exclude Far Pixels" in GlobalFog script. I do this by script when FarTerrain mod is enabled automatically - so it should not break anything when FarTerrain is not active.
you can try it out, go to a location in the mountains (so you see far terrain) and than go to the lighting tab to fog settings and change them (don't let yourself get confused by unity's message about fog and deferred shading - (the message is actually misleading since global fog still uses these values - it should have better been called "without global fog script attached to camera component these values do not have any effect on opaque objects in deferred shading" or something))
FarTerrain is now ready for foggy weather conditions in daggerfall. it should be even possible to make dynamic foggy weather now very easy by changing the fog settings dynamically

### DIFF
--- a/Assets/Game/Addons/IncreasedTerrainDistanceMod/Scripts/IncreasedTerrainDistance.cs
+++ b/Assets/Game/Addons/IncreasedTerrainDistanceMod/Scripts/IncreasedTerrainDistance.cs
@@ -607,7 +607,7 @@ namespace ProjectIncreasedTerrainDistance
                 {
                     GameObject go = currentSkyManager.gameObject;
                     string objectPathInHierarchy = GetGameObjectPath(go);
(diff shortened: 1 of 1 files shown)


Nystul-the-Magician:
found a minor problem that needs to be fixed first

Nystul-the-Magician:
there is still a minor issue with global fog - but with the current settings it is at least not completely wrong now. the minor issue is coming from the fact, that there is a jump in fog values between camera clip borders

PR #127, opened 2016-08-12
FarTerrain and ReflectionMod improvements

Nystul-the-Magician:
FarTerrain - sky texture and water height are reassigned in Update() - got lost sometimes otherwise
ReflectionsMod - global fog is used for reflection (when in deferred) - looks correct now with heavy fog


Interkarma:
Reverting PR. Cannot use UnityEditor namespace in builds. This one is a blocking issue preventing new builds from being created.

PR #141, opened 2016-10-11
Set all men's clothing that have variants

Allofich:
I tested all the men's clothing in the game, and set the ones that have variants.
I also reordered the existing variant clothing to match the ordering of the enums.

Apparently, though, the number of variants isn't being read correctly from the data files or something. The shirts only cycle through two variants, if any.

For reference, here are the testing results, including the number of clickable variants I saw.

Clothing,                                          Has clickable variants, # of clickable variants

Straps						n
Armbands					n
Kimono						n
Fancy_Armbands				n
Sash						        n
Eodoric						n
Shoes						n
Tall_Boots					        n
Boots						n
Sandals						n
Casual_pants 				        n
Breeches					        n
Short_skirt					n
Casual_cloak				        y, 6
Formal_cloak				        y, 2
Khajiit_suit				        n
Dwynnen_surcoat				n
Short_tunic					n
Formal_tunic				        n
Toga						n
Reversible_tunic			        y, 2
Loincloth					        n
Plain_robes 				        y, 2
Priest_robes				        n
Short_shirt					y, 5
Short_shirt_with_belt		        y, 4
Long_shirt					y, 5
Long_shirt_with_belt		                y, 4
Short_shirt_closed_top		        y, 4
Short_shirt_closed_top2 	                y, 4
Long_shirt_closed_top		        y, 4
Long_shirt_closed_top2		        y, 4
Open_Tunic					n
Wrap						n
Long_Skirt					n
Anticlere_Surcoat			        n
Challenger_Straps			        n
Short_shirt_unchangeable	        n
Long_shirt_unchangeable		        n
Vest						        n
Champion_straps				n

### DIFF
--- a/Assets/Scripts/Game/Items/DaggerfallUnityItem.cs
+++ b/Assets/Scripts/Game/Items/DaggerfallUnityItem.cs
@@ -334,11 +334,19 @@ namespace DaggerfallWorkshop.Game.Items
             // This list will expand as more supported items are discovered
             bool canChangeVariant = false;
             switch(TemplateIndex)
-            {
-                case (int)MensClothing.Plain_robes:
-                case (int)MensClothing.Formal_cloak:
+            {        
                 case (int)MensClothing.Casual_cloak:
+                case (int)MensClothing.Formal_cloak:
                 case (int)MensClothing.Reversible_tunic:
+                case (int)MensClothing.Plain_robes:
+                case (int)MensClothing.Short_shirt:
+                case (int)MensClothing.Short_shirt_with_belt:
+                case (int)MensClothing.Long_shirt:
+                case (int)MensClothing.Long_shirt_with_belt:
+                case (int)MensClothing.Short_shirt_closed_top:
+                case (int)MensClothing.Short_shirt_closed_top2:
+                case (int)MensClothing.Long_shirt_closed_top:
+                case (int)MensClothing.Long_shirt_closed_top2:
                 case (int)WomensClothing.Plain_robes:
                 case (int)WomensClothing.Formal_cloak:
                 case (int)WomensClothing.Casual_cloak:


Allofich:
BTW, why can't you just use the number of variants read from the data, rather than listing them out?

I assume you have a reason, so I'll see about getting the female clothing, too.

Interkarma:
Thanks for PR. :)

> BTW, why can't you just use the number of variants read from the data, rather than listing the clothing out?

This is because many items have variants defined in their template data but cannot be cycled through via "Use" button in inventory window. A good example of this are greaves which define 6 variants but cannot be "used" by player to change appearance. The variants are instead for loot variety and different material looks (leather/chain/metal).

Even in clothing, not all items with variants can be "used" by player. One example is "straps" which have 4 variants but player cannot cycle through. The variants are only used for loot variety. So just reading the total variants isn't enough, and I couldn't locate a flag to specify whether a particular item could be cycled or not, so best solution at time was to specify.

BTW, I've exported item data from .exe and it now lives in Resources\ItemTemplates.txt as a JSON file. This makes the item templates standard across all users (unlike .exe) and enables easy fixing of small errors in template data. I've occasionally had to fix minor issues (like variant count) that were exported from game data.

This is all an ongoing process, so I genuinely appreciate the time you've taken to test men's clothing in game and set those variants which user can cycle through. Happy to accept this PR based on your testing process.

Interkarma:
To complete my example above, commit 892 fixes the "Long shirt" variant count in ItemTemplates.txt for indices 167, 168, 171, 172. This should make these variants available to the random loot system and to the "Use" button thanks to your PR.

I'd say quite a few other clothing items have art available that should be part of the on-use cycle. If this is something you're interested in researching, I'm happy for you to update template data accordingly. Please just verify art exists in sequence using Daggerfall Imaging 2 and new variant count will not overflow into another image set.

Feel free to me up in email or on forums if any questions.
http://www.dfworkshop.net/contact/

Thanks for your time. Small changes like this all help make Daggerfall Unity stronger. I've added you to the Credits thread on forums. :)

Allofich:
Thanks for the welcoming response! I'm putting up a new PR to complete the male clothing, and will do the female clothing next.

I'm "R.D." on the forum, btw. I posted in automap research a few days ago. I also work on OpenMW. I may not be able to do much here since I'm already devoting time over there and don't know a thing about Unity or C#, but I hope to be able to pitch in with easy stuff every once in a while.

PR #144, opened 2016-10-12
Set up female clothing for variants

Allofich:
Sets up female clothing for variants, as I did for male clothing.

There's an issue with the WomensClothing.Short_shirt, at least on a female redguard, in both original Daggerfall and in Daggerfall Unity. On one of the variants (the first of the green vest ones), it will disappear and become inaccessible when equipped. Seems like maybe an incorrect drawing priority or something, but Daggerfall Unity looks like it sets drawing priority by item, not by variant. The item looks OK with all its variants while in the inventory in original Daggerfall. Any idea?

Also you may be aware of this, but it seems that the inventory of a female character in Daggerfall Unity is empty/inaccessible. I could only test clothing in Daggerfall Unity by having it equipped at the time of saving in original Daggerfall.


Allofich:
I guess I might as well post the in-game test findings here, too, in case someone ever decides to check back against these findings. This listing was copied from ItemEnums.cs, so it has the enum values next to it. Also "brassiere" is misspelled, unless that was intentional for some reason.
Next to the enum values are either "n" for no clickable variants, or the number of clickable variants.

        Brassier = 182, n		
        Formal_brassier = 183, n 
        Peasant_blouse = 184, n
        Eodoric = 185, n
        Shoes = 186, n
        Tall_boots = 187, n
        Boots = 188, n
        Sandals = 189, n
        Casual_pants = 190,	n
        Casual_cloak = 191, 6
        Formal_cloak = 192, 2
        Khajiit_suit = 193, n
        Formal_eodoric = 194, n
        Evening_gown = 195, n
        Day_gown = 196, n
        Casual_dress = 197, n
        Strapless_dress = 198, 2
        Loincloth = 199, n
        Plain_robes = 200, 2
        Priestess_robes = 201, n
        Short_shirt = 202, 5
        Short_shirt_belt = 203, 4
        Long_shirt = 204, 5
        Long_shirt_belt = 205, 4
        Short_shirt_closed = 206, 4
        Short_shirt_closed_belt = 207, 4
        Long_shirt_closed = 208, 4
        Long_shirt_closed_belt = 209, 4
        Open_tunic = 210, n
        Wrap = 211, n
        Long_skirt = 212, n
        Tights = 213, n
        Short_shirt_unchangeable = 214, n
        Long_shirt_unchangeable = 215, n
        Vest = 216, n

If you are interested in looking into the vanishing short shirt I mentioned above, I can attach a Daggerfall save game for that here if you want.

Interkarma:
I've not seen the issue with female inventory before, and not able to reproduce with my female character saves. I'll retest after merging changes. When the inventory is blank, this usually indicates something crashed out the inventory UI.

For any buggy behaviour, please drop a post in the Bug Reports forum. If you can, please attach/email a save game to help me reproduce the problem more quickly. :)
http://forums.dfworkshop.net/viewforum.php?f=24

Interkarma:
> Seems like maybe an incorrect drawing priority or something, but Daggerfall Unity looks like it sets drawing priority by item, not by variant. The item looks OK with all its variants while in the inventory in original Daggerfall. Any idea?

As far as I'm aware, items only have a single draw-order value from template. If the item is straight-up vanishing on my end on equip, there's probably an exception in preparing art for the paper doll. Logs might have more for me.

If you can please send me a save to reproduce (and log a bug in Bug Reports if you have time), I'll get on it. Thanks for all your efforts! :)

Allofich:
I wrote the bug reports. BTW, I see you wrote on your twitter
"And those clothing fixes by R.D. should make some art variants available that Daggerfall itself doesn't use. Loving his work!"
Well I'm glad you love my work :), but as far as I know these changes just make all the variants of Daggerfall available, they don't add any unused ones, so maybe you want to correct/remove that tweet.

PR #145, opened 2016-10-13
Identify equipment sounds and open book sound

Allofich:
This identifies some of the sounds in SoundClips.cs that weren't identified yet. I wasn't sure what 382 was so I left that undefined.

I will gladly set these to play on equipping the appropriate items if you like. Or is there a reason you would prefer to hold off on that for now? The only issue I'm aware of is that it will play the equip sound  when you load a game.


Interkarma:
Good work thanks for that. :) Sure thing, add those sounds if you like. I can iron out the on load issue later, it's not a big deal. Cheers!

PR #146, opened 2016-10-14
Play equipping sounds and correct weapon draw sounds

Allofich:
This makes items play the correct sounds when equipped, as far as I know. I tested all the weapon equip sounds, etc. I could only test one artifact that I had a save game for, but it played the correct sound, so hopefully those work correctly.

I also made the correct sound play for drawing a weapon, which is the same as the equip sounds. I didn't change the swing sound.

I also changed the spelling of "jewelry" I had added to "jewellery," to match the British spelling you were using. :) It's probably a good idea to keep references to things using the same spelling.

The "equip on load" was pretty jarring if you had a lot of items being equipped. I added in some code to stop it from happening.

Let me know if there is anything you want me to change.


Interkarma:
Looks like excellent work, thank you! :)

PR #147, opened 2016-10-15
Play correct weapon swing sounds

Allofich:
With this PR the player's weapons will play the correct weapon sounds (as far as I know.)

There are comments in the code about varying the pitch of weapon swings according to attack speed. As far as I can see, this is not what happens in Daggerfall. It just plays one of three sound effects. Two of these were labeled as "EnemySwing1" and "EnemySwing2," but these are used by the player as well, so I renamed them.

The sound effect used seems to be based on the type of weapon. I tried a few different materials of the weapons, and swinging some of them with two different characters, and I got the same results.

I'm not completely sure what all the stuff about attack speeds and sound pitches in the comments is, so I'll leave it to you Interkarma if you want to remove the related TODO comments or any code, etc.

### DIFF
--- a/Assets/Scripts/Game/FPSWeapon.cs
+++ b/Assets/Scripts/Game/FPSWeapon.cs
@@ -36,7 +36,7 @@ namespace DaggerfallWorkshop.Game
         public float Reach = 2.5f;
         public float AttackSpeedScale = 1.0f;
         public SoundClips DrawWeaponSound = SoundClips.DrawWeapon;
-        public SoundClips SwingWeaponSound = SoundClips.PlayerSwing;
+        public SoundClips SwingWeaponSound = SoundClips.SwingMediumPitch;
 
         WeaponTypes currentWeaponType;
         MetalTypes currentMetalType;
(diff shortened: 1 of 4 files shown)


Interkarma:
Cheers, and good work.

When I wrote that, I did not feel at the time I was able to reproduce the exact swing sounds of a high-speed dagger vs. a slow-speed warhammer using the default sound library alone. My impression was that Daggerfall might be varying weapon pitch based on weapon speed.

Most likely I was a bit burned out from building all the sound back-end and simply failed to notice those other sounds belonged to to the player as well. Sometimes you just need a fresh set of ears, which you have kindly provided. :)

Allofich:
You probably already know this, but there is definitely difference in attack speed, based partly if not entirely on the Speed attribute. But yeah, I can't notice any difference in the sound effect that is played. Here is a sound recording of swinging a sword with a speed in the upper 90s (first three swings) versus a speed of 50 (latter three swings).

[swings.zip](https://github.com/Interkarma/daggerfall-unity/files/531670/swings.zip)

Maybe just remove the "for swing pitch variance" part from these comments?
// TODO: Adjust FPSWeapon attack speed scale for swing pitch variance

PR #149, opened 2016-10-23
Fix parsing of PositionPrefix format token

Allofich:
Just a minor fix. I noticed that the Strength attribute description had a few stray parentheses not there in original Daggerfall. I poked around and found that the check for the "PositionPrefix" token that what comes after it is not a normal character seems to be wrong. This was causing it to show the parentheses instead of use them for PositionPrefix. As far as I saw you can use any value there other than 0x00, which causes Daggerfall to show endless gibberish and crash.

Also Daggerfall Unity was summing together the PositionPrefix to the x cursor but it should be used as an absolute position and not summed. Summing it would cause the second use of PositionPrefix on a line to be moved over too far.

Edit: I wrote various things here about how I thought there should be a default indentation of 5, since using PositionPrefix of 0x04 places text one pixel to the left of the other lines, but it seems like it's already displaying correctly in Daggerfall Unity.

### DIFF
--- a/Assets/Scripts/API/TextFile.cs
+++ b/Assets/Scripts/API/TextFile.cs
@@ -379,7 +379,7 @@ namespace DaggerfallConnect.Arena2
                     break;
                 case Formatting.PositionPrefix:
                     peek = PeekByte(ref buffer, position);
-                    if (peek != null && IsFormattingToken(peek.Value))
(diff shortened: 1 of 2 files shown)


Interkarma:
Thank you, good catch.

PR #150, opened 2016-11-05
Fixes to appearance of UI windows

Allofich:
This is a number of minor fixes to the UI windows. I'll try to explain it all here, and please let me know if you want me to change anything.

1. Identify 0xf8 format token.

This token appears in TEXT.RSC for the gold drop input message, and for two tavern room renting input messages. From testing with the gold drop message, it looks like it positions the input cursor. Without it, the cursor appears in the middle of the text window, overlapping other text, so it seems to be used for that. I put in a debug message for it that will show instead of the "Unknown formatting token" one that was showing before.

2. Change panel margins from 12 to 10.

From testing with the text in Daggerfall, I know at least that attribute popup text can go as close as 10 pixels to the left edge, and the top can be at least as close as 11 pixels. I couldn't test the top any more finely than that, because I can't add by individual pixels vertically like I could horizontally, but for now it seems reasonable to assume it is 10 all around.

3. Make message window borders be applied by multiples of 22 pixels to avoid misaligned parchment border textures, and make parchment popups be a minimum of four border tiles (44 pixels by 44 pixels).

From testing, the attribute windows and other windows in Daggerfall seem to have this as a minimum size, and the windows are always sized so that the textures line up properly.

This code is assuming 22 x 22 textures are used, which is what the vanilla ones are.

I also used the code for the reflex selection window during character creation, which was just resizing to the text size before. I've put similar code in 3 spots for the sizing of windows, so maybe it could all be consolidated somewhere to avoid code duplication.

4. Change input distance for input messages from 4 to 0. Change text panel distances for other input messages from 0 to 8 or 5.

This is to match what I see in Daggerfall, and to look correct with new border sizes from the changes to the window sizing, which seems to make all the message windows I saw vertically the same size as in Daggerfall.

5. Change label vertical alignment from top to middle.

Makes things look more correct with the resized borders. However, they seem to be positioned slightly lower than the middle, which I haven't been able to find the reason for, so it doesn't look perfect. Any idea there? Also, I think the race selection dialogs need to have the buttons at the bottom somehow factored in to the vertical text alignment in order to look right.

6. Change vertical panel sizing for input messages.

The `textBox.MaxSize.y` in `Mathf.Max(textBox.MaxSize.y, textBoxLabel.Size.y)` was causing the gold dropping dialog to be too high vertically, so I removed it as it didn't seem necessary to me (you can only enter up to 8 digits for the input in Daggerfall). I just used `textBoxLabel.Size.y.`

### DIFF
--- a/Assets/Scripts/API/TextFile.cs
+++ b/Assets/Scripts/API/TextFile.cs
@@ -99,6 +99,7 @@ namespace DaggerfallConnect.Arena2
 
             NewLine = 0x00,
             EndOfPage = 0xf6,
+            InputCursorPositioner = 0xf8,
             SubrecordSeparator = 0xff,
             EndOfRecord = 0xfe,
 
--- a/Assets/Scripts/Game/DaggerfallUI.cs
+++ b/Assets/Scripts/Game/DaggerfallUI.cs
@@ -408,7 +408,7 @@ namespace DaggerfallWorkshop.Game
                     globalFilterMode);
             }
 
-            panel.SetMargins(Margins.All, 12);
+            panel.SetMargins(Margins.All, 10);
         }
 
         void LoadDaggerfallParchmentTextures()
--- a/Assets/Scripts/Game/UserInterface/MultiFormatTextLabel.cs
+++ b/Assets/Scripts/Game/UserInterface/MultiFormatTextLabel.cs
@@ -240,6 +240,8 @@ namespace DaggerfallWorkshop.Game.UserInterface
                     case TextFile.Formatting.Text:
                         AddTextLabel(token.text, font);
                         break;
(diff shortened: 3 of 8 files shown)


Allofich:
On second thought it's pretty pointless to have a debug message for that token, since we already know where it shows up, so I removed the message. It's recognized, though, so the "unknown formatting token" message won't show for it anymore.

Interkarma:
Awesome, thank you! I've just changed the base to a new branch: 'ui-updates' as I wanted to test before pulling to master. If you had any other UI fixes to make, please send them to this branch for now.

Interkarma:
I see what you're saying about change 5 in regards to vertical alignment. Agree it looks better overall, and definitely need to factor in buttons to alignment somehow. I'll take a crack at this when I can.

PR #151, opened 2016-11-06
Use original Daggerfall text for travel message

Allofich:
Minor change to use the original Daggerfall text. I did the same thing that is done for the drop gold message to replace the token for the location name, and marked it as a temporary hack as it is marked there.

The message from TEXT.RSC is

> Do you wish to travel toý %tcn?ý

Using `GetRSCTokens` to put this into `TextFile.Token[] textTokens` the "%tcn?" is stored in `textTokens[2].text,` while `textTokens[0].text` holds the "Do you wish to travel to" part and  `textTokens[1].text` is null. I don't really understand why there is a null in between them, but this was how it was, so I tell the code to look in `textTokens[2].text` for the token to replace.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTravelMapWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTravelMapWindow.cs
@@ -33,6 +33,12 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
     /// </summary>
     public class DaggerfallTravelMapWindow : DaggerfallPopupWindow
     {
+        #region Classic Text IDs
+
+        const int doYouWishToTravelTo = 31;
+
+        #endregion
+
         #region Fields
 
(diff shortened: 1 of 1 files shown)


Allofich:
Oh and this isn't related to formatting so my assumption is it is OK to PR to master rather than the ui-updates branch.

Interkarma:
No problems and thank you.

To explain the token setup, the source text is "Do you wish to travel to[0xFD] %tcn?[0xFD][0xFE]". So the returned token array should be:

[0] = "Do you wish to travel to"
[1] = 0xfd
[2] = "  %tcn?"
[3] = 0xfd
[4] = 0xfe

So you're exactly right in replacing text in textTokens[2]. :)

I haven't done much with the %var setup yet, I'm still debating how I want to handle this. I'll probably make a decision by the time quests are further along, as quest text makes heavy use of these variables.

Allofich:
Thanks for the explanation, it always feels better to understand why something works or doesn't work.

PR #154, opened 2016-11-16
Various changes to make UI text positioning more like original Daggerfall

Allofich:
Here are various changes to make the UI text look more like original Daggerfall. Everything in the UI windows I worked on should be positioned like Daggerfall now except for buttons, which are still at the bottom of the window as before this PR, and the items in the misc skills window.

I'll go through the changes here.

1. Skills windows were changed to match what I saw in original Daggerfall, using tabs and no spaces. Tabs of 35 spaces are used, not 45.

2. I determined through experimentation that having a text entry that ends with justify left in original Daggerfall  adds the final newline into the length of the text, while ending with a justify center does not. Unity Daggerfall was adding the newline to the length in either case. Doing the same thing as original Daggerfall allows the attribute popups, which end with justify center, to be correctly sized.

3. I fixed the problem of messages with middle vertical alignment being too low. The initial row wasn't being added into `totalHeight.`

I'm pretty confident about the top three fixes. The below fixes have good results, but I'm not an experienced programmer and they might not be implemented well.

4. I added hardcoded width overrides and positioning information for the sizing of the rest, loiter and find location input windows. These windows don't use text from TEXT.RSC and I couldn't see what rules if any they followed for their positioning.

5. I added in the same max input character limits as Daggerfall for the rest, loiter, find location and gold drop windows. The find location actually allows up to 32 characters. This fixes a problem where the rest and loiter windows weren't giving their "you can only rest/loiter for ___ hours" popups if you entered in very large values.

6. I added in resizing of text labels when buttons are added. This positions the text correctly, but I couldn't figure out how to get the buttons correctly positioned 4 pixels under the text, so I just left them for now as aligned at the bottom.

### DIFF
--- a/Assets/Scripts/Game/UserInterface/MultiFormatTextLabel.cs
+++ b/Assets/Scripts/Game/UserInterface/MultiFormatTextLabel.cs
@@ -25,7 +25,7 @@ namespace DaggerfallWorkshop.Game.UserInterface
     /// </summary>
     public class MultiFormatTextLabel : BaseScreenComponent
     {
-        const int tabWidth = 45;
+        const int tabWidth = 35;
 
         PixelFont font;
         int rowLeading = 0;
@@ -79,6 +79,11 @@ namespace DaggerfallWorkshop.Game.UserInterface
         {
             get { return textAlignment; }
             set { textAlignment = value; }
+        }
+
+        public void ResizeY(float newSize)
+        {
+            Size = new Vector2(totalWidth, newSize);
         }
 
         public override void Draw()
@@ -164,7 +169,6 @@ namespace DaggerfallWorkshop.Game.UserInterface
         {
             cursorX = 0;
             cursorY += LineHeight;
-            totalHeight += LineHeight;
             tabStop = 0;
         }
 
@@ -218,6 +222,7 @@ namespace DaggerfallWorkshop.Game.UserInterface
                         break;
                     case TextFile.Formatting.JustifyLeft:
                         NewLine();
+                        totalHeight += LineHeight; // Justify left adds to height regardless of there being anything afterwards
                         break;
                     case TextFile.Formatting.JustifyCenter:
                         if (lastLabel != null)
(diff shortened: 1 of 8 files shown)


Allofich:
Regaring the misc skills window, I measured that it is positioned as follows, using the leftmost pixel as 0.
The pixel number is how many pixels there are between the leftmost pixel of the previous item and the current item. So the attributes of the first column are 85 pixels + 27 pixels, or 112 pixels, from the 0 pixel point.

0                     85 pixels   27 pixels       24 pixels           85 pixels       27 pixels
Skill                 %              attribute       skill                   %                  attribute

I started to work on this but I don't know how this should be done, since it's not in set 35 space tab increments like with the other skills.

Allofich:
Maybe the buttons can be added in to the multiformattextlabel, in which case the cursor position there could be used to correctly position them. Then I could get rid of the resizing function I'm using now. I'd have to study the code some more and see if this would work, I'm just saying an idea right now. I'll take another look later when I have some time.

Interkarma you said you had an idea for the buttons formatting, right? Is there a particular you want to do it yourself?

Interkarma:
Excellent! Looks like solid improvements overall, thank you again for taking the time.

My idea to improve the button positioning is simple, not sure if will yield the overall accuracy. Easy to try however and abandon for a different idea.

The thought was to format everything into a dedicated container panel then centre that panel inside of the popup area. Even if the buttons are at the bottom of the new container panel, and text at the top, the overall positioning will be centred to the popup area (nested panels). Will need to size the container panel and popup panel appropriately of course.

Alternatively, it should also be possible to work out the desired cursor position in pixel coordinates and align buttons using an absolute Y coordinate rather than just format against bottom.

I'm coming into some free time on Friday and (hopefully) through to end of next week. I'm happy to take a shot at this and we can improve on concept if not accurate enough.

Allofich:
Well to get the look accurate to original Daggerfall, I think we need to position the buttons 4 pixels below the last line of text. As far as I know the text in the message windows with buttons is positioned correctly with this PR after I added in the vertical size and the 4 pixel gap to the text label size, so I was thinking maybe if we can just position the button 4 pixels down relative to the cursor y position in the text label, we would have a look accurate to the original.

PR #156, opened 2016-11-19
Position buttons to be correct distance below text

Allofich:
I finally figured out how to get buttons showing at buttonTextDistance (4 pixels) below the text in message boxes.

All that is left right now as far as positioning things is the misc skills. Interkarma do you have any advice about how that should be done?

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallMessageBox.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallMessageBox.cs
@@ -120,7 +120,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             messagePanel.Components.Add(label);
 
             buttonPanel.HorizontalAlignment = HorizontalAlignment.Center;
-            buttonPanel.VerticalAlignment = VerticalAlignment.Bottom;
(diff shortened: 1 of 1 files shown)


Interkarma:
Cool! Excellent work. :)

For misc skills, I hacked together a very basic two-column setup which is only used by that popup. It works by just shoving in a tab token. The problem is my tabs are stepping a fixed number of pixels so slightly fatter numbers (like 12%) will push out the text more than slimmer values (like 6%).

I think this needs a revisit to change how the text is aligned. Rather than use a tab with fixed width, step to a fixed pixel offset instead. It doesn't even strictly have to all be handled by the text formatting, that's just the first pass solution I came up. It's also possible to add something like a tab-stop array to the popups: so tab will snap forwards to nearest stop position, or default behaviour if tab-stop array is null or empty. I will confess that I've not put a lot of thought into this one haha.

You're most welcome to take this on if you wish, you've done some amazing work so far. Otherwise, it's one of those things I'll loop back to eventually in a polishing pass. Thank you again for taking the time to do all this. :)

Interkarma:
By the way, I'm happy to merge this branch back to master when you are. Just let me know when you're ready. :)

Allofich:
Go ahead, I'd be happy to see this merged. I'm not aware of these ui changes introducing any problems.

PR #160, opened 2016-11-23
Make misc skills display alignment match classic

Allofich:
Does as the title says. I used the positioning token for this, as it seemed well-suited for it.

As far as I know, all the message windows currently in Daggerfall Unity should be aligned like original Daggerfall now.


Interkarma:
Ah, very clever solution! Thank you for the PR. :)

PR #161, opened 2016-11-25
Allow return key and keypad enter key to close message boxes

Allofich:
Edit: I had a few other changes but I decided I want to reconsider them. I left this one in that I felt was probably worthwhile. It's possible in original Daggerfall to close message windows with the big enter key and the little enter key on the keypad, which in the unity engine means the `Return` KeyCode and the `KeypadEnter` KeyCode.


Allofich:
The found gold pieces message is not in TEXT.RSC, so I added it to `HardStrings.`

Interkarma:
Good change, thank you!

PR #162, opened 2016-11-26
Basic bow combat implemented

electrorobobody:
Bows can be used now


Interkarma:
Thank you for implementing basic bow combat :) I will review and merge to a new branch as soon I have time.

electrorobobody:
Thank you, Gavin.

The project looks great. I cannot wait a playable version but there is a
long way there.

Regarding bows... You may see that the change is quite basic. I did not add
any damage calculations. Arrows are not implemented as well. It is just to
have at least something instead of "Not imlemented" message. So it is as
simple as all other weapons in the build.

I've done animation and right-click attack. Additionally bow has a cooldown
after a shot and a range much longer than a sword. :)

Cheers.

PR #165, opened 2016-12-05
Potion recipe viewer

InconsolableCellist:
Implements the potion recipe viewer so that the user can click "INFO" in the inventory window and read their potion recipes.


Allofich:
I can't test it right now, but it looks like great work!

I did notice a typo.

// When the ingredients view closes, we should close to (matchs Daggerfall behavior)

Should be "matches"

Interkarma:
This is awesome! You put that together in record time. Looking forward to seeing this in-game.

PR #166, opened 2016-12-05
Exterior automap

Nystul-the-Magician:
improvements and fixes to exterior automap
- introduced building nameplates (currently shop type is used instead of building name)
- walking to new locations works now correctly with exterior automap
- camera rotation and zoom settings are remembered and restored now when closing and reopening exterior automap while in a location (resets on new location)
- resources will be now loaded only on demand
- location layout is now only generated once - when the automap is opened for the first time for the current location


Interkarma:
Very nice work! Looking forward to playing with update. Thank you. :)

PR #168, opened 2016-12-11
Don't treat enemies as moving platforms

midopa:
If we stand just right on an enemy, the player camera will get unstable. This still allows use to stand on an enemy, however.


midopa:
Recording of the bug:
[enemy-platform.zip](https://github.com/Interkarma/daggerfall-unity/files/644279/enemy-platform.zip)

The spinning happens when I press the attack button but don't move the mouse.

Interkarma:
Very nice fix! Thank you for that. :)

PR #169, opened 2016-12-11
Add shortcut keys on the start window

midopa:


Interkarma:
Very good, thank you. :)

PR #170, opened 2016-12-11
Prevent Rat Stacks

midopa:
This isn't a perfect solution, but I figured I'd put it up and see if anyone else has ideas on how to solve this in a simple, elegant manner (without using NavMeshes because I'm not sure if that should be introduced at this stage of development).

In the original game, enemies could overlap if they got close to you. But I assume that isn't the desired behavior for Daggerfall Unity.

### DIFF
--- a/Assets/Scripts/Game/EnemyMotor.cs
+++ b/Assets/Scripts/Game/EnemyMotor.cs
@@ -10,7 +10,6 @@
 //
 
 using UnityEngine;
-using System.Collections;
 
 namespace DaggerfallWorkshop.Game
 {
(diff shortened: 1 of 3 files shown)


midopa:
Here's a video of the Rat Stack issue
[ohno.zip](https://github.com/Interkarma/daggerfall-unity/files/644887/ohno.zip)

Interkarma:
Excellent work, thank you. :)

I'm pulling this to a new branch to test quickly before I merge back to master. Changes looks solid however, I don't expect any real issues. Cheers!

PR #172, opened 2016-12-14
Fix Unity compile warnings

midopa:


Interkarma:
Thank you again!

Before committing PR, I will quickly confirm changes to DaggerfallExteriorAutomap.cs and DaggerfallExteriorAutomapWindow.cs with Nystul. I believe these files are still a work in progress, and he may have something in mind for this property that has not yet been implemented. This feature is more his baby than mine. :)

PR #174, opened 2016-12-18
Save and load weather

midopa:
Plus some cleanup (from default Rider IDE suggestions)

### DIFF
--- a/Assets/Scripts/Game/DaggerfallAutomap.cs
+++ b/Assets/Scripts/Game/DaggerfallAutomap.cs
@@ -1691,7 +1691,7 @@ namespace DaggerfallWorkshop.Game
             DestroyBeacons();
         }
 
-        void OnLoadEvent(SaveData_v1 saveData)
+        void OnLoadEvent(SaveData_v2 saveData)
         {
             DestroyBeacons();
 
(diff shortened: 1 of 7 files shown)


Interkarma:
Thank you for the PR! :)

For the change from save data v1 to v2, I would prefer this to be left as v1 at this early stage unless absolutely necessary. Reasoning for this is that save format will probably go through dozens of small changes leading up to an alpha state, and I'd rather not have "migration bloat" in the scripts until most of the game features are basically in place and players have a reasonable expectation of playing through game from start to end.

It's also possible to add new structures to the save format without direct migration, provided new structure has a reasonable default setting. For example, I added automap state saves that remained compatible with earlier structures. Default was that automap was simply initialized to defaults if no automap save data present at deserialization time. On next save, that data was serialized as normal.

Could you please revert the v2 changes back to v1, and simply add weather state to the V1 structure? I am more than happy to assist with this if needed.

Thank you for your time and contribution, it's very much appreciated. :)

midopa:
Sure thing! I was very skeptical about the incrementing the save state versions, exactly because of migration bloat! Will keep it at v1 and update this branch.

midopa:
Branch updated

midopa:
Hm. Looks like Github isn't picking it up because the PR was already merged. I'll open another PR then?

PR #182, opened 2016-12-21
Remove old meta file

midopa:
Looks like the Addons folder was removed a while ago


Interkarma:
Hmm, good spot. I thought StreamingAssets/Addons was used by Lypyl's mod system. Since it's so minor, I'll merge PR now and make a note to check this out. Thank you!

PR #190, opened 2016-12-24
Compute attacks from mouse gestures

midopa:
Mouse movement is now tracked as a gesture. The travel distance and final gesture vector is used to compute the angle of attack.

Practically, this doesn't feel much different from the original way of computing weapon attacks. Initially this was an attempt to get attacks working nicely on a trackpad, but after some thought I think the trackpad compatibility is more about tweaking the sensitivity and threshold values.

The advantage of this approach is gives a bit more control over how attacks can be initiated, how angles are computed, and gives a more natural abstraction of the user input to control (no longer have to maintain the same mouse direction over N frames/cycles).


Interkarma:
Well, this is a serious improvement to my direction over frames approach. Thank you. :)

Merging to new branch 'gesture-attack' for now.

midopa:
Sorry for the huge delay. I've updated the code and made sure it's working. I'll submit another PR.

PR #191, opened 2016-12-25
Use weather table from Daggerfall Chronicles

midopa:
Weather is loaded from a JSON table based on the table in Chronicles (pg.49).

Currently weather is determined randomly. This is a bit jarring, however. It's possible to go from snowing to sunny with no transition. Marking it as a TODO to handle smoother/believable weather transitions.

Also added fog to the possible weather states.

### DIFF
--- a/Assets/Scripts/Utility/Tuple.cs
+++ b/Assets/Scripts/Utility/Tuple.cs
@@ -1,0 +1,32 @@
+﻿using System;
+
+namespace DaggerfallWorkshop.Utility
+{
+    /// <summary>
+    /// A tuple/pair class because the version of C# that Unity uses doesn't have this yet.
+    /// </summary>
+    /// <typeparam name="T1">First/left type</typeparam>
+    /// <typeparam name="T2">Second/right type</typeparam>
+    [Serializable]
+    public class Tuple<T1, T2>
+    {
+        public T1 First;
+        public T2 Second;
+
+        public Tuple(T1 a, T2 b)
+        {
+            First = a;
+            Second = b;
+        }
+
+        public static Tuple<T1, T2> Make(T1 a, T2 b)
+        {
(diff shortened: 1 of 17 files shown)


Nystul-the-Magician:
i think fog should not be a "own" weather state but rather be a variation over existing. imho it should be a value between 0 and 100%. this would allow for more varied weather. even vanilla df had fog and fog with rain at least

midopa:
Yeah, it's definitely possible to have foggy sunny days and non-foggy rainy/snowy days.

I think a few things are needed, possibly in separate changes:

1. A management system that handles smooth transitions between weather states (still allowing for sudden changes like a sudden rain storm or scripting overrides)
2. Decoupling, managing, and driving various weather states (fog, sky, particles, etc)
3. (This is a stretch and may be over-engineering) Driving the weather with some kind of model instead of a lookup table

Assets/Scripts/Utility/Tuple.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            return new Tuple<T1, T2>(a, b);
+        }
+
+        public override string ToString()
+        {
+            return string.Format("First: {0}, Second: {1}", First, Second);
+        }
+    }
+}
```
Allofich: Not sure whether it really matters, but I notice GitHub is complaining about a lack of a new line at the end of the file here.

midopa: Fixed!

Interkarma:
Nice one! Cheers :)

And the good thing about weather is it's just atmosphere. It wasn't really a core part of gameplay (well, summoning) but I think it's fair game to re-interpret and reimplement as needed. I doubt anyone will be unhappy with smoother weather transitions vs. classic.

Interkarma:
I'll merge this into new branch 'chronicles-weather'.

PR #194, opened 2017-01-04
Implement stat-based player hand-to-hand damage

Allofich:
Simply implements hand-to-hand damage based on what the Daggerfall Chronicles says. I also did some quick testing in original Daggerfall and there was no difference in the number of hand-to-hand hits it took to kill an enemy with an up-to-down hit and a down-to-up hit, so I think all hand-to-hand hits might be the same.

### DIFF
--- a/Assets/Scripts/Game/WeaponManager.cs
+++ b/Assets/Scripts/Game/WeaponManager.cs
@@ -481,7 +481,22 @@ namespace DaggerfallWorkshop.Game
 
                 // TODO: Use correct damage based on weapon and swing type
                 // Just using fudge values during development
-                int damage = Random.Range(1, 25);
+                int damage_low = 1;
+                int damage_high = 24;
+
+                // Hand-to-hand damage formula based on Daggerfall Chronicles and testing original game.
+                // Daggerfall Chronicles table lists hand-to-hand skills of 80 and above (45 through 79 are omitted)
+                // as if they cause 2 to be added to damage_high instead of 1, but the hand-to-hand damage display in the
+                // character sheet in the original game contradicts this.
+                // From some quick testing in original Daggerfall there didn't seem to be a difference in damage from swing type
+                if (weapon.WeaponType == WeaponTypes.Melee)
+                {
+                    int skill = playerEntity.Skills.HandToHand;
+                    damage_low = (skill / 10) + 1;
+                    damage_high = (skill / 5) + 1;
+                }
+                
+                int damage = Random.Range(damage_low, damage_high + 1);
 
                 //// Check if hit has an EnemyHealth
                 //// This is part of the old Demo code and will eventually be removed
--- a/Assets/Scripts/Game/WeaponManager.cs
+++ b/Assets/Scripts/Game/WeaponManager.cs
@@ -183,6 +183,26 @@ namespace DaggerfallWorkshop.Game
             ShowWeapons(false);
         }
 
+        public int CalculateWeaponDamage(WeaponTypes weaponType, MetalTypes metalType)
+        {
+            // Fudge values, to be replaced
+            int damage_low = 1;
+            int damage_high = 24;
+
+            // Hand-to-hand damage formula based on Daggerfall Chronicles and testing original game.
+            // Daggerfall Chronicles table lists hand-to-hand skills of 80 and above (45 through 79 are omitted)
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Game/WeaponManager.cs
```diff
@@ -481,7 +481,22 @@ private void MeleeDamage(FPSWeapon weapon)
 
                 // TODO: Use correct damage based on weapon and swing type
                 // Just using fudge values during development
```
Ezren: Remove TODO

Allofich: This PR only addresses hand-to-hand damage, so the TODO is still relevant for weapon-based damage.

Ezren: Are you planning on addressing the rest of the TODO after this, or is that something I could pick up?

Allofich: I considered it but it seemed a little difficult, so go ahead! I still need to work on easier things and become more familiar with the project.

I don't think the weapon damage data is being stored anywhere yet. I wasn't sure if it would be okay to just use a source like UESP and manually enter base damage values into the item file, or if the damage data should somehow be extracted from the Daggerfall executable.

Ezren: From what I saw in the loot generator, most of the other weapons data are stored in hard coded enums in the project. Even if the goal was to import all the information from DFVanilla, I think adding more data tables is an acceptable first step just to get the damage formulas in and tested. 

@Interkarma what are your thoughts?

Allofich: I think Interkarma probably won't be around for a few hours. But I agree it probably would be fine to add data tables and I expect Interkarma will agree, too. I believe metal materials are already implemented, so  it might be fairly easy to just use the base damage values and add in the material modifier. I don't know if there's any information out there about how exactly the different swing types affect the damage type, though.

Allofich: I found in the manual for The Elder Scrolls: Arena some information.

Diagonal slash left: Pull the mouse sharply -5% to hit, +2 damage
down to the left
Diagonal slash right: Pull the mouse sharply -5% to hit, +2 damage
down to the right
Horizontal cut: Pull the mouse left or right - 0 -
Vertical chop: Pull the mouse downward -10% to hit, +4 damage
Thrust: Push the mouse upward +10% to hit, -4 damage

These are the same attack types as Daggerfall, and although there aren't exact numbers in the Daggerfall manual the descriptions match. This seems like a good starting point, and might even be the same thing.

Ezren: I've been doing research as well - I think the modifiers are slightly different for Daggerfall. The chart on [this](http://en.uesp.net/wiki/Daggerfall:Combat) page was pulled from the Daggerfall manual - I translated that to +2 diagonal, +0 horizontal, +4 vertical, and -2 stab. If we left stab as -4 it would be impossible to do damage by stabbing with an iron dagger.

Ezren: Also, since the manual refers to these bonuses as being applied to damage "potential", does that suggest that they are only applied to the upper bound of the damage range?

Allofich:
Updated to include the strength modifier, which the strength attribute description says "is factored into your hand-to-hand / weapon damage." From testing in original Daggerfall with a hand-to-hand damage of 1 and a -1 modifier, I still did damage, so I assume there is a minimum damage of 1.

Allofich:
Well, after testing in original Daggerfall, it looks like actually the strength modifier has no effect on hand-to-hand.

Allofich:
Yet another update. While the UESP says that hand-to-hand adds 2 to the maximum damage instead of 1 at skill 80 and above, apparently based on a table in the Daggerfall Chronicles, the hand-to-hand damage indicator on the character sheet in-game contradicts this. I'm going with the in-game display as most reliable, which also means there is just a single formula for any skill level.

Allofich:
Updated yet again. I wasn't thinking before when I talked about a separate function for hand-to-hand display. What we would need is functions for calculating the min and max damage, not a damage roll, so I reorganized the code for this. I assume for weapons these functions would look up the base damage from the item table and then apply the modifier from the material type.

PR #201, opened 2017-01-06
add calculations for weapon damage

Ezren:
- Adds the formulas for weapon damage
 - Adds helper methods to get modifiers for swing direction and weapon material
 - Adds weapon damage range to the weapon object

### DIFF
--- a/Assets/Scripts/Game/WeaponManager.cs
+++ b/Assets/Scripts/Game/WeaponManager.cs
@@ -13,6 +13,7 @@ using UnityEngine;
 using DaggerfallWorkshop.Game.Entity;
 using DaggerfallWorkshop.Game.Items;
 using DaggerfallWorkshop.Game.UserInterfaceWindows;
+using System.Collections.Generic;
 
 namespace DaggerfallWorkshop.Game
 {
@@ -136,7 +137,7 @@ namespace DaggerfallWorkshop.Game
                     weapon = RightHandWeapon;
 
                 // Transfer melee damage
-                MeleeDamage(weapon);
+                MeleeDamage(weapon, lastAction);
 
                 // Weapon cooldown
                 if (weapon.Cooldown > 0.0f)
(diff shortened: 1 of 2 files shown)


Allofich:
Looks good! But should we use FPSWeapon for the parameters on weapon max and min damage? I used weapon type and weapon material thinking that we could use the same functions for displaying the damage values in the info box in the inventory. Also I think monsters can use weapons. If FPSWeapon is just going to be a synonym for a weapon container, then it's fine I guess.

Also should the weapon damage values be there? I assumed you were going to add them to the ItemTemplates.txt. On the other hand, it seems like the data there is set up the same as the data for the items in FALL.exe, and I couldn't find damage values for the weapon entries in FALL.exe, so this might be fine.

On both points, I'm sure @Interkarma could give a more informed opinion then me, but I thought I'd mention these points.

Assets/Scripts/Game/WeaponManager.cs
```diff
@@ -593,6 +604,95 @@ private void ToggleSheath()
...
+                case MetalTypes.Mithril:
+                case MetalTypes.Adamantium:
+                    return 6;
+                case MetalTypes.Ebony:
+                    return 8;
+                case MetalTypes.Orcish:
+                    return 10;
+                case MetalTypes.Daedric:
+                    return 12;
+                // Covers Iron and Steel
```
Allofich: I think you mean Steel and Silver

Interkarma:
Thank you for the PR! I like where this is going. :)

Just to join the conversation, my preference longer-term would be for all calculations and formulas to be decoupled to the "FormulaHelper.cs" class as a static helper method. FPSWeapon is mainly a presentation class for displaying and animating weapon attacks. This info could otherwise be stored in WeaponManager for each hand and updated when player changes weapons.

Using a collection (List<int>) for min & max damage seems like overkill. It would be better to create a new struct containing these values. You could also store more kinds of information in a struct (such as material modifier) and skip having to check this every attack. For example:

public struct WeaponDamage
{
  public int minDamage;
  public int maxDamage;
  public int materialModifier;
}

then in WeaponManager:

WeaponDamage activeWeaponDamage;

Update activeWeaponDamage in SetWeapon() and transfer damage much as you do now.

One more thing, @midopa is working on a gesture attack system for WeaponManager that will change how weapon attack direction is computed. You might need to rework the attack direction code once that PR is merged.

I'll push this to its own branch for now. Thank you for taking the time to do this!

Ezren:
Thanks everyone for the comments! I'll get working on these suggestions tomorrow.

PR #205, opened 2017-01-14
Recreate classic Daggerfall's music playlists

Allofich:
This recreates the music playlists of classic Daggerfall, based on what look like playlists in FALL.EXE. In-game testing seems to confirm it.

Also I renamed references to "palace" throughout the code with "castle." In classic Daggerfall, the rulers' structures in Daggerfall, Wayrest and Sentinel are all called "-- Castle" on the automap, while the ones for minor rulers in other places are called "Palace." I think this also is consistent with the text of quests that send you to palaces. The only inconsistency I saw is that the castle music is called "gpalac."

Some things remain to be done. One is that "palaces" are insisting on playing the "castle" music for some reason. I can't figure this out, and I'm asking for your help Interkarma if you know what's going on. It seems to be bypassing the playlists, because even if I remove or replace the "gpalac" music from the playlists it is still played.

Also currently all guilds play mage's guild music. The guild's besides the mage's guild should all be the standard "interior" music, I think.

### DIFF
--- a/Assets/Game/Addons/ReflectionsMod/Resources/configInjectionTextures.ini.txt
+++ b/Assets/Game/Addons/ReflectionsMod/Resources/configInjectionTextures.ini.txt
@@ -183,7 +183,7 @@ useMetallicGlossMap = false
 reflectivity = 0.16
 smoothness = 0.72
 
-[044_3 {House and Palace Ground Texture (Desert)}]
+[044_3 {House and Castle Ground Texture (Desert)}]
 textureArchive = 044
 textureRecord = 3
 textureFrame = 0
--- a/Assets/Game/Addons/ReflectionsMod/Scripts/InjectReflectiveMaterialProperty.cs
+++ b/Assets/Game/Addons/ReflectionsMod/Scripts/InjectReflectiveMaterialProperty.cs
@@ -32,7 +32,7 @@ namespace ReflectionsMod
         private Texture texReflectionGround = null;
         private Texture texReflectionLowerLevel = null;
         private bool playerInside = false;
-        private enum InsideSpecification { Building, DungeonOrPalace, Unknown };
+        private enum InsideSpecification { Building, DungeonOrCastle, Unknown };
         InsideSpecification whereInside = InsideSpecification.Unknown;
         private GameObject gameObjectInterior = null;
         private GameObject gameObjectDungeon = null;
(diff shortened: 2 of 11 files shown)


Allofich:
Changed around the ordering of the music playlists to match their ordering in FALL.EXE and cleaned up the playlist selection logic a bit.

In addition to the issue with palaces that I mentioned above, there is also a problem with residences sometimes playing the temple music gneut, which was in their playlist before. I have tried to pin this down but have no idea why it is happening. It's like the playlists from before my changes are somehow still being used sometimes.

Interkarma:
This is because playlist is really stored in scene data. The code just populates default playlist first time SongPlayer component is attached. Will be necessary to reset SongPlayer components in scene to acquire updated playlist. I can do this and update scene after merging and testing this change. :)

Interkarma:
Just a few things I noticed while playing with the changes:

* Taverns no longer play tavern music (just standard interior music).
* Weather music (snowing/raining) no longer seem to play during weather events.
* Palace music no longer restarts when exiting a dungeon block back to the palace area, and ambient effects continue to play.

I'll let you know if I find anything else.

Allofich:
> Palace music no longer restarts when exiting a dungeon block back to the palace area, and ambient effects continue to play.

Hmm, not seeing this problem when I try out the latest master commit. I just did a short test and didn't notice the other problems, either, but you say they are intermittent.

Also palaces (minor ruler palaces) are still playing the "castle" music like Castle Daggerfall, etc. We'll need to figure that out at some point. Another thing is that I think the dungeon music in Daggerfall classic is set per dungeon, so Privateer's Hold will always have the same music every playthrough, etc. That's also something to figure out.

Interkarma:
Agreed, music does seem to be set by dungeon. I haven't worked out how DF defines which track is played yet, but have not spent a lot of time hunting this down.

A good spot to test palace music switch from is in Wayrest Castle. Head up stairs to royalty and turn right, bash your way through next couple of doors and you'll reach the transition doorway where ambient light and music changes (i.e. you've exited the palace block). Then cross back through doorway boundary again. Ambient light will ramp back up, but music no longer switches back to palace theme. This worked in previous builds, and I believe this is how it works in classic as well (crossing doorway boundary switches between palace theme and dungeon). I can take a closer look and test in more detail later. :)

Allofich:
I can confirm that the music doesn't switch if I go to that part of the castle. Interestingly I had done my own test of the castle where I instead went in the first door on your right as you enter, down the elevator at the end of the hall, and through the first door on your right. The music switches back and forth without a problem if you go in and out of that room.

Allofich:
It seems like the music for that part of the castle (going right from the back room with the nobility) correctly transitions if I play off of a DF Unity save, but not if I play off of an imported classic save.

PR #208, opened 2017-01-16
fixed broken fog

Nystul-the-Magician:
some recent changes to WeatherManager broke the camera fog (the kind of fog one will always want even on a sunny day).

don't disable RenderSettingsFog! Rendering Fog != Weather Fog

to see the broken fog: uncomment the line that is now commented out (and should finally be removed), go to privateer's hold (outside) on a sunny day with all mods disabled and see that there is no transition from foreground to background - this is wrong


Interkarma:
Good catch!

PR #214, opened 2017-01-28
Mesh and Texture replacement - texture consistency between models

TheLacus:
Right now model import and texture import are completely separated,
which means that if a user install some replacement for vanilla textures
these are applied only on vanilla models, and not on custom models which
use these same textures.  This commit bring some modularity as the
'textures' folder on disk available to players is now used for a global
injection, so custom and vanilla assets should blend nice together no
matter which custom textures are used. This should be especially useful
for the most vanilla-friendly models. It can also be used to import
optional bump maps where the original materials doesn't have them.


Interkarma:
Great to see this one coming along so quickly. 👍

PR #219, opened 2017-02-06
Change and save controls through classic interface

JustinSteele02:
Keyboard keys can now be changed and saved through the classic interface. Mouse, joystick, and set default button not implemented yet. Currently there is nothing preventing the player from setting duplicate keybinds, but the old keybind is removed and only the new keybind is applied. 

If the game is closed with an invalid key configuration (duplicate or empty keybind), the controls will reset to default when the game is restarted. Valid configurations are saved between restarts.


Interkarma:
Nice one! Will merge now and give it a workout. 👍

PR #221, opened 2017-02-07
Completed keyboard bindings and reset defaults

JustinSteele02:
- Set and save keyboard bindings fully functional
- Can no longer exit with invalid configuration. Duplicates are
highlighted and popups prevent closing window
- Defaults tab button returns keys to DFUnity defaults

Couldn't find the "you have multiple assignments..." message in the text.rsc dump. Probably a hard string, but it's not in our HardStrings class yet so I just added it manually for the moment.


Allofich:
Sounds like nice work!

> Couldn't find the "you have multiple assignments..." message in the text.rsc dump. Probably a hard string, but it's not in our HardStrings class yet so I just added it manually for the moment.

It's in FALL.EXE. As far as I know those are being added in to the HardStrings, so I think you can include in there.

Allofich:
I gave it a quick run, and it seems to work well. Nice job! One thing I noticed is that the key names seem to be a little high, though. In original Daggerfall they seem to be level with the text describing the action. I think lowering them down one "pixel" (as in one of the scaled up pixels from the original artwork) would do the trick.

JustinSteele02:
Yup you're right, they're off center a bit. I was planning on fixing that and merging it later along with the mouse sensitivity window.

Interkarma:
This is great! Thank you for all your work on this. I'll give this one more test and merge back to master in next 24 or so.

PR #231, opened 2017-02-20
Mesh and Texture replacement - Weapons

TheLacus:
Import textures for first person weapons, different for each kind of metal.


Interkarma:
This is great! Thank you.

PR #258, opened 2017-03-21
exterior automap - building names integration

Nystul-the-Magician:
excellent work of yours interkarma! just had to plug it in.


Interkarma:
Cheers! I tried to make it as easy as possible to plug into what you were already doing, glad you didn't hit any issues.

PR #264, opened 2017-04-08
Compute attacks from mouse gestures

midopa:
Revival of https://github.com/Interkarma/daggerfall-unity/pull/190

Mouse movement is now tracked as a gesture. The travel distance and final gesture vector is used to compute the angle of attack.

Practically, this doesn't feel much different from the original way of computing weapon attacks. Initially this was an attempt to get attacks working nicely on a trackpad, but after some thought I think the trackpad compatibility is more about tweaking the sensitivity and threshold values.

The advantage of this approach is gives a bit more control over how attacks can be initiated, how angles are computed, and gives a more natural abstraction of the user input to control (no longer have to maintain the same mouse direction over N frames/cycles).

### DIFF
--- a/Assets/Resources/defaults.ini.txt
+++ b/Assets/Resources/defaults.ini.txt
@@ -36,8 +36,7 @@ MouseLookSmoothing=True
 MouseLookSensitivity=1.0
 HeadBobbing=False
 Handedness=0
-WeaponSwingThreshold=0.8
-WeaponSwingTriggerCount=3
+WeaponAttackThreshold=0.05
 WeaponSensitivity=1.0
 
 [Startup]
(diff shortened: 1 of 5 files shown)


Interkarma:
Good work, great to see this one make a comeback! I do have one small item to fix. Please rename INI setting "WeaponSwingThreshold" to something different, as the new default value is substantially smaller than old default, and internally this setting now means something different.

INI defaults are only deployed user-side to a fresh INI or to a new value syncing into to user's INI. If we keep the same value name, the new (much lower) value won't sync down to existing users at all because they already have the older (much higher) value in place. Just renaming the value to something different like "WeaponAttackThreshold" or "WeaponMouseThreshold" will do the trick and ensure new setting is deployed to existing INIs.

midopa:
Updated!

Interkarma:
Awesome thank you! :) I'll check this in soon.

PR #275, opened 2017-04-24
Mesh, Texture and Movies replacement: improvements

TheLacus:
* Fix CHAR03I1.IMG size
* Import movies from mods
* Removed old code from MeshReplacement.
* MeshReplacement: get cached material instead of loading directly from
disk.
* TextureReplacement: Force normal and emission map if present on disk (like happens already with metallicgloss)
* Texture and Mesh replacement: use fixed size for archive (00X, 0XX or
XXX). Still uses the minimum number of digits for record and frame.
* Rename cif folder to CifRci and capitalize Img folder


Interkarma:
Looks great, awesome work as usual. 👍

PR #277, opened 2017-04-25
Clock Resource & condition

Lypyl:
Added ClockAction class which is created by the Clock resource directly.  The basics of the clock resource are now working (see __Demo06 quest for example), though the flags & random range still need to be implemented.


Interkarma:
Lovely work, thank you for this! :)

PR #278, opened 2017-04-26
support for texture arrays for terrain tiles

Nystul-the-Magician:
terrain now tries to use texture arrays (if system supports it, otherwise falls back to old approach)
- settings.ini file setting EnableTextureArrays = True added (set to False to force the use of old approach)

texture array usage has these advantages:
- no texture bleeding
- texture replacement with TheLacus mechanism (albedo, normal and metallicgloss map (last only used by reflection mod)) for textures with arbitrary sizes (all textures of an texture archive must have same dimensions and format)

new version of Distant Terrain mod (2.1.0) and Realtime Reflections mod (2.1.0) need to be used with this
to work correctly


Interkarma:
👍

PR #279, opened 2017-04-29
Changes for covens and poor homes

Allofich:
Categorizes covens as dungeons on the travel map, like Daggerfall classic. Also plays "Dungeon Exterior" music for covens and poor homes like Daggerfall classic.


Interkarma:
Very nice! Thank you. 👍

PR #280, opened 2017-04-29
Fix gargoyle sounds

Allofich:
The gargoyle's attack sound was noted as missing, but actually it was being assigned to the orc shaman. For some reason, even though Daggerfall Jukebox will extract three sounds for the orc shaman, with the attack sound being identical to that of a regular orc (the files are identical), this index seems to be skipped over in Daggerfall Unity. I reassigned the orc shaman to use the regular orc's attack sound.

### DIFF
--- a/Assets/Scripts/SoundClips.cs
+++ b/Assets/Scripts/SoundClips.cs
@@ -233,11 +233,11 @@ namespace DaggerfallWorkshop
 
         EnemyOrcShamanMove = 178,
         EnemyOrcShamanBark = 179,
-        EnemyOrcShamanAttack = 180,
+        EnemyOrcShamanAttack = 138,
 
         EnemyGargoyleMove = 181,
         EnemyGargoyleBark = 182,
-        // Gargoyle attack sound missing.
+        EnemyGargoyleAttack = 180,
 
         EnemyWraithMove = 183,
         EnemyWraithBark = 184,
--- a/Assets/Scripts/Utility/EnemyBasics.cs
+++ b/Assets/Scripts/Utility/EnemyBasics.cs
@@ -668,7 +668,7 @@ namespace DaggerfallWorkshop.Utility
                 BloodIndex = 2,
                 MoveSound = (int)SoundClips.EnemyGargoyleMove,
                 BarkSound = (int)SoundClips.EnemyGargoyleBark,
-                AttackSound = -1,
+                AttackSound = (int)SoundClips.EnemyGargoyleAttack,
                 MinMetalToHit = MetalTypes.Mithril,
                 MinDamage = 10,
                 MaxDamage = 15,


Allofich:
You should be able to mark this thread as closed.
http://forums.dfworkshop.net/viewtopic.php?f=24&t=487

I suppose you might still want to investigate why Daggerfall Unity skips over that index. Does it detect it as a duplicate and skip over it or something like that?

Interkarma:
DFTFU is not skipping over an index. Monster sounds are packed like so:
+0 (movement)
+1 (bark/i-am-here)
+2 (attack)

In the case of the gargoyle, +2 (index 183) is read just fine but it's the first wraith sound. There simply is no +2 allocation for the gorgoyle packed in BSA. It's not being skipped.

Please note that index 138 is the Orc +2 and OrcShaman appears to have all 3 sounds in order as well. Before I merge this change, can you please confirm that you verified these sound changes against classic? My research at the time was that Gargoyles simply do not have an "attack" sound.

The only reason this ended up as a bug was because another contributor accidentally removed the sanity check for invalid sounds in the audio player. This has already been resolved.

Interkarma:
This is probably one of those cases where DF doesn't have everything packed in quite the right order or there is some intended re-use in place. Your change works nicely, I just need to confirm sounds match classic. If you've checked these sounds, I'm happy to merge. :)

Allofich:
I've checked over it again. In SoundClips.cs, starting from right after sound 177 (scorpion sound) and going through to the gargoyle, we have:

178 - orc shaman walking sound
179 - orc shaman bark sound
180 - gargoyle attack sound
181 - gargoyle move sound
182 - gargoyle bark sound

I double checked the sounds in the original game. The orc shaman and the gargoyle's attack sounds match what they are in this PR.

Interkarma:
God stuff, thank you for that. I can see the bad assumption on my part more clearly now. Orc shaman is reusing the standard orc attack sound, which placed me one index off on the gargoyle. Great catch. :)

Allofich:
There's still a little bit of a mystery, at least to me. When I use your program Daggerfall Jukebox to extract all the sound effects from Daggerfall, all three of the Orc Shaman's sounds are extracted, as "10210.wav," "10211.wav" and "10212.wav," the last being the attack sound that is seemingly skipped by Daggerfall Unity's indexing and is identical to that of a regular orc. The next sounds, "10220.wav" and onward, are the gargoyle. Any idea why this happens?

Allofich:
Furthermore, the order that the gargoyle sounds are extracted by Daggerfall Jukebox matches the pattern of the other monsters, of "walk," "bark" and then "attack," but you may have noticed that to get the sounds right for the PR, I had to give the gargoyle an unusual ordering of the indexes.

Interkarma:
Thank you for the clarification, I understand better what you're telling me now. :)

What's probably happening is the sounds BSA record directory keys some shared sounds back to an identical offset. Although the orc shaman attack is the first time I'm aware of this happening in practice.

DF Unity maintains a sound clip cache keyed to index. I think DF Jukebox does the opposite and iterates through the record directory itself (ID to offset in file). This would make DF Jukebox capable of finding multiple pointers back to the same sound (like the orc clearly uses) and DF Unity not.

In hindsight, perhaps I should have used the record directory in DF Unity as well, and populated sounds enum with IDs rather than indices. I probably had a good reason for just using indices at the time, but can't recall my line of thinking right now.

I'm OK with a few twists in the sound enum. Besides a few quirks like this, the overall sound setup appears to be working quite well. There's probably not much benefit in going through and rewriting it to use IDs instead of indices.

Allofich:
Thanks for the explanation. Well if there are any other problems with monster sounds, they'll probably be noticed by someone like myself who has played the original game, so yeah I think we can focus on other things.

PR #281, opened 2017-04-30
Make Orsinium have dungeon music and lighting throughout

Allofich:
Daggerfall Unity used the castle music and lighting in Orsinium, until you got out of the first block and into the dungeon area. In classic, it has dungeon music and lighting throughout.


Interkarma:
Cheers! 👍

PR #282, opened 2017-05-05
Make cautious traveling arrive during daylight

Allofich:
With this PR the player will always arrive during daylight hours when using cautious travel, like Daggerfall classic.

From testing Daggerfall classic, I found that if the player's arrival would be between 6 p.m. and 6:10 a.m. it will be advanced to 6:10 a.m. This does the same for Daggerfall Unity.


Interkarma:
👍

PR #283, opened 2017-05-07
Set message text color to match Daggerfall classic

Allofich:
This probably shouldn't be merged yet, because there's an issue with the colors that I hope you can help me with, @Interkarma.

This is an attempt to set text colors to match Daggerfall classic. I think the text colors of Daggerfall classic correspond to these colors found in its .COL palette files.

Regular yellow color - e2 dc 00 (R226, G220, B0)
Text shadow color - 5d 4e 0e (R93, G78, B14)
Light yellow color used in quest messages ("say" command) - ff f6 67 (R255, G246, B103)

Unlike when testing FMAP_PAL.COL with the travel map, I couldn't see a difference in the game's colors from modiying these values in any of the .COL or .PAL files, so I don't know for 100% sure, but they are so close to the RGB values of the pixels seen when running in DOSBox that I think we can assume these are the values.

I tried to set these values for DF Unity. For some reason, though, the colors in-game don't match the RGB values I've set in the source code, though, and I have no idea why. My test case was using the tutorial quest and checking the font colors of the first two messages. The lighter-colored text completely loses its blue value even though it should have a blue of around 103.

Also, should I note the previous color values in a comment? Since they could be retrieved from the git history I don't suppose I need to.

### DIFF
--- a/Assets/Scripts/Game/DaggerfallUI.cs
+++ b/Assets/Scripts/Game/DaggerfallUI.cs
@@ -36,9 +36,10 @@ namespace DaggerfallWorkshop.Game
         const string splashVideo = "ANIM0001.VID";
         const string deathVideo = "ANIM0012.VID";
 
-        public static Color DaggerfallDefaultTextColor = new Color32(243, 239, 44, 255);
+        public static Color DaggerfallDefaultTextColor = new Color32(226, 220, 0, 255); // Matched to Daggerfall Classic.
         public static Color DaggerfallDefaultInputTextColor = new Color32(227, 223, 0, 255);
-        public static Color DaggerfallDefaultShadowColor = new Color32(93, 77, 12, 255);
+        public static Color DaggerfallDefaultShadowColor = new Color32(93, 78, 14, 255); // Matched to Daggerfall Classic.
+        public static Color DaggerfallQuestTextColor = new Color32(255, 246, 103, 255); // Matched to Daggerfall Classic
         public static Color DaggerfallAlternateShadowColor1 = new Color32(44, 60, 60, 255);
         public static Color DaggerfallDefaultSelectedTextColor = new Color32(162, 36, 12, 255);
         public static Color DaggerfallDefaultTextCursorColor = new Color32(154, 134, 0, 200);
--- a/Assets/Scripts/Game/Questing/Actions/Say.cs
+++ b/Assets/Scripts/Game/Questing/Actions/Say.cs
@@ -70,6 +70,7 @@ namespace DaggerfallWorkshop.Game.Questing.Actions
             messageBox.ClickAnywhereToClose = true;
             messageBox.AllowCancel = true;
             messageBox.ParentPanel.BackgroundColor = Color.clear;
+            messageBox.Label.TextColor = DaggerfallUI.DaggerfallQuestTextColor;
             messageBox.Show();
         }
     }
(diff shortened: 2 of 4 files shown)


Allofich:
Here's a comparison of the default text colors and the quest text colors from Daggerfall classic, taken through DOSBox, and the colors I get with this PR.

Default text color, classic:
classicdefaulttext

Default text color, Unity:
dfunitydefaulttext

Quest text color, classic:
classicquesttext

Quest text color, Unity:
dfunityquesttext

Interkarma:
On the first point about RPG values - I'm grabbing source colours directly from framebuffer in DOSBox (capture with Ctrl+F5 then eyedropper in paint.net). I'm also careful not to use any filtering options in DOSBox. This saves me from having to guess which index and palette Daggerfall is using. When I take a screenshot in Daggerfall and sample the pixels for default text, I'm getting 243, 239, 44 as per my source code. Here's my output in DOSBox.

df in dosbox

I am happy with the RPG values as currently set and don't see any reason to change these. :)

For your second point, I agree the GUI output in Unity does not quite match the source values. The text above renders like so in Unity.

df in unity

The output is 232, 224, 8. Not quite what is set. I noticed this slight muting of colours a while back. Not sure where this problem originates from, it might be a Unity quirk. It's something I've been meaning to chase down but have not gotten around to it yet.

I'd consider the second point a straight bug. But for selecting the RGB values in first point, are you using any filtering in DOSBox? This could influence your output as well.

Allofich:
The only thing in my DOSBox configuration I know of that would affect the RGB values of pixels is the "machine" variable. I'm using "machine=vgaonly," because back when I was working on the travel map colors, I found that this seemed to produce the closest RGB values (as checked using the eyedropper in Windows Paint) to those defined in the .COL files, more than the default machine value.

I don't know, though, which value of machine makes colors most like a real computer running in DOS in those days.

The main thing I was aiming for with this PR, though, was introducing the lighter text for quest messages. I wonder why the color is changed so much when run through Unity... If I or you or someone can't figure it out in a while, I'll close the PR.

Interkarma:
You're right, popup text does have a brighter colour. It's also different to the "default text" colour used on character sheet, etc. That's not something I've noticed before...

When I eyedropper a quest popup from DOSBox framebuffer, I get 255, 247, 101. Here's a snip from first popup player sees when starting a new game. Out of curiosity, what RGB values do you get when sampling the same output from your DOSBox?

fall_002

For the slightly muted colours in Unity, I just consider this a bug either on my side or something in Unity itself. I'd rather tackle this down the road than change all the colour values right now.

Allofich:
From the DOSBox screen, I get 248, 244, 96. From the framebuffer, I get 255, 247, 99.

When I was working on the palette for the travel map I noticed the colors were slightly off from the RGB values in the .COL file. I didn't know about the framebuffer capture feature, so I didn't use that. But based on that experience I suspect Daggerfall uses a value of 255, 246, 103, a very close value found as ff f6 67 in the .COL files. My eye can't tell the difference between these very tiny differences anyway, though, like between the value I get from the DOSBox screen and the framebuffer.

Do you want to just close this PR to work on the Unity color muting later or shall I try to approximate the quest color text now, while leaving the other colors as they are? Or if you want to approximate it yourself that is fine with me, too.

Allofich:
Also, regarding muted colors in Unity, IIRC this issue was not present with the travel map. Rather, there the RGB values of the location dots were exactly the same as those defined in the source. Maybe that's useful (or not) for figuring this out.

Interkarma:
That is useful information regarding travel map, thanks! I think it's best to fix the muting problem first, or we're just estimating the outcome and it may vary by platform. My hunch (and it's a big hunch, don't hold me to it) is that it could be related to Unity's gamma colour handling and my runtime glyph textures.

With your observation re. travel map, I'll take a look at other texture systems to see which are most affected. This won't be a priority right now though. I'm starving for free time lately and try to spend what I have working on quest stuff.

I will close PR for now and we can revisit later. Thank you for the enlightening chat around this! :)

PR #284, opened 2017-05-08
Implement basics of status window

Allofich:
This implements the basics of the status window (I key). It shows the place name and date. Right now the legal reputation just shows "a common citizen."

Some variables in TEXT.RSC that start with % are interpreted now. Anything with % followed by a letter or number is considered a variable that needs interpreting. This seems like the case in Daggerfall classic, because something like %a or %1 will just display as a blank in the game. If there's a space or a period or an exclamation point, and probably other things, after the % it will show in Daggerfall classic as it is, like "% " or "%." or "%!"

### DIFF
--- a/Assets/Scripts/API/TextFile.cs
+++ b/Assets/Scripts/API/TextFile.cs
@@ -89,6 +89,7 @@ namespace DaggerfallConnect.Arena2
             PullPreceeding = 0x02,
 
             FirstCharacter = 0x20,
+            VariablePrefix = 0x25,
             LastCharacter = 0x7f,
 
             PositionPrefix = 0xfb,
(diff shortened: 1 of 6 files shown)


Allofich:
For the wilderness, currently a blank name is returned for the location. In Daggerfall classic the region name is returned in this case, so that will need to be added somehow.

Interkarma:
Hey mate! :) This is good work and gets the immediate result, but it's at odds with the design that I'm working towards, I'm sorry.

I do not wish for %vars to be handled by the token stream. They often cross over with other game systems, and I'm working towards a unified solution for this alongside the quest system. One of the reasons I have so little handling for %vars in the game right now is because I want to handle them in a specific way. I do not see them as a formatting token like alignment, etc. I see them as text macros to expand based on end system consuming that text.

These macros often need very contextual information (e.g. item player just selected), and I don't want to be parsing that information in UI classes. The job of classes like MultiFormatTextLabel is just to render lines of text. I don't want it checking in on game systems, reading player position, etc. That's not its job.

I'm sorry for rejecting this PR. You do good work and I respect the contribution and time taken on this. It might be worth checking with me first before making core design decisions. I often do things a certain way because I have a plan for it down the road. Please use the developer forum or drop me a quick email with your proposed work and I'll let you know if any problems.

Cheers mate.

Allofich:
That's all right, I thought something like this might be the case.

PR #287, opened 2017-05-13
Adjust attack-related timings to resemble classic

Allofich:
Here are some changes to make the combat feel a little more like Daggerfall classic.

Daggerfall Unity's swing sound played at the start of an attack, played whether an enemy was hit or not, and the hit didn't register until after the animation finished.

In classic the swing seems to play at just about the frame after the middle frame. It plays if the player misses or the player hits a door, but not if they hit an enemy. The hit registers at about this same time.

This PR tries to make the above behavior like classic.

### DIFF
--- a/Assets/Scripts/Game/FPSWeapon.cs
+++ b/Assets/Scripts/Game/FPSWeapon.cs
@@ -131,10 +131,7 @@ namespace DaggerfallWorkshop.Game
 
             // Do not change if already playing attack animation
             if (!IsPlayingOneShot())
-            {
-                PlaySwingSound();
                 ChangeWeaponState(state);
-            }
         }
 
         public void ChangeWeaponState(WeaponStates state)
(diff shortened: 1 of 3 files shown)


Interkarma:
This PR has a conflict with another PR issued the day before. Let me know once resolved and I'll review and merge. Sorry for the hassle. :)

Allofich:
Ah, OK. Let me take a look.

Allofich:
Looks like this messes up the bow attacks. Please don't merge yet.

Allofich:
The bow is working correctly now.

I also did a little bit of renaming.

bool isAttacking to bool isStartingAttack, due to the changes I made.

MeleeDamage() to WeaponDamage(), since it's also used for bows. Same with CalculateMeleeDamage to CalculateWeaponDamage.

Another thing, I noticed that the swing left melee attack (kick) doesn't have all its animation frames in Daggerfall Unity. Adding to the number of frames in WeaponBasics.cs didn't work, though, it caused an unrelated frame to show up. It might be that for that one the beginning frames are played back in reverse. What do you think?

PR #288, opened 2017-05-13
Add book sounds

Allofich:
Adds book opening and page turning sounds.

Unrelated to this PR, there is an error in the console when turning book pages, although the page turning seems to work fine.


Interkarma:
Cool! Thanks for heads up on page turning error.

PR #289, opened 2017-05-13
Add "no spells in your spellbook" message

Allofich:
Kind of pointless, but I added the "You have no spells in your spellbook." message so something happens when you click use on your spellbook. I also moved a few text IDs to the top of the file, which is where I've seen them in other files.


Interkarma:
No problems. Here's my style rule for consts:

* If a const is used only once, declare close to where it is used (preferably top of method where used).
* If a const is used multiple times in class, or will be public to other classes, declare at top of class.

There isn't really a problem moving these from a functional point of view, it's just a style preference. On the whole, you guys do a good job keeping the style consistent with me. :)

PR #291, opened 2017-05-13
Fix enumerator error

Allofich:
As mentioned in https://github.com/Interkarma/daggerfall-unity/pull/288, an error appears in the console when turning pages in books.

I borrowed a solution for the same error from here:
http://answers.unity3d.com/questions/290595/enumeration-operation-may-not-execute-problem-with.html

The error is gone, and I can't see any difference in behavior when turning pages or opening and closing the various panels.


Interkarma:
That's the correct fix. Cheers!

PR #293, opened 2017-05-16
Make player look straight when leaving dungeons

Allofich:
A simple fix for http://forums.dfworkshop.net/viewtopic.php?f=24&t=325. Any reason to not just set the player's view horizontal like this?


Interkarma:
The SetFacing(Vector3 forward) overload should be left as-is. It's job is to set a literal facing vector to the player's camera.

You do have the right idea to fix dungeon exit problem though. :) I'd just prefer it if there was a new method in PlayerMouseLook called SetHorizontalFacing(Vector3 forward) that uses your fix.

Then change line 1122 in StreamingWorld to call SetHorizontalFacing() instead of SetFacing().

Thanks for looking into this bug.

Interkarma:
Oops, meant to just comment and keep PR open. I'll quickly make the changes suggested and update.

PR #294, opened 2017-05-16
Fix string.Format usage

Allofich:
There are a number of incorrect calls to `string.Format`.


Allofich:
Also removed an unused field from `DaggerfallBankingWindow.cs`.

Interkarma:
Some good catches, thanks.

PR #295, opened 2017-05-16
Show hand-to-hand damage in character sheet

Allofich:
Shows the hand to hand damage in the character sheet, like classic Daggerfall.

In classic, this text is colored a light blue. I don't know off the top of my head if there are any other places with other-colored text like this, but if I'm not mistaken at the moment Daggerfall Unity only supports one color of text per text label/message box.  I want to change just the color of this one token, but the only way I thought of was to add a color property to `TextFile.Token,` and I thought you might not want that, so for now the text is the default color.


Interkarma:
I've noticed that light blue as well. Not sure what it's in aid of tbh. Thanks for the PR though! 👍

PR #297, opened 2017-05-22
Set same ambient dungeon effects as classic

Allofich:
Sets the same ambient dungeon effects as classic, which are all next to each other in the list of sounds.

Alternate ambient sounds might be fun, too, but I figure we want to at least have the same ones as classic ready for use.


Interkarma:
👍

PR #299, opened 2017-05-22
Implement ineffective weapon materials

Allofich:
Implements ineffective weapon materials, as well as a little reorganizing. Later we will probably want to use the weapon damage code for AI attacking the player or each other, too.

### DIFF
--- a/Assets/Scripts/API/Save/CharacterRecord.cs
+++ b/Assets/Scripts/API/Save/CharacterRecord.cs
@@ -70,6 +70,7 @@ namespace DaggerfallConnect.Save
             doc.currentFatigue = parsedData.currentFatigue;
             doc.skillUses = parsedData.skillUses;
             doc.startingLevelUpSkillSum = parsedData.startingLevelUpSkillSum;
+            doc.minMetalToHit = parsedData.minMetalToHit;
             doc.armorValues = parsedData.armorValues;
 
             return doc;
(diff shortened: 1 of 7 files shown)


Allofich:
Sorry, I'm not sure about some of the reorganizing I did here, so I want to hold off on this for now.

Allofich:
Resurrected this old PR.

The off-by-one difference in DF Unity material enums and those of classic threw me off for a little bit. Why are they changed? Anyway I used the DF Unity enums.

Minimum metal to hit is loaded from classic saves. This variable changes when the player becomes a vampire or werewolf/wereboar. It's not saved and loaded in DF Unity saves.

Interkarma:
The MetalTypes enum isn't exactly off by one - it's a DFTFU-specific enum designed to be accessible from an editor script in the Inspector window. It predates the full game, back to when I only had a general toolset in mind.

WeaponMaterialTypes in ItemEnums.cs came along later and should have the same values as classic. I'd be more than happy for enemies to use the classic WeaponMaterialTypes as MinMetalToHit - although +1 does the job as well. :)

Allofich:
Ok, I changed it to use WeaponMaterialTypes. Now I subtract one from the metals defined for enemies, since they use MetalTypes. Maybe I'll change them to use WeaponMaterialTypes as well later.

Interkarma:
Cool! Thank you for all these PRs, I didn't have time to review properly last night. Will work through them today and merge to master.

PR #300, opened 2017-05-22
Set gargoyle to have blood splash

Allofich:
I can see why it was set without, but the gargoyle should have a blood splash like in classic, because it has blood in its corpse texture.

### DIFF
--- a/Assets/Scripts/Utility/EnemyBasics.cs
+++ b/Assets/Scripts/Utility/EnemyBasics.cs
@@ -665,7 +665,6 @@ namespace DaggerfallWorkshop.Utility
                 HasRangedAttack1 = false,
                 HasRangedAttack2 = false,
                 CanOpenDoors = true,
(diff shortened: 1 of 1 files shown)


Interkarma:
At the time I wrote this, I was not able to produce any blood splashes on gargoyles in combat. Could have just been unlucky. I'll accept PR as you're confident they have blood splashes in game.

Allofich:
> I was not able to produce any blood splashes on gargoyles in combat

Really? Well before submitting the PR I checked in classic just to be sure and they did appear. If you care enough about it I have a save game before encountering a gargoyle that I can send you, but for now I'll consider this done.

Allofich:
Just checked in classic again and yes, it does appear for me at least.

Interkarma:
I'm happy this is done and trusted you here enough to merge. I was probably just unlucky at the time I tested this. Cheers!

PR #301, opened 2017-05-22
Track fatigue at 64 times character sheet value

Allofich:
Track fatigue internally at 64 times the value shown in the character sheet, the same as classic does. This will allow for implementing loss of fatigue values in the same way.

### DIFF
--- a/Assets/Scripts/Game/Entities/DaggerfallEntity.cs
+++ b/Assets/Scripts/Game/Entities/DaggerfallEntity.cs
@@ -73,7 +73,7 @@ namespace DaggerfallWorkshop.Game.Entity
         public ItemEquipTable ItemEquipTable { get { return equipTable; } }
         public int MaxHealth { get { return maxHealth; } set { maxHealth = value; } }
         public int CurrentHealth { get { return currentHealth; } set { SetHealth(value); } }
-        public int MaxFatigue { get { return stats.Strength + stats.Endurance; } }
+        public int MaxFatigue { get { return (stats.Strength + stats.Endurance) * 64; } }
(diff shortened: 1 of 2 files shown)


Interkarma:
Could you link me a source for this? I've never seen that fatigue is tracked internally at 64x. I've not spent much time with this however, won't be getting to this until the spells & effect update.

Allofich:
In the Daggerfall Chronicles it says

"The PC's maximum fatigue is: ( Strength + Endurance ) * 64"

Also, using the save game editor DaggeD, the fatigue for the game you select will show roughly 64 times what the in-game character sheet shows. It will probably be a little off from when it was divided by 64 into an integer and the remainder was dropped.

Also I'm trying to reverse-engineer the game and have found that it multiplies (strength + endurance) by 64 and, when forming the string for the character sheet, it divides them by 64.

Allofich:
Also the values listed in the Daggerfall Chronicles for the fatigue loss from various activities are

Do nothing: 11 points per minute
Swing weapon: 11 points per swing
Climb: 22 points per minute
Running: 88 points per minute
Swimming: 44 points per minute (except Argonians)
Jump: 11 points per jump

and these wouldn't make sense using the fatigue value from the Character Sheet. I loaded a game, jumped, then re-saved and checked the value in DaggeD, and it had fallen by 11 points, so these are probably accurate.

Interkarma:
Awesome thank you. 👍

Allofich:
This also means when you load any of your existing DF Unity save games, your fatigue bar will be nearly empty at first. But it's fine after you restore your fatigue by resting or when starting a new game, or when loading a classic save.

Interkarma:
That's not a problem, thanks again for this. :)

PR #302, opened 2017-05-23
Add fatigue loss for some actions

Allofich:
Adds fatigue loss for jumping and attacking, as well as per minute of game time based on whether the player is running or not. The values are from the Daggerfall Chronicles and I confirmed by testing in classic.

### DIFF
--- a/Assets/Scripts/Game/Entities/DaggerfallEntityBehaviour.cs
+++ b/Assets/Scripts/Game/Entities/DaggerfallEntityBehaviour.cs
@@ -42,9 +42,9 @@ namespace DaggerfallWorkshop.Game.Entity
         //private int SwimmingFatigueLoss = 44;     // According to DF Chronicles
 
         private int JumpingFatigueLoss = 11;        // According to DF Chronicles and verified in classic
-        bool AppliedFatigueLossThisJump = false;
+        private bool CheckedCurrentJump = false;
 
-        bool gameStarted = false;
+        private bool gameStarted = false;
 
         #endregion
 
@@ -106,16 +106,17 @@ namespace DaggerfallWorkshop.Game.Entity
                     else
                         Entity.DecreaseFatigue(DefaultFatigueLoss);
                 }
-                // Reduce fatigue when jumping
-                if (!AppliedFatigueLossThisJump && playerMotor.IsJumping)
+                // Reduce fatigue when jumping and tally jumping skill
+                if (!CheckedCurrentJump && playerMotor.IsJumping)
                 {
                     Entity.DecreaseFatigue(JumpingFatigueLoss);
-                    AppliedFatigueLossThisJump = true;
+                    Entity.TallySkill((short)Skills.Jumping, 1);
+                    CheckedCurrentJump = true;
                 }
                 // Reset jump fatigue check when grounded
-                if (AppliedFatigueLossThisJump && playerMotor.IsGrounded)
+                if (CheckedCurrentJump && playerMotor.IsGrounded)
                 {
-                    AppliedFatigueLossThisJump = false;
+                    CheckedCurrentJump = false;
                 }
             }
         }
--- a/Assets/Scripts/Game/Entities/DaggerfallEntityBehaviour.cs
+++ b/Assets/Scripts/Game/Entities/DaggerfallEntityBehaviour.cs
@@ -32,6 +32,20 @@ namespace DaggerfallWorkshop.Game.Entity
         EntityTypes lastEntityType = EntityTypes.None;
         DaggerfallEntity entity = null;
 
+        private int lastGameMinute = -1;
+        PlayerMotor playerMotor = null;
+
+        // Fatigue loss per in-game minute
+        private int DefaultFatigueLoss = 11;        // According to DF Chronicles and verified in classic
+        //private int ClimbingFatigueLoss = 22;     // According to DF Chronicles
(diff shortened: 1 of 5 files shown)


Allofich:
> The values are from the Daggerfall Chronicles and I confirmed by testing in classic.

I didn't check the climbing and swimming values in classic yet, since those aren't in DF Unity yet anyway.

Allofich:
Reworking this.

Interkarma:
No probs. Looking forward to merging when you're happy with it. This is great.

Allofich:
Shoot, I can't reopen this because I force-pushed. I'll have to make a new PR.

Well, if you're around, maybe you can give me advice. :)
Here it is, uploaded again.
https://github.com/Allofich/daggerfall-unity/tree/fatigue

One is that I found a problem after I sent the PR. I check the current game minute and reduce the fatigue when the game minute changes, but I found that on starting up DF Unity and loading a saved game this was always being triggered once at the beginning. I did a Debug.Log on the current minute and for some reason it seems it was being set to 30 once before being updated to the real value. I don't see where in the code it would be set to 30.

I also thought that maybe I put the code in the wrong place. I was thinking of trying to move it to DaggerfallEntity.cs but now I don't know. Where should these types of things go? I also noticed that the PlayerHealth.cs that I added to says "Temporary class" at the top so maybe I shouldn't be expanding it. Any advice is appreciated.

Interkarma:
Happy to help. :) Let me back to you in the morning, if I may. Just after 9pm here and I'm a bit shattered from today haha.

Allofich:
Found a trick to reopen after force-pushing. :)

Interkarma:
Some feedback so far:

I would prefer this not to be in PlayerMotor at all. It's only job is to move player controller around scene and it shouldn't be involved in gameplay logic. There's a hack in place right now to apply fall damage (which is where PlayerHealth comes in) but this will all go away when I get around to cleaning it up. I also prefer not to use SendMessage for new gameplay scripts. There's still a rare couple of places SendMessage is used in the old scripts but things have moved to delegate events since then.

I'd move your fatigue tracking to DaggerfallEntityBehaviour.Update(). From there you can access both the parent entity class with DecreaseFatigue() and cache PlayerMotor reference to check the IsRunning property. As you just want this to apply to player entity, check the entity type first in update before managing fatigue.

For the minute property being set to 30 at start, I'm guessing that's from the WorldTime > DaggerfallDateTime Inspector in editor (i.e. scene data). I'm not sure where else this might be getting set at startup. Maybe a better way of tracking game time passed is needed (like deltas) rather than checking if minute value has wrapped.

That's a lot to unpack. Let me know if any help needed and I'll be happy to assist. :)

Allofich:
I moved the game-minute tracking from PlayerMotor to DaggerfallEntityBehaviour.Update(). I found that the initial fatigue loss happened because these update functions were being called before the game time was set to the value read from the classic save or to the new game value. I fixed it by checking for GameManager.Instance.StateManager.GameInProgress.

I thought about adding in the event for when the player runs out of fatigue, but I haven't used events and delegates yet and am not sure how that should be done. There are events already for OnDeath, OnExhausted and OnMagickaDepleted, but again there is a comment about these being temporary. What should happen on player fatigue <= 0 is:

if (enemiesNearby) // same check as for when resting
       messagebox, no background color // you drop to the ground, you are easy prey and never awaken, textID 1072
       killPlayer
else
       messagebox, no background color // you drop to the ground, you awaken an hour later textID 1071
       rest (60 minutes) // same as resting normally for one hour, so health, fatigue and magicka heal

Assets/Scripts/Game/Entities/DaggerfallEntityBehaviour.cs
```diff
@@ -52,10 +66,19 @@ public DaggerfallEntity Entity
         void Start()
         {
             SetEntityType(EntityType);
+            playerMotor = GameManager.Instance.PlayerMotor;
         }
 
         void Update()
         {
+            // Wait until game has started and the game time has been set.
```
Interkarma: Would prefer this to come after check to change Entity type please. :)

Interkarma: Ideally move game started check inside your if (EntityType == EntityTypes.Player) block. Putting the game started check at front of method actually blocks the intended entity set and update, which should happen whether game is running or not.

Interkarma:
It's OK to use the OnExhausted event for now.

The reason many of these events and methods are temporary is because it's just placeholder. Work on these player systems aren't scheduled to start in earnest until 0.5. I've intentionally done the minimum possible on these systems because it's not quite time to flesh them out yet.

I'm happy to take in these enhancements to fatigue and tallying skills though. Just keep in mind this will all get reworked in the 0.5 chain. There's going to be a complete framework built around entities, effects, and attributes. :)

Interkarma:
Just a heads up - I'm doing some other work in PlayerEntity around factions right now. Will likely need to rework PR once that's complete. Apologies for the inconvenience.

Allofich:
> Just keep in mind this will all get reworked in the 0.5 chain.

That's fine. I'm just trying to figure out how classic does things, and I think there are benefits to implementing it functionally into DF Unity. Like how it just allowed me to notice the rest healing calculation was incomplete.

Allofich:
Moved the game started check to where you suggested.

Interkarma:
Great stuff, thank you for all the work on this!

PR #303, opened 2017-05-23
Tally weapon skill uses

Allofich:
Tallies weapon skill uses. I've managed to figure out a simple function in FALL.EXE used for tallying skills, and I added in calls to tally weapon usage. I also added in enums for skills based on their ordering in FALL.EXE.


Interkarma:
This looks good. Have to get to work, will review properly and merge when I can later today.

PR #305, opened 2017-05-25
Fix rest health rate calculation

Allofich:
The problem with the resting function was with the endurance modifier. If you had a negative endurance modifier you could regain no health or even lose health. In classic you regain a minimum of 1 no matter how low your endurance modifier is, and this PR makes DF Unity the same.

### DIFF
--- a/Assets/Scripts/Game/Formulas/FormulaHelper.cs
+++ b/Assets/Scripts/Game/Formulas/FormulaHelper.cs
@@ -94,7 +94,7 @@ namespace DaggerfallWorkshop.Game.Formulas
         // Calculate how much health the player should recover per hour of rest
         public static int CalculateHealthRecoveryRate(int medical, int endurance, int maxHealth)
         {
-            return (int)Mathf.Floor((HealingRateModifierMedical(medical) * HealingRateModifierMaxHealth(maxHealth)) + HealingRateModifier(endurance));
+            return Mathf.Max((int)Mathf.Floor(((HealingRateModifierMedical(medical) * HealingRateModifierMaxHealth(maxHealth)) + HealingRateModifier(endurance))), 1);
         }
 
         #endregion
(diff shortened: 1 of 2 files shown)


Allofich:
Also removed redundant casts.

Interkarma:
Thank you for looking into this so quickly. Please restore the casts however, they are not exactly redundant and I follow this coding style for a reason.

https://www.aubrett.com/article/information-technology/web-development/net-framework/csharp/csharp-division-floating-point

https://www.dotnetperls.com/divide

I would prefer if my existing code is not modified casually, especially when it has no bearing on the contribution being added. Cheers. :)

Allofich:
Oh, sorry. Visual Studio marked them as redundant and I assumed they were as well, since the parameter is an int, but sure, you're the project maintainer so I won't argue with you about anything like this. :)

Berzeger:
However, since one of the operands already is a float (10f for example), the cast actually is redundant. I believe Allofich is right.

Allofich:
What am I talking about, they are int in the parameter. Just ignore me. :)

Allofich:
Well I'll leave the call to Interkarma. I don't mean to get caught up on this. For now I'll put them back.

Allofich:
OK the cast changes are out, or I can put them back in if you change your mind Interkarma, whatever.

I also put back a Math.Floor that I had taken out because it seemed redundant to me with the (int) cast, but I see there is (int)Mathf.Floor elsewhere in the file, too, so I put it back.

Interkarma:
Thank you Allofich. :) I agree it shouldn't matter in this case because one side is an explicit float. But I've been caught out enough around this issue between what VS suggests as an optimisation and what Unity's Mono C# compiler actually does. I had to work around unexpected outcomes with the old compiler several times prior to 5.3.

Anyway, it might be unnecessary here, but it's still part of my coding style now to be defensive when dividing around ints. I prefer for the casts to be left in this case. Thank you mate. :)

PR #307, opened 2017-05-25
Implement zero-fatigue behavior

Allofich:
This implements the game behavior when the player runs out of fatigue. There are two cases, depending on whether there are enemies nearby:

Enemies nearby: "you are easy prey for monsters" message, and player dies
No enemies nearby: "you awaken one hour later" message, and player is forced to rest for one hour

I redid the CanPlayerRest() function into a AreEnemiesNearby() function, because I needed this functionality separated out from the rest function.

There are a few issues that could be ironed out.
1. When the player dies with the "you are easy prey for monsters" message, he can still move the view and move slightly, unlike when you die from damage. I don't know why.
2. The player death sound plays while "you are easy prey for monsters" message is up.
3.  If you lose your last fatigue points from jumping, you will start the jump before getting the message box, whereas in classic you just lose the fatigue points and get the message box with no jump. I use cancelMovement() so the jump doesn't continue after the message box, but ideally I'd like to copy classic and stop the jump before it starts, but I know you don't want game logic in PlayerMotor.cs.


Interkarma:
Great stuff! I'll merge this now and we can look at those issues one at a time.

PR #308, opened 2017-05-25
Fix issue with more than one songplayer in scene and tdbg command

TheLacus:
A while ago i reported that the songplayer value in tdbg always shows that no song is playing.
This is caused by having more than one songplayer in scene, as two of them are detected but the correct one seems to be always the second. 
I don't know if this information can be trusted, so i have implemented a foreach to to seek wich one is playing.

On an unrelated note, i noticed that you started updating the copyright date, do you wish contributors to do the same when they submit a PR?


Interkarma:
Awesome, thank you!

Interkarma:
About the copyright date - yes, please update if you need to change a file and notice it's out of date.

I forgot to update this at the start of year. Will just do files as they get changed for now.

PR #310, opened 2017-05-27
Implement lockpicking

Allofich:
Here's an implementation of lockpicking.

I've found what I think is the formula for lockpicking interior doors in FALL.EXE. It involves randomness so it's hard to test in-game. It seems like I have higher success in the original game but that might have just been luck.

I also made an educated guess that the lock value of doors should be divided by 8 compared to what it is in DF Unity. DF Unity was marking a lock of 160 or more as magically locked, while original Daggerfall shows the magically-locked message for a value of 20 or more. Dividing the DF Unity lock values by 8 produces what seems like correct values for the locked doors in the Castle Daggerfall throne room.

The basic lockpicking behavior also replicates this behavior from the original game:
1. If you click on a locked door outside of steal mode, you get a message that the door is locked.
2. If you click on a non-magically locked door in steal mode, you make a lockpicking attempt. If you succeed the door opens and a click sound plays, which DF Unity was using for when you clicked on a locked door.
3. If you fail to pick a lock, you can no longer attempt to pick it unless you leave and re-enter the area. In original Daggerfall this persists across save and load.

### DIFF
--- a/Assets/Scripts/Game/Formulas/FormulaHelper.cs
+++ b/Assets/Scripts/Game/Formulas/FormulaHelper.cs
@@ -115,6 +115,17 @@ namespace DaggerfallWorkshop.Game.Formulas
                 return 0;
         }
 
+        // Calculate chance of successfully lockpicking a door in an interior. If this is higher than a random number between 0 and 100 (inclusive), the lockpicking succeeds.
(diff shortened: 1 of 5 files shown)


Allofich:
I think there might be a separate lockpicking function used when you try to enter a door from outside, so I've labeled this as being for interiors.

Interkarma:
This is really excellent work, thank you. 👍

PR #311, opened 2017-05-27
Improvements to FaceWall

TheLacus:
* Rotate the GameObject to align it to non-straight dungeon walls
* Additionally to bring near the go if is distant from the wall, now it
also move it away if it's clipping in it.


Interkarma:
Your dedication to this is impressive and the results are worth it. You're laying the groundwork for some early mods that could not have existed by this stage without you picking up the ball and running with it. Great work, and cheers. 👍

TheLacus:
Thank you, i appreciate your kind words.

PR #312, opened 2017-05-29
Implement lockpicking difficulty messages

Allofich:
Implements lockpicking messages. I noted what seems to be an oversight in classic: the lockpicking difficulty messages in classic are always for the harder of two lockpicking formulas. That harder formula is used (I think) on exterior doors that lead into interiors, while an easier one is used for lockpicking the animating doors in interiors. I don't know whether the oversight is that the difficulty messages only account for one of the formulas, or if its that there are two different formulas.

I have hooked up the text to use the interior door formula (since DF Unity right now only has locked doors in interiors), so you will get accurate messages instead of the ones you will get in classic.

BTW, to be like classic, the text should be in the middle of the screen, the same spot currently being used for interaction mode changes. This spot is also used in classic for some other messages like "You have no arrows." when you try to equip a bow. If it's okay I'd like to change its name from ModeText to "MidscreenText" or something and use it for these other messages.

### DIFF
--- a/Assets/Scripts/Game/DaggerfallUI.cs
+++ b/Assets/Scripts/Game/DaggerfallUI.cs
@@ -376,11 +376,11 @@ namespace DaggerfallWorkshop.Game
             }
         }
 
-        public static void SetModeText(string message, float delay = 1.5f)
+        public static void SetMidScreenText(string message, float delay = 1.5f)
         {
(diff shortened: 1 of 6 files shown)


Interkarma:
No problems changing to "MidscreenText". Great work. :)

Allofich:
Updated.

PR #314, opened 2017-05-31
Identify and import more data from classic saves

Allofich:
Through a combination of testing with the save editor DagHex and confirming in-game, new values have been identified in classic saves. Current health, spell points and fatigue are imported as they are into DF Unity, as are skill use counts.

The player's reputation with commoners, nobility, scholars, merchants and underworld is based on the message you get about your reputation being higher or lower during character creation. DagHex has nobility and merchants switched compared to this, and DaggeD, another save editor, has the reputations in the opposite order of how they're shown in-game. I'm going with the in-game display as most reliable until it can be confirmed somehow.


Interkarma:
Looks like excellent work, thank you! 👍

PR #315, opened 2017-06-01
Fix player dying on starting new game

Allofich:
Looks like I shouldn't have changed the thing. I didn't test my changes in the last PR on starting a new character. This PR should fix it.

### DIFF
--- a/Assets/Scripts/Game/Entities/PlayerEntity.cs
+++ b/Assets/Scripts/Game/Entities/PlayerEntity.cs
@@ -87,7 +87,7 @@ namespace DaggerfallWorkshop.Game.Entity
         /// <summary>
         /// Assigns player entity settings from a character document.
         /// </summary>
-        public void AssignCharacter(CharacterDocument character, int level = 1, int maxHealth = 0)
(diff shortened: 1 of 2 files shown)


Allofich:
Hold on, I think I can simplify the fix.

Interkarma:
Haha all good. Thanks for the fix. :)

And a quick thing I just thought of then - does fatigue loss respect god mode when 'tgm' is toggled on? I didn't think to look at that earlier.

Allofich:
Oh, I just simplified the fix, but you merged it. Well, you can add the simplification to it yourself if you want. Just change the assignments to health, fatigue and magicka in AssignCharacter() to be lower case instead of upper case (PlayerEntity to DaggerfallEntity). Then you can get rid of the Mathf max statements, because the PlayerEntity OnDeath and OnExhausted events won't run anymore. Well, it's up to you.

Allofich:
Fatigue loss does not respect god mode, no.

Interkarma:
Please send another PR for the simplified fix. Will be happy to merge. :)

Sorry for the fast turnaround. I don't normally do merges around this time of day, but working from home and can be a little more proactive.

PR #317, opened 2017-06-01
Implement skill raising

Allofich:
Implements skill raising. From testing classic, I think the only factors that affect how many skill uses are needed to raise a skill are the skill value (it's percentage) and the player's level. Different skills have different modifiers to how many uses they need. Running has such a high one because it is tallied very quickly while moving.

### DIFF
--- a/Assets/Scripts/Game/Entities/DaggerfallSkills.cs
+++ b/Assets/Scripts/Game/Entities/DaggerfallSkills.cs
@@ -367,5 +367,72 @@ namespace DaggerfallWorkshop.Game.Entity
         {
             return GetPrimaryStat((DFCareer.Skills)index);
         }
+
+        // These are based on observations from testing classic and may not be completely accurate.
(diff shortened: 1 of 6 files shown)


Allofich:
There seems to be a bit of discrepancy between what the Daggerfall Chronicles says and what the game actually does.

The Daggerfall Chronicles says:
"If at least six game hours have passed since the last skill increase (of any skill)" to then increment the skill.

But if I'm not mistaken the game actually does the check when hours passed is > 6, not >= 6, and the time of the last skill increase is updated at this point whether skills qualify for advancement or not. So it would actually be a timer of the last skill check, not the last skill increase.

For now, I'm doing it the way the Daggerfall Chronicles says.

Allofich:
Actually.. I think I'm going to implement it the same was as the original game for now, because this could affect gameplay.

Allofich:
Okay, it's implemented like classic now.

Interkarma:
This is fantastic mate. I've not started down this path much at all, and to have this sort of research and contribution made is incredibly valuable. Thank you. :)

PR #318, opened 2017-06-02
Import some more information from classic saves

Allofich:
This imports a couple more things from classic saves:

1. Race for player to return to after they have become a vampire, werewolf or wereboar.
2. Hit points to gain on leveling up

Also corrected the race enums to match classic (they were off by 1) and added in vampire, werewolf and wereboar.

### DIFF
--- a/Assets/Scripts/API/Save/CharacterRecord.cs
+++ b/Assets/Scripts/API/Save/CharacterRecord.cs
@@ -94,7 +94,7 @@ namespace DaggerfallConnect.Save
             parsedData.armorValues = reader.ReadBytes(7);
 
             reader.BaseStream.Position = 0x58;
-            parsedData.levelUpModifier = reader.ReadInt32(); // the starting total of all the primary skills, the two top major skills and the top minor skill
+            parsedData.levelUpModifier = reader.ReadInt32();
             parsedData.opponentUnknown = reader.ReadInt32();
 
             reader.BaseStream.Position = 0x5c;
@@ -245,7 +245,7 @@ namespace DaggerfallConnect.Save
             public UInt16 transportationFlags;
             public Races race;
             public Byte[] armorValues;
-            public Int32 levelUpModifier;
+            public Int32 levelUpModifier; // The starting total of all the primary skills, the two top major skills and the top minor skill
             public Int32 opponentUnknown;
             public Int16 baseHealth;
             public Int16 currentHealth;
--- a/Assets/Scripts/API/Save/CharacterRecord.cs
+++ b/Assets/Scripts/API/Save/CharacterRecord.cs
@@ -126,10 +126,13 @@ namespace DaggerfallConnect.Save
 
             parsedData.skills = ReadSkills(reader, out parsedData.skillUses);
 
-            reader.BaseStream.Position = 0x016f;
+            reader.BaseStream.Position = 0x16f;
             parsedData.equippedItems = ReadEquippedItems(reader);
 
-            reader.BaseStream.Position = 0x01fd;
+            reader.BaseStream.Position = 0x1f2;
+            parsedData.race2 = ReadRace(reader);
+
+            reader.BaseStream.Position = 0x1fd;
             parsedData.timeStamp = reader.ReadUInt32();
 
             reader.BaseStream.Position = 0x230;
@@ -262,6 +265,7 @@ namespace DaggerfallConnect.Save
             public DaggerfallSkills skills;
             public Int16[] skillUses;
             public UInt32[] equippedItems;
+            public Races race2; // Stores character's original race for when returning from being a vampire, werewolf or wereboar
             public UInt32 timeStamp;
             public DFCareer career;
         }
--- a/Assets/Scripts/Game/Entities/EntityEnums.cs
+++ b/Assets/Scripts/Game/Entities/EntityEnums.cs
@@ -43,6 +43,9 @@ namespace DaggerfallWorkshop.Game.Entity
         WoodElf = 6,
         Khajiit = 7,
         Argonian = 8,
+        Vampire = 9,
+        Werewolf = 10,
+        Wereboar = 11,
     }
 
     /// <summary>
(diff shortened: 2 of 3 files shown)


Allofich:
Added import of resistance flags

Allofich:
Never mind, the resistance flags and immunity flags are already imported I see. Let me take that out.

Interkarma:
Interesting how I've performed (int)parsedData.race + 1 to force value read up to match one-based Races enum. I feel that I would have had a specific reason for doing this at time of writing, but cannot recall it right now.

I'll hold merging this for a bit until I go back through my past processes to determine why it was does this way.

Interkarma:
I can't find my reasoning for this right now. I'm willing to accept it's a mistake, but does seem very intentional on my side. Note the other enums are zero-based.

I'm still hesitant to change however. It would mean that a deserialized RaceTemplate from existing Daggerfall Unity save will no longer carry the correct race ID. This could change morphology behaviour for newly generated equipment and other downstream problems.

I'll look at this in more detail over the weekend. :)

Allofich:
Okay. The enums in this PR are correct as far as how classic uses them, at least.

Interkarma:
I have to agree for now as I can't divine my original reasoning. :) And by itself, I don't see a problem with your change. All new characters and freshly imported classic saves will be OK.

But to play nice with existing Daggerfall Unity saves would involve an uptick of save version and a migration step to decrement restored race IDs. Otherwise player race will say one thing but be handled differently internally based on the restored ID. Item generation is the first thing to come to mind as the race ID is ultimately used to select correct texture set fitting race body shape.

All else being equal, this really only impacts one-way classic save import, which the "+1" already takes care of. I guess it's a question of whether deleting a "+1" is a good enough reason to break existing saves, uptick save version, and write save migration handlers. That's the full impact, I'm afraid.

Considering classic save import is a one-way step and Daggerfall Unity saves are otherwise completely standalone, I'm inclined to leave it how it is for now. I'll come back to it over weekend. It's getting late here. :)

Allofich:
Sure, take your time. I didn't notice any problems during quick testing but if you have a particular case you can think of that might break in a Unity save, I can try it out. If saves are indeed broken then I can rewrite this to retain the +1 part.

Allofich:
Ah, yeah. I tried with an Argonian saved with master and when I clicked on a dead enemy I got the unsupported body morphology error. I'll undo the change then.

Interkarma:
Another case would be a Nord (old ID=3, new ID=2). After loading an existing DFU save game, the body morphology for subsequently generated items would still be made for ID=3, which is now Elf morphology. This would cause new armour/weapons not to align properly to character.

Thank you for undoing change for now. I agree you're matching the correct values to classic, and I can't for life of me recall why I did it this way (apart from a stupid mistake). It might come to me later... or it could just be a stupid mistake. :)

Allofich:
Oh, but we should still have the vampire, werewolf and wereboar race IDs. Let me add those back in.

Interkarma:
Awesome. Will need handling for body morphology of those races too. I haven't tested with vamp/were characters at all - I believe the body stays the same just the head is swapped.

Maybe in the case of vamp/were, assign morphology on base race?

Allofich:
Well werecreatures can't access inventory in the first place, but for vampires I think it's the base race morphology with the vampire face put on it. This is just speaking from memory.

Allofich:
Interesting, I just tried clicking on dead human enemies with a wereboar character, and although I get the "You cannot access the inventory while shapeshifted" message, the bodies say "This body has no treasure" if I keep clicking them. After reverting to regular form, they do have treasure when I click them.

Allofich:
Oh wait, I think the ones that I already clicked on as a wereboar still say "This body has no treasure" but the one or two I hadn't clicked on yet do have treasure? Unfortunately I killed the enemies on top of each other so it's hard to differentiate the bodies.

Interkarma:
When you import a vamp/were character from classic, what happens when GetBodyMorphology() is called? If it throws exception, will need to add vamp/were support to switch block. Can probably just use base race for purposes of defining body type.

Definitely more work needed way down the track when vamp/were gameplay implemented. For now, I'm just happy if these characters import and get loot fitting their body type (shapeshift form excluded for now obvs). :)

Allofich:
I tried the test above again, and I can confirm that clicking a dead human body will give the "You cannot access the inventory while shapeshifted" message and make the body have no treasure, even if you turn back into regular form. The body will have treasure if you turn back into normal form without having clicked it as a werecreature.

I'll try importing the character into DF Unity now.

Allofich:
Both master and this PR freeze with "KeyNotFoundException: The given key was not present in the dictionary." when I try to load the character, who is in wereboar form.

Interkarma:
Ah yes, will need a RaceTemplate instance for them as well after Argonian in RaceTemplate.cs.

Like I said, I haven't tested with vamp/were at all. Quite the can of worms. :)

Interkarma:
Merging this one now. Interesting how sometimes the small changes can result in the longest conversations. :)

One of us can add the vamp/were RaceTemplate stuff later, and then check the morphology setup for them. It's not like vamp/were characters are importable right now anyway, which avoids a lot of other problems.

Assets/Scripts/API/Save/CharacterRecord.cs
```diff
@@ -262,6 +265,7 @@ public struct CharacterRecordData
             public DaggerfallSkills skills;
             public Int16[] skillUses;
             public UInt32[] equippedItems;
+            public Races race2; // Stores character's original race for when returning from being a vampire, werewolf or wereboar
```
Interkarma: @Allofich I'm hoping you might be able to help with some trouble I'm having with importing lycanthrope saves. When importing a classic character infected as either a werewolf/wereboar, I've noticed that race/race2 will both be the same value (e.g. Breton) unless our lycanthrope is actively transformed at the time save was made. If the character is not transformed at time of save, I can't seem to find a way to identify if they are a werewolf/wereboar.

Is this something you might know more about, so I can check weretype on import even if not transformed at time of save? Thank you. :)

Allofich: I commented in https://github.com/Interkarma/daggerfall-unity/commit/dd4980dd828bbaabc872199f066bd4c955aaec07

PR #319, opened 2017-06-02
Restore some settings from classic saves

Allofich:
Does as the title says.


Allofich:
Added another one for god mode.

Interkarma:
Nice find, I wasn't aware there was a god mode enabled value stored in classic saves. 👍

PR #320, opened 2017-06-02
ModSettings: Tuples, improvements on Cancel button

TheLacus:
* Added support for tuples in ModSettings and ModSettingsWindow.
* Do not write to disk automatically when a preset is applied, let the
user Save or Cancel all changes.
* Minor changes to colors: use a brighter blue for section titles.


Interkarma:
Great stuff 👍

PR #322, opened 2017-06-03
Implement player leveling

Allofich:
Implements player leveling. As far as I know everything that should be implemented is implemented. Unity saves will at some point need to store the sum of starting skills used in calculating levels. For now this value is estimated based on the current level.

### DIFF
--- a/Assets/Scripts/API/Save/CharacterRecord.cs
+++ b/Assets/Scripts/API/Save/CharacterRecord.cs
@@ -69,6 +69,7 @@ namespace DaggerfallConnect.Save
             doc.currentSpellPoints = parsedData.currentSpellPoints;
             doc.currentFatigue = parsedData.currentFatigue;
             doc.skillUses = parsedData.skillUses;
+            doc.startingLevelUpSkillSum = parsedData.startingLevelUpSkillSum;
 
             return doc;
         }
@@ -94,7 +95,7 @@ namespace DaggerfallConnect.Save
             parsedData.armorValues = reader.ReadBytes(7);
 
             reader.BaseStream.Position = 0x58;
-            parsedData.levelUpModifier = reader.ReadInt32();
+            parsedData.startingLevelUpSkillSum = reader.ReadInt32();
             parsedData.opponentUnknown = reader.ReadInt32();
 
             reader.BaseStream.Position = 0x5c;
@@ -245,7 +246,7 @@ namespace DaggerfallConnect.Save
             public UInt16 transportationFlags;
             public Races race;
             public Byte[] armorValues;
-            public Int32 levelUpModifier; // The starting total of all the primary skills, the two top major skills and the top minor skill
+            public Int32 startingLevelUpSkillSum; // The starting total of all the primary skills, the two top major skills and the top minor skill
             public Int32 opponentUnknown;
             public Int16 baseHealth;
             public Int16 currentHealth;
--- a/Assets/Scripts/Game/Entities/PlayerEntity.cs
+++ b/Assets/Scripts/Game/Entities/PlayerEntity.cs
@@ -43,6 +43,11 @@ namespace DaggerfallWorkshop.Game.Entity
         protected short[] skillUses = new short[34]; // TODO: Save to and load from DF Unity saves
         protected uint timeOfLastSkillIncreaseCheck = 0; // TODO: Save to and and load from DF Unity saves, load from classic saves
 
+        protected int startingLevelUpSkillSum = 0;
+        protected int currentLevelUpSkillSum = 0;
+
+        protected bool readyToLevelUp = false;
+
         #endregion
 
         #region Properties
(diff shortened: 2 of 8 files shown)


Interkarma:
Holy crap dude you are killing this, awesome work. 👍 

I can add these items to DFUnity saves tomorrow. Just to summarise, you need the following saved and loaded:

* PlayerEntity.skillUses[]
* PlayerEntity.timeOfLastSkillIncreaseCheck
* PlayerEntity.startingLevelUpSkillSum
* PlayerEntity.currentLevelUpSkillSum

Please just let me know if I've missed anything. I'm about to head out for the night and didn't have long to review code.

Allofich:
> Holy crap dude you are killing this, awesome work. 
Thanks. I happen to have a lot of free time at the moment to work on this, but maybe not so much in the near future.

> PlayerEntity.skillUses[]
> PlayerEntity.timeOfLastSkillIncreaseCheck
> PlayerEntity.startingLevelUpSkillSum

Yes for these. If you want to change the data type, like if you don't want to use an array or something, that is fine.

> PlayerEntity.currentLevelUpSkillSum

This can be calculated from the current player skill values so it isn't necessary to save it.

Interkarma:
I've added those three items to save/load now. Please test and let me know if anything else I can do.

Great work. :)

Interkarma:
Hey just some feedback on levelling. I started a level 1 warrior and went for a run through PH in both classic and DFUnity. Here's a few UI quirks I noticed on our end - nothing major at all. :)

* Once ready, level-up did not trigger by resting like it does it classic. In fact it didn't trigger until I fast travelled to a city.
* I heard the level-up chime but character window didn't open automatically.
* Stat rollout was not visible on character window so impossible to assign points or click exit (screens below)
* Can still close with ESC key though without assigning points. Will need to prevent ESC from closing this window when a level-up in progress.

Otherwise, the level-up skill increases and notifications seemed to work perfectly as per classic. Excellent work!

Please just let me know I anything I can do to assist with the above. If you're not available to fix, I will loop back and look at these as soon as I have some time away from quest system.

levelup-norollout1
levelup-norollout2

Allofich:
At the moment I can only confirm the ESC key issue. Did you have this issue with the current master?

playerEntity.RaiseSkills(), which sets in motion the level-up progress, is called in DaggerfallRestWindow's OnPop(). It's also called when fast traveling, so I don't see why fast travel would work for you while resting wouldn't. Do you have the save file?

Interkarma:
Don't have a save file for this from before issue. My test process was new character > clear PH > rest a lot. Level didn't ding until fast travel. Let me try and reproduce and I'll send it through when I can. :)

Allofich:
> My test process was new character > clear PH > rest a lot. 
I tried the same, but didn't have a problem. I just tried again (latest master), and still didn't have a problem, so unless it was caused by some change you had in a test branch or something, there must be something more to it. If you can reproduce the problem and the cause remains elusive maybe we can each record a video of our attempts and find out where our behavior is diverging.

Interkarma:
Thanks for taking a look! I haven't been able to reproduce the level-up not happening until fast travel. That was probably just an edge-case of some kind. I'm not too worried about that one right now. :)

I have been able to reproduce the stat spinner issue each time however. The difference I think is that I'm testing in a build. It works fine inside the editor, but doesn't display in build. No exceptions in output log either >.<

Interkarma:
I can't see anything wrong with your code to add/remove the stat spinner either. Must be something up with the UI loading on my end. I never tested that spinner outside of character creation either, so probably not surprising it turned out to be less robust than expected.

I'm happy to see this as a problem on my end if you like. I'll dive into it when I can ahead of the next builds dropping.

Allofich:
Thanks, I think I would like if you would take on the task of trying to fix it, then.

Allofich:
I can reproduce when testing in a build, too. A few other things I noticed:
1. The problem will happen with a classic save, too, so it's not related to having used the stat spinner in character generation.
2. If you build an executable (regardless of whether you then run it), and then if you run in the editor after that, the problem will happen when running in the editor. If you close and re-open Unity, or if you double-click the DaggerfallUnityGame.unity scene to "restart" it, then the problem won't be there when you run in the editor.

Allofich:
Another thing, I just tried with the latest Unity, 5.6.1f1, and the problem still exists there.

Allofich:
Looks like I fixed the problem by just more closely copying the way the statsRollout is made in character creation. I don't know why I didn't try that before. I don't quite understand the difference between instantiating something with a new object in the class definition and doing it in the setup() method, but that seems to be what does it. I'll make a PR.

Interkarma:
Great work, thank you for persisting with this. You've saved me a loop back later to look into this. 👍 

I'm about to shut down for the night. Will merge in change in morning when I'm a bit fresher. Cheers!

PR #323, opened 2017-06-04
ModSettings: support multiple pages

TheLacus:
Places settings in a list of panels instead of a single one, in order to
support more than 38 keys.
A button and the keyboard Tab key can be used to switch among pages.


Interkarma:
This is a good idea! 👍

PR #324, opened 2017-06-04
Correct stat raising calculations

Allofich:
Stat raising calculations should now be correct with this PR. This works for all the test cases I've used.

In a werewolf save I'm using (downloaded from the Internet), it looks like the skills raised by being a werewolf also have their advancement modifier halved, so they rise more easily. The character in the save is in human form but gets this benefit it seems. There may be something similar with vampires. This will need some more testing but I'm not going to bother with it for now.


Interkarma:
Thank you for continuing to work on this! 👍

PR #325, opened 2017-06-04
Fix for null reference on Unity saves

Allofich:
In an earlier PR I fixed null reference to skillUses[] on a new character, but it also happens with Unity saves that were saved before the additional data. This PR initializes the array in that case. It also sets EstimateStartingLevelUpSkillSum() to only be called if the skillsum is 0 because nothing was in the save file.

I guess stuff will continue to be added to the save file, so is this the way to handle missing data? Maybe defaults should be set first and things only assigned from the loaded file if they are non-null/non-0.


Interkarma:
Thank you. Yep need to make sure data not otherwise recovered from saves is initialised with some sensible defaults. This is fairly automatic for field values but gets more interesting around arrays. Lists can also help avoid this as their internal array is always initialised to 0 members provided list itself is created at construct time. I think an array is perfect in this case however. :)

I'm happy for this to be something that happens on a case-by-case basis, as "sensible defaults" might mean something a bit different each time.

PR #326, opened 2017-06-07
Fix input text overlapping

Allofich:
This is sort of a patched-on-top fix for overlapping text when you click on the banner/statue at Shedungent. This problem has been around since one of the changes I made some time ago to text formatting.

What classic seems to actually do is add these input text prompts to the same "popup text" lines used for things like "-- just died."


Interkarma:
Cheers mate.

PR #331, opened 2017-06-12
Do nothing if rest key is pushed while in air

Allofich:
Currently in DF Unity if you push the rest key while in the air you get the "There are enemies nearby." message. With this PR instead nothing happens, the same as classic.


Allofich:
Or, hold on. Let me rewrite it a bit.

Allofich:
Ok, rewrote it.

PR #332, opened 2017-06-12
Import player orientation from classic saves

Allofich:
This imports the pitch and yaw values from classic saves, resolving http://forums.dfworkshop.net/viewtopic.php?f=24&t=424&sid=3e361409f513f292e15381a7752cc11f.

After the pitch and yaw data is what seems to be a copy of the player position data found in the header. It is this position data, and not the player position data in the header, that determines where the player is when a save is loaded in classic. There is additional data after that, but I don't know what it is.


Interkarma:
Woo! Thanks for this one! :)

PR #333, opened 2017-06-13
Fix 0-hour rest/loiter taking an hour

Allofich:
For http://forums.dfworkshop.net/viewtopic.php?f=24&t=529&p=5281#p5281.


Interkarma:
Great, thank you for fixing. 👍

Interkarma:
Just testing this one, unfortunately rest will end instantly if player specified "until healed". Going to revert this one for now. :)

PR #334, opened 2017-06-14
Raise foe max spawn limit

Allofich:
I saw the comment about foe spawn limits and decided to test it. I found no limit to the number of spawns classic can do of the same foe at a time.

I know for 100% sure it can spawn 20 at a time, as I was able to count them. I then raised it to 50, and what looked like 50 (I didn't actually count this time) spawned. An additional 50 (or what looked like 50) of another enemy spawned at the same time. If there is any limit it is very high.

### DIFF
--- a/Assets/Scripts/Game/Questing/Foe.cs
+++ b/Assets/Scripts/Game/Questing/Foe.cs
@@ -25,8 +25,8 @@ namespace DaggerfallWorkshop.Game.Questing
     /// </summary>
     public class Foe : QuestResource
     {
-        // Arbitrary limit to spawn count, need to check this is at/above max classic expects
-        const int maxSpawnCount = 10;
+        // Testing shows classic can spawn at least 50 of the same enemy at a time, and probably many more
+        const int maxSpawnCount = 50;
 
         int spawnCount;
         MobileTypes foeType;


Allofich:
Just to be sure I went ahead and counted, and all 50 were there. We probably don't even need to put a limit, actually.

Interkarma:
Oh cool, did you test that by injecting a quest into classic? 👍

Allofich:
Yeah, the one that spawns ghosts and wraiths in Daggerfall.

Interkarma:
I like the limit - it would be very easy for a quest scripter to crash someone's game by creating a quest with this set to an impossible number. A sanity check of some kind is always good.

When I wrote the comment, I wasn't so much thinking about what Daggerfall could handle, but what was the maximum used by the canonical quests. That was the number I was leaving myself a reminder to find. :)

I'm pretty happy with 50 for now though.

Interkarma:
An update on this. I've set the limit back to 20, and I'm likely to reduce to the initial 10 again.

As an FYI, this value is completely unrelated to "send aFoe every m minutes n times with k% success", which I believe is where you did the testing. Rather, this value sets a sanity limit on "Foe  _symbol_  is  nn  anEnemy" action. The most Daggerfall ever uses in canonical quests is 6.

I'll make a call on this later in testing after building out quest enemy handling further.

Allofich:
That's fine.

PR #336, opened 2017-06-14
Fix 0-hour rest and loiter taking 1 hour

Allofich:
Let's try this again.

I also noticed that the stupid healing rate when resting is acting up again, or maybe I didn't fix it correctly the last time. I'm only healing 1 point with a character who should heal more. I'll take a look at that next.


Allofich:
Oh, never mind, the character I was testing with has "rapid healing" in classic. That plus the -1 endurance modifier not being applied in classic.

PR #337, opened 2017-06-14
Implement rapid healing

Allofich:
Implemented rapid healing (none, in light, in darkness, always) and cleaned up the earlier health recovery formula. This is checked against classic and should be correct.

### DIFF
--- a/Assets/Scripts/Game/Entities/PlayerEntity.cs
+++ b/Assets/Scripts/Game/Entities/PlayerEntity.cs
@@ -398,7 +398,7 @@ namespace DaggerfallWorkshop.Game.Entity
             {
                 // TODO: Duplicates rest code in rest window. Should be unified.
                 DaggerfallUnity.Instance.WorldTime.Now.RaiseTime(1 * DaggerfallDateTime.SecondsPerHour);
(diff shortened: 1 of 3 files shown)


Assets/Scripts/Internal/WorldTime.cs
```diff
@@ -46,6 +46,12 @@ public DaggerfallDateTime Now
             get { return DaggerfallDateTime; }
         }
 
+        public bool isDayTime()
```
Interkarma: WorldTime is just a value container. This doesn't belong here, I'm sorry. :) This functionality already exists in DaggerfallDateTime.IsDay. You can check from anywhere through the singleton like so:
    if (DaggerfallUnity.Instance.WorldTime.DaggerfallDateTime.IsDay) { }

Interkarma:
Awesome, thank you for adding that and fixing up formula. 👍

Allofich:
I didn't notice DaggerfallUnity.Instance.WorldTime.DaggerfallDateTime.IsDay. The PR uses it now.

Interkarma:
Cheers mate

PR #338, opened 2017-06-15
Identify more data from classic saves

Allofich:
I identified quite a lot more data from classic saves with the help of Donald Tipton's save-viewing program chunktcl. I also found on my own that pitch and yaw seem to be default properties of all records.

The player's location is now gotten from the type 0x04 record, which I named "CharacterPositionRecord." I renamed the old CharacterPositionRecord to  HeaderCharacterPositionRecord.

Some of the record root data already being read into SaveTreeBaseRecord was reading data from the wrong location, but they should be reading from the correct spots now.

One thing that I think would now be possible is to recreate nearby monsters from save games, since I found the monster record and it includes their position. They have their name at the beginning of their data, so we could use that to determine the type. I'm not real sure about how you would want to handle spawning those monsters from a saved game, though.

### DIFF
--- a/Assets/Scripts/API/Save/SaveTree.cs
+++ b/Assets/Scripts/API/Save/SaveTree.cs
@@ -30,7 +30,6 @@ namespace DaggerfallConnect.Save
         public SaveTreeHeader Header;
         public SaveTreeLocationDetail LocationDetail;
         public SaveTreeBaseRecord RootRecord = new SaveTreeBaseRecord();
-        public PlayerDirectionRecord DirectionRecord = new PlayerDirectionRecord();
         public Dictionary<uint, SaveTreeBaseRecord> RecordDictionary = new Dictionary<uint, SaveTreeBaseRecord>();
 
         // Private fields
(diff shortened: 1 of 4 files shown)


Interkarma:
This looks like solid work, great stuff. It's been well over a year since I last dived into classic save, I'll merge this one into its own branch and check over weekend.

Is the new position record always identical to what you've now renamed to header position? The current world placement while outside seems to be quite accurate already using the header position.

Allofich:
> Is the new position record always identical to what you've now renamed to header position?

That seems to be the case.

Interkarma:
This PR breaks item equip table data. Any characters imported under this branch come in starkers now. :)

Are any other offsets affected?

Current vs new save import below:
current-save-import
new-save-import

Allofich:
I double-checked the offsets, and they should be correct. They were off by one before.

It looks like this off-by-one was being compensated for in the logic for determining if something is equipped or not. Removing that fixes the problem.

PR #342, opened 2017-06-17
Fix items in classic saves not being equipped

Allofich:
Fixes problem mentioned in https://github.com/Interkarma/daggerfall-unity/pull/338.

### DIFF
--- a/Assets/Scripts/Editor/SaveExplorerWindow.cs
+++ b/Assets/Scripts/Editor/SaveExplorerWindow.cs
@@ -410,7 +410,7 @@ namespace DaggerfallWorkshop
             uint[] equippedItems = characterRecord.ParsedData.equippedItems;
             for (int i = 0; i < equippedItems.Length; i++)
             {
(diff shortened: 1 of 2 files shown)


Allofich:
I haven't touched the save explorer until now, but it looks like it uses the same logic for equipped items. Let me add it to this PR.

Allofich:
Also updated the Save Explorer to use the actual player position data, not the player position data in the header. I've checked and double-checked this, btw. Changing this value (in the 0x04 type record) in SAVETREE.DAT changes where the player is located, while changing the position values in the header does nothing that I can tell.

Interkarma:
Looks like a great outcome. Thank you for all that. 👍 

I wonder if the header position is used for a quick initial check on load (e.g. what region textures to load, what chunks are needed) before loading the actual player record at same position later in the process. I guess it's fairly moot at this stage as all we really need is the player position and either gets the job done.

I often use the Save Explorer to play with visualising save data, and probably just left it wherever I was at the time. I seem to recall I was playing with something around vertical placement and water? It's been a while... I wouldn't worry too much about it. :)

PR #343, opened 2017-06-17
Correctly read cheat variable for godmode

Allofich:
Currently in master, if you load a classic save that has both the invulnerability cheat and the cheat to reveal all map locations active, godmode will not be activated.

Also includes code for reading the cheat to reveal all map locations. This is commented out for now until DF Unity supports hidden map locations, to avoid the unused variable warning.

### DIFF
--- a/Assets/Scripts/API/Save/SaveVars.cs
+++ b/Assets/Scripts/API/Save/SaveVars.cs
@@ -29,19 +29,34 @@ namespace DaggerfallConnect.Save
         const string filename = "SAVEVARS.DAT";
         const int weaponDrawnOffset = 0x3BF;
         const int gameTimeOffset = 0x3C9;
-        const int godModeOffset = 0x173B;
+        const int cheatsOffset = 0x173B;
         const int lastSkillCheckTimeOffset = 0x179A;
         const int factionDataOffset = 0x17D0;
         const int factionDataLength = 92;
 
         bool weaponDrawn = false;
         uint gameTime = 0;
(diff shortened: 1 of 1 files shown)


Interkarma:
Good fix, thanks - and bonus points for suppressing warning for now. :)

I'll be adding location discovery soon in DFUnity as part of quest system (both travel map and location buildings). I'm not that concerned about importing the classic building nameplates right now (if at all, I hate the way DF handled this), but getting the player's discovered world locations from classic saves would be nice.

I haven't even started down the path of researching how discovered locations are stored in classic saves yet. Is that something you'd be interested in helping with?

Allofich:
> Is that something you'd be interested in helping with?

Sure, if I can. I actually was taking a shot at that yesterday, but I ran up against a problem.

What I found was that the data must be stored in (or dependent in some way on) MAPSAVE.SAV in a given save's folder. Exchanging the MAPSAVE.SAV between two saves changed the revealed locations.

Unfortunately, while I was hoping that simply uncovering a single location and comparing MAPSAVE.SAV before and after would show only a single byte somewhere changing, instead a fairly large chunk of the file changed, and I couldn't pin down in what manner it had changed.

Interkarma:
I thought since you were diving into classic saves lately it might me a subject you were interested in. :)

Don't spend any more time on it than you want to, though. I'm implementing discovery my own way in DFUnity, and it's not the end of the world if we can't import classic discovery state right away. This will come over time. It won't be an issue for players starting new characters anyway, and imported characters can workaround with cheat/console.

PR #344, opened 2017-06-17
MeshReplacement: name, optimization

TheLacus:
* Added name as discussed
* Resource-import only on dev builds, import only from mods on released game.
* Optimization on method call to avoid repeating some code. Basically  some minor changes to integrate more default and custom import.

### DIFF
--- a/Assets/Scripts/Internal/DaggerfallInterior.cs
+++ b/Assets/Scripts/Internal/DaggerfallInterior.cs
@@ -222,7 +222,7 @@ namespace DaggerfallWorkshop
                     doors.AddRange(GameObjectHelper.GetStaticDoors(ref modelData, entryDoor.blockIndex, entryDoor.recordIndex, modelMatrix));
 
                 // Get GameObject
-                if (MeshReplacement.ImportCustomGameobject(obj.ModelIdNum, modelMatrix.GetColumn(3), node.transform, GameObjectHelper.QuaternionFromMatrix(modelMatrix)) == null)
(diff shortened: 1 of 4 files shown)


TheLacus:
I'm going to create another PR for flat action, but i still haven't found anything to test it.

On a similar topic, billboard -> model replacement for NPCs is going to have troubles to implement all the progress you're doing on questing since they don't have a DaggerfallBillboard component.

Maybe it's possible to add it without camera rotation and animations and only benefit for the relevant parts?
I know this is the last thing you want to think about now, just pointing it out for the future.

Interkarma:
Awesome thank you!

You raise a good point about the current usage of DaggerfallBillboard and flat involvement in quest system.

Rather than keep the flat around, I'd rather offload the required information (really just factionID and some other small stuff) to a new component. It really just needs to be a data container. I'll see what I can do today.

Interkarma:
I've just a quick pass over the code, and fortunately DaggerfallBillboard is not highly involved in the quest side of things. The Person resource itself doesn't use it at al, and QuestNPCClickHandler (which I'm thinking of expanding to handle more roles) has all the data it needs to be standalone as well. The only times I really use DaggerfallBillboard is when adding a quest NPC to scene.

From your end, the main challenge will be assigning correct Quest*Handler component to replacement model. This can be done in a similar way to now, but will have to happen at the point of GameObjectHelper.AddQuestNPC() for example.

Let's wait a bit for all this to settle down into a final form. Then we can safely experiment with some ideas.

TheLacus:
Thank you, I'll just add the action component for now then.

PR #347, opened 2017-06-19
Identify more save variables

Allofich:


Interkarma:
Looks good! It's great to see more of these values become identified.

PR #350, opened 2017-06-20
Identify classic save tree header data

Allofich:
Identified data in the classic save tree header and modified the SaveExplorer a little to match.

To keep compatibility with the current loading behavior, which uses the byte at 18 as the record length of SaveTreeLocationDetail, I reset the stream here for that class, but that byte is actually the current environment of the player.

### DIFF
--- a/Assets/Scripts/API/Save/SaveTreeHeader.cs
+++ b/Assets/Scripts/API/Save/SaveTreeHeader.cs
@@ -25,15 +25,16 @@ namespace DaggerfallConnect.Save
     public class SaveTreeHeader
     {
         // Consts
-        public const int HeaderLength = 18;
-        public const int VersionNumber = 0x26;
+        public const int HeaderLength = 19;
+        public const int VersionNumber = 0x126;
 
         // Public fields
         public long StreamPosition;                             // Starting position in file stream
         public byte[] RawData;                                  // Raw byte data read from stream
-        public Byte Version;                                    // Version - must be 0x26
+        public int Version;                                     // Version - must be 0x126
         public HeaderCharacterPositionRecord CharacterPosition; // Character position
-        public UInt16 Unknown;                                  // Unknown
+        public UInt16 MapID;                                    // ID of current map. 0xFFFF if not in a location.
+        public Byte Environment;                                // Player environment. 1 = outside, 2 = building, 3 = dungeon
 
         /// <summary>
         /// Reads header from file.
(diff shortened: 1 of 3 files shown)


Allofich:
The environment is shown in the save explorer. 1 is outside, 2 is in a building, and 3 is in a dungeon. Where would be a good place to make an enum for this?

Assets/Scripts/API/Save/SaveTreeHeader.cs
```diff
@@ -25,15 +25,17 @@ namespace DaggerfallConnect.Save
     public class SaveTreeHeader
     {
         // Consts
-        public const int HeaderLength = 18;
-        public const int VersionNumber = 0x26;
+        public const int HeaderLength = 19;
+        public const int VersionNumberNew = 0x126; // Save from Daggerfall 2.17
```
Interkarma: Just a note: latest Daggerfall version is .213 (fully 1.07.213). :)

Assets/Scripts/API/Save/SaveTreeHeader.cs
```diff
@@ -64,25 +66,20 @@ void ReadRawData()
...
-            CharacterPosition = new HeaderCharacterPositionRecord();
-            CharacterPosition.RecordType = reader.ReadByte();
-            if (CharacterPosition.RecordType != (int)RecordTypes.CharacterPosition)
-                throw new Exception("Expected CharacterPosition in SaveTreeHeader has an invalid record type, must be 0x01.");
-
-            // Read CharacterPosition.Unknown
-            CharacterPosition.Unknown = reader.ReadUInt16();
+            // Read Version - must be 0x125 or 0x126
+            Version = reader.ReadInt32();
+            if (Version < VersionNumberOld || Version > VersionNumberNew)
```
Interkarma: I would prefer behaviour to match original design and support only .213 save versions. My early testing found a lot of inconsistencies in unpatched Daggerfall saves and I consider .213 to be the baseline for classic save import. Cheers. :)

Interkarma:
Since it's only going to be used by importers, please pop the enum as public in SaveTree within "Structs & Enums" region.

Assets/Scripts/API/Save/SaveTreeHeader.cs
Allofich: Whoops. I'll fix that.

Assets/Scripts/API/Save/SaveTreeHeader.cs
Allofich: Ok. The latest Daggerfall still accepts 1.0 saves, but I doubt many people have 1.0 saves lying around. :)

Assets/Scripts/API/Save/SaveTreeHeader.cs
Allofich: Well actually since it will just be one version number, as before, I'll remove the comment. I wonder what version number saves from Daggerfall versions between 1.07.213 and 1.0 were.

Assets/Scripts/API/Save/SaveTreeHeader.cs
Interkarma: I found a few on my first release haha. There's definitely people out there who never patched. :) That's why I put in the check, as these saves don't always load properly (slightly different format).

Classic might accept 1.0, but it has the luxury of 100% knowing the file format. We're kind of piecing it together and it's enough just to support .213 I think.

Interkarma:
This looks OK, will merge straight to master and let you know if any issues.

PR #351, opened 2017-06-20
Match shop quality text thresholds to classic

Allofich:


Interkarma:
Awesome, thank you for taking the time to test this and refine. 👍

PR #352, opened 2017-06-20
Fix hit point calculations

Allofich:
For level up, I was just adding the Career.HitPointsPerLevelOrMonsterLevel. Actually it is a random number between this value and half of it, plus the endurance modifier.

### DIFF
--- a/Assets/Scripts/Game/Entities/EnemyEntity.cs
+++ b/Assets/Scripts/Game/Entities/EnemyEntity.cs
@@ -86,7 +86,7 @@ namespace DaggerfallWorkshop.Game.Entity
 
                 // Enemy class is levelled to player and uses same health rules
                 level = GameManager.Instance.PlayerEntity.Level;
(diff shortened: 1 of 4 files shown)


Allofich:
Also fixed the formula for determining the player's initial health. This is not affected by the endurance modifier and does not involve any randomness. I'm pretty sure this is what is used for human enemies' health, too.

Interkarma:
This seems very nice on a quick playthrough. Monsters felt a bit too paper-thin before and now they feel more like Daggerfall. This will be awesome once all the proper to-hit and other combat equations are accounted for. Great work! 👍 

Speaking of which, are melee formulas also something you're considering building out? I have no complaints if that's a direction you want to head down.

Allofich:
> Speaking of which, are melee formulas also something you're considering building out?

Yes, although the damage and to-hit calculations seem to have a lot of different factors to them and I haven't figured it all out yet. By the way, I thought that some time ago another contributor added damage based on weapon type, but was that dropped for some reason?

Interkarma:
I vaguely recall the PR for this. All I can remember is that it needed some work and contributor never returned to complete, so PR was closed. It was a far cry from real combat formulas though.

PR #353, opened 2017-06-22
Implement weapon damage calculations

Allofich:
This puts in the basic weapon damage calculations for the player. The weapon type, material and strength modifier are all that are considered so far.

I plan to keep adding bits to this as I figure out the formulas, but first of all here are the basics.

### DIFF
--- a/Assets/Scripts/Game/FPSWeapon.cs
+++ b/Assets/Scripts/Game/FPSWeapon.cs
@@ -62,6 +62,12 @@ namespace DaggerfallWorkshop.Game
         int currentFrame = 0;
         Rect curAnimRect;
 
+        #region Properties
+
+        public WeaponStates WeaponState { get { return weaponState; } }
+
+        #endregion
+
         void Start()
         {
             dfUnity = DaggerfallUnity.Instance;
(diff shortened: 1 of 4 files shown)


Interkarma:
This looks quite good. Will review properly over weekend and merge. Cheers for the work. 👍

Interkarma:
Just one bit of feedback. I would prefer if the following were moved elsewhere and instead accepted the DaggerfallUnityItem as an input parameter to method.

* GetEquipSound
* GetSwingSound
* GetWeaponSkillUsed

DaggerfallUnityItem itself is intended to hold the item itself. Things you "do" with that item should be elsewhere. A more fitting place for these would the ItemHelper class, which already has other methods along these lines. You can access the ItemHelper from anywhere using DaggerfallUnity.Instance.ItemHelper.

Interkarma:
On second thought, this is OK.  They could be seen as just properties of the item itself under the right circumstances. I take it back, these are fine. :)

Allofich:
Great, I'll keep adding bits to the damage calculations as I can.

Allofich:
If you change your mind and want to move anything around, that's fine. You have the best understanding of the code structure, of course.

Interkarma:
BTW, I had a play with this earlier and am very happy so far. Thank you for your work on this. :)

PR #354, opened 2017-06-23
Add enums for furniture items

Allofich:
The group that was marked Error is furniture, according to the save game editor Daghex, which seems to be accurate about these sorts of things.

### DIFF
--- a/Assets/Scripts/Game/Items/ItemEnums.cs
+++ b/Assets/Scripts/Game/Items/ItemEnums.cs
@@ -37,7 +37,7 @@ namespace DaggerfallWorkshop.Game.Items
         Artifacts,
         MensClothing,
         Books,
-        Error,                                      // Not sure what this is for - from Dagged
+        Furniture, // According to Daghex
         UselessItems2,
         ReligiousItems,
         Maps,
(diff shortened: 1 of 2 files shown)


Interkarma:
I've known about these for a while and I'm of two minds about this. While it might seem nice to round out the enums, those item indices:

* Are never used in the game.
* Do not have valid texture properties.
* Have no meaningful purpose.

I think "error" is a pretty accurate term for them haha. Maybe a compromise could be "FurnitureNotImplemented" or "FurnitureDoNotUse". :)

Allofich:
Okay, well, the comment in the source code made it sound like you didn't know what they were, so I added them in for documentation. I won't mind if you change the name or something.

Interkarma:
I apologise, I did not mean to cause offence if any was taken :)

The original author of that file is Lypyl, most of the notes in there are his. At least with your changes, Daggerfall Unity should be able to instantiate a blank item of that specific group/subgroup. That's better handling than Daggerfall itself which just errors out (and I believe why Lypyl tagged it that way).

I'll just leave it as is. Someone might actually build a furniture mod on top of this someday.

Allofich:
No offense taken!

PR #355, opened 2017-06-25
Fix damage values

Allofich:
Here are some fixes to damage values from the player's weapon hits.

It turns out there were a number of inaccurate things in the information Daggerfall gives about damage values, which were throwing me off until I figured them out. One is that the diagonal slash down to the right is actually a weaker attack than the horizontal attacks. The other diagonal attack is stronger, and the manual groups them together as both being stronger.

Another was that the strength damage modifier is different from what classic shows in the character sheet. It's actually (STR - 50) / 5, roughly twice the value that classic shows, in either negative or positive direction.

And then another was that the damage display for weapons with material modifiers (those other than steel and silver) is also incorrect! The material modifier to damage is half of what is shown, in both positive and negative direction.

This is backed up by a whole lot of testing and I'm pretty sure about these.

### DIFF
--- a/Assets/Scripts/Game/Formulas/FormulaHelper.cs
+++ b/Assets/Scripts/Game/Formulas/FormulaHelper.cs
@@ -25,7 +25,7 @@ namespace DaggerfallWorkshop.Game.Formulas
 
         public static int DamageModifier(int strength)
         {
-            return (int)Mathf.Floor((float)strength / 10f) - 5;
+            return (int)Mathf.Floor((float)(strength - 50) / 5f);
         }
 
         public static int MaxEncumbrance(int strength)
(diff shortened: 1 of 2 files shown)


Interkarma:
I appreciate all your work on this, thank you :)

Allofich:
No problem. It's interesting finding this stuff that doesn't work as described. I'm going to add the bonus / penalty against monster type next, but it seems that in classic these don't work as character advantages and disadvantages in the original game, making those useless. But they do work, I think, when used as item enchantments. Need to do a bit more confirming.

Also, now that I figured out the strength modifier, I think I understand the way the endurance modifier works, which seemed to be adding +1 to the negative modifiers. Maybe all the modifiers are like this, but it seems they are based around 50 as the central point. The strength modifier goes by steps of 5, so you get +1 from strength levels of 55-59, and -1 from 45 down to 41, etc., while the endurance modifier goes by steps of 10, so you get +1 from endurance levels of 60-69, and -1 from 40 down to 31. This is different from how the attribute popups on the character sheet calculate it (40-49 is -1 modifier, 50-59 is 0 modifier, 60-69 is +1 modifier), but it does make sense in its own way, and fits with how the manual says a score of 50 is average. Based on the way the popups work, 55 would instead be "average." I think I will restore endurance behavior to classic rather than fixing what I saw as a bug, unless you object. Probably all the modifiers will have this same issue.

PR #356, opened 2017-06-26
Add more damage modifiers

Allofich:
More work on damage modifiers. The damage modifier stuff should be almost done.


Interkarma:
This is awesome, you're smashing it! I'll merge now because it looks good from what I can tell. Will test more thoroughly over next day or so. Just send me another PR if you spot any problems. Cheers.

PR #357, opened 2017-06-26
Action component for flat->3d object

TheLacus:
I moved my code in AddFlat() and hid the quest part under an if, so action and sound are applied in the same way for vanilla and custom. Let me know if you have other preferences.

### DIFF
--- a/Assets/Scripts/Utility/AssetInjection/MeshReplacement.cs
+++ b/Assets/Scripts/Utility/AssetInjection/MeshReplacement.cs
@@ -151,30 +151,6 @@ namespace DaggerfallWorkshop.Utility.AssetInjection
 
             return null;
         }
-        
-        /// <summary>
-        /// Checks if dungeon flat should have a torch sound.
-        /// </summary>
-        static public bool HasTorchSound (int archive, int record)
-        {  
-            if (archive == TextureReader.LightsTextureArchive)
-            {
-                switch (record)
-                {
-                    case 0:
-                    case 1:
-                    case 6:
-                    case 16:
-                    case 17:
-                    case 18:
-                    case 19:
-                    case 20:
-                        return true;
-                }
-            }
-
-            return false;
-        }
 
         #endregion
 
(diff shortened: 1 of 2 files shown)


Interkarma:
Hey mate. :)

This one looks like it will break the quest handling of NPC if billboard if replaced. The SpecialNPCClickHandler and StaticNPC components still need to be attached to the replacement GameObject if this is an NPC. These components do not use the Billboard at all and are essential for Quest system to operate on those objects. It's OK to replace GameObject with something other than a Billboard - but the StaticNPC and SpecialNPCClickHandler still need to be assigned.

Here's the general process:

* Use obj.Resources.FlatResource to determine if this is an NPC.
* Inject either standard Billboard or replacement GameObject.
* If this is an NPC, assign StaticNPC (always) and SpecialNPCClickHandler (for special named NPCs). I'll probably get rid of the latter at some point now I'm more comfortable with things. It's a legacy from an earlier direction.

I'm going to merge this one and make the necessary changes quickly so the quest scripts still get attached to replacement GameObject. Can you confirm if replacement GameObject is always given a trigger collider? This is necessary for player clicks on object to be captured.

Cheers!

Interkarma:
OK, reworked now. Change should support either original billboard or replacement, and still add the required NPC quest stuff to either one. As long as replacement GameObject has a nice trigger collider, NPC object should OK in both cases.

Let me know if any problems, I'm happy to try and help how I can. :)

Interkarma:
Cleaned this up a little more. The billboard is now only added if ImportCustomFlatGameobject() returns null. I've completely decoupled NPC check and factionID from billboard.

TheLacus:
Hey, thank you for everything!

> Can you confirm if replacement GameObject is always given a trigger collider? This is necessary for player clicks on object to be captured.

Adding a BoxCollider with `isTrigger = true` will accomplish or there is something more specific to take care?

Interkarma:
That should do it! :)

PR #358, opened 2017-06-27
Add more enemy stats and fix health calculation

Allofich:
This adds the stats for monsters' level, armor value and multiple attacks. I need to look into the attacks a little more but I think a monster with multiple damage values gets multiple damage rolls each time they attack. I think that Daggerfall actually supports (or at least there is room in the character/monster record) for up to 5 attacks, but no monsters have more than 3. I also fixed the health calculation for enemy classes.

### DIFF
--- a/Assets/Scripts/DaggerfallUnityStructs.cs
+++ b/Assets/Scripts/DaggerfallUnityStructs.cs
@@ -188,10 +188,16 @@ namespace DaggerfallWorkshop
         public int SightModifier;                   // +/- range of vision for acute/impaired sight
         public int HearingModifier;                 // +/- range of hearing for acute/impaired hearing
         public MetalTypes MinMetalToHit;            // Minimum metal type required to hit enemy
-        public int MinDamage;                       // Minimum damage per hit
-        public int MaxDamage;                       // Maximum damage per hit
+        public int MinDamage;                       // Minimum damage per first hit of attack
+        public int MaxDamage;                       // Maximum damage per first hit of attack
+        public int MinDamage2;                      // Minimum damage per second hit of attack
+        public int MaxDamage2;                      // Maximum damage per second hit of attack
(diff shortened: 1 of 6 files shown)


Allofich:
I accidentally messed up the player's starting health calculation. Let me fix that before this is merged.

Allofich:
Fixed.

Interkarma:
Thank you for the update 👍

Allofich:
Looks like there's a problem with the order that things are being initialized when loading a game. If you load a DF Unity save where there are enemy classes around you, they will all be instantiated with Level 1 health regardless of the player level. If you load an area by walking through the entrance, they will be correctly instantiated.

Interkarma:
Thanks for letting me know. When loading a game, the enemy is instantiated first then saved properties are deserialized onto the enemy entity. This could mean resetting the enemy type (for random enemies), changing their position, or disabling them in the scene if dead. If the expected properties are visible in SaveData.txt, this is probably a deserialization issue.

A few things to check:
* Confirm save game was created after level and health changes went into place.
* Save a new game and use the Debug Inspector and compare live data to what gets written into SaveData.txt.
* Close and open game, reload saved game, check data loads properly. Is the data in the save game wrong, or did it just not load correctly?
* Determine if this effects static or random enemies (or both). They can have different bugs.

I need to take another pass at the enemy serialization setup. If you feel this is just a save/load bug, I'm happy to fix when I loop back to that ahead of stable 0.4 release (hopefully in a few months).

I'm not too concerned if a few things are broken right now, I wasn't even expecting all of the work you're doing on combat formulas until possibly 0.5 or 0.6 cycles. So it's all great progress. :)

Allofich:
Just to let you know, I was going to finally go back and take a look at this problem with enemies having low health when a saved game is loaded, but I can't seem to reproduce the issue anymore, so it seems to be fixed.

PR #359, opened 2017-06-28
Implement hit/miss calculations for player

Allofich:
These are hit and miss calculations for the player. It's reverse-engineered from the executable so I think it's correct. There's no easy way to test it since it uses probabilities, but the results seem to match classic well.


Interkarma:
This looks good, very excited to see all this coming together! :)

PR #363, opened 2017-07-01
Identify building data in classic saves

Allofich:
I figured out the data following the save tree header. Also, with this PR a classic save I had that didn't load correctly now does. Treating the "environment" variable as the number of LocationDetail records would cause it to move the reader past the character position record (record type 0x04), and the save would fail to load.


Interkarma:
An interesting find! Were you looking for the tavern room times specifically?

Allofich:
Kind of. The tavern room times helped tip me off to what was going on.

PR #364, opened 2017-07-01
Minor fail-safe to video loading / playing

murderbeard:
This prevents a stream of errors in case someone didn't set up their daggerfall directory properly and is missing video files.

If the vidFile is unable to Load endOfFileReached is set to true so the video player just skips it and won't stay on a black screen forever.


Interkarma:
Hi, welcome and thank you for the contribution! 👍

murderbeard:
Hey @Interkarma , hope to be of some help to the project. Love the work so far!

PR #365, opened 2017-07-02
Replaced any remaining non generic FindObjectOfType calls with generic version

murderbeard:
There were a couple of stragglers that used the old style of FindObjectOfType. These were not trying to get an interface for example so the non generic version wasn't required.

Did a similar search for GetComponent but those were all generic.


Interkarma:
Cheers! :)

PR #366, opened 2017-07-03
Smoothed out camera

murderbeard:
As the PlayerMotor uses FixedUpdate to calculate movement; the camera can feel choppy if your FPS is over 50. Increasing the fixed time step much higher will result in a lot of physics calls that affect FPS in general so that's not a good option. 

This adds the option to smooth out a transform child of PlayerMotor. This child will smoothly lag behind the motor so there is no perceivable stutter. It detects if the player has teleported when the distance travelled is larger than the velocity of motor allows, including a fairly large margin. When teleported the smoothing is reset so you don't see the camera lerping through the entire scene.

A few notes; smoothing reset also happens when the player switches to or from crouching. Currently this change is a hard jump anyway so I don't think this is noticeable. One option is to add this height difference as a minimum distance before it can even be regarded as a teleport.

One thing that might be an issue is that all children of the PlayerMotor are now grouped together under a new GameObject. This means that those children are hidden in the ProjectView as Unity only shows one level of Prefab depth. Trying to group them on runtime Start just adds confusion or issues because the Transform children cannot be found as they're disabled.

### DIFF
--- a/Assets/Prefabs/Player/PlayerAdvanced.prefab
+++ b/Assets/Prefabs/Player/PlayerAdvanced.prefab
@@ -36,7 +36,7 @@ GameObject:
   m_IsActive: 1
 --- !u!1 &100002
 GameObject:
-  m_ObjectHideFlags: 0
+  m_ObjectHideFlags: 1
   m_PrefabParentObject: {fileID: 0}
   m_PrefabInternal: {fileID: 100100000}
   serializedVersion: 5
@@ -58,7 +58,7 @@ GameObject:
   m_IsActive: 1
 --- !u!1 &100004
 GameObject:
-  m_ObjectHideFlags: 0
+  m_ObjectHideFlags: 1
   m_PrefabParentObject: {fileID: 0}
   m_PrefabInternal: {fileID: 100100000}
   serializedVersion: 5
@@ -74,7 +74,7 @@ GameObject:
   m_IsActive: 0
 --- !u!1 &100006
 GameObject:
-  m_ObjectHideFlags: 0
+  m_ObjectHideFlags: 1
   m_PrefabParentObject: {fileID: 0}
   m_PrefabInternal: {fileID: 100100000}
   serializedVersion: 5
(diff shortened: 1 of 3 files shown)


Interkarma:
Thanks for the PR. :) I can see how this works, but I can't say I experience much camera chop at all. My camera feels very smooth to me and I run at a consistent 60fps (with vsync enabled). I'll try without vsync to see if I can reproduce the problem first.

How well does this handle floating origin updates when running through wilderness? That's like a teleport so would cause the smooth follower reset, but it needs to be perceived by the player as a smooth continuous movement. And how well does it deal if player uses set_runspeed 300 and bolts through the wilderness for example?

I'll definitely check this out, but I'm always wary of introducing more mechanical complexity to fix a minor (and in my case imperceptible) problem. :)

Interkarma:
My impression so far: NICE! :)

I had to test this on my higher-end VR PC to really see a major difference. On my mid-range development PC the camera smoothness is about the same, even at 60fps. But there really is a quite noticeable difference on the more specced PC.

I haven't been able to break this so far. It's worked perfectly with floating origin and everything else I can throw at it. Very good work. 👍 

I've merged with a new branch for now. I just want to put it through a few more paces. If you have any more changes on this please throw it to the 'smooth-camera' branch.

Once I'm done, I'll merge back to master. Thanks again. :)

murderbeard:
Hey. Great to hear it's working on your end as well. 

I've noticed that it does register the floating origin changes as a teleport so it can produce a minor camera stutter when the terrain gets updated. 

It's not immediately clear how to account for this in a simple way. One thing worth trying is to add a function like PlayerMotor.Offset(Vector3 offset). Which gets called from FloatingOrigin.OffsetPlayerController; in which it could account for teleporting in sections with a floating origin. Or maybe adding a Teleport function to the motor but that would maybe make it more obtuse about where a Teleport should be applied if not by manipulating the gameobject/transform directly.

Will be trying that out; and throwing any results into the 'smooth-camera' branch! :)

Interkarma:
From my experiences so far, I'm not too concerned about the minor camera stutter on floating origin updates. There's a lot of other stuff going on as well (besides shifting the world sideways) like streaming in new locations and terrain data. That can produce a bit of a stutter too, and anything coming from the smooth follower is practically invisible to me.

I've left that branch open in case you have any more refinements to send through. I've already merged the other changes back to master.

PR #369, opened 2017-07-08
Add player armor calculations and enemy combat sounds

Allofich:
Added player armor to the hit calculations, and miss sounds and collapse sound on death for enemies.

I wanted to add sounds for when enemies hit the player (two types, when the attack is done without a weapon and with), but I wasn't sure how that should be handled as far as what audio source to use.

### DIFF
--- a/Assets/Scripts/Game/Formulas/FormulaHelper.cs
+++ b/Assets/Scripts/Game/Formulas/FormulaHelper.cs
@@ -380,7 +380,7 @@ namespace DaggerfallWorkshop.Game.Formulas
             // The in-game display in Daggerfall of weapon damages with material modifiers is incorrect. The material modifier is half of what the display suggests.
             if (weapon != null)
             {
-                damageModifiers += weapon.GetMaterialModifier();
+                damageModifiers += weapon.GetWeaponMaterialModifier();
             }
 
             // Check for a successful hit.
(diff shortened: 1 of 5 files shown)


Interkarma:
Looks like good work! Will check properly and merge when I can. Cheers :)

Allofich:
When I have the time and the drive, I still need to add in the armor value from shields. Not sure when that will be. You can go ahead and merge this PR when you want, or if you haven't done so by the time I add shields, I'll put them in this PR.

Assets/Scripts/Game/Formulas/FormulaHelper.cs
```diff
@@ -421,25 +421,44 @@ public static bool CalculateSuccessfulHit(Entity.DaggerfallEntity attacker, Enti
...
-            int[] randomMods = { 0, 1, 1, 1, 1, 2, 2, 2, 3, 3, 3, 3, 4, 4, 4, 4, 5, 5, 5, 6 };
-            int randomMod = randomMods[(UnityEngine.Random.Range(0, 19 + 1))];
+            int armorValue = 0;
 
-            chanceToHit += randomMod;
+            if (target == player)
+            {
+                // Choose struck body part
+                int[] bodyParts = { 0, 0, 1, 1, 1, 2, 2, 2, 3, 3, 3, 3, 4, 4, 4, 4, 5, 5, 5, 6 };
+                int struckBodyPart = bodyParts[(UnityEngine.Random.Range(0, 19 + 1))];
```
Interkarma: This is perfectly OK for now, but consider using bodyParts.Length instead of 19+1 for upper value of Range(). This ensures the correct range will still be used if the array ever changes size.

Interkarma:
Excellent work. It's great to see those armour calculations coming together.

Thank you again for all your efforts on this. :)

Assets/Scripts/Game/Formulas/FormulaHelper.cs
Allofich: I'll fix that when I add shields or otherwise modify this function again.

PR #370, opened 2017-07-10
Identify more classic save data

Allofich:


Interkarma:
Cool! You're smashing through that unknown save data. 👍

Allofich:
Thanks! There are a number of other pieces of data, where I think I know what it is or know what its involved with, at least. I'll try to keep adding in pieces when I can.

PR #372, opened 2017-07-13
Action component for flat->3d object /2

TheLacus:
Sorry for the long delay :P

Add a trigger to imported gameobjects for NPCs. Flats with an action are already given a trigger on [AddAction()](https://github.com/Interkarma/daggerfall-unity/blob/master/Assets/Scripts/Utility/RDBLayout.cs#L853) so custom flats should be already covered for that.


Interkarma:
Cheers! And no worry about the delay, we're not saving lives here or anything. :)

PR #373, opened 2017-07-15
More work on combat calculations

Allofich:
Implement armor values for player shields.
Use correct value for monster level and relabel old one as number of
hit point die rolls for monsters.
Read armor values from classic as signed bytes.

### DIFF
--- a/Assets/Scripts/API/ClassFile.cs
+++ b/Assets/Scripts/API/ClassFile.cs
@@ -170,7 +170,7 @@ namespace DaggerfallConnect.Arena2
             cfg.Unknown2 = reader.ReadBytes(8);
 
             // Hit points per level
-            cfg.HitPointsPerLevelOrMonsterLevel = reader.ReadUInt16();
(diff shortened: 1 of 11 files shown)


Interkarma:
Dude, that's a ton of work. Looks good on first pass, but I'm a bit fried right now. Will be sure to look at it properly and merge in next in next 24.

Thank you for all of that. :)

Allofich:
Go ahead and take your time. There's nothing too complicated in there. The one thing that seems like it could maybe cause a problem would be the playerEntity.reset() I added when loading a Unity save, since otherwise the armor values from before load would persist. I didn't notice any problem with it, though.

I made add a little more to this PR, but it won't be anything complicated. It would also work as a separate PR if I don't get to it before you merge this.

Interkarma:
You know me too well, that entity.Reset() was one of the things I wanted to look at more closely. :) It *should* be OK, but I wanted to look closer when I was fresh.

PR #374, opened 2017-07-15
Match monster encounter lists to classic

Allofich:
The lists were mostly correct but lacked duplicate entries (except for one duplicate Mage entry). All lists actually have 20 entries, and duplicates fill this out.


Interkarma:
This is excellent, well done!. 👍

PR #375, opened 2017-07-15
Identify more classic data

Allofich:


Assets/Scripts/API/BlocksFile.cs
```diff
@@ -584,8 +584,7 @@ private void ReadRmbFldHeader(BinaryReader reader, int block)
                 blocks[block].DFBlock.RmbBlock.FldHeader.BuildingDataList[i].Unknown3 = reader.ReadUInt32();
                 blocks[block].DFBlock.RmbBlock.FldHeader.BuildingDataList[i].Unknown4 = reader.ReadUInt32();
                 blocks[block].DFBlock.RmbBlock.FldHeader.BuildingDataList[i].FactionId = reader.ReadUInt16();
-                blocks[block].DFBlock.RmbBlock.FldHeader.BuildingDataList[i].Sector = reader.ReadInt16();
```
Interkarma: Please do not change this. The LocationID is actually used by quest system - the "sector" value is useless to me.

Interkarma: Just to expand - I can see they belong together, but I like having them separated this way. The "sector" value, or those 16 bits of the ID are highly unknown at this time. I had to work around this while building the quest system, and now I'd rather not pollute the known good locationID value with unknown data. I'd only have to shift out to the parts I needed anyway. I'm happy for this to remain separated for legacy purposes. :)

Allofich: Ok. The player/ship data stored in the save file seems to store these together as a single value, but we can probably work around that.

Interkarma: Thank you. :) I've ended up using a different building ID setup in Daggerfall Unity than Daggerfall uses, and I'd like to keep the two systems distinct (rather than have two things called buildingID that don't equate to each other).

If the value is encountered elsewhere, just read as 2x Int16 values like now (Sector/LocationID). Cheers mate.

Interkarma:
Awesome! Thank you again for all your work on this 👍

PR #376, opened 2017-07-16
Initialize new character armor values to 100

Allofich:
100 is the player character armor value for no armor equipped.


Interkarma:
Thank you for that.

PR #377, opened 2017-07-17
Initialize player armor values to 100 when loading

Allofich:
Directly and only resets the armor values to their starting value of 100 when loading DF Unity saves, since the previous attempt of calling PlayerEntity.reset() also reset faction data, which wasn't wanted.

Another approach would be to save the armor values in DF Unity saves, and then we could just copy without worrying about resetting the values or calculating the armor values when loading a game. Loading from classic saves is already working this way, since classic saves have the armor values in them.


Interkarma:
Sounds like a good case for including with saves! I'll merge this PR for now and setup the armorValues array to save with the player data when I can.

PR #378, opened 2017-07-17
Identify more classic item data

Allofich:
Identifying more of the item record data led me (kind of forced me) to remove some unneeded complexity from potion recipe code. I also was able to implement importing stack count for arrows.


Interkarma:
Looks good, thank you mate! :)

PR #381, opened 2017-07-24
Identify more classic save data

Allofich:
Details:

1) Corrected some magic effect enums (just comments in the CharacterRecord right now)
2) Verified UnknownTownLink records are doors
3) Identified Creature2 as record for dead mobiles
4) Renamed some fields and records to match what the Fixsave utility calls them. Fixsave seems to call recordID "mapID" or "map identifier" but I didn't change to that because at the moment recordID seems like a better name to me. Also it calls "container" an "item holder" but I didn't change to that because container was already being referenced and had a file named after it.
5) A little bit of other renaming of as-yet unused records to try to make the descriptions more accurate.

### DIFF
--- a/Assets/Scripts/API/Save/CharacterRecord.cs
+++ b/Assets/Scripts/API/Save/CharacterRecord.cs
@@ -282,8 +282,8 @@ namespace DaggerfallConnect.Save
             public Byte level;
             public PlayerReflexes reflexes;
             public UInt32 physicalGold;
-            public Byte magicEffects1; // x1 = paralyzed, x4 = invisible, x8 = levitating, x10 = chameleon, x20 = shade, x40 = open
-            public Byte magicEffects2; // x1 = silenced
+            public Byte magicEffects1; // x1 = paralyzed, x4 = invisible, x8 = levitating, x20 = lock, x40 = open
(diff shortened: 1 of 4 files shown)


Interkarma:
This great, thank you.

For the UnknownTownLink/Door/DoorStop (as Tipton calls it in some places), it's definitely the header of some building data, and seems to link with more details on interior capabilities - such as available NPCs, quest markers, etc. It would be awesome to know how this data relates to the Sector/BuildingID as well.

It's fairly academic by this point, though. Because this data wasn't very well understood, I had to develop my own ways of enumerating buildings and collecting their markers, etc. This is working quite well in the quest system currently and I'm unlikely to overhaul it all now even if/when this data is well understood.

Still, it's awesome to see your work proceed on the classic data so nicely. This is valuable to everyone. 👍

Allofich:
Yeah, I don't expect you to tear everything down just to copy classic, don't worry. :) Figuring out the classic data will be needing for importing as much as possible from classic saves, and it helps for figuring out how Daggerfall does things. It also just feels good to put together the pieces of the puzzle! Hopefully I can keep making progress.

Interkarma:
I really respect that. :)

PR #382, opened 2017-07-24
Change lockpicking to be closer to classic

Allofich:
I found out a little more how lockpicking works in classic.
It turns out that when the player fails to pick a lock (not counting magic locks), the door stores the player's lockpicking skill, and subsequent attempts fail if the player's lockpicking skill matches that value. I thought you couldn't try again at all if you failed once (unless you left and re-entered the area), but you can try again if your lockpicking skill rises.

### DIFF
--- a/Assets/Scripts/Game/PlayerActivate.cs
+++ b/Assets/Scripts/Game/PlayerActivate.cs
@@ -153,12 +153,9 @@ namespace DaggerfallWorkshop.Game
                         DaggerfallActionDoor actionDoor;
                         if (ActionDoorCheck(hits[i], out actionDoor))
                         {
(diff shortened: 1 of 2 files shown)


Interkarma:
Good update! Would you like me to add LockPickingAttempted and FailedSkillLevel  to the data saved/load by SerializableActionDoor?

Allofich:
> Would you like me to add LockPickingAttempted and FailedSkillLevel to the data saved/load by SerializableActionDoor?

For FailedSkillLevel, go ahead. LockPickingAttempted is removed now, so no need for that.

Interkarma:
Whoops, of course. Not the biggest mistake I've made today haha. :)

I've added this short value to ActionDoorData_v1 package now.

Allofich:
Good to hear!

PR #385, opened 2017-07-25
Add day time logic for imported models

TheLacus:
If you want a test model to try, [this is a town lamp](https://drive.google.com/open?id=0B_JvVNlSG3cRWHpzZ0txaUF6NDA).


Interkarma:
Good update, thank you! :)

PR #386, opened 2017-07-25
Implement pickpocketing

Allofich:
The basics of pickpocketing are implemented here. More work needs to be done on it, but first of all I want to be sure you're OK with the ray tracing checks. It's using a mix of a "motor" for mobile NPCs and a "DaggerfallEntityBehaviour" for mobile enemies, but it works.

The most notable problem at the moment is that many of the messages when you don't find any gold on the target are too long and run off the screen. Looks like the text formatting still needs some work.

Also shortened the lockpicking formulas.

### DIFF
--- a/Assets/Scripts/Game/Formulas/FormulaHelper.cs
+++ b/Assets/Scripts/Game/Formulas/FormulaHelper.cs
@@ -120,23 +120,15 @@ namespace DaggerfallWorkshop.Game.Formulas
         // Calculate chance of successfully lockpicking a door in an interior (an animating door). If this is higher than a random number between 0 and 100 (inclusive), the lockpicking succeeds.
         public static int CalculateInteriorLockpickingChance(int level, int lockvalue, int lockpickingSkill)
         {
-            int lockpickingChance = (5 * (level - lockvalue) + lockpickingSkill);
-            if (lockpickingChance > 95)
-                lockpickingChance = 95;
-            else if (lockpickingChance < 5)
-                lockpickingChance = 5;
-            return lockpickingChance;
+            int chance = (5 * (level - lockvalue) + lockpickingSkill);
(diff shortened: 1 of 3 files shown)


Allofich:
I made a few changes. In the latest Daggerfall you can pickpocket creatures as well as enemy classes. According to PATCHED.TXT, in version 1.07.212 "Pickpocketing monsters has been fixed. It is more consistent now." The change seems to be that you can pickpocket creatures from then on when you could only do so to classes in the earlier versions.

I don't know why they considered that a "fix." It might be kind of funny but it's pretty ridiculous to be able to take not only gold but get messages about finding various things in your victim's "pockets" when you try pickpocketing a rat or something. But I decided for now to set DF Unity to be like the earlier Daggerfall behavior and only allow pickpocketing from enemy classes.

Another minor change is I added a non-plural string for when you only find one gold piece, so the message won't say "1 gold pieces" like in classic.

I usually try not to diverge from classic, but I was thinking of maybe allowing pickpocketing for vampires and vampire ancients, since it might be kind of cool to find stuff on them that gives the feeling that they are walking among regular people. It fits with the lore, too.

Another thing to consider is whether to allow spam pickpocketing enemies. Currently you can do so in this PR, as you also can in classic. To me, the only pickpocketing scenario in the original game that seems "fully implemented" is that of pickpocketing mobile NPCs, and for them, you only get one shot at it, similar to lockpicking doors. Maybe we should do the same thing for enemy mobiles.

Interkarma:
Yay pickpoket! Awesome work mate. :)

I agree it makes sense to store a pickpocket attempted flag on the serializable enemy object. Easy enough to roll this into save data.

Looking forward to watching this one evolve. I'll add it to the roadmap as a completed feature with your name on it.

PR #389, opened 2017-08-03
Wrap long text messages over multiple lines

Allofich:
Fixes long text messages going off screen, which could happen with many of the pickpocketing messages where you find nothing useful.

It doesn't intelligently wrap by word, but by letter, but Daggerfall is the same way and it doesn't really bother me.

### DIFF
--- a/Assets/Scripts/Game/UserInterface/MultiFormatTextLabel.cs
+++ b/Assets/Scripts/Game/UserInterface/MultiFormatTextLabel.cs
@@ -144,6 +144,7 @@ namespace DaggerfallWorkshop.Game.UserInterface
             textLabel.AutoSize = AutoSizeModes.None;
             textLabel.Font = font;
             textLabel.Position = new Vector2(cursorX, cursorY);
+            textLabel.MaxWidth = 288;
             textLabel.Text = text;
             textLabel.Parent = this;
 
--- a/Assets/Scripts/Game/UserInterface/TextLabel.cs
+++ b/Assets/Scripts/Game/UserInterface/TextLabel.cs
@@ -155,7 +155,11 @@ namespace DaggerfallWorkshop.Game.UserInterface
 
             // First pass encodes ASCII and calculates final dimensions
             int width = 0;
+            int greatestWidthFound = 0;
+            int lastEndOfRowByte = 0;
             asciiBytes = Encoding.ASCII.GetBytes(text);
+            List<byte[]> rows = new List<byte[]>();
+
             for (int i = 0; i < asciiBytes.Length; i++)
             {
                 // Invalid ASCII bytes are cast to a space character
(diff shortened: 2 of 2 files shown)


Interkarma:
My only concern with this one is that MaxWidth is always set to 288 by AddTextLabel() so the old behaviour will never be called in the event of overflow. Adding new lines will break what I'd consider the "layout contract" from the text input stream. A few other systems (e.g. quest popups) expect blocks of text to be laid out line-by-line without any wrapping logic. Adding an arbitrary wrap that applies to all multiformat labels might inject new orphan characters where they aren't intended (and perhaps risk flowing off bottom of screen rather than off the side for very large popups).

My preference in this would be the default behaviour to observe the text stream layout so the text author has full control over line breaks (and things flow off screen it's their fault). Then selectively choose which systems (e.g. pickpocket) use the wrapping logic.

The other option is just to change the input stream so that it doesn't need to wrap automatically (as when writing quests).

Please let me know your thoughts. :)

Allofich:
> Please let me know your thoughts. :)

Sure. I made it work this way because I tested another text popup (an attribute description popup on the character sheet) in classic and it also wrapped around, so my assumption is that all text popups in classic do this. Are you sure quest popups don't use any wrapping?

That said, I realized today that implementing wrapping as I did would probably mess up the text in books, which I assume also use text labels. I wasn't able to test books (There seems to be an issue with books in DF Unity. I couldn't open the book in my save)

Anyway, the only text pop ups I'm aware of that run off the screen in DF Unity are these pickpocketing ones, so think now that the simplest and most foolproof solution is to just replace these with hard strings in DF Unity that have line breaks added to them in appropriate spots. I'll change the PR to do it this way.

Allofich:
I may have changed my mind again.  Replacing the text with hard strings wasn't as straightforward as I thought. I'm looking at optionally specifying a maximum text width when making a message box, so the pickpocketing messages could set a width only used for them.

Interkarma:
Yeah, that would work well I think. Then the wrapping behaviour can be applied where needed, and everything else just follows the input format.

I wasn't aware of my problems with books however, this was working very well last time I looked. Maybe a recent problem. I'll take a look when I have time for a coding session.

Allofich:
I hadn't gotten around to changing this PR, but on the plus side, books are working fine when I started a new game in DF Unity. The problem I had earlier was with an imported classic save. Are books from imported saves supposed to work?

I took a look earlier at passing max text width as a parameter for use only with pickpocketing, but I wasn't sure how to do it gracefully. If you don't mind, I think I'd prefer to just move on to other things (working on item spawning stuff currently) since there don't seem to be any obvious functional problems from this PR.

Interkarma:
Books should import OK, but been a while since this was tested - could have been a regression somewhere along the line. Something to look at when I can.

Happy for you to move on. The new text wrapping behaviour doesn't break anything I could find. I put it through as much punishment as I could. If we find any issues down the track, can worry about it then. :)

Interkarma:
A user has found an issue with new text wrapping behaviour.

http://forums.dfworkshop.net/viewtopic.php?f=24&t=602

I also noticed this changes my intended behaviour of the text being cut off MaxWidth, characters now overflow rather than stop.

I'll issue a change soon to fix and we can look at reimplementing the text wrap another way. :)

Interkarma:
Just merged a change that should accommodate text wrapping (always enabled for multi-format labels as this seemed to work fine before) and single line text with cutoff (such as ListBox used in save window).

Let me know if any problems.

PR #390, opened 2017-08-03
Match monster loot lists to classic

Allofich:
Matches monster loot lists to classic. Also a few minor cleanups to the enemyBasics.cs file.

Looks like Bethesda probably decided to cut back on magic item spawns in between the time of the data Chronicles uses and the game's release. Similar to how they disabled magic item repair.

This PR doesn't touch loot from enemy classes.


Interkarma:
This is awesome. Thanks for finding and refining that data.

PR #391, opened 2017-08-04
Make loot generation chance closer to classic

Allofich:
This makes loot generation chance closer to classic.

I had noticed that in DF Unity tons of loot would be found by higher level characters. DF Unity multiplies the chance of finding all item types by the player's level, but classic only does this for the first four categories of items (ingredients). Everything else is just the chance seen in the loot matrices regardless of player level.

Also, classic's compare is (random value from 0 to 99) <= chance, so even a 0% chance category can create an item. This also means the values in the matrices are effectively 1 higher. My guess is that this is probably unintended, so I didn't recreate it.

### DIFF
--- a/Assets/Scripts/Game/Items/LootTables.cs
+++ b/Assets/Scripts/Game/Items/LootTables.cs
@@ -98,7 +98,7 @@ namespace DaggerfallWorkshop.Game.Items
             List<DaggerfallUnityItem> items = new List<DaggerfallUnityItem>();
 
             // Random gold
-            int goldCount = Random.Range(matrix.MinGold, matrix.MaxGold) * playerEntity.Level;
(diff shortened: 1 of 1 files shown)


Nystul-the-Magician:
just out of interest - how do you find out all this stuff?
i mean there is no source code or anything that helps you getting that info ;)

Allofich:
Fixed some of the ranges. Maximum gold was never given and the loot chances would never succeed for a chance of 1 %.

> just out of interest - how do you find out all this stuff?
>i mean there is no source code or anything that helps you getting that info ;)
I don't have the source code, but I have a "decompiled" version of FALL.EXE, which I have been studying, identifying values and functions based on known data, in-game behavior, looking at FALL.EXE and save game files with a hex editor, etc.

Nystul-the-Magician:
wow you do amazing work!

PR #392, opened 2017-08-07
Implement enemy equipment

Allofich:
This PR implements enemy equipment. Enemy classes and some monsters receive it, and the equipment can be found on their bodies when they die. Enemies with weapons use it for their attack damage now. I haven't yet implemented the effect of the armor on enemies' armor values. Also, right now the materials are using the random material methods in ItemBuilder, but I haven't looked into how classic chooses material.

Also, enemy classes now use the same loot table keys as classic. Their equipped items are added on top of this.


Interkarma:
Looks good! Will get this merged today when I can.

PR #393, opened 2017-08-09
Update enemy equipment calculations

Allofich:
Updates to enemy equipment.

This mainly deals with armor values from equipment. I diverged somewhat from classic behavior for this. I'm trying to keep that to a minimum, but the classic behavior just seemed incorrect to me. In classic both monsters with equipment and enemy classes get their armor values from their equipment, overriding the armor values from the enemy definitions. This generally makes the monsters that use equipment (orc variants and centaurs) have worse armor than any other monsters, making them very easy to hit. The armor values they have in the enemy definitions work well, so in this PR they use whichever is better.

I did something similar for weapons. Here there's not as much of a problem with classic's behavior, but orc warlords in particular tend to do significantly less damage than they have in their enemy definition, because their weapon stats override the damage values. It seemed to me to be best if weapons are also only used if they are better than the enemy definition values.


Interkarma:
I agree with your conclusions, and totally happy with the outcome. Great work as usual. 👍

PR #394, opened 2017-08-11
Add DoorText action

Allofich:
This partially implements Action 99, which displays text at the top of the screen when clicking on some doors in info mode. Cases I know of that exist:
1) Doors in Castle Daggerfall off of throne room.
2) Doors in Castle Wayrest off of entrance area and in basement area before dungeon.
3) Doors in back of Orsinium throne room.

There's a few other texts in the used area of TEXT.RSC that might be unused or there might be a few other doors somewhere with this action.

There's an issue with this that I wanted your help with Interkarma. This action displays text at the top of the screen, but the text used is multi-line text. Is there currently a way to display this? Right now the text just runs off the screen.

This action can also cause castle guards to go hostile but this isn't implemented yet.

I found a few places where classic calls the wrong ID, creating an error message. I put in patches for these doors. Also, for fun, I recreated the Daggerfall error message if a wrong ID is called. :) Although I already patched the only wrong ID calls I know of. If you want me to remove the Daggerfall-style message and do a normal exception message I can do so.

Also, incidentally, do you know why the textIDs here are numbered as they are? From 7701 to 7705 are the first IDs, but then it suddenly jumps to 7717 for the orsinium one, and the next one is ID 7764.

Also, I display the popup text for 2 seconds. I measured the time in classic and it was 2 seconds or very close to it.

### DIFF
--- a/Assets/Scripts/API/DFBlock.cs
+++ b/Assets/Scripts/API/DFBlock.cs
@@ -714,8 +714,9 @@ namespace DaggerfallConnect
             ///<summary>50 Unknown </summary>
             Unknown50 = 0x32,
 
-            ///<summary>99 Unknown - seems to be releated to enemy hostility in special blocks</summary>
-            Unknown99 = 0x63,
+            ///<summary>99 Displays text at the top of the screen when clicked in info mode.
+            ///            Can cause castle guards to go hostile if clicked outside of info mode.</summary>
+            DoorText = 0x63,
 
             ///<summary>100 Unknown, only on 2 objects</summary>
             Unknown100 = 0x64,
--- a/Assets/Scripts/Game/PlayerActivate.cs
+++ b/Assets/Scripts/Game/PlayerActivate.cs
@@ -40,6 +40,11 @@ namespace DaggerfallWorkshop.Game
 
         public float RayDistance = 2.4f;        // Distance of ray check, tune this to your scale and preference
 
+        public PlayerActivateModes CurrentMode
+        {
+            get { return currentMode; }
+        }
+
         void Start()
         {
             playerGPS = GetComponent<PlayerGPS>();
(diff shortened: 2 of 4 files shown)


Interkarma:
> There's an issue with this that I wanted your help with Interkarma. This action displays text at the top of the screen, but the text used is multi-line text. Is there currently a way to display this? Right now the text just runs off the screen.

Happy to help. I'll take a look and get a solution in for you this weekend.

Interkarma:
I've added a new DaggerfallUI.AddHUDText() overload that accepts a token array. You can just grab the tokens for text ID with DaggerfallUnity.Instance.TextProvider.GetRSCTokens() then pass them into AddHUDText().
This is fairly simple as I don't believe scrolling popup text actually uses the justification tokens and is always centred. But it will create one new line per full text token and add them all to the popup at once.
This seems to handle text like 7701, etc. fairly well.
Let me know if any feedback or problems.

Allofich:
Thanks. I'll add that in to this PR soon.

Allofich:
Updated. Some additional changes:

1) The replacement text I was using for the Orsinium throne room lacks a justify center, so it's all one long line. It's the only one of the door texts that does this, so rather than implementing text wrapping for just this one message, I patched over it with two replacement strings.

Is this OK? I just thought now, any Daggerfall translations out there might depend on using TEXT.RSC (even though this is text that doesn't normally appear). But we would have to handle all the strings in the .exe to support translations anyway, right? So maybe it doesn't really matter.

2) Changed door triggered actions to trigger for player. Otherwise the door text would appear when enemies opened doors. Also this makes it so unique sound effects only play when the player is the one to activate a door. I don't know if we want that or not but it's the same behavior as classic.

Interkarma:
Yep, that's all OK. Good work. :)

PR #396, opened 2017-08-13
Flesh out armor and weapon generation closer to classic

Allofich:
Currently in master, DF Unity is far more generous giving out weapons and armors of high materials than classic is. This PR replicates what I've found of the armor generation function in classic.

It uses random numbers so it's hard to say for sure, but it seems like classic is a little more generous with materials than the results from this PR. I don't think I'm missing anything from the generation function, so maybe it's a difference in the random number generators of classic and Unity, or maybe there's another factor somewhere that affects material generation.

Also, this PR sets weight, gold value, condition and enchantment points by material type. Variants aren't set yet and are always variant 0. as they are in master.

### DIFF
--- a/Assets/Scripts/Game/Items/ItemBuilder.cs
+++ b/Assets/Scripts/Game/Items/ItemBuilder.cs
@@ -27,35 +27,22 @@ namespace DaggerfallWorkshop.Game.Items
         const int firstFemaleArchive = 245;
         const int firstMaleArchive = 249;
 
-        static ArmorMaterialTypes[] armorMaterialList = new ArmorMaterialTypes[]
-        {
-            ArmorMaterialTypes.Leather,
-            ArmorMaterialTypes.Chain,
-            ArmorMaterialTypes.Iron,
-            ArmorMaterialTypes.Steel,
-            ArmorMaterialTypes.Silver,
-            ArmorMaterialTypes.Elven,
-            ArmorMaterialTypes.Dwarven,
-            ArmorMaterialTypes.Mithril,
-            ArmorMaterialTypes.Adamantium,
-            ArmorMaterialTypes.Ebony,
-            ArmorMaterialTypes.Orcish,
-            ArmorMaterialTypes.Daedric,
-        };
+        // This array is used to pick random material values.
+        // The array is traversed, subtracting each value from a sum until the sum is less than the next value.
+        // Steel through Daedric, or Iron if sum is less than the first value.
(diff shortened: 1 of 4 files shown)


Allofich:
Updated to include weapons.

I think I could clean this up by merging weapon and armor metal types into one, which I think is how classic does it. "Leather", "Chain" and "Plate" are their own separate thing. But I don't want the PR to be too big and I may have to figure out any problems with importing from classic saves so for now I'm working around the existing enums.

Assets/Scripts/Game/Items/ItemBuilder.cs
```diff
@@ -357,27 +354,45 @@ public static DaggerfallUnityItem CreateRandomArmor(int playerLevel, Genders gen
             int groupIndex = UnityEngine.Random.Range(0, enumArray.Length);
             DaggerfallUnityItem newItem = new DaggerfallUnityItem(ItemGroups.Armor, groupIndex);
 
-            // Adjust for gender
-            if (gender == Genders.Female)
```
Interkarma: I notice this has been removed and PlayerTextureArchive is no longer being set by race and gender any more. Can you let me know where this is being done now? Cheers.

Interkarma:
Happy to see some improvements to loot generation over my temp solution - which was definitely too generous in terms of quantity and materials. I'm grateful to see you working your magic on these.

My only real concern is the PlayerTextureArchive assigned based on race and gender appears to have been removed. I'm assuming this is now a redundant call and is being done elsewhere, but haven't had time to follow results in details across multiple races. Are you able to let me know where the PlayerTextureArchive is being assigned for newItem now?

Allofich:
My mistake. I originally did this PR by having CreateRandomArmor() work by just picking a piece of armor and then calling CreateArmor(), and I forgot to put the race and gender parts back in when I changed it back.

Allofich:
Added spawning of all armor variants, except for helms, which I haven't figured out yet.

Allofich:
I haven't been able to figure out exactly how classic handles helms yet, so I just enabled the different varieties and used the FixLeatherHelm function. If I figure it out I'll fix it later. I'm fine for this to be merged, btw.

Nystul-the-Magician:
I never understood why there are no metal variants on chain mail - there should be in my opinion - maybe classic df is not the way to go here?

Allofich:
Do you mean as part of baseline Daggerfall Unity?  I think a change like that would be better as a mod.

Nystul-the-Magician:
I am not sure, have no idea how difficult it would be to implement this as a mod - but I would fear that those systems are so low-level that it would be hard to mod this ;)

PR #397, opened 2017-08-14
travelmap location discovery implemented

Nystul-the-Magician:
Hi Interkarma!
I added static function checkIfLocationTypeAlwaysKnown to PlayerGPS - reason is because I added a check to location discovery function DiscoverCurrentLocation which checks if the location type is one of those that should always be known to the player like town, village, temple, common graveyard - I did add the check so that discovery dictionary (and savegame data) remains as small as possible - imagine a player that travels a lot will have lots of locations in there that could be saved
BUT: if you don't like it JUST remove the check (line 604, 605) - then the function checkIfLocationTypeAlwaysKnown could be moved over to DaggerfallTravelMapWindow.cs...

added 2 console commands (in DaggerfallTravelMapWindow.cs): map_reveallocations and map_hidelocations (it is only a temporal reveal - discovery state is of course not changed)

best regards!

### DIFF
--- a/Assets/Scripts/Game/Questing/Actions/RevealLocation.cs
+++ b/Assets/Scripts/Game/Questing/Actions/RevealLocation.cs
@@ -15,13 +15,15 @@ using FullSerializer;
 namespace DaggerfallWorkshop.Game.Questing
 {
     /// <summary>
-    /// Incomplete. Just stubbing out action for now so quest will compile.
+    /// Reveals location on travelmap
     /// </summary>
     public class RevealLocation : ActionTemplate
     {
+        Symbol placeSymbol;
+
         public override string Pattern
         {
-            get { return @"reveal"; }
+            get { return @"reveal (?<aPlace>\w+)"; }
         }
 
         public RevealLocation(Quest parentQuest)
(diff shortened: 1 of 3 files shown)


Nystul-the-Magician:
Undiscovered locations can be searched for right now, this is unintended and I will fix this soon...

Nystul-the-Magician:
@reveal action should work now for dungeons in current region.
note: reveal aPlace in province aProvince# at magicNumber does currently not work - this will be next (might need some help with this)

Interkarma:
I recall that it was common graveyards that are always known (and forgotten are the ones that must be discovered). I could have named them the wrong way around in enum though. Could you confirm for me? Thanks mate.

Nystul-the-Magician:
I checked back in vanilla df, since I thought the same - but vanilla always reveals both

Allofich:
I just checked, and I see cemeteries on the map in DF Unity that I don't see in the original game. Can you check again? I'm looking at Daggerfall province.

Nystul-the-Magician:
yeah will check again ;)

Nystul-the-Magician:
well - this is mysterious - it seems like classic shows some of the forgotten graveyards by default, but not all.
to test: make fresh game or go to a region you are sure you did not discover any dungeons, e.g. i looked up dungeons on Isle of Balfiera (travelmap filter for dungeons), compare to dfunity one time with line "locationType == DFRegion.LocationTypes.GraveyardForgotten ||" active and one time inactive in function PlayerGPS.checkIfLocationTypeAlwaysKnown()

btw. quest CUSTOM01 does not correctly return place resource (null is returned from function ParentQuest.GetPlace(placeSymbol) in RevealLocation.Update()), not sure why this happens here.

although lines in CUSTOM01.txt don't seem to match (dungeon is not matching graveyard):
Place graveyard remote ancientgraveyard
reveal dungeon

but even if I change it to reveal graveyard it does not work - ParentQuest.GetPlace(placeSymbol); still returns null

all other quests correctly work with (they all seem to use this format "Place mondung remote dungeon" - "Place graveyard remote ancientgraveyard" seem to be use a remote type currently not known (ancientgraveyard)

PR #398, opened 2017-08-17
improvements to discovery implementation of travelmap

Nystul-the-Magician:
added console command map_reveallocation [regionName] [locationName]
reverted behavior: forgotten graveyards are not known automatically (more investigations are needed of classic daggerfall behavior with graveyard - seems like common graveyards are always known, some forgotten graveyards are always known and some forgotten are not...)


Interkarma:
Awesome, thank you for doing this. 👍

PR #401, opened 2017-08-23
Approximate classic player movement speeds

Allofich:
This PR approximates classic player movement speeds.

First I recorded the length of time it took to walk the same distance with the same speed stat in classic and Unity. Comparing the movement values used in FALL.EXE for to those of DF Unity I found the ratio was about 39.5 to 1, meaning a speed value of 39.5 in classic is about the same as a speed value of 1 in DF Unity. This is probably slightly off. For one thing movement is classic seems to be instant stop-and-go but there is acceleration and deceleration in DF Unity.

The walk, crouch and run formulas are matched to classic. The run formula is also in DF Chronicles and matched what I saw in the assembly code.

One little thing I found: the "maxSpeed" value in z.cfg can be used to modify walk speed in Daggerfall version 1.0 but as of the latest Daggerfall version maxSpeed now seems to be hardcoded at 200 and is no longer read from z.cfg. Also I'm not sure why they named it "maxSpeed" because that doesn't seem to be what it does. Maybe it was used in a different way at some point or maybe I'm missing something. Walk speed in classic is calculated as (Player.Speed - 50 + maxSpeed).

Also I modified the get_walkingspeed and get_runningspeed commands to fit with the new code. If you set a value with them it's used as an override. You can return to the classic formula-based speeds by setting a value of -1.

### DIFF
--- a/Assets/Game/Addons/UnityConsole/Console/Scripts/DefaultCommands.cs
+++ b/Assets/Game/Addons/UnityConsole/Console/Scripts/DefaultCommands.cs
@@ -210,7 +210,7 @@ namespace Wenzil.Console
         {
             public static readonly string name = "set_walkspeed";
             public static readonly string error = "Failed to set walk speed - invalid setting or PlayerMotor object not found";
-            public static readonly string description = "Set walk speed. Default is 3";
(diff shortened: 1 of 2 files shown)


Interkarma:
More awesome work, cheers! 👍 

Will get these merged soon.

PR #402, opened 2017-08-24
Correct set_weather usage and add check

TheLacus:


Interkarma:
This makes it clearer, well done.

PR #403, opened 2017-08-25
Implement building opened/closed hours

Allofich:
Implements:

- All buildings except houses should have the same hours for the doors being locked/unlocked. You can get a message for stores' opening and closing hours by clicking info mode, as in classic.
- Houses 1 through 4 are open during the day and closed at night, and you get a greeting when you enter, like in classic.

To be done:
- There's something more to how the houses' open hours are calculated, but I haven't figured it out yet.
- Since exterior doors don't have lock values and can't be picked or bashed yet, you can get in through any exterior door right now by clicking in steal mode.

### DIFF
--- a/Assets/Scripts/Game/PlayerActivate.cs
+++ b/Assets/Scripts/Game/PlayerActivate.cs
@@ -40,6 +40,10 @@ namespace DaggerfallWorkshop.Game
 
         public float RayDistance = 2.4f;        // Distance of ray check, tune this to your scale and preference
 
+        // Opening and closing hours by building type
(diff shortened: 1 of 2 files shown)


Allofich:
@Interkarma There are two small changes I was thinking about adding but I thought that maybe you're intentionally leaving them out. One is that buildings are only identified in classic, and store times are only shown, if you click on them in info mode. The other is that buildings of the "house for sale" type are named "House for sale" and this appears when you click on them in info mode and on the automap. Would you rather these not be in at the moment?

Interkarma:
Thank you for making a start on this. :)

I'd like building discovery to trigger when entering buildings. Happy for opening times to display in Info mode (I hadn't started on this yet) and "house for sale" to show on automap.

PR #404, opened 2017-08-26
Make building discovery closer to classic

Allofich:
Does these changes:

1. Houses on sale now identified as "House for sale" when clicked in info mode, and on automap
2. Clicking on buildings only discovers them (unless door is clicked) and shows shop times if in info mode
3. Buildings are discovered when their doors are clicked, in any mode

This all matches classic behavior as far as I can tell.


Interkarma:
Very good, thank you mate.

PR #406, opened 2017-08-27
travel time countdown implemented for worldmap

Nystul-the-Magician:
this is done via coroutines
had to implement it with a helper class that inherits from MonoBehaviour so that coroutines could be used (since DaggerfallTravelPopUp class is not inheriting from MonoBehaviour)
this works reasonable well

unfortunately there are these limitations:

- location loading and countdown cannot be done in parallel - since unity does not allow different threads with calls to unity api functions (this would be needed since the gui needs to be updated and location loading also calls unity api funcions like gameobject creation etc., current approach with coroutines does  no concurrent processing

- countdown is done before actual fast travel actions right now

- since coroutines are involved now: camera updates happen now that did not before (I guess in between coroutines processing when yielded) - leading to "empty" scene loading in the background when new location is loaded (or in case of of farterrain and realtime reflecions mod some artifacts - camera glitches, and bottom side of reflection plane with green debug color)

- just disabling cameras and other camera update prevention measures did not work or made mods throw exceptions - so I decided to restrain from this for now - will further look for a solution

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTravelMapWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTravelMapWindow.cs
@@ -604,6 +604,12 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
                 if (pixelBuffer == null)
                     pixelBuffer = new Color32[width * height];
 
+                // note by Nystul: check necessary to prevent exception which could happen if pixelBuffer is 320x160 instead of 320x200 -
+                // otherwise marked line below will throw exception (e.g. after fast travel to a location in Wrothgarian Mountains and reopening map)
+                // not sure why this happens, but it happens, maybe Lypyl can take a look
+                if (pixelBuffer.GetLength(0) != width * height)
+                    return;
+
                 for (int y = 0; y < height; y++)
                 {
                     for (int x = 0; x < width; x++)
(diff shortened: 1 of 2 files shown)


Interkarma:
My preference would be not to use coroutines for this task and work within the UI system instead. Both DaggerfallTravelMapWindow and DaggerfallTravelPopUp inherit from UI classes that receive an Update() tick update every frame via the UI framework. It's not necessary for Unity to do the timing here, you can just run the text countdown through UI system then load the destination as normal.

For an example, have a look at how DaggerfallRestWindow implements the rest countdown through Update(), which calls the TickRest() method. You can setup custom timing because the UI will tick even when game time is paused (such as when UI is open). You just need to time with Time.realtimeSinceStartup as Time.gameTime is paused when UI is open.

This will still have the first two limitations, but that's OK. I'm not even 100% sure Daggerfall is parallel loading while countdown runs. But at least the camera issues won't manifest using this method.

I'm not sure of a way to load the location in parallel while countdown runs. This would likely involve a significant redesign of world loading and UI to engineer a mechanism for this. I'd rather just remove the countdown than undertake this. :)

Nystul-the-Magician:
ok i will take a look and change the code accordingly

Interkarma:
Cheers Nystul. Let me know if any help needed. :)

Nystul-the-Magician:
implemented now as suggested by you - much easier now ;) thx for the hints

Interkarma:
That was fast! :)

Thanks for all that, I'll review and merge soon.

Nystul-the-Magician:
last commit is fixing bug in travelmap unrelated to travel time countdown, but i added it to this pull request

PR #408, opened 2017-08-28
TextureReplacement: minor optimization and logs for terrain

TheLacus:
I removed a method used before my recent changes. I noticed it was also called for terrain but a more specific method exists for the purpose, so i swapped them. I took the occasion to add a few error messages for the log.

Nystul may want to review this PR but these should be pretty minor changes; all features, including normal map import, are preserved.


Nystul-the-Magician:
looks good to me ;)

Interkarma:
Great stuff, thanks for this. 👍

PR #409, opened 2017-08-28
Added transport menu and horse riding.

ajrb:
- Transport popup menu added. 
- Options enabled depending on having horse or cart in inventory.
- Horse and cart transport modes change height and speed. (cart and horse are the same to minimise changes to PlayerMotor class)
- Allow jumping since it makes sense that a horse can jump. (should be disabled for cart when PlayerMotor is rewritten)
- Play sounds for movement (clip-clop) and random neighing. (every 2-30secs)
- Transport state is correctly saved & loaded. (not imported from classic saves)
- Automatically dismount when entering buildings or dungeons.
- Fixed palette bug in CfaFile.

### DIFF
--- a/Assets/Prefabs/Player/PlayerAdvanced.prefab
+++ b/Assets/Prefabs/Player/PlayerAdvanced.prefab
@@ -27,6 +27,7 @@ GameObject:
   - component: {fileID: 114374306957635026}
   - component: {fileID: 114683130944510392}
   - component: {fileID: 114121517922970790}
+  - component: {fileID: 114764040474619316}
   m_Layer: 0
   m_Name: PlayerAdvanced
   m_TagString: Player
@@ -1106,6 +1107,18 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   PlayerTorch: {fileID: 100004}
+--- !u!114 &114764040474619316
+MonoBehaviour:
+  m_ObjectHideFlags: 1
+  m_PrefabParentObject: {fileID: 0}
+  m_PrefabInternal: {fileID: 100100000}
+  m_GameObject: {fileID: 100000}
+  m_Enabled: 1
+  m_EditorHideFlags: 0
+  m_Script: {fileID: 11500000, guid: e59a84727916a12479a199612a450dcb, type: 3}
+  m_Name: 
+  m_EditorClassIdentifier: 
+  RidingVolumeScale: 1
 --- !u!198 &198167633220797900
 ParticleSystem:
   m_ObjectHideFlags: 1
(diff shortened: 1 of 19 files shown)


ajrb:
One thing I should note is that I didn't add any support for modding any of this. I will look at adding this later once I see how the functionality is received.

Hopefully you like my code, I tried to fit into existing systems & conventions as much as possible. Please bear in mind I only started looking at the codebase 3 days ago and I'm not familiar with C# but it's pretty easy to pick up new languages after so many years. Been doing mainly Java for 15 years so capitalised method names kinda wierds me out... :-)

Interkarma:
Just merged this and had a quick play. Awesome work! I'll make time for a deeper check and merge to master as soon as I can make time.

Two very minor items so far:

1. Horse might still be a touch high. In classic, I'm able to (barely) ride under the stone arch leading into Castle Daggerfall, but in DFUnity player capsule is enough to block riding under arch. If you're confident horse height is correct, could also work around this by shrinking capsule a bit more (like crouching) during riding state.

2. If player opens UI (e.g. local automap) while still moving forwards, the clip-clop sound will continue playing. This is a truly minor thing, just need to pause the audio source when game is paused (can check from GameManager).

But yeah, this is good stuff all round. Excellent work. 👍

ajrb:
Cheers, glad you like what I've done. Hopefully closer scrutiny wont change your mind.

I thought it was a bit high still but it's still lower than classic (which is ridiculous unless it's some stilt-horse), I've lowered it to 2.8m which is the highest that will fit through the arch. This is 1.8m (large horse) + 1m (sitting rider) so is realistic. However it looks a lot lower than classic. I think the difference is probably that classic doesnt change the collision box height and DFU does. Going through the arch in classic your eyes are almost scraping the ceiling.

I found another issue, have you opened console while on a horse? :-)

Interkarma:
Sounds reasonable on the height. Agree it's probably a collider issue more than anything else, because your height "felt" really right to me as well. I'd be just as happy for the riding collider to be slightly smaller while camera remains at same height. There's likely a happy medium in there somewhere.

Ha, didn't notice the console. All good mate, no rush. And just let me know if any help needed. :)

TheLacus:
> One thing I should note is that I didn't add any support for modding any of this. I will look at adding this later once I see how the functionality is received.

I believe Cfa files can be considered CifRci for this purpose.

ajrb:
> I believe Cfa files can be considered CifRci for this purpose.

What do you mean exactly? In what respect? I'm new to this so you might need to join a few dots for me.

TheLacus:
From what i see in the code, cfa files are defined by filename, record (only one) and a number of frames, similar to cfi and rci.

In practice i'm saing that adding LoadCustomCif() in ImageReader should be ok to import textures as name.cfa_0-frame.png, as opposed to create another import system for these files.

If you don't know what i'm talking about, refer to [this](http://www.dfworkshop.net/projects/daggerfall-unity/modding/textures/).

ajrb:
Thanks, I get what you mean now. Replacements don't need to follow the same format. Makes total sense, thank you for pointing that out.

PR #411, opened 2017-08-29
Fix issues with horse riding

ajrb:
- Capsule height and camera height now set correctly. Height when riding is now 2.6m and camera is always set at 9cm below the top of the capsule. (execpt for when crouching cos that is strange and I didn't want to mess with it)
- Speed is calculated using the base units from DF provided by Allofich (thanks!) and affected by speed skill level. (moved numbers into constants)
- Sounds and animations stop when game is paused.


Interkarma:
Cheers mate! Will merge for now and test when I can later today once I'm free from work.

Interkarma:
Going to merge this branch back to master now. If you have any further fixes/refinements to this, just send straight to master. Good job. 👍

PR #412, opened 2017-08-30
fix for incorrectly shown border on travelmap (world map)

Nystul-the-Magician:
fix for incorrectly shown border on travelmap (world map) -
border is only shown in region map now


Interkarma:
Speedy work and much appreciated :)

PR #415, opened 2017-08-31
Assign correct info popup window based on item

Allofich:
This hopefully helps lay the groundwork for info popups on items. This doesn't parse anything, it just assigns the correct popup, or in the case of potions, inserts a TODO with information for implementing it later.


Interkarma:
This is a very valuable start, thank you!

PR #416, opened 2017-09-01
Fixes

Allofich:
A few minor fixes. I fixed up a few comments from my last PR, fixed being able to equip the buckler with 2-handed weapons (it's reported in the bug reports forum) and added the popup for when you can't travel due to nearby enemies.

I used the CUSTOM01 quest to test the travel fix, and I noticed that the quest doesn't reveal the location of the graveyard on the travel map. The quest file says "reveal dungeon" which I think should be "reveal graveyard", but even after doing that the graveyard won't show up on the map, so it's not in this PR. I did a quick investigation but couldn't find out what the problem is.


Nystul-the-Magician:
i already told interkarma that custom01 is broken - it has to do with quest machine not resolving the symbol for place correctly in this case

Allofich:
> btw, "reveal dungeon" should be correct, there is no such thing as "reveal graveyard"

I said that because looking at the official quests, they seem to have a pattern of declaring a dungeon location with a name, and then using reveal with that same name. And, looking at 20C00Y00 right now, it has two dungeons, declared as "mondung" and "hideout", and later uses "reveal mondung" and "reveal hideout". We need to be able to handle multiple dungeons separately, right?

Interkarma:
Hey guys - the "reveal _placeSymbol_" action works fine. I've just used the wrong symbol name in CUSTOM01 by mistake. It's a simple script error, don't sweat it. :)

Interkarma:
Thanks for the fixes Allofich. Will review and merge when I can.

Nystul-the-Magician:
Are you sure Interkarma? What would be the right symbol name in CUSTOM01?

Interkarma:
The target symbol has to match the Place resource symbol. So:

    Place _graveyard_ remote ancientgraveyard
    reveal _graveyard_

NOT

    Place _graveyard_ remote ancientgraveyard
    reveal _dungeon_

The Place resource itself is agnostic of how it was created, but the target symbol was wrong in my script. Now the symbol name is correct, the reveal action should add it to discovery as expected. Let me know if you still have issues with this.

Nystul-the-Magician:
I am asking because I tried to change it to _graveyard_ already - but this did not work...
when i tried to debug ancientgraveyard was somehow not resolved correctly as valid place in quest machine

Interkarma:
No worries, I can check that as well. But the script was most certainly wrong. :)

Interkarma:
For the second issue, problem is with this bit of code in RevealLocation:

    if (place.SiteDetails.siteType == SiteTypes.Dungeon)
    {
        // do reveal
    }

Exterior Places are actually Town Sites (internally), not dungeons. It doesn't make sense to filter RevealLocation to dungeons only, as it can be used for other things as well. I'll issue a change shortly to fix this.

PR #417, opened 2017-09-01
Add sounds when enemy hits player or swing misses

Allofich:
Adds hit sounds when enemy hits player based on their weapon, or lack of, and also plays an appropriate miss sound if the player isn't in range of their attack.

### DIFF
--- a/Assets/Scripts/Game/EnemyAttack.cs
+++ b/Assets/Scripts/Game/EnemyAttack.cs
@@ -126,9 +126,9 @@ namespace DaggerfallWorkshop.Game
                         if (dfAudioSource)
                         {
                             if (weapon == null)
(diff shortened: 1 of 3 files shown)


Allofich:
Adjusted the volumes to what I believe are the values being used in classic. I see it passing 110 as an argument for these, and 105 for the body falling sound, while it normally passes 100 to the sound function.

I've refrained from changing the volume for the player hitting opponents or them parrying, because I think these will be a bit too loud since they're playing as 2d sounds now. I think they should be 3d sounds played on the enemy that is hit, rather than being played from the FPSWeapon.

Interkarma:
Cheers, looks good. I agree that 3D sound makes more sense for these sounds.

PR #418, opened 2017-09-01
Fix horse casule height and prevent horse crouching

ajrb:
Thanks for pointing out why the y axis capsule center needs repositioning.

Also realised I had re-allowed crouching at some point. Horses cant crouch right?


Interkarma:
Cheers. 👍

Allofich:
> Horses cant crouch right?

Right.

PR #419, opened 2017-09-02
Prototype macro handling system

ajrb:
Please let me know what you think of this. If you like it I will finish it and tidy it up and submit again.

Info status popup complete. (press 'I')
Weapon info almost complete.
Armour info partially complete.
Quest macros handled.

### DIFF
--- a/Assets/Game/Addons/ModSupport/ModSettings/ModSettingsWindow.cs
+++ b/Assets/Game/Addons/ModSupport/ModSettings/ModSettingsWindow.cs
@@ -1,4 +1,4 @@
-﻿// Project:         Daggerfall Tools For Unity
+// Project:         Daggerfall Tools For Unity
 // Copyright:       Copyright (C) 2009-2017 Daggerfall Workshop
 // Web Site:        http://www.dfworkshop.net
 // License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
@@ -588,7 +588,7 @@ namespace DaggerfallWorkshop.Game.Utility.ModSupport.ModSettings
                 string description = presets[currentPresetIndex][ModSettingsReader.internalSection]["Description"];
                 if (description != null)
                 {
-                    messageBox.SetText(description, "", message);
+                    messageBox.SetText(new string[] { description, "", message });
                 }
                 else
                 {
--- a/Assets/Scripts/Game/DaggerfallUI.cs
+++ b/Assets/Scripts/Game/DaggerfallUI.cs
@@ -376,6 +376,9 @@ namespace DaggerfallWorkshop.Game
                 case DaggerfallUIMessages.dfuiOpenQuestJournalWindow:
                     uiManager.PushWindow(dfQuestJournalWindow);
                     break;
+                case DaggerfallUIMessages.dfuiStatusInfo:
+                    MessageBox(22);
+                    break;
                 case DaggerfallUIMessages.dfuiExitGame:
 #if UNITY_EDITOR
                     UnityEditor.EditorApplication.isPlaying = false;
@@ -877,7 +880,7 @@ namespace DaggerfallWorkshop.Game
             return texture;
         }
 
-        public static DaggerfallMessageBox MessageBox(string message)
+        public static DaggerfallMessageBox MessageBox(string message, MacroDataSource mds = null)
         {
             DaggerfallMessageBox messageBox = new DaggerfallMessageBox(Instance.uiManager, Instance.uiManager.TopWindow);
             messageBox.SetText(message);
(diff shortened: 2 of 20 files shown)


Interkarma:
OK, I like the general design here. :)

I don't think we need to make too many changes, but it would be a better fit for MacroDataSource to be changed to an interface and become IMacroDataSource. The reason for this is that C# allows single inheritance only, but can have multiple contracts attached through interfaces. This would allow you to add IMacroDataSource contract even to classes that already inherit from something else. And it frees classes that might want to implement macro data source to have some other base class in their design.

As the inheriting class already needs to implement all the members in abstract base class, an interface is a perfect match for this. You'll be able to remove all the "not implemented" exceptions and those methods will be forced to implement the required methods by interface contract.

I'm going to merge to branch and play with this some more today. Will let you know if I have any more feedback.

ajrb:
I know C# only allows single inheritance - it's the same as Java. (and a principle I agree with ever since I tried to wrap my head around some horrible multiple inheritance C++ class hierarchy of microsoft's)

I was originally going to have an interface for MacroDataSource classes to implement, but I figured until we actually have a class that already extends another and also needs to be a MDS there was little point. (I assume that VS can easily extract an interface for us when the time comes anyway) The abstract base class is there so that MDS classes don't need to implement all of the methods. Unless I have misunderstood something about C# that differs from Java, I think that your statement "As the inheriting class already needs to implement all the members in abstract base class" is incorrect.

I would also preferred to stick to composition rather than inheritance but as there are going to be a lot of methods inheritance seemed the better option. Having to write all the delegate functions (not C# delegates) for every MDS would have been a pain.

If I am missing some C# way of structuring this because I essentially think in Java do let me know. :-)

I'll get back on with developing this tomorrow then since you seem to be okay with the approach. Not planning to define all the macros, but would like to have the inventory & info messages all working. I'll also write some instructions for adding new macro definitions, because the entire point of this is to ensure that no one needs to expend much thought on them going forward.

ajrb:
Actually been thinking a little more and might have a way to avoid the need for inheritance by using composition instead. It's late but I'll try it out tomorrow morning.

Interkarma:
>Unless I have misunderstood something about C# that differs from Java, I think that your statement "As the inheriting class already needs to implement all the members in abstract base class" is incorrect.

You mistake my meaning. I say this because the only purpose of the base methods right now is to throw an exception if called when not implemented. They serve no actual purpose whatsoever that couldn't be better accomplished with an interface. :)

I like what you're doing with this, but I don't want these classes inherited from MacroDataSource and consuming their only possible ancestor. I might want them to inherit from another base class later, or you might need to attach macro handling to a class that already inherits from something else.

if you're going to keep this general design, please just change MacroDataSource to an interface called IMacroDataSource. It's a much better way to accomplish what you're doing here.

ajrb:
I was under the impression that implementations of interfaces have to provide implementations for all methods defined by that interface. That's certainly true in Java, is it not for C#?

Interkarma:
>I was under the impression that implementations of interfaces have to provide implementations for all methods defined by that interface. That's certainly true in Java, is it not for C#?

Correct, that's required to fulfil the contract. :) My point is that I don't want inheritance used on these classes, except as an interface. Right now the base class does nothing more than throw exceptions, it's not really contributing anything an interface couldn't do better.

The class implementing the interface can then handle not implemented members as required. Preference would be for just an empty method stub or logged informational message in these cases rather than throw an exception.

Interkarma:
>Actually been thinking a little more and might have a way to avoid the need for inheritance by using composition instead. It's late but I'll try it out tomorrow morning.

Composition could work too, and would also solve my issue with using inheritance on these classes.

Interkarma:
Thinking about this more - I agree it doesn't make much sense for objects like items to implement macros for pronouns or religion (even as stubs), but in that case it probably doesn't make sense for them to be in the base class either. I think you're on the right track, but we just need to agree on how best to refine, and a way that works even if object requiring a macro data source already inherits from something else. :)

Composition might just be a better design. Let's see how that rolls and go from there!

ajrb:
I just tried what you were suggesting and it does give compilation errors. I do agree that it's not ideal to have each context provider class inherit for macros but I considered the trade off to be worth it so when adding new handlers in future people didn't need to go through all of the implementing classes adding a new method. Anyway, going to use composition to get the best of both worlds now. Guess I was just happier to live with the inheritance. (my design coleagues at work would be scandalised, haha)

Thinking about it, not sure there will be a large number of context provider classes like I imagined. There's items & quests but what else? Well hopefully more will turn up over time otherwise my point about avoiding them having to implement all methods seems rather weak...

Interkarma:
The main sources for text will be item data, quests and quest resources, and dialog (of which the best dialog text is actually from quests anyway).

I've been chewing over this as well in the back of my mind today. You're right, there's a only a few mainline parametric macros. Could possibly just handle this with a central class that has a few overloaded methods. Input references for each overload will determine what macros are available to expand. Most of the time this will be an item, npc, or quest.

But let's see the composition design next and go from there. :)

ajrb:
Aha, spells is another I think? :-)

It's almost done now - just writing steps to add new macro handlers. Since I want to make it easy to add new ones going forward. Much happier with it not requiring objects to inherit.

ajrb:
Thanks komoku, I hadn't realised. I just got the titles from http://en.uesp.net/wiki/Daggerfall:Reputation and threw them into the code. I'll refine it to match classic.

PR #420, opened 2017-09-02
Implement weapons and armor taking damage

Allofich:
Weapons and armor take damage now.

For now at least, I've copied classic behavior of only damaging non-magical weapons and armor when a damage-causing hit is made with a weapon. I know Morrowind has the same thing where creature hits don't damage armor, so this may have been a conscious decision.

I found what seems to be an oversight in classic, in that there is no code for damaging shields, only the weapons and the armor on the body part that gets hit, so I introduced an attempt at adding shield damage in. Going on the assumption that an attacked target first tries to avoid an attack completely with agility, luck, etc. and then uses their shield and then any armor equipped under the shield, I put in a "what if" calculation that looks at whether the attack gets avoided at the shield stage, in which case the shield takes damage. If the attack is not avoided, the armor worn underneath takes damage.

As part of putting that in, I needed to refactor some stuff for getting shield armor values (or else there would have been duplicate code), which makes the PR a little bigger.

I named the damage function DamageThroughPhysicalHit. Classic uses a separate function for damaging magic items when you use them, and we might, too.

The damage to items is done with the same calculation as classic. A message appears for the player and and also for enemies when an item breaks, which I think would also happen in classic, but I didn't test enemy. Item is unequipped from both player and enemy when it breaks.

I improved the message when an item breaks to show the long name of the item rather than the short name, and to be grammatically correct ("have" for boots, gauntlets and greaves) and say "has/have" broken rather than "is" broken. I noted the difference from classic in a comment.

### DIFF
--- a/Assets/Scripts/Game/Items/DaggerfallUnityItem.cs
+++ b/Assets/Scripts/Game/Items/DaggerfallUnityItem.cs
@@ -810,6 +810,127 @@ namespace DaggerfallWorkshop.Game.Items
             }
         }
 
+        public int GetShieldArmorValue()
+        {
+            switch (TemplateIndex)
+            {
+                case (int)Armor.Buckler:
+                    return 1;
+                case (int)Armor.Round_Shield:
+                    return 2;
+                case (int)Armor.Kite_Shield:
+                    return 3;
+                case (int)Armor.Tower_Shield:
+                    return 4;
+
+                default:
+                    return 0;
+            }
+        }
+
+        /// <summary>
+        /// Get body parts protected by a shield.
+        /// </summary>
+        public BodyParts[] GetShieldProtectedBodyParts()
+        {
+            switch (TemplateIndex)
+            {
+                case (int)Armor.Buckler:
+                    return new BodyParts[] { BodyParts.LeftArm, BodyParts.Hands };
+                case (int)Armor.Round_Shield:
+                case (int)Armor.Kite_Shield:
+                    return new BodyParts[] { BodyParts.LeftArm, BodyParts.Hands, BodyParts.Legs };
+                case (int)Armor.Tower_Shield:
+                    return new BodyParts[] { BodyParts.Head, BodyParts.LeftArm, BodyParts.Hands, BodyParts.Legs };
+
+                default:
+                    return new BodyParts[] { };
+            }
+        }
+
+        /// <summary>
+        /// Get the equip slot that matches to a body part.
+        /// Used in armor calculations.
+        /// </summary>
+        public static EquipSlots GetEquipSlotForBodyPart(BodyParts bodyPart)
+        {
+            switch (bodyPart)
+            {
+                case BodyParts.Head:
+                    return EquipSlots.Head;
+                case BodyParts.RightArm:
+                    return EquipSlots.RightArm;
+                case BodyParts.LeftArm:
+                    return EquipSlots.LeftArm;
+                case BodyParts.Chest:
+                    return EquipSlots.ChestArmor;
+                case BodyParts.Hands:
+                    return EquipSlots.Gloves;
+                case BodyParts.Legs:
+                    return EquipSlots.LegsArmor;
+                case BodyParts.Feet:
+                    return EquipSlots.Feet;
+
+                default:
+                    return EquipSlots.None;
+            }
+        }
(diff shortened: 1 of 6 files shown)


Allofich:
I redid the shield damage logic, simplifying things and making it more consistent with the rest of the damage calculations by just damaging the shield rather than the armor of the hit body part if the hit body part is protected by the shield.

Assets/Scripts/Game/Items/DaggerfallUnityItem.cs
```diff
@@ -810,6 +810,127 @@ public int GetMaterialArmorValue()
...
+        /// <summary>
+        /// Get body parts protected by a shield.
+        /// </summary>
+        public BodyParts[] GetShieldProtectedBodyParts()
+        {
+            switch (TemplateIndex)
+            {
+                case (int)Armor.Buckler:
+                    return new BodyParts[] { BodyParts.LeftArm, BodyParts.Hands };
+                case (int)Armor.Round_Shield:
```
ajrb: Round shield protects legs?

ajrb:
Looks good stuff to me. Do shields & heavy armour reduce dodge chance in daggerfall at all?

Allofich:
What do you mean by dodge chance? They increase your chance to avoid taking a hit, which is what armor does in Daggerfall.

Personally I prefer a model where armor reduces damage rather than just increasing your chances to avoid taking damage, but changing that would be a major change to the game so I think it would be better as a large-scale mod, assuming we are able to mod game systems some day.

ajrb:
I meant is the chance of avoiding hits due to agility constant no matter if you're wearing speedos or heavy plate with a tower shield.

Also should a tower shield not protect the chest as well? Possibly the round & kite could too with round not protecting legs? Just wondering, feel free to ignore me. :)

Allofich:
There's no penalty to your ability to dodge things through agility from wearing heavy stuff.

For that, and for the shields, maybe it would be better if they were different, but for this and other things I'm generally just copying classic for now, except for fixing what I judge as bugs or unintended oversights. I don't want to get too sidetracked from working on figuring out how classic does stuff. Also some fans of the original probably won't like it if we change too much, if it isn't optional.

There's a few different things I've thought of while researching classic that I think would be cool to see:

- Different armor model, as I said above.
- Wandering monsters, monsters that fight each other.
- Adjustments to pickpocketing.
- Human enemies using appropriate equipment. As it is they all just draw from the same pool, so mages can have swords and armor, etc.
- Use the unused voice files for the different player races when getting hit, dying, maybe when attacking, and maybe use these for human enemies, too.

But, first of all, I'm going for recreation of classic, minus bugs. I do hope we can have some robust way to mod game systems at some point.

ajrb:
That's cool, I agree completely. I'm just not that familiar with classic systems so was curious if there was any downsides to using a shield or not. I like all of those ideas for enhancements, especially #4.

Interkarma:
Not much to add other than this looks like the usual Allofich magic. :)

I had scheduled item damage to go with guilds & services, as the player would then have the ability to repair items. I'm happy to merge this now, but players are very likely to be upset that items will decay with no way to repair them.

Perhaps a "repairallitems" console command that iterates through player's item collections and sets back to full condition would be an easy way to keep the decay for testing but give players the ability to quickly and simply repair their gear?

Allofich:
I had thought about that, but I don't think it's a huge problem. For one thing, item decay is pretty slow in this game. I don't know the formulas for later games in the series but I think it's a lot slower than Morrowind and Oblivion. For another, broken items are not deleted from the game world, so once we have repairing in, players could turn in any broken items they have for repairing. I believe Hazelnut is working on condition popups in the inventory screen, too, which will show condition and so people shouldn't be caught by surprise by an item breaking.

A repair all command might be good anyway but maybe we can wait for if people want it.

Allofich:
Another idea, and maybe the best one to be sure to avoid any problems or pressure on us, is to just comment out the calls to the DamageToPhysicalHit (there are 3 of them) until repairing is in. I'm completely fine with doing that.

Interkarma:
I'm happy for people to experience your work a bit first as well. Let's just run with it for a bit to get some feedback. We can then disable the damage or add a console command to repair all. Cheers. :)

PR #422, opened 2017-09-03
Refactor macro context system

ajrb:
Here you go, no inheritance for context providing objects anymore. Also added documentation, tidied up and added item condition & worth handlers.

This is now complete and ready for merging with master from my point of view. I plan to add a few more handlers, and also write instructions on 'Developer Discussion' forum so it's easy for anyone to add new handlers in the future when the see a %abc[tag] appear in game.


Assets/Scripts/Game/Items/DaggerfallUnityItemMCP.cs
```diff
@@ -0,0 +1,89 @@
...
+                return parent.weightInKg.ToString();
+            }
+
+            public override string WeaponDamage()
+            {   // %wdm
+                int matMod = parent.GetWeaponMaterialModifier();
+                return String.Format("{0} - {1}", parent.GetBaseDamageMin() + matMod, parent.GetBaseDamageMax() + matMod);
+            }
+
+            // Armour mod not correct yet.. need to refactor DaggerfallEntity.UpdateEquippedArmorValues()
```
Allofich: Looks like the armor value displayed by classic in the popup is half the value shown floating around your character on the inventory screen (which is the value GetMaterialArmorValue() returns). It's rounded down in case it doesn't divide evenly, like with leather, which is "3" and so shows as 1.

Allofich:
Also, just FYI, I found during damage calculations that the damage display for weapons in classic is incorrect. The values given in classic are what the damage would be if the material modifier was twice what is actually applied during combat. This is the base damage, before anything like strength modifiers are factored in.

Assets/Scripts/Game/Items/DaggerfallUnityItemMCP.cs
Allofich: This does not seem to be the case for shields, which show the same value as gets added to the floating numbers around the character.
This seems like a slip up in the original game, so I think we should just show the values as you have them now.

ajrb: Okay, I assumed DFU was mistaken. :-)

You're the oracle for this stuff as far as I'm concerned!

ajrb:
So does that mean my code is correct already then? I wasn't sure why there was a discrepancy.

Ebony dagger shows:
DFU = 5 - 10
Classic = 9 - 14

Allofich:
Yeah, that should be right.

ajrb:
Spent several hours adding macro handling for stats (attributes). Was more effort than I expected TBH.

Allofich, I moved the shield armour values into DaggerfallEntity.GetMaterialArmorValue() so that it's automatically correct when displayed on the info popup. This method is now slightly misnamed since for shields it ignores material, but this is a more logical place for these numbers IMO.

Interkarma:
Awesome, looks really good mate! 👍 

I'll merge to branch and have a play with it today. Will try to get everything merged to master by this evening.

ajrb:
Thanks, I've put a lot of work into this now. Once it's in master I'll write a post of instructions to add new macro handlers on the forum for people (including me) to refer to once I've forgotten all about it.

Interkarma:
Yeah, that hard work really shows. I've had a good play with this before work and everything looks good so far, I just want to test quest text a bit more before I merge to master.

I'll let you know if any issues found, but all good so far. :)

PR #425, opened 2017-09-04
Adjust monster speeds

Allofich:
Monster speed is the same formula as the player has when walking.

Also bumped up monster attack animation to be closer to classic. The speed of the swing in classic is not dependent on the monster's speed, and I think it may vary depending on the speed of the computer (or the "cycles" in DosBox), but I put it at what seems to me like a reasonable value that doesn't look overly fast.

I was also looking at adjusting the point in the animation swing where damage is applied. Interestingly, it seems like it varies by enemy in classic. Also I noticed, for example, that the imp plays its attack animation forward and then backward in classic. I don't know where this information is stored, though.

### DIFF
--- a/Assets/Scripts/Game/EnemyMotor.cs
+++ b/Assets/Scripts/Game/EnemyMotor.cs
@@ -10,6 +10,7 @@
 //
 
 using UnityEngine;
+using DaggerfallWorkshop.Game.Entity;
 
 namespace DaggerfallWorkshop.Game
 {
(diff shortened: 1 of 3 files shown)


Interkarma:
A nice snappy change, thanks. 👍 

I didn't notice the imp was playing animations forwards then backwards. The animation setup needs a bit of work for the flipping issue anyway. Can look into this later, it's not affecting gameplay right now.

PR #426, opened 2017-09-04
Removed length limit on popups.

ajrb:
Also tidied comments on MacroHelper


ajrb:
Unrequired change, my bad.

PR #427, opened 2017-09-05
Multi popups and item name fixes

ajrb:
- Added identified state to items. (first use of flags)
- Most item names now match classic, including unidentified magic items.
- Artifacts not currently identified by item system but names will be correct once that has been done.
- Tooltips also now match item titles on popups.
- Book title & author now shown on popups.
- DaggerfallMessageBox can now have a next message box added which will be displayed when user clicks. The last message box will close all of them. As far as I know Daggerfall only ever uses 2, but I wrote it so you can nest to any level. (of course) Actually found a couple of the artifacts do use 3! (skele key & mask)
- Add second info popup for health status.
- Implement magic item powers popup using multiline macros. Just displays "Power number N" for each power for now until magic systems implemented and a mapping for the DF spell effect numbers is availiable.

### DIFF
--- a/Assets/Scripts/Game/DaggerfallUI.cs
+++ b/Assets/Scripts/Game/DaggerfallUI.cs
@@ -85,7 +85,6 @@ namespace DaggerfallWorkshop.Game
         DaggerfallBookReaderWindow dfBookReaderWindow;
         DaggerfallQuestJournalWindow dfQuestJournalWindow;
         QuestMachineInspectorWindow dfQuestInspector;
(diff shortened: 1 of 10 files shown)


Interkarma:
Looks like excellent work, thank you. :)

Will complete check and merge as soon as possible.

ajrb:
You're welcome. One thing leads to another, and before you know it it's the early hours of the morning... I's sure you know how that goes.

Interkarma:
Merging! Will play with this more when I get home from work tonight. Good stuff mate!

PR #429, opened 2017-09-06
Add player gender macros

ajrb:


Assets/Scripts/Utility/MacroHelper.cs
```diff
@@ -367,31 +375,33 @@ private static string LocalReputation(IMacroContextProvider mcp)
...
+                return "a scoundrel";
             else if (rep < -20)
-                return "Criminal";
+                return "a criminal";
             else if (rep < -40)
-                return "Villain";
+                return "a villain";
             else if (rep < -60)
-                return "Pond Scum";
+                return "pond Scum";
```
seanohue: Shouldn't `Scum` be lowercased as well? (`scum`)

ajrb: Well spotted.

PR #430, opened 2017-09-07
Pass down wrapText parameter

Allofich:
This just passes down the wrapText parameter from where pickpocketing happens. It ends up passing a bool through a bunch of functions, which might not be to your liking. Feel free to reject this if you have something better in mind.

I removed maxWidth as a parameter for now, since it's not currently used.

### DIFF
--- a/Assets/Scripts/Game/DaggerfallUI.cs
+++ b/Assets/Scripts/Game/DaggerfallUI.cs
@@ -888,9 +888,9 @@ namespace DaggerfallWorkshop.Game
             return texture;
         }
 
-        public static DaggerfallMessageBox MessageBox(string message, IMacroContextProvider mds = null)
+        public static DaggerfallMessageBox MessageBox(string message, IMacroContextProvider mds = null, bool wrapText = false)
         {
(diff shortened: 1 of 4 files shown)


ajrb:
If this isn't going to have widespread use, maybe it could be restricted to just the SetText() method. So if you want to wrap you can't use the quick constructors but have to create messagebox, get tokens for id and then call SetText(tokens, mcp, wrap)

That's already done in some places. Just a suggestion.

Interkarma:
Allofich, would you have any issues with airjb's suggestion? It might be nice to use more direct route if possible. Otherwise, I have no real problem with merging as-is.

Allofich:
I redid it by passing it in as a bool to the messagebox constructor. This seems better to me than before, and it will make it easy to use wrapping again if there's another set of messages somewhere that needs it.

As for ajrb's suggestion, we could do that, but unless I'm missing something it would still involve passing the wrap bool through around 3 or 4 functions before making its way down to the AddTextLabel function. If you would prefer I try that way, Interkarma, I'll give that a shot, though.

Interkarma:
I'm happy with this. :)

PR #431, opened 2017-09-07
talk window - first stage of implementation

Nystul-the-Magician:
I create this pull request now since important parts of the basic window functionality are working and I already touched a lot of files - so I want to do several "smaller" pull-request (in fact the are not that small) instead of one huge in the end...

things changed in other files than DaggerfallTalkWindow that are worth mentioning:

- MobilyPersonNPC now holds all npc data and its data is propagated to/consumed by MobilePersonMotor and MobilePersonBillboard class

- Found and fixed a bug in HorizontalSlider class preventing correct slider operation in many cases

- extended Listbox (and its dependency) TextLabel class to support labels starting from a character other than the first (if it is not clear what I mean check out talk window list box on the left - it can be scrolled to left/right to show entries that are longer than the list box's window frame size)

### DIFF
--- a/Assets/Scripts/Editor/SetupMobilePersonEditor.cs
+++ b/Assets/Scripts/Editor/SetupMobilePersonEditor.cs
@@ -38,7 +38,7 @@ namespace DaggerfallWorkshop
             DrawDefaultInspector();
             if (GUILayout.Button("Apply Person Type"))
             {
-                setupMobilePerson.ApplyPersonSettings();
+                setupMobilePerson.ApplyPersonSettingsViaInspector();
             }
             if (GUILayout.Button("Align To Ground"))
             {
--- a/Assets/Scripts/Game/DaggerfallUI.cs
+++ b/Assets/Scripts/Game/DaggerfallUI.cs
@@ -83,6 +83,7 @@ namespace DaggerfallWorkshop.Game
         DaggerfallAutomapWindow dfAutomapWindow;
         DaggerfallExteriorAutomapWindow dfExteriorAutomapWindow;
         DaggerfallBookReaderWindow dfBookReaderWindow;
+        DaggerfallTalkWindow dfTalkWindow;
         DaggerfallQuestJournalWindow dfQuestJournalWindow;
         QuestMachineInspectorWindow dfQuestInspector;
 
@@ -166,10 +167,16 @@ namespace DaggerfallWorkshop.Game
             get { return dfBookReaderWindow; }
         }
 
+        public DaggerfallTalkWindow TalkWindow
+        {
+            get { return dfTalkWindow; }
+        }
+
         public DaggerfallAutomapWindow AutomapWindow
         {
             get { return dfAutomapWindow; }
         }
+
         public DaggerfallExteriorAutomapWindow ExteriorAutomapWindow
         {
             get { return dfExteriorAutomapWindow; }
@@ -203,6 +210,7 @@ namespace DaggerfallWorkshop.Game
             dfAutomapWindow = new DaggerfallAutomapWindow(uiManager);
             dfBookReaderWindow = new DaggerfallBookReaderWindow(uiManager);
             dfQuestJournalWindow = new DaggerfallQuestJournalWindow(uiManager);
+            dfTalkWindow = new DaggerfallTalkWindow(uiManager);
 
             dfExteriorAutomapWindow = new DaggerfallExteriorAutomapWindow(uiManager);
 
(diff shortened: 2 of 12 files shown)


Allofich:
Hey Nystul, I want to help but it's hard to keep up! I did look at the talk stuff yesterday and uncovered a function for determining the addition of buildings to the "regional" tab. I'll lay it out below.

Allofich:
First the data, there is an array of bytes at offset 1B32FA in FALL.EXE.
{ 06 00 07 00 08 00 0A 00 05 11 12 14 15 16 17 2B 33 37 }
This is used to identify which knightly orders are in which regions.
At offset 1B328E is an array of shorts:
{1A 00 15 00 1D 00 1B 00 23 00 18 00 21 00 16 00 9E 01 70 01 9D 01 98 01 9A 01 9B 01 99 01 9F 01 A0 01 A1 01 28 00 29 00 0F 00 0A 00 0D 00 02 00 00 00 03 00 05 00 06 00 08 00 }
These are 28 possible building types you can ask about. The values are either faction numbers (first values are temples, then knightly orders, etc.) then shop types (see the building enums in DF Unity) at the end.

Allofich:
Calling the first array KnightlyOrderRegions[] and the second one BuildingTypes[], function looks like this.

`CheckIfStoresAndFactionBuildingsExist()
{


    for (i = 0; i < 28; ++i)
    {
        if (i >= 8 && i <= 17) // is Knightly Order
        {
            if (CurrentRegion == KnightlyOrderRegions[i] && !DoesBuildingExistLocally(BuildingTypes[i], false)) // false means search building by faction
                AddRegionalBuildingTalkOption(i);
        }
        else if (i >= 20) // store
        {
            if (!DoesBuildingExistLocally(BuildingTypes[i], true)) // true means search by building enum, 
                AddRegionalBuildingTalkOption(i);
        }
        else if (!DoesBuildingExistLocally(BuildingTypes[i], false)) // temples, guilds
        {
            AddRegionalBuildingTalkOption(i);
        }
    }
}`

With DoesBuildingExistLocally being an interation over all local buildings in the location.

Allofich:
Other than that, I've found some stuff regarding how reactions to the player are handled. It will be necessary to access the NPC's faction data for this, so that's something to perhaps keep in mind.

Allofich:
If you don't feel like trying to implement the above function, don't worry about it. I might try to do it myself later.

Nystul-the-Magician:
thanks for your efforts! This will come handy sooner than later!
next thing I want to do is a TalkManager class that manages talk topics, answers etc and will be used by the DaggerfallTalkWindow to get the contents (topics and answers after selecting a topic) for the different talk sub-windows (I mean the left, top, right frame).
I want to to split presentation and logic as much as possible so this is the reason to write a manager class here.
About the different talk modes/topics:

asking for locations would be the first system I planned to implement (here your insights will be super valuable)

topics coming from quest systems should also be possible this way

completely unknown in terms of how does it work and how to do it are:

-  tone options for all kind of topics (where are they stored)

- tell me about topics (same as above)

- news (same as above)

- work (this seems a bigger system behind the scenes as all open quests from shopkeepers are known in vanilla daggerfall in the talk window how it seems)

Allofich:
> tone options for all kind of topics (where are they stored)

As far as this one goes, the different player tones seem to be located next to each other in TEXT.RSC, going by these enums

0 = Polite tone
1 = Normal tone
2 = Blunt tone

So for asking for the location of something, it is (SpeechStyle + 7225).

Interkarma:
Very excited to see this here. I'm on my way to bed, so nothing to add right now. Looking forward to going through this when I can however. :)

Allofich:
Hey Interkarma, do you happen to know what the difference between Daggerfall native units of distance and those of DF Unity would be?

There is a frequently used function in the code that I've determined finds the 2d distance between two sets of coordinates, but I don't know how that distance would convert to DF Unity.

Allofich:
It's relevant here because it affects whether NPCs will mark a location on your map or not.

Interkarma:
Hey Allofich, check the top of MeshReader.cs for a breakdown of ratio between Daggerfall world units (inches) units and Unity units (metres). There's also some good unit conversion stuff in CityNavigation to convert between absolute world position and local scene position (SceneToWorldPosition() and WorldToScenePosition()). Daggerfall has a few different sets of 2D coordinates, but usually uses world coordinates (absolute inches) at ground level.

Hit me up in the coredev forums if any help needed, happy to help however I can. I'd probably need more info on that function you've found though,

Allofich:
OK great, so we can probably convert it then.

Interkarma:
This looks good mate, well done! Since there's no major impact to any other gameplay systems, I'll just push this straight to master.

PR #433, opened 2017-09-07
Artifact handling added

ajrb:
DFU now recognises artifacts, now correctly showing name and powers.
Also fixed item template hitpoints for some ingredients.


Interkarma:
Cheers mate, looks like good work. 👍

PR #434, opened 2017-09-08
Identify more save vars, use avoidHit mod

Allofich:
Identifies several values in savevars, and uses the mod to avoiding hits that can be set during character generation in the combat formula.

I also identified the other mods that can be set during character generation (you get one of these as the result of the "You have the most trouble ___." question.) The only one I haven't found yet is the reaction modifier, which I think affects how much NPCs like you. The modifiers I found were all next to each other but the reaction one didn't show up.

### DIFF
--- a/Assets/Scripts/API/Save/SaveVars.cs
+++ b/Assets/Scripts/API/Save/SaveVars.cs
@@ -27,39 +27,67 @@ namespace DaggerfallConnect.Save
         #region Fields
 
         const string filename = "SAVEVARS.DAT";
+        const int biographyModifiersOffset = 0x30;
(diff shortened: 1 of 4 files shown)


Allofich:
I've arranged the order the offsets are declared and the order the values are read from the SAVEVARS file to be in increasing order. This included moving the global vars part up. If you want the global vars one to be listed last for whatever reason, I can put it back, but I like it if the discovered data is in the order of the original file, and we can fill in the missing pieces as we go until hopefully we have a complete reconstruction of it.

Interkarma:
This is great - happy to see some of those biography modifiers come into play. We'll be able to have complete chargen working before too long. :)

PR #435, opened 2017-09-08
MeshReplacement: Treasures

TheLacus:
Imports 3d models for treasures and lootable corpses. 
I don't think this is the best way to deal with the fact that the treasure prefab is loaded from resources, but if I replace the entire gameobject (and add lootable script and sphere collider) i have troubles preserving serialization. I'm open for suggestions.


Interkarma:
No better suggestion right now. This looks like it would get the job done for the time being, so no complaints here. :)

PR #437, opened 2017-09-08
Change %reg to be same as %crn, new macro %hrg for home region

ajrb:


Interkarma:
Thanks for fixing this!

PR #438, opened 2017-09-09
Updates to activating objects

Allofich:
Changes:
1) Implements "You are too far away..." message, shown when player tries to activate various objects out of range.
2) Implements "You see a/an %s." when you click on an enemyMobile in info mode.
3) Activation ranges match classic. Buildings and enemy mobiles can be identified from far away like in classic.
4) Fixed being able to activate things through walls (ex. pickpocketing guards in Daggerfall castle through the wall)

To do this, I upped the raycast length when activating things to a long one. There was a problem with this though: I was able to click on things behind walls. There was a comment that RayCastAll was being used because when things block the raycast, it feels unresponsive to the player, but changing to RayCast fixed my problem and also fixed being able to activate through walls, and I don't notice any obvious unresponsiveness. Is there a particular case you were thinking of Interkarma? Could using a layer mask on the ray cast to filter out certain colliders solve the issue?

### DIFF
--- a/Assets/Scripts/Game/Entities/EnemyEntity.cs
+++ b/Assets/Scripts/Game/Entities/EnemyEntity.cs
@@ -32,6 +32,7 @@ namespace DaggerfallWorkshop.Game.Entity
         int careerIndex = -1;
         EntityTypes entityType = EntityTypes.None;
         MobileEnemy mobileEnemy;
+        bool pickpocketByPlayerAttempted = false;
 
         #endregion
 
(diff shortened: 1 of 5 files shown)


Allofich:
Also, I wanted to implement the message for when you click on a corpse in info mode, but unless I missed it the corpse doesn't have the enemy type any more. Would we have to reverse-lookup the type from the texture being used?

Allofich:
Added updated pickpocketing. You can only do one per target now. This matches classic for mobileNPCs, but differs from classic for enemy mobiles, but that behavior in classic seems to be an oversight/exploit to me, since you can spam pickpocket attempts and trivially raise your pickpocket skill.

Interkarma:
Awesome work Allofich, thank you. :) I've been debating removing the raycast all for a while now. I think the problems it added were likely greater than those it solved, so agree that it can go.

I don't have a good info-click solution for monster corpse loot containers right now. My feeling is just to add a simple field value to loot containers which tracks what monster type the container dropped from (if any). The info click can then just read this from loot container on ray hit. And will need to serialise this new value with loot containers as well.

PR #439, opened 2017-09-09
Implement weight and encumbrance

ajrb:
Inventory, carry, wagon, gold.
Added message for gold too heavy from bank over classic.
Fixed skeletons key artifact popups.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/HardStrings.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/HardStrings.cs
@@ -35,6 +35,8 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
         public const string usingLeftHand = "Using weapon in left hand.";
         public const string itemHasBroken = "%s has broken.";
         public const string itemHasBrokenPlural = "%s have broken.";
+        public const string cannotCarryAnymore = "You cannot carry anymore stuff.";
+        public const string cannotCarryGold = "You cannot carry that much gold.";
 
         public const string enterSaveName = "Enter save name";
         public const string selectSaveName = "Select a save";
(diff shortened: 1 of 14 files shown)


ajrb:
Wanted to implement following changes:
- message for trying to exceed wagon limit (classic just does nothing)
- allow equipping straight from wagon (classic allows but not current implementation in DFU.. I didn't in case there's a good reason)

Allofich:
If you want to, there's another message from classic you can add:

"Your cart cannot hold anymore stuff." - When you try to overload your wagon/cart

Also, if you meant to recreate the message from classic when you try to withdraw more gold from your bank account than you can carry, it's "You cannot carry so much gold."

Allofich:
> message for trying to exceed wagon limit (classic just does nothing)

Really? A string exists for it, as I mentioned.

Allofich:
At the bottom of DaggerfallCharacterSheetWindow(), in UpdateSecondaryStatLabels(), you'll see a TODO comment for updating the encumbrance display when you are leveling up and add or remove a point to a related attribute. Think you could add the encumbrance line in there?

Allofich:
> Also, if you meant to recreate the message from classic when you try to withdraw more gold from your bank account than you can carry, it's "You cannot carry so much gold."

Actually, the message you used is probably better, because it also covers the case where the player is weighted down by non-gold items and cannot withdraw and carry even a small amount of gold. Well, I'm fine with either one, I guess.

ajrb:
Hmm, I tested classic and that message was not displayed when trying to put something in the cart when it was overloaded. happy to implement the message, just wanted to check with others first.

Same with gold, in my version of classic no message is displayed. I'll try again, and use your quoted message if I still can't get it on my classic.

Thanks for pointing out the leveling event, I missed that. Will fix it.

Interkarma:
This looks like really solid work, thank you for adding all of this. 👍

Assets/Scripts/Game/UserInterfaceWindows/HardStrings.cs
```diff
@@ -36,6 +36,7 @@ public class HardStrings
         public const string itemHasBroken = "%s has broken.";
         public const string itemHasBrokenPlural = "%s have broken.";
         public const string cannotCarryAnymore = "You cannot carry anymore stuff.";
+        public const string cannotHoldAnymore = "You cart cannot hold anymore stuff.";
```
Allofich: Small typo :) Should be "Your" not "You"

PR #440, opened 2017-09-10
Implement clicking in info mode on dead mobiles.

Allofich:
Does as title says.

I think that classic still has a monster's stats within a corpse, so I did it the same way (adding an EnemyEntity field to the loot container) instead of just adding a name string and isEnemyClass bool.

Corpses EnemyEntity field isn't saved in DF Unity saves yet, so clicking on them after loading a DF Unity won't show anything.


Interkarma:
Thank you. No problems storing that entity data, it definitely gets the job done. Are you planning to add to save game, or would you prefer I work this in?

PR #441, opened 2017-09-10
Implement starting crime guild quests

Allofich:
This starts the thieves guild and dark brotherhood quests in the same manner as classic. The necessary data is also imported from classic saves.

Problems:
- I don't know Unity and its "components" and stuff very well and couldn't figure out how to make a blood splash appear when a citizen is killed. Copying the approach used in the file for hitting an enemyMobile didn't work. I guess that mobileNPCs need to have "EnemyBlood" assigned to them somehow.
- The population manager may need to be told when a mobileNPC is killed. I don't know much about it and haven't touched it.
- The thieves guild quest and dark brotherhood quest might not necessarily work. Are we only activating quests that are curated? Should these be commented out from starting for now?
- Variables aren't saved in Unity saves yet.

### DIFF
--- a/Assets/Scripts/Game/WeaponManager.cs
+++ b/Assets/Scripts/Game/WeaponManager.cs
@@ -646,6 +646,16 @@ namespace DaggerfallWorkshop.Game
                         hitEnemy = true;
                     }
                 }
+
+                // Check if hit a mobile NPC
+                MobilePersonNPC mobileNpc = hit.transform.GetComponent<MobilePersonNPC>();
+                if (mobileNpc)
+                {
+                    // TODO: Create blood splash.
+                    weapon.PlayHitSound();
+                    mobileNpc.Motor.gameObject.SetActive(false);
+                    GameManager.Instance.PlayerEntity.TallyCrimeGuildRequirements(false, 5);
+                }
             }
         }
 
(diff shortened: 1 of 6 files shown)


Assets/Scripts/Game/WeaponManager.cs
```diff
@@ -646,6 +646,16 @@ private void WeaponDamage(FPSWeapon weapon, out bool hitEnemy)
...
                     }
                 }
+
+                // Check if hit a mobile NPC
+                MobilePersonNPC mobileNpc = hit.transform.GetComponent<MobilePersonNPC>();
+                if (mobileNpc)
+                {
+                    // TODO: Create blood splash.
+                    weapon.PlayHitSound();
+                    mobileNpc.Motor.gameObject.SetActive(false);
```
Interkarma: Will probably need to handle this through population manager so it can recycle the NPC. I'm happy to work on this if you like, and add the blood splash. I've added this to my queue.

Interkarma:
This is awesome! I don't know if those quests will work properly, and the guild structure definitely won't be ready for a while, so please just disable the quest deployment for the time being. I would like to revisit this soon though - once main quest is mostly playable. Good to have this ready to go for later. :)

PR #444, opened 2017-09-11
Reaction mod, serialize data, disable crime quests

Allofich:
Identifies and imports the biography reaction mod, serializes new player entity data (minMetalToHit, biography mods, thieves guild and dark brotherhood) and disables starting dark brotherhood and thieves guild quests or tallying their requirements.

The only piece of data I didn't serialize was that for the enemy entity in corpses. I gave it a quick go, just saving and loading the EnemyEntity in SerializableLootContainer, but not all the data was being saved, as it would just show "You see a dead ." after killing an enemy, saving, loading the save and clicking the corpse again. Interkarma, do you know what needs to be done there?


Interkarma:
Cheers mate, will review properly and merge in morning.

For EnemyEntity might just need to add [Serializable] to class and [SerializeField] to private members you want to save/load. Other option is to make a storage struct for the required data to marshall across save and live states. I tend to use the second option rather than save classes directly, but either should work.

Allofich:
The name now works. I switched from trying to save the entity to just saving a string with the name and a bool if it was an enemy class or not, to keep it simple.

Interkarma:
Often the simplest solutions are the best - good work!

PR #445, opened 2017-09-11
Apply TerrainDistance at startup

TheLacus:


Interkarma:
Thank you for this. :)

PR #446, opened 2017-09-13
Implement holiday text

Allofich:
Holiday text implemented according to reversed function from classic.

I put in a timer to keep the text from immediately appearing upon game load or repeatedly happening every time you cross back and forth over a city border. Actually the city border (for when the "You are entering" text and holiday text appear) should be further out to match classic, which would also alleviate this problem, but that can come later. The timer is just something I hacked together and could be removed or adjusted.

The holiday text will need to have additional text macros filled out before it all looks correct.


Nystul-the-Magician:
you rock! You really motivate me to also contribute. I love your passion for putting all this stuff in ;)

Allofich:
Really? Thanks, that's great to hear!

ajrb:
I'm starting to suspect he has a secret copy of the daggerfall source code! Either that or mega-ninja reverse engineering skills. :)

Interkarma:
This is awesome Allofich, great work!

For expanding the city border, I did some work around this with PopulationManager (as NPCs can spawn a ways outside the tight city rect). While it's still useful to have the precise rect for other reasons, I can use the approach in PopulationManager to enlarge boundary point where player is considered within city limits.

PR #449, opened 2017-09-13
Fix merge conflict with shield armour values

ajrb:
Caused by me changing Allofich's code while he was still working on it I think.

Needs Allofich to confirm I fixed this correctly, but basically removed the bit I added and was duplicated by his work.


Interkarma:
Thanks mate. I'll just wait for Allofich to give the thumbs up before I merge this one.

Allofich:
Looks good to me. In fact looks like currently shields are giving double their correct armor amounts due to the code duplication, but this PR should stop that.

Interkarma:
Thanks guys :)

PR #450, opened 2017-09-13
next stage of implementation of Talkwindow

Nystul-the-Magician:
@Interkarma: these things might be of particular interest for you:

added GameObject "TalkManager" with attached script to scene

extended TextLabel and Listbox (all existing functionality should still work as intended):
Listbox now has horizontal scroll modes entry-wise or pixel-wise now. Vertical scroll mode can be set to entry-wise (old mode) or pixel-wise as well. Pixel-wise scroll modes go hand in hand with a newly introduced restricted render area for listbox's content so that everything outside this Rect won't be drawn.

TextLabel also has the restricted render area so that listbox can propagate its restriction area to its child textboxes. TextLabel can now use a property WrapWords in conjunction with existing property WrapText to wrap text while respecting word wrapping (so words don't get teared apart if possible)

I had to change VerticalScrollBar class to allow for setting the scroll index without raising a scroll event - otherwise I had an infinite loop when I implemented scrollbar scrolling for listbox content and mouse up/down scrolling inside the listbox for listbox content (not on the scrollbar which works as well of course).
I did it so that existing code still works, hope this fulfills everyone's needs


Interkarma:
This is great stuff Nystul! I look forward to reviewing and merging when I can.

Interkarma:
Looks good from a reading. I'll merge and let you know if I can find any issues in testing. Thanks mate.

Interkarma:
One of the changes cause ListBox.TextColor not to be used correctly. For example, DaggerfallUnitySaveGameWindow sets a custom colour on savesList, and this now showing up as Daggerfall's default yellow colour instead of what was set. Please look into this when you can. :)

Interkarma:
This is really great work Nystul. You're amazing. :) 👍

Nystul-the-Magician:
thx for fixing ;)

PR #451, opened 2017-09-14
Initial version of trade window

ajrb:
Finished initial version of trade window

- Cost & price calculations are rubbish
- Selling items only for now
- Building name macro needs doing


ajrb:
Lots of duplicate commits in this branch - seem to come from updating to master to get code I needed to use. Sorry about that, I have no idea why it happens. I'm very familiar with git branch management & rebasing but GitHub & forked repos seems to have some wierdness I don't understand.

If it's a problem I will create a diff and try to put everything into a single commit. I did try and it's going to take some time as it didn't 'just work'. :(

Interkarma:
No problems mate, I expected this to be a large one. Thanks for all the hard work!

I'll send this to a new branch on my end to start with, then merge back to master as soon as I'm done testing.

Interkarma:
Looks good on first read and with a quick test. Nothing to add right now except that you're amazing! :)

I'll test this more thoroughly when I can and send to master.

ajrb:
Aww, thanks for the kind words. 

I do expect there to be some minor issues... like not being able to sell something somewhere you should be able to etc. There are so many permutations! Anyway I plan to set the forumites on testing it...

PR #452, opened 2017-09-14
Add armour values to paperdoll

ajrb:
Well that was surprisingly easy... which is a first for any of the DFU tasks I've set myself.

Note I moved a couple of the values a little to improve on classic. Spent much more time on the positioning than the code!! lol


Interkarma:
Good job again! :)

PR #454, opened 2017-09-15
Boots & shoes count as leather armour

ajrb:
Tiny little tweak I made last night - just had to give it a test before submitting


Allofich:
I can confirm they give the same armor as leather in classic.

PR #455, opened 2017-09-15
Basic repair - instant

ajrb:
Cost calculation is rubbish
Fixed broken skills popup


ajrb:
Could do with help getting the colour right so it matches INVE02I0.IMG

Interkarma:
Nice work!

PR #456, opened 2017-09-15
TextureReplacement: support for cfa files

TheLacus:
I changed an array of Texture2D to an array of ImageData in TransportManager in order to maintain the correct size on screen with different resolutions. This is often an issue with daggerfall hud images where texture size is used as rect. Width and Height of Imagedata always refers to Daggerfall file, while Texture2D properties are the actual resolution of texture.

### DIFF
--- a/Assets/Scripts/Game/TransportManager.cs
+++ b/Assets/Scripts/Game/TransportManager.cs
@@ -1,5 +1,5 @@
 // Project:         Daggerfall Tools For Unity
-// Copyright:       Copyright (C) 2009-2016 Daggerfall Workshop
+// Copyright:       Copyright (C) 2009-2017 Daggerfall Workshop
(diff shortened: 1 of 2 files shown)


ajrb:
I'm not sure I understand the distinction.

So do your changes mean people can replace the horse animations with mods now?

TheLacus:
Yes, it means they can import horse textures with higher resolution than vanilla. In this case, ImageData.width  and ImageData.height still refer to the resolution of the original texture, which is what is used to determine the rect. Otherwise a texture with 2x width would also be 2x wider on screen.

ajrb:
Fantastic thanks! I was planning to figure out how to add modding if anyone expressed an interest on the forums but no one did so I left it... long enough for you to do it. Win-win. :-)

Interkarma:
Now we just need someone to make a horse armour mod. :)

PR #459, opened 2017-09-16
Handle building name macro & fix quest item stack combination

ajrb:
Couple of fixes for items.


Interkarma:
Short and sweet, thanks for fix on stacking issue.

PR #460, opened 2017-09-16
Banking window on bankers and letters of credit implemented

ajrb:


Interkarma:
Great to see more work on the banking system. At this rate, you're going to fill out most of shops and services before I can get to spells. :)

PR #461, opened 2017-09-17
Implement classic cost formula for selling

Allofich:
Implements classic cost formula for selling.

Prices in classic are influenced by a value, set for each region in the game, that changes every in-game day. This value is imported from classic saves now. It's in a big chunk of data I called RegionData, but I'm just assuming it's all region-related. Most of it is 0 in my saved game I was using for research, but a few other values are set. For now I'm storing the regional price adjustments in the player entity for convenience, but we might want to have some sort of storage for regional data in the future, so I didn't include it in DF Unity saves. When starting a new game or loading a DF Unity save the values will be randomized to the same range they are randomized to in a new game in classic.

I haven't implemented the daily change in the values (either up 2% or down 2%, depending on something to do with the merchants faction).

I also put in the buying formula, which isn't used being used yet and probably needs more work.

### DIFF
--- a/Assets/Scripts/API/Save/SaveVars.cs
+++ b/Assets/Scripts/API/Save/SaveVars.cs
@@ -46,6 +46,9 @@ namespace DaggerfallConnect.Save
         const int cheatFlagsOffset = 0x173B;
         const int lastSkillCheckTimeOffset = 0x179A;
 
+        const int regionDataOffset = 0x3DA;
+        const int regionDataLength = 4960;
+
         const int factionDataOffset = 0x17D0;
         const int factionDataLength = 92;
 
@@ -86,6 +89,8 @@ namespace DaggerfallConnect.Save
         bool godMode = false;
 
         uint lastSkillCheckTime = 0;
+
+        ushort[] priceAdjustmentsByRegion = new ushort[62];
 
         List<FactionFile.FactionData> factions = new List<FactionFile.FactionData>();
 
@@ -341,6 +346,14 @@ namespace DaggerfallConnect.Save
             get { return globalVars; }
         }
 
+        /// <summary>
+        /// Gets array of how much prices are adjusted in each region.
+        /// </summary>
+        public ushort[] PriceAdjustmentsByRegion
+        {
+            get { return priceAdjustmentsByRegion; }
+        }
+
         #endregion
 
         #region Constructors
@@ -398,6 +411,7 @@ namespace DaggerfallConnect.Save
             ReadUsingLeftHandWeapon(reader);
             ReadCheatFlags(reader);
             ReadLastSkillCheckTime(reader);
+            ReadRegionData(reader);
             ReadFactionData(reader);
 
             return true;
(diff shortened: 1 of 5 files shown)


Allofich:
Implemented getting a letter of credit instead of physical gold if the gold from a sale would pass the encumbrance limit, as in classic.

Allofich:
Mercantile skill now tallied when a transaction is agreed to.

Allofich:
The "yes/no" transaction messages with merchants are matched to classic. The repair message may still be off until the repair cost calculation is matched to classic.

ajrb:
Wow, great stuff. I didn't even know shops would give a letter of credit if you couldn't carry the gold.

One question, is GetTradeAmount supposed to be the same for selling and repair like you have it? I assume so, but just want to make sure.

Allofich:
No, I haven't worked on repair yet.

Allofich:
>  I didn't even know shops would give a letter of credit if you couldn't carry the gold.

Pretty cool, isn't it? I always thought the letters of credit were an interesting bit of realism in this game.

ajrb:
Yeah, and it also makes gold having weight viable. Anyway really nice work man!

I think we make a pretty good team. Although I do occasionally wonder if I should try and figure out how to reverse engineer classic, but unless you're getting tired of doing it I'm perfectly happy to leave you to it. Just say if you want me to start doing my own damn reversing. :-) (although I would want some guidance)

Allofich:
I'm not tired of it, it's fun for me. I'm not a professional coder, so I appreciate it that you, Nystul and Interkarma can get game systems up and running, and I can then pitch in with adjustments or additions to it afterwards.

ajrb:
That's great to hear, because I'm loving doing the game programming. Just didn't want to assume anything.

Allofich:
Repairing costs now match classic.

There is a very minor difference (around 1 to 3 gold pieces) when selling/repairing multiple items compared to the classic price, but this seems more like a bug in classic than an intended thing, because the price is that those few gold pieces off from the summed price of the individual items. DF Unity is almost always just the sum.

I say "almost" because even in DF Unity, when selling a few valuable items, the amount is 1 gold piece more than the sum of their individual costs for some reason. Not sure what's happening there.

ajrb:
I found that classic had a rounding error for letter of credit commission and was out by 1gp often. I don't think a few gp either way will make any difference but I may take a look and see if changing precision helps. But not going to lose any sleep over it.

The next release of DFU is going to be great - with correct formulas. :-)

Allofich:
Yeah, I noticed you had just fixed a 1gp error and wondered about that. But when I took a quick look at the trade window code it seemed like it was just adding together integers for the sum, so I'm not sure what's happening there.

ajrb:
I would say not to worry about it, I fixed the LoC because it looks dumb when you get 1000gp and are charged 9 gp commission. :-)  (in classic it's actually 11gp)

Interkarma:
Awesome work Allofich. Having not looked into the selling formula at all, I'm amazed how involved it is. Very grateful to have you on board. :)

PR #462, opened 2017-09-17
Correct the LoC commission to be 1% not 10

ajrb:
It was set to 10% but should be 1%. Had to use Math.Round to stop it being 1gp off.


ajrb:
Actually didnt need Math.Round just needed a double

ajrb:
This is included in my latest PR so closing this one.

PR #464, opened 2017-09-17
Read trapped souls from classic saves and display info

ajrb:
Not sure if the TrappedSoulRecord is right since the data with the soul monster id is at 0x1B which is part of the root record - SpriteIndex in fact.

All the records seem to have above the root record is a single byte, value = 150

Anyway after mapping to the corresponding MobileEnemy, the soul names are now displaying correctly. Guess there will be more to do once magic is implemented.


ajrb:
Also contains a fix to correct the LoC commission to be 1% not 10%

PR #465, opened 2017-09-17
Talkwindow - location hints and revealing

Nystul-the-Magician:
new things:
text record viewer in unity editor (to have an easy way to show text records)
dialog macro expansion started
location hints implemented (direction and map reveal) - currently chances are hardcoded and not like in vanilla since this has to be found out, furthermore npcs always use the same hardcoded "mood" set for answers


Nystul-the-Magician:
yeah that is the reason
funny aspect I was thinking quite some time why I don't need a MCP class and why one would need it - finally I realized as long as there is only one instance it works without, as soon as you have instances you need MCP I think

Nystul-the-Magician:
I might have to implement MCP anyway I just realized since some dialog texts use pronouns like %g2 which are resolved with a contextual macro handler

ajrb:
Ah well give me a shout when you do and I'll set up the infrastructure for ya.

PR #466, opened 2017-09-18
Implement daily price adjustments

Allofich:
Implemented mostly. There's still a little bit to figure out, perhaps related to adjusting regional power.

I've tried to keep there from being a noticeable pause when the function has to process for a lot of days. I think it's OK but Interkarma, can you test doing a long fast travel and see if there is a noticeable pause for your computer? If there is maybe the function can be streamlined further.

### DIFF
--- a/Assets/Scripts/Game/Entities/DaggerfallEntityBehaviour.cs
+++ b/Assets/Scripts/Game/Entities/DaggerfallEntityBehaviour.cs
@@ -33,7 +33,7 @@ namespace DaggerfallWorkshop.Game.Entity
         EntityTypes lastEntityType = EntityTypes.None;
         DaggerfallEntity entity = null;
 
-        private int lastGameMinute = -1;
+        private uint lastGameMinutes = 0;
         PlayerMotor playerMotor = null;
 
         // Fatigue loss per in-game minute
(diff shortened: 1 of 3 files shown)


ajrb:
I can't see anything that would cause any modern cpu to pause even for a travel time of several years. Maybe I'm missing something but looks linear time to me.

Allofich:
Perhaps it's fine now. Initially I had done it by running the function from start to finish for each day that had passed, which caused a noticeable pause after I done a travel of like, over 100 days. I then changed it to accept a "times" parameter and just do the adjustments to the pricing that many times, without redoing the search, and this seemed to fix it. Sometimes, though, it still seemed like maybe? there was a brief hiccup after fast traveling, but I couldn't tell if it was just because of something else like loading the scenery of a big city. Anyway I think it's probably fine.

ajrb:
The only optimisation I can see is to pull out (merchantsFaction.power - regionFaction.power) / 5 out of the for loop, but saving a subtract and division per day*regions is unlikely to be noticeable... still it would make it slightly more efficient if you wanted to do that.

Allofich:
Yeah, but I think it's possible that power values might also be changed during the loop, in which case I would need to put them back in. But I may do that when I finish out the function. I'm happy for this to be merged as it is, though.

Interkarma:
Thanks mate, I'll merge as-is and let you know if any issues. It doesn't look that expensive on just a read of the code. Will let you know if it seems unusually slow when I test. :)

Interkarma:
Allofich, found an infinite loop styled bug with this one. Seems to happen when loading two saves in a row and second save exists in the past.
I've just disabled the call from from DaggerfallEntityBehaviour for now until you have a chance to look at this.
Once this is fixed, I'll also have a look at the performance and let you know if I have any feedback.

PR #467, opened 2017-09-18
Implement correct greetings based on reaction

Allofich:
This should hopefully give correct greetings based on reputation. A few other things, (masque of clavicus, something to do with guilds) in classic can affect NPC reaction to player, which isn't in yet.

The reputation modifiers with the general social groups (commoners, merchants, etc.) are imported from classic saves. They aren't saved in DF Unity yet because I want to be sure that the manner I'm storing them in the player entity is OK, which is as an array, to make it simple to access using the sgroup values from faction.txt.

### DIFF
--- a/Assets/Scripts/API/Save/CharacterRecord.cs
+++ b/Assets/Scripts/API/Save/CharacterRecord.cs
@@ -67,6 +67,11 @@ namespace DaggerfallConnect.Save
             doc.currentHealth = parsedData.currentHealth;
             doc.maxHealth = parsedData.maxHealth;
             doc.currentSpellPoints = parsedData.currentSpellPoints;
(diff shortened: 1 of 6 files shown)


Nystul-the-Magician:
wow great work! thanks for putting that in

Allofich:
No problem. Finding the reaction within PlayerActivate might not be quite how you were organizing things, but I needed to account for the "You get no response." message if you click on an NPC when their reaction to you is too negative.

Nystul-the-Magician:
should be good for now - I would love to extend the mobilenpc class anyway.
Ultimately it would be nice to have both static and mobile npcs inherit from the same parent npc class which will hold all necessary information (conversations are also possible with static npcs - this is currently not implemented but I will work on it soon)

PR #468, opened 2017-09-18
TextLabel now supports text scaling (property TextScale)

Nystul-the-Magician:
this can be used to set text scaling, e.g. 0.5 is half text size
text and word wrapping will adapt accordingly


Nystul-the-Magician:
need to fix some issues that i found

Nystul-the-Magician:
fixed the issues

ajrb:
If I commit these changes in my branch will it conflict later?

Interkarma:
This is really quite clever. Great work!

Nystul-the-Magician:
@arjb: unlikely the changes i needed to do are exclusive to the restricted area render drawing

PR #469, opened 2017-09-18
Item hover info panel in inventory

ajrb:
Enabled by a setting, disabled by default.
Thanks to Nystul for the text scaling.

### DIFF
--- a/Assets/Scripts/SettingsManager.cs
+++ b/Assets/Scripts/SettingsManager.cs
@@ -1,4 +1,4 @@
-﻿// Project:         Daggerfall Tools For Unity
+// Project:         Daggerfall Tools For Unity
 // Copyright:       Copyright (C) 2009-2016 Daggerfall Workshop
 // Web Site:        http://www.dfworkshop.net
 // License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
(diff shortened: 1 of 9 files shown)


ajrb:
Should mention I disabled potion popup info, temporarily. Will be reimplementing using the macro handler system.

Assets/Scripts/SettingsManager.cs
```diff
@@ -155,6 +156,7 @@ public void LoadSettings()
             ToolTipDelayInSeconds = GetFloat(sectionGUI, "ToolTipDelayInSeconds", 0, 10);
             ToolTipBackgroundColor = GetColor(sectionGUI, "ToolTipBackgroundColor", DaggerfallUI.DaggerfallUnityDefaultToolTipBackgroundColor);
             ToolTipTextColor = GetColor(sectionGUI, "ToolTipTextColor", DaggerfallUI.DaggerfallUnityDefaultToolTipTextColor);
+            EnableInventoryInfoPanel = GetBool(sectionGUI, "EnableInventoryInfoPanel");
```
Interkarma: Don't forget the matching SaveSettings() for this value. :)

Interkarma:
This looks good, happy to have this change. Up to you about the potion popup, I think it's OK to just let this be non-functional for a while.

Interkarma:
This works really well - good job! :)

The only refinement I'd suggest is for the info panel to clear when mouse not hovering on an item. It feels a touch strange to have the item info still displaying when mouse no longer over an item.

But otherwise feels very modern and snappy.

Assets/Scripts/SettingsManager.cs
ajrb: Is that so it can be added to the settings window?

Dimillian:
I love it, I was thinking about that ! 
For now the settings is not exposed into the launch window, also if enable I think it should disable the tooltips. And yes, it should clear up on mouse out.

PR #471, opened 2017-09-19
Talkwindow - implementation of talking to static npcs started

Nystul-the-Magician:
currently only a few portraits are in (e.g. quest giver of fighters guild of daggerfall)
portraits are matched to billboards texture archive and record index
for currently unknown combinations (almost all static npcs) the portrait image with the text "oops! ..." is chosen to make it easily recognizable that the mapping needs to be done for this npc's billboard

### DIFF
--- a/Assets/Scripts/Game/Questing/Components/SpecialNPCClickBehaviour.cs
+++ b/Assets/Scripts/Game/Questing/Components/SpecialNPCClickBehaviour.cs
@@ -32,6 +32,14 @@ namespace DaggerfallWorkshop.Game.Questing
     /// </summary>
     public class SpecialNPCClickHandler : MonoBehaviour
     {
+        StaticNPC npc; // needed for talk window setup
+
+        public StaticNPC StaticNPC
+        {
+            get { return npc; }
+            set { npc = value; }
+        }
+
         public int IndividualFactionID { get; set; }
 
         /// <summary>
@@ -45,6 +53,13 @@ namespace DaggerfallWorkshop.Game.Questing
 
             // Check active quests to see if anyone has reserved this NPC
             ulong[] questIDs = QuestMachine.Instance.GetAllActiveQuests();
+
+            // if currently not part in an active quest allow conversation with npc
+            if (questIDs.Length == 0)
+            {
+                GameManager.Instance.TalkManager.TalkToStaticNPC(npc);
+            }
+
             foreach (ulong questID in questIDs)
             {
                 // Get quest object
--- a/Assets/Scripts/Game/PlayerActivate.cs
+++ b/Assets/Scripts/Game/PlayerActivate.cs
@@ -783,7 +783,10 @@ namespace DaggerfallWorkshop.Game
             // Handle special NPC in home location click
             SpecialNPCClickHandler specialNPCClickHandler = npc.gameObject.GetComponent<SpecialNPCClickHandler>();
             if (specialNPCClickHandler)
+            {
+                specialNPCClickHandler.StaticNPC = npc;
                 specialNPCClickHandler.DoClick();
+            }
             else
             {
                 FactionFile.FactionData factionData;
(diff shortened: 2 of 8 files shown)


Allofich:
You should be able to get the faces for unique NPCs by, after getting the npc's faction, checking if the npc is of type 4 ("facINDIVIDUAL" in /arena2/faction.txt), and then getting the number face in the faces.cif file that matches the faction record's "face" value.

The faction of unique NPCs is already gotten in some way, right? Because there is that debug feature that lets you click on an npc to get your reputation with them.

Nystul-the-Magician:
will take a look, faces.cif was the next thing i will work on, but many portraits are in a different file - so will have to take a look

Nystul-the-Magician:
is there an existing way in code to get the "face" value (the one that is also in the factions.txt file)?

Nystul-the-Magician:
yeah I have found this out already.
Portrait resolving works for special npc already - looking for a way to automatically resolve common static npcs as well

Nystul-the-Magician:
@Interkarma - pls especially review the changes to SpecialNPCClickBehaviour.cs - I hope I managed to make it so that only for static npcs that are not part of an active quest the talk window will open...

Assets/Scripts/Game/Questing/Components/SpecialNPCClickBehaviour.cs
```diff
@@ -32,6 +32,14 @@ namespace DaggerfallWorkshop.Game.Questing
     /// </summary>
     public class SpecialNPCClickHandler : MonoBehaviour
     {
+        StaticNPC npc; // needed for talk window setup
+
+        public StaticNPC StaticNPC
```
Interkarma: Shouldn't need to store this here. StaticNPC is another MonoBehaviour-derived component peered with SpecialNPCClickHandler. As you already have the GameObject, just use something like:

StaticNPC npc = npcObject.GetComponent<StaticNPC>();

Or from within SpecialNPCClickHandler itself:

StaticNPC npc = this.GetComponent<StaticNPC>();

Just a heads-up, future of SpecialNPCClickHandler is a little uncertain. I'm happy with this here for now, but wouldn't want to build out this class too much. :)

Assets/Scripts/Game/Questing/Components/SpecialNPCClickBehaviour.cs
```diff
@@ -45,6 +53,13 @@ public void DoClick()
 
             // Check active quests to see if anyone has reserved this NPC
             ulong[] questIDs = QuestMachine.Instance.GetAllActiveQuests();
+
+            // if currently not part in an active quest allow conversation with npc
+            if (questIDs.Length == 0)
```
Interkarma: Yep, this should be OK. 👍

Interkarma:
Looks good so far!

I've noticed it's possible to initiate talk with NPCs injected as part of a quest to a new location. A good example is Lady Brisienna (to repro follow setmqstage 1 and get to the point where you visit her in a random inn).

One short-term strategy is to not allow talk with any NPC carrying the QuestResourceBehaviour component. This is a sure sign they've been injected by quest system.

Assets/Scripts/Game/Questing/Components/SpecialNPCClickBehaviour.cs
Nystul-the-Magician: ok, will do so ;)

Nystul-the-Magician:
should we prevent it or allow it? does it break the quest? i hope not.
shouldn't be to hard to follow your instructions if you want me to change it

Interkarma:
Doesn't break the quest, but talk window pops up first. The quest click then fires when player closes talk window.

I believe player can speak with Lady B once quest click has fired and she returns to a normal NPC. It might just be easier in short term to suppress talk on these injected NPCs until we smooth out process so talk window doesn't fire when a quest click trigger is pending. Not sure of an easy way to do that right now so just suppressing talk window when QuestResourceBehaviour present is probably easiest to begin with. We can refine from there. :)

Nystul-the-Magician:
ok thanks for clearing this up for me

Nystul-the-Magician:
actually this would be exactly what I need - an extension to QuestResourceBehaviour class that allows me to query if a quest click trigger is pending ;)
for now i will just disable talk window as suggested

Interkarma:
That's not a bad idea mate, I think that will have some legs. :)

I'm aiming to release next live build tomorrow sometime. If we can just sidestep problem for now we can work on this more once update in the wild.

Nystul-the-Magician:
ok then i will issue pull request with fixes

PR #472, opened 2017-09-19
Reimplement potion recipes and potion info

ajrb:
Also refined the info panel to wrap long lines, had to change MultiFormatTextLabel and the calculation is a bit of hack. Hope the changes don't break anything else. Would be great if someone else could look at TextLabel.LineHeight and check/fix it.

Removed old potion window class.

Added two new helper statics to TextFile for creating text and formatting tokens.


ajrb:
Can GetPotionRecipeTokens() be replaced with a new text RSC entry? Not sure if we have a mechanism yet.

Nystul-the-Magician:
is multiformattextlabel really needed here - wouldn't a normal textlabel do the job as well here. I am just asking out of curiosity. TextLabel supports text and word wrapping and is tested a bit

ajrb:
Multiformat is required because of the line breaks from the RSC text in token arrays.

Nystul-the-Magician:
ah i see

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
```diff
@@ -257,14 +258,17 @@ protected override void Setup()
             // Setup native panel background
             NativePanel.BackgroundTexture = baseTexture;
 
-            // Character portrait
-            SetupPaperdoll();
+            // Suppress tool tips if info panel is active
+            suppressTooltips = DaggerfallUnity.Settings.EnableInventoryInfoPanel;
```
Interkarma: Not sure how I feel about this. There's already an INI option to enable/disable tooltips. Shouldn't the end user be the one to decide if they want neither, either, or both?

Interkarma:
Looks like a good update, cheers mate. :)

I'd just consider letting the user decide on tooltip enable/disable based on their own preference, rather than make the decision for them.

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
ajrb: I was thinking that disabling tooltips disables them everywhere, which finding those bugs indicates that people don't really do. I didn't want to create yet another setting for tooltips in inventory on/off if info panel is being used, so I suppressed them automatically if it is.

I'm not 100% convinced it's the right approach, but they are very useful for trade window which doesn't have the info panel.

I do have an idea that the info panel could be moved to between the action buttons in between the local & remote items scrollers - then it could be on trade windows as well. (albeit slightly higher) I was thinking this because I noticed the satchel icon in the data files and started to think this could actually be used as the target icon for player inventory, and have the weight/encumbrance label over it like the wagon does. It's a real pain to not be able to see how close you are to being over-encumbered in the inventory screen.

The info panel would be quite a bit smaller this way however. I might try it and see what t looks like. Thoughts?

Interkarma: That's fine mate, thank you for extra detail. Now I see where you're coming from, I'm happy with how you've done this. Cheers. :)

ajrb: Any thoughts about moving it and using satchel icon to represent player inventory? I don't think it's actually used anywhere in classic is it?

Interkarma: All I can say is give it a go and see how it looks. For a post hoc feature it's already quite nice.

PR #473, opened 2017-09-19
Fix map windows bug with tooltips off and save souls

ajrb:
Both exterior and interior map windows locked up and broke when tooltips are turned off, so I fixed it. Just clumped all of the tooltip delay set statements in a conditional null check.

Also added the trapped soul variable set from importing classic to be saved in DFU.


Nystul-the-Magician:
thx for fixing - did not know that tooltips can be turned off globally

ajrb:
No problem, guess this means no one really plays with them off... :-)

PR #474, opened 2017-09-20
progress and fixes for talk functionality

Nystul-the-Magician:
in PlayerActivate.cs: put the hit test for quest npcs first and added return - so no need to check for attached script
changed SpecialNPCClickBehaviour.cs as suggested
added FlatsFile.cs class (flats file reader for file FLATS.CFG) and uses this information to derive correct common static npc resolving


Interkarma:
Great work all around Nystul. 👍

PR #475, opened 2017-09-20
MeshReplacement: tweaks

TheLacus:
* Avoid NullReferenceException on block/city viewer in editor
(ModManager is not present); import models from resources while in
editor instead.
* Print an error instead of throwing an exception when a
material is not assigned; do not print an error if maintexture is not
present (transparency etc.)
* Assign filtermode using sharedmaterials.


Interkarma:
I love the progress on this. It's just as valuable to the overall project as core gameplay features.

PR #476, opened 2017-09-20
Implement inventory icon and move info panel

ajrb:
- Info panel now appears on trade windows as well.
- Inventory tooltops no longer suppressed if info panel enabled.
- Backpack icon & weight displayed for carried items, i.e. player inventory.

### DIFF
--- a/Assets/Scripts/Game/Items/ItemHelper.cs
+++ b/Assets/Scripts/Game/Items/ItemHelper.cs
@@ -518,12 +518,11 @@ namespace DaggerfallWorkshop.Game.Items
         // TODO: can this be replaced with a new text RSC entry?
         private static TextFile.Token[] GetPotionRecipeTokens()
         {
-            TextFile.Token[] tokens = new TextFile.Token[5];
+            TextFile.Token[] tokens = new TextFile.Token[4];
             tokens[0] = TextFile.CreateTextToken(HardStrings.potionRecipeFor);
             tokens[1] = TextFile.CreateFormatToken(TextFile.Formatting.JustifyCenter);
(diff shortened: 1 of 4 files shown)


Interkarma:
Those screenshots on the forums looked good. Looking forward to playing with this!

Dimillian:
Was the setting in the launch window added?

ajrb:
No, I've not added that. Was waiting to see if people thought it warranted adding to settings window.

Dimillian:
Well I love it, and as it is now, non dev cannot test it. Would be cool to have the toggle for the upcoming release so any user can test it :)

I would love to do it,  but on my dev env I can only use the 2017 unity version, and it add too much changes for me to commit on master.

ajrb:
Very happy you love it. Anyone can test it once the release is out, they simply have to change a setting in the ini file. I will see if @Interkarma wants it adding to advanced settings, and if so I will add it.

TheLacus:
I probably need to add a second page to settings window, maybe separate graphics from gameplay.

ajrb:
Yes I think that would be a good idea, prob not before next release though. So would have basic, advanced (enhancements etc) and graphics settings?

Interkarma:
@TheLacus you might also be able to use the new text scaling feature to fit more on a single page. Just a suggestion though, do whatever you feel is best here. :)

Nystul-the-Magician:
currently class TextLabel is the only class that supports text scaling through property TextScale - but at least text labels can be shrinked a bit with this

ajrb:
Actually MultiFormatTextLabel also supports TextScale, mainly by passing along to TextLabel. It does have some limitations like center justification not working too well. I used it for the hover panel so it's set to work for that.

PR #477, opened 2017-09-21
Talkwindow - fixes: female faces derived from factions and faces defined in FLATS.CFG.txt

Nystul-the-Magician:
fix for common female faces derived from factions
fix for blue-cloaked woman at court wayrest using FLATS:CFG.txt in resources folder (gets loaded instead of FLATS.CFG in Arena2 folder)
one can add missing billboard to face mappings in there for npcs


Interkarma:
Good job Nystul, this lets us map the custom faces even closer than classic can. 👍

PR #478, opened 2017-09-21
Make iventory info panel enabled by default

ajrb:
Also ensured the info is cleared each time inventory accessed. Seems the on start event is not getting called.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
@@ -581,6 +581,10 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             if (remoteItemsScrollBar != null)
                 remoteItemsScrollBar.ScrollIndex = 0;
 
+            // Clear info panel
+            if (itemInfoPanelLabel != null)
+                itemInfoPanelLabel.SetText(new TextFile.Token[0]);
+
             // Refresh window
             Refresh();
         }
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
```diff
@@ -1815,6 +1761,7 @@ private void UpdateItemInfoPanel(DaggerfallUnityItem item)
             itemInfoPanelLabel.SetText(tokens);
         }
 
+        // This never gets called AFAICS, do we even need it with OnPush() working? - Hazelnut
```
Interkarma: This event is raised when a new game is started (e.g. load game, new character). The purpose is reset to reset certain global inventory state (such as the currently selected tab) that should not carry over from game to game. This is distinct from OnPush() which is used to set things up whenever window is opened. It is being called, but it's not intended for the use you had here, OnPush() was always the correct place for that. :)

Interkarma:
Couple of bits of quick feedback.

* Info text doesn't clear when player moves mouse away from an item to non-item area of UI (or empty item slot)
* Info text doesn't update when scrolling with mouse wheel (tooltips support this). This leads to different statuses between the two bits of information.

But this is looking good. I like the placement position.

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
ajrb: Yeah I get what its supposed to do now, although I put my clear info panel in there by mistake. That's how I caught that it's not being called. I put a log statement in there and after loading a new game - nothing is logged to console. Worried I broke something I checked out DFU from before I ever committed anything and the same thing so at least it wasn't me.

Not sure why it's not being called, the event works fine in other places. Just wanted to bring to your attention.

Interkarma:
I'd also consider removed the decimal places on the encumbrance so only whole numbers are shown (or maybe limit to a single decimal place). This text could also be centred inside the subpanel area. It looks a bit off-kilter left aligned. I love the satchel icon though, good use of this. :)

ajrb:
Those points are by design, although we will see how it plays. 

The info box shows the last item moused over, it should only clear when inventory is opened.
I didn't want the info message generation firing every time the mouse moves so I used the mouse enter event.

Could change it and store item & compare to see if it's a different one to avoid generation I guess. then use a different event.

Regarding the decimal places - this was copied from the wagon label used in classic and I quite like the precision considering many items are 0.25kg. It also shows gold weight as well. The alignment is also copied from classic. Looks better on wagon icon than bag, may experiment with different placements.

Interkarma:
OK, no problems on the weight text then. :)

I still feel the same about the first two points though. It just adds an unresponsive feeling for me, which is a shame because this is really quite snappy. Can you use OnMouseExit (to clear info) and OnMouseScrollUp/Down (to update info)? This will avoid needing to store items or check every mouse move.

ajrb:
Was all the same issue caused by a copy paste error in extracting the list component. It was checking the wrong thing was defined. Fix incoming...

PR #479, opened 2017-09-22
Fix info panel scroll reaction

ajrb:
Also fixed broken message & don't display for arrows


ajrb:
Shame I missed the build, but couldn't quite finish last night - had to get some sleep. ;-)

Interkarma:
Thank you for that mate. Still have to do a blog post on the weekend. I'll refresh builds when I do. :)

ajrb:
Really? I didn't think you'd make another so soon. I've been working on a major refactoring of the inventory window, extracting the item scrollers into a component. This is to:
- Reduce the ridiculous size of the inventory window class
- Allow same code to be reused twice in inv and once in item enchanter when we do that
- Enable me to create a version that displays 16 items and have it switchable by config

I wont submit PR until you release build due to the size of change.

PR #480, opened 2017-09-22
Identify gender flag from flats.cfg

Allofich:
Identified the gender flag in flats.cfg, which was marked as unknown. It's either 1 for male or 2 for female, but if it's preceded by a ?, this marks the flat to be censored (not displayed) in ChildGard mode.

I think this gender flag may be used for determining the gender of flats for quests and filling out some gender-based macros, like getting an appropriate name. I don't know if it might help you in any way for the quest system, Interkarma.


Interkarma:
Awesome, thank you Allofich! 👍

Nystul-the-Magician:
a thanks. yeah the ? was basically the unknown part in this ;)

PR #481, opened 2017-09-22
Inventory and Trade windows complete using new ItemListScroller component

ajrb:
I've been working on a major refactoring of the inventory window, extracting the item scrollers into a component. This is to:

-Reduce the ridiculous size of the inventory window class
-Allow same code to be reused twice in inv and once in item enchanter when we do that
-Enable me to create a version that displays 16 items and have it switchable by config

I wasnt going to submit PR until you released build due to the size of change - with another build so soon I think I will. Sooner it's tested by players the better I think.


Interkarma:
Cool! looking forward to playing with it soon. I will push through to a branch for testing at first.

I opted to use panel colour tinting in the UI rather than fixed bitmap tiles like classic, as it looks very similar and gives me more flexibility in colour range. Also needs slightly less memory for UI slices, not that it's much to begin with.

You shouldn't have much trouble setting the panel nesting required, as you seem to have a good mental map of the UI by now. But please don't hesitate to ask if there's anything I can help with.

ajrb:
I was thinking about the panel border stuff wondering if that would do it, but just nesting an extra panel would be much easier - thanks for the suggestion.

Pulling out the item list into it's own component was a bit of work, but it just started to come together once I got past a certain point with everything slotting into place. Its improved the code quite nicely I think.

PR #483, opened 2017-09-23
Animation fixes and adjustments for enemies

Allofich:
Fixes and adjustments for enemy animations. I'll go through everything below.

1. (not animation related) Fixed an incorrect null assignment (from my earlier PR) that made a warning in Unity.
2. Fixed rat being flipped horizontally. For some reason there was specifically code to flip the rat. Instead, the same code was used to flip the giant bat and the imp, who both needed it. There is also existing code to flip the scorpion, but I haven't tested it yet to see if that is functioning properly.
3. Fixed enemies not playing the first frame of their attack animation. currentFrame was incremented immediately in AnimateEnemy() of DaggerfallMobileUnit, so the first frame didn't show.
4. Adjusted animation speeds. Slightly sped up enemy attacks, lowered enemy movement animation and added faster animation for flying enemies. The animations should be somewhat closer to classic, and I think they look better.
5. Added code for enemies to hit on a particular frame of their attack, rather than after finishing. I don't know how classic handles this, and I don't see myself figuring it out anytime soon. Instead I went through all the enemy animations (your Daggerfall Imaging 2 software is so helpful for this stuff) and set each enemy to hit on the frame that it looks like their attack should connect on.
6. Enemies can continue to move some (1/4 speed) while attacking. I'm not 100% sure about this because the animations for enemies do look like they should be standing still and not "sliding" along the floor, but the player can move while attacking and IMO it's too easy to defeat enemies in DF Unity currently because you can outmaneuver them. Since people are making quests now I think getting game difficulty problems under control is starting to take on some more importance. Enemies can also move while attacking in classic.

I experimented with them moving at 1/2 speed, but I felt like it would maybe be too difficult for new characters in privateer's hold, and I didn't like having the enemies constantly up in my face, so for now I settled on 1/4 speed as making it more difficult but still possible to dodge hits. I would like to alleviate the "walking backwards while swinging your weapon" playstyle somehow. Maybe we can improve the AI at some point to not rush the player so much.

6. Enemies have some randomness to the time between weapon swings, to make enemies a bit less predictable and robotic.

### DIFF
--- a/Assets/Scripts/DaggerfallUnityStructs.cs
+++ b/Assets/Scripts/DaggerfallUnityStructs.cs
@@ -202,6 +202,7 @@ namespace DaggerfallWorkshop
         public bool ParrySounds;                    // Plays parry sounds when attacks against this enemy miss
         public int MapChance;                       // Chance of having a map
         public string LootTableKey;                 // Key to use when generating loot
+        public int HitFrame;                        // Frame of attack animation at which hit on target is attempted
     }
 
     /// <summary>
--- a/Assets/Scripts/Game/EnemyAttack.cs
+++ b/Assets/Scripts/Game/EnemyAttack.cs
@@ -44,7 +44,7 @@ namespace DaggerfallWorkshop.Game
         void Update()
         {
             // Handle state in progress
-            if (mobile.IsPlayingOneShot())
+            if (mobile.IsPlayingOneShot() && (mobile.LastFrameAnimated < mobile.Summary.Enemy.HitFrame))
             {
                 // Are we attacking?
                 if (mobile.IsAttacking())
@@ -54,7 +54,7 @@ namespace DaggerfallWorkshop.Game
             }
 
             // If an attack was in progress it is now complete and we can apply damage
-            if (isAttacking)
+            if (isAttacking && mobile.LastFrameAnimated == mobile.Summary.Enemy.HitFrame)
             {
                 MeleeDamage();
                 isAttacking = false;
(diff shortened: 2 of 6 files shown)


ajrb:
All sounds awesome. I did some dungeoneering with my level 2 mage char from 1996 and I could just about get by before these changes without magic... suspect that will not be the case anymore. :-)

Very curious about how this change will affect the feel of combat.

Allofich:
I removed a fix that had been put in for female thieves. It adjusts the scale of the female thief for some directions, but the fix is clearly making the sprite too large. See the attached image, the fix makes her feet go through the floor.

Interkarma, if you did this fix, can you remember what it was meant to fix?
withscaleadjust
withoutscaleadjust

Allofich:
A few more adjustments. I've played some more with it and I went back to half speed instead of quarter speed. Also it is now random whether or not an enemy will track the player's left/right movement while attacking. I had left it so they wouldn't, in order that the player can see enemies' sideways animations when they dodge to the left or right, but I realized you could just continually go in circles around an opponent and avoid getting hit, so now they might track you and hit you. I believe they always track you in classic.

Interkarma:
Thanks for all of this Allofich, looks like some great stuff in here.

The female thief texture set has (had?) too-small scaling values for some of the directional sprites. These needed to be enlarged or she would shrink (by around 1/3) in those directions. Not sure what else might have changed to correct this since I added code, but I can confirm she is now too large in those directions and happy for this code to be removed.

PR #484, opened 2017-09-23
Fixes for inventory window

ajrb:
- Items can now be colour highlighted in enhanced mode
- Fixed broken acrtions when info panel disabled

Thanks for pointing out that last one. I forgot to test that.


Dimillian:
Fix this? Got that when testing an old save with master and enhanced inventory enabled.
 
screen shot 2017-09-23 at 21 07 03

ajrb:
Yes that is what I fixed. I copied a null check and forgot to change the name to the right event. :)

Just added a fix for an armour exploit. When equipping armour to a slot which already has armour, the unequipped value was not taken off the slots armour value allowing players to increase armour by repeatedly swapping armour.

Interkarma:
Thank you mate! 👍

PR #485, opened 2017-09-23
SettingsWindow: visual improvements

TheLacus:
Reduced font size to get space for more settings. @ajrb 

@Nystul-the-Magician 
Had to add a property to ListBox because label is not publicly accesible.

### DIFF
--- a/Assets/Scripts/Game/UserInterface/ListBox.cs
+++ b/Assets/Scripts/Game/UserInterface/ListBox.cs
@@ -41,6 +41,7 @@ namespace DaggerfallWorkshop.Game.UserInterface
         bool wrapWords = false;
         int rowsDisplayed = 9;
         int rowSpacing = 1;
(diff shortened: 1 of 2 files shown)


TheLacus:
@Interkarma 
Not completely IT but i'm happy to add post processing in core if you want. The only thing that stop me is that toggles for individual effects are more options to add; also, for some effects, i included a performance/quality "slider" that merge all options provided by API and i would prefer not to lose it. 

If you're ok with this i'm going to add a postprocessing page to settings window, otherwise please add v0.2 to the released mods list. I'm fine either way :)

Nystul-the-Magician:
I usually did it like this (you can set textScale for every listbox item independently:
listbox.AddItem(listItem, out textLabelOut);
textLabelOut.textLabel.TextScale = 0.75f;

but setting it through listbox may be easier if you need same size for all, yes

PR #487, opened 2017-09-24
fixes to TextLabel and MultiFormatTextLabel - infopanel values updated

Nystul-the-Magician:
sry hazelnut for screwing up things - hope this works now good enough
LineHeight in MultiFormatTextLabel fixed


Interkarma:
Nystul, I just reverted this one. It breaks like spacing quite badly. Screenshot below and in the revert commit. Sorry mate.
line-spacing

Nystul-the-Magician:
is it really necessary to revert the pull request? isn't just rejecting also good!
every revert is causing me HUGE amount of work since i need to backmerge - if you can pls don't do it again!

Nystul-the-Magician:
sry for the inconvenience. Have overseen this. Should not work late at night even if hazelnut is needing help ;)
new pull request should work better

PR #490, opened 2017-09-25
MeshReplacement: Early implementation of StreamingWorld trees

TheLacus:
Instantiate trees from mods on streaming world. It needs some good optimization and i've yet to get the correct positions, but we're almost there.

trees


Nystul-the-Magician:
wow amazing

ajrb:
O M G

Dude, that is fantastic!

Interkarma:
I just spotted this. This has a lot of potential!

PR #492, opened 2017-09-26
Trees: fixes and tweaks

TheLacus:
It's still a bit off, don't know if due to terrain resolution or something i can fix.

treeposition 1
treeposition 2

### DIFF
--- a/Assets/Scripts/Terrain/TerrainHelper.cs
+++ b/Assets/Scripts/Terrain/TerrainHelper.cs
@@ -381,6 +381,7 @@ namespace DaggerfallWorkshop
 
             // Remove exiting billboards
             dfBillboardBatch.Clear();
+            MeshReplacement.ClearNatureGameObjects(terrain);
 
             // Seed random with terrain key
             Random.InitState(MakeTerrainKey(dfTerrain.MapPixelX, dfTerrain.MapPixelY));
--- a/Assets/Scripts/Utility/AssetInjection/MeshReplacement.cs
+++ b/Assets/Scripts/Utility/AssetInjection/MeshReplacement.cs
@@ -120,9 +120,9 @@ namespace DaggerfallWorkshop.Utility.AssetInjection
             if (!LoadGameObjectFromMods("Flats", name, out prefab))
                 return false;
 
-            // Get instance properties (TODO: position is different than original)
+            // Get instance properties
             TerrainData terrainData = terrain.terrainData;
-            Vector3 position = new Vector3(x / (float)terrainData.heightmapWidth, 0.0f, y / (float)terrainData.heightmapHeight);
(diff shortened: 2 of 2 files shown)


Interkarma:
I had a similar issue with aligning tree billboards. The secret sauce for me was to use Terrain.SampleHeight() rather than the raw terrain data itself. Unity returns a much more accurate result within the scene. Have a look at TerrainHelper.LayoutNatureBillboards() around line 460 for where I do this.

Let me know if there's anything I can do to help. :)

TheLacus:
Thanks for the tip, but i tried working on the y axis and it didn't work. Whatever value i assign i get the same result in game, only x and z affects the position. Tree is always placed correctly on terrain on height.

This 2d position on terrain plane is assigned as values in range 0-1, relative to terrain; this means that i'm calculating position as currentTile/detailResolution. Don't know if there is a way to get a more precise value than this, i'll get to do more tests and report again.

TheLacus:
Anyway, i'd like to include a WindZone. I tested it as a component of WeatherManager go and it works fine, do you think is ok there? Or you prefer i add it via script only if modding is enabled?

Interkarma:
My misunderstanding then, I thought you were having troubles with the vertical placement. I will formulate a better answer for you when I can. :)

TheLacus:
I'll keep an eye on this but i don't think is an important issue for now.
What about wind? No rush but i wonder if you missed my second comment :)

Interkarma:
I'm sorry for the delayed reply. I got a bit distracted and forgot to loop back to you.

Happy for WindZone to be part of WeatherManager if that makes things easier for you. If there's any performance impact, I only ask the component is disabled at startup if mods not enabled. :)

Interkarma:
For the layout process - I'm not sure but if you need to provide a range of 0-1, could you keep things within the same domain as LayoutNatureBillboards()? The X/Y loop is stepping through tiles relative to MapsFile.WorldMapTileDim, so u=x/(MapsFile.WorldMapTileDim-1) and v=y/(MapsFile.WorldMapTileDim-1) should come very close as the tile grid is already considered to cover the whole terrain tile edge to edge.

The terrain terrainData.detailResolution you're using in ImportNatureGameObject() is set by the terrain sampler's heightmap dimension, which by default is 129 (close, but slightly off from MapsFile.WorldMapTileDim). It's also possible for a replacement terrain sampler to use a higher resolution heightmap, which would really throw things out of kilter as the heightmap dimension would no longer be close to WorldMapTileDim.

I hope that helps. If you still get stuck, I'd be happy to try and help. I'd just need some quick steps on setting up a test environment to reproduce problem properly. :)

TheLacus:
Thanks for the detailed answer. You're right, i had the solution in front of my eyes the whole time, thank you for pointing out WorldMapTileDim :)

PR #494, opened 2017-09-27
Guild services - initial implementation (identification)

ajrb:
- Can now identify things at mages guild. (Allofich, can you check the new formula please?)
- Can now purchase random items from magic items seller in mages guild. (Allofich, please check I used formulas correctly)
- Items to be purchased have a yellow highlight.
- All mage guild & fighters guild npcs have popups now.
- Two temple npcs have popups.
- Quest popup will be assimilated at some point. (once Gav is happy)

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallGuildServicePopupWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallGuildServicePopupWindow.cs
@@ -11,8 +11,11 @@
 
 using UnityEngine;
 using DaggerfallWorkshop.Utility;
+using DaggerfallWorkshop.Game.Entity;
+using DaggerfallWorkshop.Game.Items;
 using DaggerfallWorkshop.Game.UserInterface;
 using DaggerfallConnect;
+using DaggerfallConnect.Arena2;
 
 namespace DaggerfallWorkshop.Game.UserInterfaceWindows
 {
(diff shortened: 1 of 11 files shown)


Interkarma:
Very happy to see this here. :) Will merge and test soon.

Allofich:
Sounds like good stuff! I'll try to figure out the formulas if/when I can.

ajrb:
Yeah, no real rush just whenever you can mate.

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallGuildServicePopupWindow.cs
```diff
@@ -0,0 +1,194 @@
...
+        // Fighters Guild:
+        FG_Training = 849,
+        FG_Repairs = 850,
+        FG_Quests = 851,
+
+        // Temples:
+        Make_Donation = 810,
+        Cure_Diseases = 813,
+    }
+
```
Allofich: I'm thinking of adding enums for all the faction IDs, to be accessible from anywhere, as things are called by faction ID for all kinds of things. Is it OK with you do to that and remove this section?

Allofich: Never mind, I changed my mind, and I don't think I would have needed to delete this section anyway.

ajrb: I missed this, sorry. Glad you change your mind, because I looked as how Gav was thinking of doing it, and realised how big the case statement would be if the guild and the service are separated for this. This way, the enum uses faction id tell the popup which service it is for, and the guild is a separate piece of data.

PR #496, opened 2017-09-28
Purchase UI refinements & some bugfixes

ajrb:
- Fix race bug with createRandomClothing
- Fix weight handling when purchasing
- Correct equip behaviour when purchasing
- Fix for material weight calculation. (Was causing Dwarven and Ebony weapons and armour to weigh nothing)


Interkarma:
Some good stuff in there, cheers! 👍

PR #498, opened 2017-09-30
Fix inventory window not being refreshed after trying to equip prohibited items

Allofich:
For http://forums.dfworkshop.net/viewtopic.php?f=24&t=735&p=8659#p8659

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
@@ -1010,6 +1010,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
                     messageBox.ClickAnywhereToClose = true;
                     messageBox.Show();
                 }
+                Refresh();
                 return;
             }
             // Try to equip the item, and update armour values accordingly


ajrb:
I don't think that refresh is necessary. Did you reproduce the issue, because I can't.

Allofich:
Weird, I thought I did, but now I didn't have the issue...

I'll dig around a bit more.

ajrb:
It may have been connected to the bug when dead enemies were leaving their items marked equipped? If someone was equipping from corpse and it was prohibited that would give the symptoms but would be same bug. (which is fixed)

Allofich:
Yeah, I can't reproduce now. Maybe I just thought I saw the issue because I was expecting to and didn't notice the item in my inventory. I'll close this, then.

ajrb:
I think there may be an issue with prohibited code with classic saves imported. Can't confirm because I only have saves that are hacked which have the issue. A Daedric item generated in DFU has native material of 9, but imported from classic (hacked) has a material 521 for Daedric.

This breaks the prohibited code, but not completely sure it isn't from the wierd saves. Just in case you had the info.

Allofich:
Thanks, I see the problem. Yes, the material enums are different in DF Unity and classic, and DF Unity should probably be changed to match classic.

In DF Unity, leather, chain and plate are included in the material enum, whereas in the classic saves there is the material enum (0 through 9) followed by the armor type enum (0 = leather, 1= chain, 2=plate). For daedric armor, classic saves have 09 (daedric) 02 (plate). Weapons don't have the second part, so a daedric weapon is just 09 00.

I tried to compensate for this in the prohibited item check but I shifted a number the wrong direction. I'll put up a PR.

Allofich:
> A Daedric item generated in DFU has native material of 9

Are you sure? For a weapon, that should be right, but I think DF Unity items will use the ArmorMaterialsType enum and assign ArmorMaterialsType.Daedric, which is 0x0209. If not, my prohibited armor material check probably won't work for DF Unity-generated armor.

I said the DF unity enums should probably be matched to classic but I wonder if it would cause havoc with existing items.

ajrb:
Sorry that was for a sword. Should have said. I assume prohibited materials apply to weapons as well?

Allofich:
They do. So the DF Unity you checked was a sword, but the classic item you checked was an armor piece, right?

ajrb:
Yes. Armour matches, DFU weapons don't.

Allofich:
Sorry to keep going with this, but what do you mean that DFU weapons don't match? Is there a problem that still needs to be fixed?

ajrb:
No problem about keeping going with this, I've just been quite ill with flu. There's definitely still an issue with it (I think) because it stops a char with forbidden mats of leather, iron & steel from equip daedric, but until I am fully back on my feet and can think clearly I don't have any chance of figuring out what exactly is wrong.

Allofich:
Just submitted a PR to fix it, thanks.

PR #507, opened 2017-10-05
Implement tallying of running skill

Allofich:
Implements running skill tallying.

In classic, the function to tally the running skill is called by what seems to be a constant update loop, and because of this the rate at which it is tallied varies by how fast the game is running. I did a few tests and found that in both a town and dungeon environment, in my DosBox setup (running at 30000 cycles) the tally rate is about 16 per second that the run key is held down as long as you aren't riding a horse, even if you aren't moving. I set it so that PlayerMotor.isRunning has to be true for the tally to occur.


Interkarma:
Awesome work, thank you! 👍

PR #508, opened 2017-10-06
Fixes for training

Allofich:
Fixes and adjustments to training to make it more like classic.

1. Fixed black screen background on message popups.
2. Added "You don't have enough gold" popup and prevent training if player doesn't have enough gold.
3. Fatigue loss is now defaultFatigueLoss per minute * 180 minutes, which I think is what it is in classic.
4. Number of skill tallies gained matches classic.
5. Time of last training is imported from classic saves.

Hazelnut, I think I can add in all of the trainers and their lists of offered skills pretty easily, too, unless you want to do it.

### DIFF
--- a/Assets/Scripts/API/Save/CharacterRecord.cs
+++ b/Assets/Scripts/API/Save/CharacterRecord.cs
@@ -77,6 +77,7 @@ namespace DaggerfallConnect.Save
             doc.startingLevelUpSkillSum = parsedData.startingLevelUpSkillSum;
             doc.minMetalToHit = parsedData.minMetalToHit;
             doc.armorValues = parsedData.armorValues;
+            doc.lastTimePlayerBoughtTraining = parsedData.lastTimePlayerBoughtTraining;
             doc.timeForThievesGuildLetter = parsedData.timeForThievesGuildLetter;
             doc.timeForDarkBrotherhoodLetter = parsedData.timeForDarkBrotherhoodLetter;
             doc.darkBrotherhoodRequirementTally = parsedData.darkBrotherhoodRequirementTally;
@@ -167,6 +168,9 @@ namespace DaggerfallConnect.Save
 
             reader.BaseStream.Position = 0x1fd;
             parsedData.timeStamp = reader.ReadUInt32();
+
+            reader.BaseStream.Position = 0x209;
+            parsedData.lastTimePlayerBoughtTraining = reader.ReadUInt32();
 
             reader.BaseStream.Position = 0x211;
             parsedData.timeForThievesGuildLetter = reader.ReadUInt32();
--- a/Assets/Scripts/Game/Entities/DaggerfallSkills.cs
+++ b/Assets/Scripts/Game/Entities/DaggerfallSkills.cs
@@ -368,7 +368,7 @@ namespace DaggerfallWorkshop.Game.Entity
             return GetPrimaryStat((DFCareer.Skills)index);
         }
 
-        public int GetAdvancementMultiplier(DFCareer.Skills skill)
+        public static int GetAdvancementMultiplier(DFCareer.Skills skill)
         {
             switch (skill)
             {
--- a/Assets/Scripts/Game/Entities/PlayerEntity.cs
+++ b/Assets/Scripts/Game/Entities/PlayerEntity.cs
@@ -71,10 +71,10 @@ namespace DaggerfallWorkshop.Game.Entity
         protected ushort[] priceAdjustmentByRegion = FormulaHelper.RandomRegionalPriceAdjustments();
 
         // Fatigue loss per in-game minute
-        private int DefaultFatigueLoss = 11;        // According to DF Chronicles and verified in classic
-        //private int ClimbingFatigueLoss = 22;     // According to DF Chronicles
-        private int RunningFatigueLoss = 88;        // According to DF Chronicles and verified in classic
-        //private int SwimmingFatigueLoss = 44;     // According to DF Chronicles
+        public const int DefaultFatigueLoss = 11;
+        //public const int ClimbingFatigueLoss = 22;
+        public const int RunningFatigueLoss = 88;
+        //public const int SwimmingFatigueLoss = 44;
 
         private float runningTallyTimer = 0f;
         private float runningTallyInterval = 0.0625f; // Tally every 1/16 second of running. The rate at which the running skill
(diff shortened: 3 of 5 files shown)


ajrb:
Wow, didn't expect this. I wasn't done, but not been well enough to work on it any more yet. That's fine though and I am happy to do the rest of the trainers and lists, I just wanted to be sure it was a workable scheme before doing that.

What was the black screen background on message popups problem? I didn't see that when testing.

Allofich:
For a couple messages, I think it was the one that you have spent three hours training and the one when you try to train again too soon, the background would be pitch black when the message box appeared. I was running from the editor, in Unity 5.5.0f3, in case that makes a difference.

> I am happy to do the rest of the trainers and lists
OK, I'll move onto something else then.

Probably the next thing I'll do is finish up something I was working on the other day and import region data from classic saves. It has the price adjustment data in it, which I'm currently storing as its own thing in the player entity, but I want to go ahead and bring in the whole data structure, although I don't know what all of the data in it means. Part of the data is the regional legal reputations, so I'll try to hook it up to the status info popup.

I'd like to start saving this data in Unity, so shop prices won't change when you save and load a game, but I would probably just do so as a copy of the structure used in classic. Would that be fine with you, Interkarma? Would that be a problem if you wanted to later change the data to some kind of DF Unity structure?

ajrb:
If you are wanting something to work on, there's figuring out how shelves in stores are stocked with items. Will be needed once shelves are active for purchasing.

I've done some work on identifying shelves during loading the shop assets but still want to ask Gav about how he intended the container stuff to be used for it before I start doing it and hooking it to the purchase UI. One thing I am quite puzzled about is how the shelves are stocked and when they're restocked. You can buy from a shelf and the state is persisted after entering another store and coming back. If you clear a shelf it seems to get restocked. Maybe it's confusing because I'm just looking at end behaviour.

Allofich:
> I've done some work on identifying shelves during loading the shop assets but still want to ask Gav about how he intended the container stuff to be used for it before I start doing it and hooking it to the purchase UI. One thing I am quite puzzled about is how the shelves are stocked and when they're restocked. You can buy from a shelf and the state is persisted after entering another store and coming back. If you clear a shelf it seems to get restocked. Maybe it's confusing because I'm just looking at end behaviour.

Yeah I haven't been able to figure that one out yet, either.

ajrb:
Hi, now I'm home from work I've checked the black screen issue - not sure why I didn't notice that before.

Anyway it's because I had not set a parent window correctly, so I don't think the background clear settings are necessary... I was quite puzzled why they should only be needed here.

Interkarma:
> Would that be fine with you, Interkarma? Would that be a problem if you wanted to later change the data to some kind of structure or structures unique to DF Unity?

I think using the classic data is OK here, and I'm OK with it being saved in PlayerEntity. I'll let you know if any problems with it, but can't think of any right now. :)

> I've done some work on identifying shelves during loading the shop assets but still want to ask Gav about how he intended the container stuff to be used for it before I start doing it and hooking it to the purchase UI. One thing I am quite puzzled about is how the shelves are stocked and when they're restocked. You can buy from a shelf and the state is persisted after entering another store and coming back. If you clear a shelf it seems to get restocked. Maybe it's confusing because I'm just looking at end behaviour.

Following is a rough overview on how I planned to do this before you came along. I'm happy for you to go in a different direction if you need to, but you should find some of the setup is already done if you stick roughly to the same plan. :)

The shelves should already be split out to individual models by layout classes (if not let me know). Attach a loot container to the shelves and create a loot table specific to the shop type to populate contents of shelf loot container. There's a "player owns" flag on the loot container object waiting to be used for this, but you might have to add another bit of info for shelf/shop type here. When player opens a non-owned, shelf-based loot container, just open the purchase window and use the appropriate contents for the buy interface. Tangentially this is roughly how containers can to be stocked in homes, etc. for the player to steal from.

Right now, I don't persist interior state of any buildings (Daggerfall will persist a few interiors for a while - until player leaves world cell I think). I'm perfectly happy for shelves to restock when player exits and enters shops again. Container persistence is mostly unrelated to the buying interface and can be built later.

Let me know how you go, I'm happy to field any questions or help if I can. :)

Interkarma:
Oh, and thank you for the PR Allofich! 👍

Allofich:
`Anyway it's because I had not set a parent window correctly, so I don't think the background clear settings are necessary... I was quite puzzled why they should only be needed here. I will wait for this PR to be merged and submit mine changing it to set a parent window.`

Another place where it happens is the "Do you want to want to access your wagon" message when you click dungeon exits.

Allofich:
> If you are wanting something to work on, there's figuring out how shelves in stores are stocked with items. Will be needed once shelves are active for purchasing.

I'm in a position now to help out with this. I've largely figured out the code for stocking the shelves, which happens when you click on a shelf.

First we need to add code for the player activating a shelf. If the clicked object's TextureArchive is 71 or 72, and if the texture record is less than 4 but not 3 in TextureArchive 72 (looks like the texture for a safe. Might be unused in-game), then it's a shelf.

I'm not quite sure what all is involved in setting up a new object to test for raycasts against. Interkarma would probably know quickly how to do that. 

Once we have clicking on the object and it ready to hold an item collection, the next thing classic does is differentiate between a store shelf and a bookshelf. This depends on the building type.

If the building type is Bookseller, Library, Guild or Temple, it's a bookshelf. Guild and Temple should require you to be a member to access.
If the building type is HouseForSale, FurnitureStore, or anything above Temple (14), I think maybe you just can't click it. I'll try to test it in classic. There seem to be equipment lists for these, though, seen below.
Otherwise it's a store shelf.

Populating the store shelves is a bit complicated to write out here, but I can fill it out if there's a framework for adding items in place. All I need is the building data (building type and shop quality).

I don't know yet about when shelves are restocked. From testing classic all I know that until you click a shelf, it's contents aren't set and can change between save/load. Once you've clicked it, the contents are set and will be the same if you then reload the game.

Interkarma:
Something else to keep in mind is the mod system - shelves are likely to be replaced by other models. Rather than identify by texture as in classic, it should be equally workable to identify shelf by model ID at layout time.

The shelf needs to be identified before the model replacement takes place, and the loot container attached to resultant object (classic or replacement) after model injection has run.

I don't mind if shelf contents are generated when player enters store (layout time) or when player first clicks on shelf. End result for the player is effectively the same. Loot containers are already serializable, just need to make sure the containers are registered and they should load/save with minimal effort. But might just need a little support work for 3D objects carrying linked containers.

It would be nice for save games to keep the last few interiors player has visited (for shelves to persist, etc.) but this can definitely be added later.

Great work guys, I look forward to seeing your progress. Please just let me know if anything I can do to help support this. :)

Allofich:
> Something else to keep in mind is the mod system - shelves are likely to be replaced by other models. Rather than identify by texture as in classic, it should be equally workable to identify shelf by model ID at layout time.

Anything that works. I'm not sure how model replacement works, but even if model replacement was used, does the original data read from the .rdb (or whatever is used for shop interiors) include the texture archive/record definitions? If so, identifying by texture archive and record would still work, right?

Also, I just checked classic, and a shelf in a house I tried did nothing when clicked on, which confirms what I thought.

I don't think I'm going to try to set up the shelves to be clicked on or have an associated loot container myself. I don't really know how to go about that and I'm lazily hoping that Hazelnut or you Interkarma will do it. :)

Interkarma:
Happy to help. :) If you guys can wait until the weekend, I can wire up some basics for you to build on as needed.

ajrb:
Moved this to core devs forum so I can pick it back up when I am able. Thanks guys.

PR #509, opened 2017-10-07
Import and save region data structures from classic

Allofich:
Imports and saves region data structures from classic. A lot of the data is still unknown, but I've been able to determine the data size of them. I think a lot of them are flags for various region events, things like a ruler dying or some group becoming a public enemy, the stuff you see in rumors or on the signboards at city entrances.

Right now, the game play benefit is that regional price variations are saved and loaded, and so will no longer be re-randomized every time you open a DF Unity save.

The region data includes the legal reputation per region. I went ahead and hooked this up to the status pop-up (and fixed a problem that caused all legal reputations under 0 to be shown as "undependable"), but I see that some groundwork has already been laid in place for storing legal reputation as a part of the persistent faction data, rather than in the regional data. Interkarma, I know you're busy with the quest stuff, but can you take a look and decide how to proceed here? We could just use the legal reputation in the region data records like classic does, and it will keep things simple. But maybe you have plans for managing the legal reputation from within the persistant faction data? In that case, should all the regional data stuff (just price adjustments, currently) be in there, too?

### DIFF
--- a/Assets/Scripts/API/Save/SaveVars.cs
+++ b/Assets/Scripts/API/Save/SaveVars.cs
@@ -16,6 +16,7 @@ using System.Collections;
 using System.Collections.Generic;
 using DaggerfallConnect.Arena2;
 using DaggerfallConnect.Utility;
+using DaggerfallWorkshop.Game.Entity;
 
 namespace DaggerfallConnect.Save
 {
@@ -47,7 +48,8 @@ namespace DaggerfallConnect.Save
         const int lastSkillCheckTimeOffset = 0x179A;
 
         const int regionDataOffset = 0x3DA;
-        const int regionDataLength = 4960;
+        const int regionDataLength = 80;
+        const int regionCount = 62;
 
         const int factionDataOffset = 0x17D0;
         const int factionDataLength = 92;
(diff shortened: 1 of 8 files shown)


Allofich:
Added in quest gold reward calculation from classic. We need faction information to complete it (faction's power level and player's rank), which doesn't seem to be available yet.

ajrb:
Haha, that was really dumb of me on the rep conditionals - what a rookie error. Thanks for fixing it.

Allofich:
No problem. :) I didn't notice it at first glance, and was wondering why a "scoundrel" character from classic was being called "undependable" in DF Unity. Anyway, fixed now.

Interkarma:
>Interkarma, I know you're busy with the quest stuff, but can you take a look and decide how to proceed here? We could just use the legal reputation in the region data records like classic does, and it will keep things simple. But maybe you have plans for managing the legal reputation from within the persistant faction data? In that case, should all the regional data stuff (just price adjustments, currently) be in there, too?

I just stubbed legal reputation so I had data to receive changes from "legal repute" quest action. I don't have any further plans with it right now. If you'd like to move into the region data records, that makes sense and is completely OK with me.

Interkarma:
Thank you for all your work on this. 👍

PR #510, opened 2017-10-07
Development of guild training

ajrb:
Also fixes for
- Library booksellers
- Sale of religious items
- Black background on exit dungeon message


Interkarma:
Cheers mate, looking forward to taking this for a spin. 👍

PR #511, opened 2017-10-07
Fix classic save items being wrongly put in wagon

Allofich:
Fixes http://forums.dfworkshop.net/viewtopic.php?f=24&t=371.


Interkarma:
Interesting approach, I did not think to approach the problem this way. Clever. :)

PR #512, opened 2017-10-07
Fix prohibited armor material check

Allofich:
Fixes prohibited armor material checking. My earlier fix attempt by changing the shift direction was incorrect. First, I wasn't thinking straight -- if you have two bytes and want to shave off the right one and just use the left one, you can shift to the right, but if you do the same thing in the other direction to try to use only the right byte, you would need to again shift back after shaving off the left byte or the number would be too large.

Second, I didn't pay attention to the fact that the DF Unity material enum is a 4-byte int, so my shift didn't even shave off the left bit, it just made the number bigger.

For the fix this time I just cancel out everything but the rightmost byte. Tested on a character with prohibited leather, iron and steel equipping daedric armor (which Hazelnut said was wrongly prohibited from equipping), and it works now.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
@@ -988,9 +988,10 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
                 // Check for prohibited armor type (leather, chain or plate)
                 else if ((1 << (item.NativeMaterialValue >> 8) & (int)playerEntity.Career.ForbiddenArmors) != 0)
                     prohibited = true;
+
                 // Check for prohibited material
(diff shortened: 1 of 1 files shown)


ajrb:
That looks correct logic, however I was wondering whether these checks should be using the methods in DFCareer like IsMaterialForbidden() etc?

Allofich:
I don't know. What's better

`if (item.IsShield && ((1 << (item.TemplateIndex - (int)Armor.Buckler) & (int)playerEntity.Career.ForbiddenShields) != 0))`

or

`if (item.IsShield && playerEntity.Career.IsShieldForbidden((DFCareer.ShieldFlags)(1 << (item.TemplateIndex - (int)Armor.Buckler))))`

If it's up to me I'd just leave it as it is since it's a little shorter and exposes what's happening a bit more without having to look at the called function.

ajrb:
I think I agree with you. It doesn't make it make any more readable.

PR #514, opened 2017-10-09
Fix for clicking on other regions from Cybiades

Allofich:
This is a very manually, trial-and-error type of fix (found that upper right corner of Cybiades in its zoomed-in view and the less zoomed-in view from Sentinel seemed to match, and went from there), but the results seem to work. Fixes http://forums.dfworkshop.net/viewtopic.php?f=24&t=764.


Interkarma:
Thanks for taking a look at that. :)

PR #516, opened 2017-10-11
Reorganize texture fix

Allofich:
Reorganized the texture fix from the other day to make it fit in better with the existing code. I put it where previously doors were being set to not support winter. I also removed setting winter support to false since it is false by default and doesn't need to be set to false, as is done for the interior textures.


Allofich:
Never mind, I don't like that I added a function parameter for just this one case and that will usually be passing an unused value.

PR #518, opened 2017-10-13
Implement classic travel time calculations

Allofich:
Fixes http://forums.dfworkshop.net/viewtopic.php?f=24&t=770.

Travel and cost calculations are reversed from classic. Travel that includes travel over ocean is slightly off due to TerrainHelper.DilateCoastalClimate changing the values read from CLIMATE.PAK. It's not a huge deal but we're so close to having the correct values from classic it would be nice to fix. What do you think, maybe keep a second, unaltered copy of the CLIMATE.PAK buffer for using in travel calculations? Or adjust the DilateCoastalClimate function so it doesn't alter the buffer?

Also, ocean-traveling costs are free currently because all players have ships. My suggestion is we disable ships until buying them is implemented, so that we can test for now with normal travel costs.

The call to the time calculation function from Clock.cs is using the same parameters as before. I don't know how quests are affected.


Interkarma:
This looks excellent, thank you! Will review and merge soon. 

I don't want to remove ship as a transport target just yet as it might be a while before this is implemented in banks completely. But I'm happy for ship to be ignored for the purpose of calculations for now so we can test the basic costs.

I will take another look at the coastal dilation and store and unmodified buffer. I didn't consider the impact on travel calculations when I wrote this.

Interkarma:
This works great! :)

PR #521, opened 2017-10-14
Implement castle guards going hostile

Allofich:
This implements castle guards going hostile. The main thing I was aiming for was finishing up the "DoorText" action I worked on earlier. Guards will go hostile now when you open the doors to restricted areas in the castles, like in classic. Also, guards will go hostile when you attack any one of them that was passive. All guards in the area will be set to hostile. I think in classic it might just be those that are within loaded range or something, but this behavior seems fine to me, rather than potentially having some guards standing around and ignoring the fight.

As for the DoorText trespassing action, the classic behavior is emulated, but it is somewhat broken IMO.
- The descriptive messages that warns you about areas being off limits are only seen if you click the doors in info mode, so they are easy to miss.
- If you click the door in info mode, it leads to an exploit, because the door opens but the trespassing check doesn't trigger.

You know I in general want to copy classic, but I think broken game mechanics like this should be fixed. What do you think about doing these:
- Always show the descriptive messages. This makes sense since they include things like guards issuing you warnings not to enter.
- On first click, the message is shown. After that, regardless of interaction mode, the door will be opened, and trespassing check performed.

I've gathered over the time watching your work, Interkarma, that you like avoiding extra clicks wherever possible, so maybe you won't like that second suggestion, since it would mean two clicks to open these doors. These doors aren't very common, but if you have another suggestion, maybe we could do that.

### DIFF
--- a/Assets/Scripts/Game/GameManager.cs
+++ b/Assets/Scripts/Game/GameManager.cs
@@ -497,6 +497,26 @@ namespace DaggerfallWorkshop.Game
             return areEnemiesNearby;
         }
 
+        /// <summary>
+        /// Make all enemies in an area go hostile.
+        /// </summary>
+        public void MakeEnemiesHostile()
+        {
+            DaggerfallEntityBehaviour[] entityBehaviours = FindObjectsOfType<DaggerfallEntityBehaviour>();
+            for (int i = 0; i < entityBehaviours.Length; i++)
+            {
+                DaggerfallEntityBehaviour entityBehaviour = entityBehaviours[i];
+                if (entityBehaviour.EntityType == EntityTypes.EnemyMonster || entityBehaviour.EntityType == EntityTypes.EnemyClass)
+                {
+                    EnemyMotor enemyMotor = entityBehaviour.GetComponent<EnemyMotor>();
+                    if (enemyMotor)
+                    {
+                        enemyMotor.IsHostile = true;
+                    }
+                }
+            }
+        }
+
         #endregion
 
         #region Public Static Methods
--- a/Assets/Scripts/Game/PlayerActivate.cs
+++ b/Assets/Scripts/Game/PlayerActivate.cs
@@ -448,7 +448,7 @@ namespace DaggerfallWorkshop.Game
                                             break;
                                         }
                                         enemyEntity.PickpocketByPlayerAttempted = true;
-                                        Pickpocket(enemyEntity);
+                                        Pickpocket(mobileEnemyBehaviour);
                                     }
                                     break;
                             }
@@ -895,14 +895,17 @@ namespace DaggerfallWorkshop.Game
         }
 
         // Player has clicked on a pickpocket target in steal mode
-        void Pickpocket(EnemyEntity target = null)
+        void Pickpocket(DaggerfallEntityBehaviour target = null)
         {
             const int foundNothingValuableTextId = 8999;
 
             PlayerEntity player = GameManager.Instance.PlayerEntity;
+            EnemyEntity enemyEntity = null;
+            if (target != null)
+                enemyEntity = target.Entity as EnemyEntity;
             player.TallySkill(DFCareer.Skills.Pickpocket, 1);
 
-            int chance = Formulas.FormulaHelper.CalculatePickpocketingChance(player, target);
+            int chance = Formulas.FormulaHelper.CalculatePickpocketingChance(player, enemyEntity);
 
             if (UnityEngine.Random.Range(0, 101) <= chance)
             {
@@ -934,6 +937,19 @@ namespace DaggerfallWorkshop.Game
             {
                 string notSuccessfulMessage = HardStrings.youAreNotSuccessful;
                 DaggerfallUI.Instance.PopupMessage(notSuccessfulMessage);
+
+                // Make enemies in an area aggressive if player failed to pickpocket a non-hostile one.
+                EnemyMotor enemyMotor = null;
+                if (target != null)
+                    enemyMotor = target.GetComponent<EnemyMotor>();
+                if (enemyMotor)
+                {
+                    if (!enemyMotor.IsHostile)
+                    {
+                        GameManager.Instance.MakeEnemiesHostile();
+                    }
+                    enemyMotor.MakeEnemyHostileToPlayer(gameObject);
+                }
             }
         }       
     }
--- a/Assets/Scripts/Game/WeaponManager.cs
+++ b/Assets/Scripts/Game/WeaponManager.cs
@@ -652,6 +652,11 @@ namespace DaggerfallWorkshop.Game
                         EnemyMotor enemyMotor = hit.transform.GetComponent<EnemyMotor>();
                         if (enemyMotor)
                         {
+                            // Make enemies in an area aggressive if player attacked a non-hostile one.
+                            if (!enemyMotor.IsHostile)
+                            {
+                                GameManager.Instance.MakeEnemiesHostile();
+                            }
                             enemyMotor.MakeEnemyHostileToPlayer(gameObject);
                         }
                     }
--- a/Assets/Scripts/Internal/DaggerfallAction.cs
+++ b/Assets/Scripts/Internal/DaggerfallAction.cs
@@ -737,7 +737,6 @@ namespace DaggerfallWorkshop
         /// <summary>
         /// 32
         /// Shows text at the top of the screen when player clicks on associated door in info mode.
-        /// TODO: Can cause castle guards to go hostile if player clicked door outside of info mode.
         /// </summary>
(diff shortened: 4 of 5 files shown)


Interkarma:
I like your solution. You're right these doors aren't common, and the player can land in hot water too easily with classic behaviour, so a couple of chances starting with showing the info text is a pretty great idea.

If you don't mind working it up, I'd love to see how your idea plays out. :)

Interkarma:
Will get this one tested and merged to master soon. :)

Allofich:
OK I implemented like I said. It took a lot longer than I thought, because I had to figure out how to get it working with all the cases. I think it's all working now. I also removed the Orsinium replacement text I'd put in and my guesses for where some unused text was used, because I seem to have found out. There's a few doors in the Daggerfall throne room that have unused text defined for them, and the Orsinium one seems to be set up in a floor trigger when you enter. We could try to get those working, but it might require even more bending over backwards to patch around all the different cases. And maybe they were intentionally not used, who knows.

I also added guards going hostile from door bashing in the three castle throne room areas, and guards going hostile after a failed pickpocket on a non-hostile one.

Interkarma:
Something I'm not sure about - how does the quest system inform world player is supposed to enter a restricted area, such as to visit Mynisera as part of main quest chain in Sx013?

I just tested in classic and the player is definitely set upon by guards when entering Mynisera's chambers, you've emulated this behaviour perfectly. But I don't recall having this issue when player was supposed to visit a particular questor, and there's nothing in the quest scripts to explicitly make guards docile.

Do you have any ideas here? If there's nothing obvious, it might be worth relaxing the penalty for opening restricted doors until main quest is fully tested. I'd rather not have players being punished for going where the game tells them to at the moment.

But excellent work mate. 👍  I like the immediate feedback on the doors when clicking on them - it might be nice to increase the text delay on those messages a little though so that slower readers have more of a chance to parse the information.

Interkarma:
I've just had a longer play with classic and can confirm the guards will attack player no matter what when entering Mynisera's quarters, even when she is involved in quests. I guess that means this is feature perfect to classic, and we don't need to look for something that should be mitigating guard hostility.

It still seems rather unfair player is punished for doing that the game wants them to do, though. However, I feel its best to wait for feedback and see how others feel once they get the changes in next build. Daggerfall is a harsh game by nature and I don't want to sand down too many of those sharp edges.

Allofich:
Yeah, I know about that. I felt the guards attacking in that part was a bit strange when I played through the main quest in classic.

The trespass/door text action seems very rough, even for Daggerfall, with unused messages, invalid text IDs, the only way to see them (info mode) also being an exploit to avoid the trespassing check, etc. I wonder what the story behind it is. There's this unused message in TEXT.RSC in the section of texts used for this action:

> The door restrains an unnatural silence,ü devoid of even the chittering of smallü insects. You hear Popudax say  Our storeroomü is only open to those whom I trust.

This suggests to me that there might have been an intention that these off-limits doors could change state and become available, but I'm not aware that this ever actually happens in classic. This also seems to be supported by the guard at the entrance to the castle saying "Do not stray from the audience chamber lest ye have royal blessing to do so," and the fact that there are unique characters (like Popudax) in the side-rooms that can only be met by trespassing in there.

I also thought maybe the guards attacking you for going in Mynisera's room could make sense, since Mynisera and Aub-ki are at odds and maybe the guards are attacking you for trying to talk to Mynisera. But, then, if that was supposed to be the case it seems like it would be mentioned or hinted at in some way, which as far as I know it is not. Also, in Mynisera's Letters they say Mynisera leaves to go to that dungeon with the letters, and she's actually not around for a while, so apparently she's not being kept prisoner in her room or anything. If I had to guess, I'd say the guards attacking you when you go in Mynisera's room is probably an oversight. 

I think that maybe the best thing to do for now is just hard-code the message for Mynisera's door to not trigger the trespass check. If it would be useful for custom quests, maybe a new quest command that removes the trespass check on doors could be added. That could allow for peacefully accessing the side-rooms in Daggerfall castle, for example. It could also be added to the main quest to make the door to Mynisera available. Going by later TES games, I'd say access to Mynisera would probably stay forever once it was granted, even when the quests with her were done.

Interkarma:
That's a great write-up, thanks for all that. I agree it seems very rough and probably part of an unimplemented feature.

A new quest action would be nice, but probably easier just to leave it as is or hard-code out trespass check for Mynisera's door. Whatever you decide to do is OK with me. You've captured classic perfectly as far as I'm concerned. Anything else from here is just icing on top. :)

Dimillian:
From a purely gameplay perspective, a role-player could understand that as he try to access a private room, and that a guard, unaware of your duties might attack you because his duty his to protect the castle at all cost. 

However, I believe that for a better gameplay and sense, we might want to fix those classic "rushed" mechanisms, and actually not guard those doors during quests. We should also make any illegal action (like killing guard) affect your reputations within the castle. So guard could catch you inside the castle, send you to jail etc...

Allofich:
I took a quick look over the quests around the time you meet Mynisera. It doesn't seem like you're ever formally invited to see her (no letter or anything), you just go to her yourself after meeting Cyndassa. So there might not be a logical point for the trespass check to disappear. But Mynisera's greeting when you first meet her, "Hello %pcn. I have heard much about you. Now you come to me with a very interesting tale of deception..."  is too calm and doesn't seem right if you are/were just in a life-or-death struggle with the guards... Well, I'll sleep on it, but at the moment I'm still thinking it makes more sense to remove the guards attacking the player for this door.

Allofich:
> We should also make any illegal action (like killing guard) affect your reputations within the castle. So guard could catch you inside the castle, send you to jail etc...

Yes, I also think those features would be nice to see in some form (I don't know if they should be optional mods or not) at some point.

Dimillian:
IMO, it looks like a bug in classic, so it depends to where Interkarma want to go for the V1. Classic experience is classic experience :) Maybe we'll have a "Daggerfall Unofficial patch" as a mod which fix all the classic quirks such as this one.

Allofich:
It's yet more complicated. I disabled the trespass check for Mynisera's door but was still attacked when I entered the room. Something in there, the floor or an invisible wall or something, also has this action on it, just with no text message defined. This also triggers in classic, and the info mode exploit applies for it as well. This additional check doesn't seem to have a function in Mynisera's room, since you have to trigger the door's trespass check to even get in the room... I don't think there's anywhere in the game where you can trespass without opening a door first. So I tried requiring the check be on a door, but then the side-rooms from the main room no longer caused trespassing. The doors to them don't have the action on them, so for them the floor/invisible wall or whatever it is what handles the trespass check.

Different options I can think of are:
1) Just leave it as it is. This is what I think I'm going to do.
2) Disable the Mynisera door check and the floor checks.
3) Disable the Mynisera door check and add a new check to make guards hostile if you pick a lock, so you'll be attacked for the locked side rooms. (You're already attacked if you try to bash the doors)

I think it's good that if you're not involved in a quest that you're attacked for entering Mynisera's room, so that's a plus for not disabling that check entirely. The note that it is marked "as the private chambers of a member of the royal family" suggests you shouldn't be intruding in there.  So, I think I'm just going to move on to other stuff for now.

Interkarma:
Completely agree mate, don't spend any more time on this. The behaviour is obviously intended in the classic data. I'll explain this if anyone even brings it up. Thankd for taking a look at it. :)

Interkarma:
I thought of a way to handle this that turns the problem back into gameplay. Here's the overview on what could happen when player triggers the "make guards hostile" action either through opening doors on stepping on forbidden parts of the castle:

* Find all enemies within a certain range (like the rest check). In the context of this action, enemies are likely be guards, but this could be more specific as well.
* Perform a quick ray check (based on enemy facing and line of sight) to see if any guards in range can actually see the player.
* Perform a local stealth check of some kind.

If no guards are in range or can't see the player OR player passes a stealth check (maybe a roll is required if guards can see player) then don't trigger the action. The player should be safe until they next step on forbidden areas. It's like the don't stand in lava game with stealth.

In the context of Mynisera, there's a giant pillar they can hide behind while opening door, which effectively gives even low-stealth characters an in-game sneak option with the above handling.

We don't need to do anything with this right now, but I like the basic idea. Any feedback is welcome. :)

Allofich:
Sounds like a decent idea. With the current setup, though, I think the player might never trigger the guards for Mynisera's door, though, because enemies seem to have a pretty narrow field of view for noticing the player. The guard over by the side-room is looking sideways from the player so he might never notice.

Interkarma:
It's true, the north-by-default facing makes it a bit hard without practicality giving guards a huge fov. Could always just ignore fov, but that wouldn't give player a sporting chance elsewhere.

Anyway, just throwing ideas out there. I'm happy with things working like classic, you've done a great job with all this.

PR #523, opened 2017-10-20
Modsettings config

TheLacus:
* Create presets from inspector. Properties are cloned from main config file for creation and changes sync.
* MultipleChoice and Color support for inspector config.


Interkarma:
Thank you for your continued work on this 👍

PR #524, opened 2017-10-21
Get door lock value like classic

Allofich:
Gets the lock value for doors in the same manner as classic.

For regular locks the value will be the same, but apparently magical doors can have a few different values, which wouldn't be correct before this PR. Maybe those values matter when using lock-opening spells.

### DIFF
--- a/Assets/Scripts/API/DFBlock.cs
+++ b/Assets/Scripts/API/DFBlock.cs
@@ -962,7 +962,7 @@ namespace DaggerfallConnect
             /// <summary>Index into ModelReferenceList array.</summary>
             public UInt16 ModelIndex;
 
-            /// <summary>Unknown.</summary>
+            /// <summary>Trigger flag and starting lock for doors.</summary>
             internal UInt32 TriggerFlag_StartingLock;
 
             /// <summary>ID of sound to play when action is executed. Also used for spell & text index.</summary>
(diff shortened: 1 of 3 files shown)


Allofich:
Fixed a problem with the Castle Daggerfall entrance doors. They were being found as a "DaggerfallActionDoorSpecial" and not being unlocked, so if you closed them they locked again, and you could lock yourself in the castle.

I don't know much about the "DaggerfallActionDoorSpecial" so this fix (checking for a regular door first) might interfere with what you used the special door for.

Allofich:
Although I haven't seen a problem with the CloseDoor action (where is it used?) I imagine it would have the same problem of being detected as a special door and not being locked, so I put in the same fix there.

Interkarma:
Interesting, thank you for the fix. The "DaggerfallActionDoorSpecial" component is for doors that are opened by an action, but do not otherwise have the DaggerfallActionDoor component attached. A good example are false bits of pillars in Lysandus' Tomb which are not actually doors but still need to receive an "open" action. I created a new component just for these objects rather than refactor the entire door/lock setup. I'll need to see why the "DaggerfallActionDoorSpecial" was applied to these doors, that will be an error of mine.

It looks like your fix should do the job. Will test and let you know if any issues.

Interkarma:
This one breaks the "DaggerfallActionDoorSpecial" usage, unfortunately. Rather than revert, I'll see if I can resolve issue from PR.

PR #526, opened 2017-10-22
Emulate classic selection of dungeon textures

Allofich:
I've got, hopefully, the correct selection of textures emulated on classic. Thanks for putting in the recreation of the Watcom random functions. That and also notes from the blog on DaggerXL helped me get this one together.

I'm not completely sure about the sewer textures, but I expect they are correct. At any rate
http://forums.dfworkshop.net/viewtopic.php?f=24&t=739, which was a problem with sewer textures being selected out of range, should be fixed,

Some reorganization should probably be done at some point. The same array that is used for getting indices in the travel calculator turned out to be used for the dungeon textures as well, but it's kind of weird to be using a value from the travel calculator class for dungeon textures. Also the name "StaticTextureTables.cs" is now a misnomer since I went ahead and removed the pre-defined texture tables.


Allofich:
Went ahead and renamed StaticTextureTables

Nystul-the-Magician:
oh nooooo - like the dfunity variety - now it will be boring same texture all the time - can we put in old behaviour via ini config pls

Allofich:
An .ini option sounds reasonable, in which case I guess we would want the static texture tables back for the old behavior. The previous texture selection should be changed to treat sewer textures separately from the other textures, though, because they have more records than the other archives and can't have non-sewer ones simply substituted in for them. That was causing the bug I linked to.

Interkarma:
Oh man, this is incredible! You have no idea how crazy this used to drive me. I had part of the solution but never the whole thing. Thank you for doing this, I'm looking forward to checking it out.

Nystul, the dungeons in DFUnity were seeded by their ID and would have the same textures each time you visited them. I just hand-matched the textures for each special dungeon locations so they matched classic. It should be the same here, just the seed is correct. Or was DFUnity previously throwing a wider range in its random selection? I spend so much time in this version, it's hard to remember sometimes. :)

Interkarma:
This works great as far as I can tell. Fantastic effort! 👍

Nystul-the-Magician:
I just remember the brown-yellow ugly texture in vanilla all the time. There literally was no other textures used with a few exception or if you went to a different climate.
dfunity seemed to use more randomness - at least it seemed to me as it would ;)

Allofich:
> Or was DFUnity previously throwing a wider range in its random selection?

As far as I can see, yes, because it would pick randomly from all the available texture sets. This includes mixing different climate textures together or mixing sewer and non-sewer textures, which classic doesn't do.

The classic algorithm is pretty odd, though, and I want to create a fixed version of it. It seems classic never uses the swamp or mountain texture sets, just the desert and woodland ones. The usage of the { 0, 0, 0, 1, 2, 3, 4, 5, 5, 5 } array to get indices to  { 0, 0, 1, 4, 4, 0, 3, 3, 3, 0 } is strange. As you can see, only up to index 5 can be called. Also, if you compare  { 0, 0, 1, 4, 4, 0, 3, 3, 3, 0 } to the climate enums, you'll see this matches up to them and assigns desert textures to desert, swamp to swamp, etc. There is one "3" (swamp) for "mountain2", but other than that it matches.

As classic does it, the textures don't match their supposed climates (as labeled in the TEXTURE.??? files).
No idea why classic does what it does, and I wonder if the usage of the  { 0, 0, 0, 1, 2, 3, 4, 5, 5, 5 } array, which is also used in the travel calculations, was some sort of mistake? They both have to do with climate/terrain, so maybe it was accidentally used because of its name or something?

I want to experiment with assigning textures sets based on climate as it seems like they should be, together with the classic algorithm for picking variants among them, and have this as a fixed version. Maybe we can have the original classic, the fixed version, and maybe the old version too for the most completely shuffled textures. I like the idea of texture "themes" by climate, so I'm partial to the "fixed" version (and also because it's my idea :)), although you the player might not see much variety if they only go to dungeons within a particular region or two.

What do you guys think? Also, assuming we have all 3 options
1) What should be the default?
2) Should main story dungeons use the classic formula in all 3 cases?

Allofich:
Also, I should have named "textureRecordOffset" as "textureArchiveOffset." I got confused while I was writing that part. I'll fix that.

Allofich:
I got mixed up while writing my above post, and edited out that part.

It looks like the { 0, 0, 1, 4, 4, 0, 3, 3, 3, 0 } array matches up to the "ClimateType" settings in MapsFile.cs, so I'm thinking of applying the dungeon climate textures according to these values. (0 being desert textures, 1 being mountain, etc.)

Interkarma:
I'm happy for dungeons to have a wider range of textures if that's not too hard to work in. I'd still like key dungeons to match classic though (the ones I statically mapped earlier).

Allofich:
This is still somewhat inaccurate. The Mantellan Crux textures are off because it's an ocean square but in that case classic sets it to a swamp climate index. Also sewer textures are off. I'll fix these and do some clean up before moving on to adding any new options.

PR #529, opened 2017-10-25
Talkwindow - progress and new actions implemented

Nystul-the-Magician:
new stuff:

- defaults.ini  entry EnableModernConversationStyleInTalkWindow can be set to true for improved conversation/dialog style
- added actions "dialog link" and "add dialog": first implementation pass (there will be cases where it does not work as vanilla, but we need a start)
- introduced TalkManagerMCP partial class for correct macro resolving for different quest topics
- save serialization of quest talk topics
- improvements with textlabel positioning
- answers do use message "envelope" like in vanilla
- answers use npc social group now to select answer "envelope"


Interkarma:
Wow huge update! Looks good, but more than I can take in on a read alone. I'll merge to master and let you know if any problems found.

Thanks for your continued work on this. It's incredible to see a feature not even technically scheduled until next release cycle to grow before my eyes. :)

PR #530, opened 2017-10-25
Modsettings

TheLacus:
* Implemented MultipleChoices as sliders (take up a single line, not affected by number of items)
 Something like [this](http://support.amd.com/PublishingImages/Support/KB/Witcher4.png). Basically hides the index with the corresponding string set in scriptableobject config, no more need to write all possibilities in tooltip :)
* Organization in settingswindow and reader classes
* Added method Deserialize() in ModSettings which parse known settings values (int, float, tuple, color etc., the same that can be chosen in SO inspector to show in window) of a section in a class at once.

### DIFF
--- a/Assets/Game/Addons/ModSupport/ModSettings/ModSettings.cs
+++ b/Assets/Game/Addons/ModSupport/ModSettings/ModSettings.cs
@@ -10,10 +10,9 @@
 //
 
 using System;
-using System.IO;
 using System.Globalization;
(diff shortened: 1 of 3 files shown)


Interkarma:
I don't say it enough - the modding in Daggerfall Unity wouldn't be anywhere near as advanced without you. :)

Nystul-the-Magician:
great stuff!

PR #532, opened 2017-10-27
Optionally show material for helms/shields

Allofich:
Addresses request in http://forums.dfworkshop.net/viewtopic.php?f=4&p=9231&sid=a0d0a83058d2068e130a0061395aacb0#p9231.

Possible values are:
0 - Don't show material (like classic). Default.
1 - Show for non-leather and chain materials 
2 - Same as 1 but show chain
3 - Show all


Interkarma:
Ditto on this one. Great work, and I will add your options to manual in next build. :)

PR #533, opened 2017-10-27
Add dungeon texture options

Allofich:
Adds settings.ini options for dungeon textures.

0 = classic textures
1 = climate-based textures, using classic algorithm for picking among them. Main story dungeons have classic textures.
2 = climate-based textures for all dungeons
3 = Random selection from all textures (DF Unity implementation from before). Main story dungeons have classic textures. Sewer texture fix added so out-of-range textures shouldn't be used.
4 = Random selection from all textures (DF Unity implementation from before) for all dungeons. Sewer texture fix.

### DIFF
--- a/Assets/Resources/defaults.ini.txt
+++ b/Assets/Resources/defaults.ini.txt
@@ -14,6 +14,7 @@ ShadowResolutionMode=1
 UseLegacyDeferred=False
 DungeonLightShadows=False
 EnableTextureArrays=True
+RandomDungeonTextures=0
 
 [ChildGuard]
 PlayerNudity=False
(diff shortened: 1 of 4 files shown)


Nystul-the-Magician:
awesome thanks! ;)

Allofich:
I fixed the oddity of dungeons in classic skipping over the assignment of dungeon textures in rainforest climates, since I can't see any reason for this and it must just be a mistake in classic.

I had been using the default textures for that case, which is what happens if it's the first dungeon you visit since starting the program, like if you load your game in one, but in classic it's actually just not set to anything for the rainforest climate, so the textures will be whatever is leftover from the last dungeon, or if you haven't been to any dungeons, it's the default dungeon textures.

Anyway you'll now get a normal variety of textures in rainforest climate dungeons when using the classic algorithm.

Interkarma:
This is really awesome. More choices than anybody could ask for. I've noted your INI options and will add to manual update in next builds. Cheers!

PR #543, opened 2017-11-03
Identify dungeon water level and castle block data

Allofich:
This identifies the dungeon water level value, which is contained in the "SoundIndex" value of a block's start marker.

I don't know if I'm propagating the value up in a way you would like, Interkarma, so feel free to change that. I just poked around at it until I found a way to get it through the various data structures up to the PlayerEnterExit class, which is where I suppose it is wanted.

Start markers also contain the value that determines whether a dungeon block is a main story castle block with NPCs, initially non-hostile guards, etc. Since we can use that I put that in and removed the check based on block name, but functionally it should be the same. If you would prefer it continue to be done by block name that is fine.

### DIFF
--- a/Assets/Scripts/API/DFLocation.cs
+++ b/Assets/Scripts/API/DFLocation.cs
@@ -642,6 +642,12 @@ namespace DaggerfallConnect
 
             /// <summary>Name of RDB block.</summary>
             public String BlockName;
+
+            /// <summary>Height level of dungeon water. 10000 means no water.</summary>
+            public Int16 WaterLevel;
+
+            /// <summary>Whether this block is a main story castle area.</summary>
+            public Boolean CastleBlock;
         }
 
         #endregion
--- a/Assets/Scripts/API/Save/SaveVars.cs
+++ b/Assets/Scripts/API/Save/SaveVars.cs
@@ -46,6 +46,7 @@ namespace DaggerfallConnect.Save
         const int currentRegionIdOffset = 0x173A; // Not bothering to read right now.
         const int cheatFlagsOffset = 0x173B;
         const int lastSkillCheckTimeOffset = 0x179A;
+        const int dungeonWaterLevelOffset = 0x17A8; // Not bothering to read right now.
 
         const int regionDataOffset = 0x3DA;
         const int regionDataLength = 80;
--- a/Assets/Scripts/Game/AmbientEffectsPlayer.cs
+++ b/Assets/Scripts/Game/AmbientEffectsPlayer.cs
@@ -35,9 +35,13 @@ namespace DaggerfallWorkshop.Game
         SoundClips[] ambientSounds;
         AudioClip rainLoop;
         AudioClip cricketsLoop;
+        AudioClip waterSound;
         float waitTime;
         float waitCounter;
+        float waterSoundWaitTime = 0.0625f; // Approximation of classic's main update loop. About 1/16 of a second (varies in classic by frame rate). TODO: Unify with runningTallyInterval in PlayerEntity. Where should this value go?
+        float waterWaitCounter;
         AmbientSoundPresets lastPresets;
+        PlayerEnterExit playerEnterExit;
 
         public enum AmbientSoundPresets
         {
@@ -62,12 +66,14 @@ namespace DaggerfallWorkshop.Game
         {
             rainLoop = null;
             cricketsLoop = null;
+            waterSound = null;
         }
 
         void OnEnable()
         {
             rainLoop = null;
             cricketsLoop = null;
+            waterSound = null;
         }
 
         void Update()
@@ -79,6 +85,7 @@ namespace DaggerfallWorkshop.Game
                 lastPresets = Presets;
                 rainLoop = null;
                 cricketsLoop = null;
+                waterSound = null;
 
                 // Stop playing any loops
                 if (dfAudioSource.AudioSource.isPlaying)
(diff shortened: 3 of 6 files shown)


Interkarma:
Great work! Will be very exciting to get water in soon, and nice to know how those castle blocks are identified. Awesome. :)

Allofich:
I'm pretty sure the dungeon water level be a Y coordinate, as used in classic. Do you know how classic Y coordinates are translated into DF Unity Y coordinates?

I'm testing in a water block in Castle Wayrest. The dungeon water level here is -488. When I'm standing about where the water would be, the player's Y coordinate in DF Unity is about 13 or so.

I was looking at adding the water sound that randomly plays when you're in a block with water, but I think it's supposed to play at the player's X and Z, but the Y coordinate of the water level, so I need to get the DF Unity Y coordinate.

Interkarma:
There's no mapping between the Y coordinates in dungeons between classic and DFUnity yet. It's something which needs more research and perhaps changes to layout classes. Same for outside, as the two games have different concepts of vertical space. It's why importing a classic save teleports player outside of dungeon or snaps player to ground level.

The dungeon side of things should be solveable though. Will just need to find a way to baseline the dungeon and then the usual unit scaling applies. It was something I was going to play with later.

Allofich:
I think I've pretty much figured it out. Luckily it looks like the Y coordinates in classic and DF Unity are already in a constant relation to each other.

One thing is I forgot that in classic, negative Y is up, while in DF Unity positive Y is up.

Here's some Y values I measured:
Entrance to Privateer's Hold (when entering from outside):
Classic: -1536
DF Unity: 39.35 (x40 = 1574)

Entrance to Lysandus's Tomb
Classic: -768
DF Unity: 20.15 (x40 = 806)

Entrance to Castle Wayrest
Classic: -1024
DF Unity: 26.55 (x40 = 1062)

Sewer section of Castle Wayrest, standing on ledge above water:
Classic: -512
DF Unity: 13.75 (x40 = 550)

After multiplying by 40, and ignoring the negative/positive difference, looks like it's always 38 off.

Allofich:
Which means I could add the dungeon water sound, but that brings me to another problem, if you have time to answer. The easiest way to add the sound would be to use AudioSource.PlayClipAtPoint(), but unfortunately this plays in just one spot and the sound quickly drops off you move away. Even if we solved the drop off issue it would still always sound like the water sound was playing from behind you where you just had been, which isn't what it's like in classic. I think what would be best is an AudioSource that moves with the player's X and Z but keeps it's own constant Y at the water level, but I don't know how to add that.

It's kind of pointless to have the water sound before the water is in, so this can wait, but I wondered if you could offer any help. If not, don't worry about it.

Allofich:
Went ahead and added water sounds, including what should be approximation of how classic calculates the frequency of it. Just using PlayClipAtPoint for now, which has too steep of a falloff for my liking. I also used it for the sound when enemies die, and there too I felt the falloff was too extreme.

The water sound will need improvement, but it's an initial implementation. If you'd rather not have it, I can remove it from the PR, or if you can selectively remove that commit, that's fine.

Allofich:
To convert from classic player Y position to DF Unity player Y position would be
 (playerEnterExit.blockWaterLevel * -1 * MeshReader.GlobalScale) + 0.95f,
with the .95f covering the difference that was 38 when going the other way.
I suspect this is probably because DF unity and classic measure the player's Y from different parts of the player, so I left this out of the conversion from classic's water height level to DF Unity, making the conversion just simply
(playerEnterExit.blockWaterLevel * -1 * MeshReader.GlobalScale)

Interkarma:
Excellent work! Happy to have the audio stuff. It's something we can improve as water goes through development.

Really looking forward to seeing more with water soon. Is there any more you'd like to do on this? My time is a little limited, but would be happy to help if you're interesting in continuing work on water and need anything from me to progress. Just let me know and I'll do what I can. :)

Interkarma:
Regarding sound, if you don't want the drop-off, why not play it as a 2D one-shot sound like the other ambient effect clips? That way it doesn't have a position in space that falls off when player moves away. This is probably simpler than translating the audio source to follow player.

Interkarma:
In classic, do the other dungeon ambient effects (rattling, wind, clanking, etc.) stop when player is underwater? I can't recall. It might be a good idea to setup a whole new ambient effect player for underwater and switch to it when player's head height drops below water level. But definitely not urgent for now. :)

Allofich:
>  Is there any more you'd like to do on this? My time is a little limited, but would be happy to help if you're interesting in continuing work on water and need anything from me to progress.

I will probably keep adding to the gameplay side of things with water as I can. I don't think I need anything from you on that. I'm not intending to try any rendering, I wouldn't know where to start and I imagine you've probably had some ideas kicking around for a long time. So if I add anything before water is rendered it will be like invisible water (ex. splash sound when you go past water level), which is fine with me but are you OK with it? As long as things are going in the right direction I find it fun to have things come together piece-by-piece like that.

> In classic, do the other dungeon ambient effects (rattling, wind, clanking, etc.) stop when player is underwater?

I just checked and they still play while you're underwater. Btw I know what you meant, but if you were talking about this PR when you mentioned switching the ambient effects player, this PR was about water sounds while you're above water, so it wouldn't involve switching the player in that case.

> Regarding sound, if you don't want the drop-off, why not play it as a 2D one-shot sound like the other ambient effect clips?

I just tested the sound in classic, and it does depend on your distance to the water. It does not, however, change as the player moves. So if it begins while you're at the top of some stairs away from the water, it will be quiet through the entirety of the clip even if you descend midway and get close to the water (maybe all sounds in classic are like this, don't know). Therefore we can probably get the same effect as classic by just playing a 2D one-shot with the volume adjusted based on y-distance to the water. If we could get it to dynamically change as the player moved that would probably be an improvement, but as far as just getting baseline classic quality, the 2d one-shot with adjusted volume may work.

Interkarma:
> I just checked and they still play while you're underwater. Btw I know what you meant, but if you were talking about this PR when you mentioned switching the ambient effects player, this PR was about water sounds while you're above water, so it wouldn't involve switching the player in that case.

I just meant while underwater (I couldn't recall). Sorry for being unclear. You're way more across this than I am right now. :)

PR #544, opened 2017-11-04
Get most pre-discovered locations like classic

Allofich:
Gets most pre-discovered locations in the same manner as classic.

Some DFRegion.LocationTypes.GraveyardForgotten type locations that start the game known in classic are not gotten yet, but this is the same in the current implementation in master. I haven't been able to find how these are gotten yet.

As far as I know this is functionally identical to master, but it's a step toward correctly reading the map files and could be helpful as reference in figuring out how to import discovered location data from classic saves or how to get the remaining locations that start the game known so I think it's worth merging, but I'll leave the call to you Interkarma.

I also removed what looked like a couple out-of-date TODO comments.

### DIFF
--- a/Assets/Scripts/API/DFRegion.cs
+++ b/Assets/Scripts/API/DFRegion.cs
@@ -66,44 +66,36 @@ namespace DaggerfallConnect
         /// </summary>
         public enum LocationTypes
         {
-            /// <summary>Large dungeon.</summary>
-            DungeonLabyrinth = 132,
-            /// <summary>Medium dungeon.</summary>
-            DungeonKeep = 135,
-            /// <summary>Small dungeon.</summary>
-            DungeonRuin = 138,
-
-            /// <summary>Graveyard visible.</summary>
-            GraveyardCommon = 172,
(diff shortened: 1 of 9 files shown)


Allofich:
In researching the way classic uses the data I did a cleanup and changed data to be largely be gotten more like classic. Enums for dungeons and map locations make more sense now. Functionally things are still the same, and I still haven't been able to figure out what makes some graveyards show up at the start of the game. As far as I can tell btw, classic does not differentiate between different types of graveyard. I believe the previous enum for it in DF Unity was just including the "discovered" flag as part of the identifying enum.

Interkarma:
That's a nice tidy-up. You're working from new information than I had at the time, so I'm just going to trust everything works the same. :) I don't believe these enums are serialized anywhere (such as quest system, save data, etc.). Hopefully no reason for it to impact live games either. Will test and let you know if any feedback.

Allofich:
Looks like something I did messed up dungeon type selection. I'm getting incorrect monsters (undead in a dungeon that's not supposed to be an undead one). I'll try to figure out what's wrong.

PR #547, opened 2017-11-06
Remove no longer needed bit shifts

Allofich:
I missed this in my PR on the map data. A leftover shift was causing all dungeons to have crypt monsters.


Allofich:
Another one, which would have given all dungeons crypt treasure piles.

I went through the whole code looking for ">> 8" so this is hopefully all of them.

Interkarma:
Thanks, and well spotted. I had forgotten I'd done that in a few spots around the layout code.

PR #548, opened 2017-11-07
Water movement, sounds

Allofich:
New water stuff.

- Sounds for walking in shallow water, swimming in deeper water, underwater ambient sound
- Swim movement, borrowing the FakeLevitate class, since levitation and swimming movement are pretty much the same thing in Daggerfall. Swim speed based off of classic formula.
- Swim fatigue and skill tallying, matching classic.
- "You cannot sleep now." message if you try to rest while swimming.

### DIFF
--- a/Assets/Scripts/API/Save/SaveVars.cs
+++ b/Assets/Scripts/API/Save/SaveVars.cs
@@ -38,6 +38,7 @@ namespace DaggerfallConnect.Save
         const int isDayOffset = 0x391;
         const int typeOfCrimeCommittedOffset = 0x3A3;
         const int inDungeonWaterOffset = 0x3A6;
+        const int breathRemainingOffset = 0x3AB;
         const int weatherTypesOffset = 0x3B7;
         const int weaponDrawnOffset = 0x3BF;
         const int gameTimeOffset = 0x3C9;
@@ -83,6 +84,7 @@ namespace DaggerfallConnect.Save
         bool isDay = false;
         byte typeOfCrimeCommitted = 0;
         bool inDungeonWater = false;
+        int breathRemaining = 0;
         byte[] weatherTypes; // Probably the possible weather types for the player's location
         bool weaponDrawn = false;
         uint gameTime = 0;
@@ -266,6 +268,14 @@ namespace DaggerfallConnect.Save
         public bool InDungeonWater
         {
             get { return inDungeonWater; }
+        }
+
+        /// <summary>
+        /// Gets breath remaining when underwater from savevars.
+        /// </summary>
+        public int BreathRemaining
+        {
+            get { return breathRemaining; }
         }
 
         /// <summary>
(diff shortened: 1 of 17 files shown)


Allofich:
The random pitch effect is quite noticeable on the water swimming sound, and maybe sounds a little odd. Classic doesn't have this random pitch effect for footsteps, instead footsteps alternate between a left foot/right foot sound, only one of which is used by DF Unity currently, with that one sound receiving pitch variation. I've done work on making DF Unity footsteps like classic already and can submit that as a PR after some more work, but I held off on it because I wasn't sure if it would conflict with your preferences Interkarma, if you have a liking for the random pitch footsteps. That would address the swim sounds, though.

Interkarma:
Wonderful work as ever, thank you. 👍 

I don't have a problem with the footstep pitch setup being changed to alternating. At the time, the variable pitch sounded less mechanical to my ears, but it's a very minor thing. Please drop and use the classic approach if that makes things easier for you. :)

Allofich:
Foot step stuff is in.

If it still sounds too mechanical, we could put the pitch variation back in, but I would suggest toning it down to .1f (was .3f) or something. Even with .1f the swimming sound pitch variation was kind of distracting, though, probably because it's a high-pitched sound to begin with and changes in the pitch are more noticeable. If we put pitch variation in again we could disable it for the swim sound... but if we allow sound replacement at some point, that would need to be reworked. For now I'm happy to go with classic.

Footsteps in classic can vary when you're in a dungeon depending on the floor material, but I don't know where that value is. For example the dirt at the bottom of the chasm in Scourg Barrow has a soft footstep sound. If we can get that in, maybe we can also make the outdoor footstep sound be a harder-sounding one for the walkways in cities or the 3d geometry of castle-type buildings, which would be a nice improvement over classic.

One improvement that would

Allofich:
Implemented holding breath underwater, and dying if it runs out, matched to classic behavior.

Right now I just assign the spell point bar's texture to the breath meter. Is there some way to use a color value like classic? But we should also allow for texture replacement. Also I think the existing vitals bars all have their textures just loaded once, but the breath meter should change from yellow to red when it gets low.

Interkarma:
Apologies for delay merging, I've been out sick for a while. I will get this merged when I can.

For flat colour bars, just use a panel with Background colour set as needed. For texture tinting, overlay a semi-transparent panel to modify underlying texture.

Allofich:
No rush, hope you feel better.

Allofich:
The breath meter's colors are implemented now, using the colors from classic (from ctrl-F5 screenshots in DOSBox) and the same threshold.

I don't think I handled setting the color in the way you said, Interkarma. I wanted to use the existing vertical progress bar code. I expanded that class to accept a color and not just a texture, and for the color it creates a 1 by 1 Texture2D out of it (borrowing code from the BackgroundColor in BaseScreenComponent), which lets it use the same GUI.DrawTextureWithTexCoords function as the texture uses. Functionally it works, but if you object to it because it's hacky or something, feel free to change it.

Interkarma:
That's a perfectly good solution. 👍

Interkarma:
The work you've done on dungeon water here is just incredible. Just had a fun time splashing around and swimming. Everything feels and sounds just about spot-on. Great work! 👍

PR #550, opened 2017-11-09
SettingsWindow: support for multiple pages

TheLacus:
As showed on forums, settings are grouped in tabs or pages. You can move to the next one with the tab key or selecting a title in top bar.


Interkarma:
Thanks for this change, it should make it a lot easier to stack in more options in future. 👍

PR #552, opened 2017-11-16
Implement map item

Allofich:
Works basically like classic. Although UESP says that map items reveal a large undiscovered dungeon, it's actually any undiscovered location, including small dungeons and covens, which I confirmed through testing.

Since I found the flag in my earlier PR that determines whether a location starts the game discovered (not counting some of the graveyard locations which are still a mystery), we now have "discovered" data in two places, that flag in the map tables, and a discovered locations dictionary. Maybe we could get by just using the bit in the map tables? Also there's what feels like a little bit roundabout code in this PR in order to discover the location using the region and location names, which are the parameters for the to be the DiscoverLocation function, but it works fine. If you would like to reorganize anything, please go ahead.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
@@ -1084,6 +1084,15 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
                 return;
             }
 
+            // Handle map items
+            if (item.IsOfTemplate(ItemGroups.MiscItems, (int)MiscItems.Map))
+            {
+                RecordLocationFromMap(item);
+                from.RemoveItem(item);
+                Refresh(false);
+                return;
+            }
+
             // Handle quest item transfer
             if (item.IsQuestItem)
             {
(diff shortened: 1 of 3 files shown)


Nystul-the-Magician:
nice! already wanted to ask you if you could take a look into this ;)

Allofich:
Thanks!

Interkarma:
Awesome, thank you for this! :)

> we now have "discovered" data in two places, that flag in the map tables, and a discovered locations dictionary. Maybe we could get by just using the bit in the map tables?

The live discovery system in Daggerfall Unity should always be considered authoritative at runtime. The bit should only be used to determine if a location is "pre-discovered" for purposes of populating travel map. Otherwise just use the location discovery system in PlayerGPS.

> Also there's what feels like a little bit roundabout code in this PR in order to discover the location using the region and location names, which are the parameters for the to be the DiscoverLocation function, but it works fine.

It would be easy enough to add overloads to support MapPixelID, Latitude/Longitude, or indices if preferable. But internally, it would be doing much the same kind of thing as you're already doing, so not much difference performance-wise.

Allofich:
OK, thanks for the response. Do you want me to change anything in the PR?

Interkarma:
It should be fine. I will merge when I can. :)

PR #553, opened 2017-11-17
Potion generation and enemy inventory changes

Allofich:
Creates potions randomly from the 20 varieties, like in classic, with correct names and values. Also works with potions imported from classic.

### DIFF
--- a/Assets/Scripts/Game/EnemyDeath.cs
+++ b/Assets/Scripts/Game/EnemyDeath.cs
@@ -92,9 +92,6 @@ namespace DaggerfallWorkshop.Game
                 mobile.Summary.Enemy.CorpseTexture,
                 DaggerfallUnity.NextUID);
 
-            // Generate items
-            loot.GenerateItems();
-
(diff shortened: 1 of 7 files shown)


Allofich:
I'm using an array of the prices in the function and a string array of the potion names now for simplicity, but these are actually in some kind of potion definitions in the .exe, starting at 1BA142. Each is 109 bytes long, with the first two bytes being the value.

Interkarma:
Well done! Thank you. :)

Allofich:
Updated. I would have made a separate PR but it would have had a merge conflict.

Now all inventory items (loot table items, maps, potions and potion recipes) are added to enemies on instantiation, not death. This is the same as classic and better in my opinion, because you would get different results if you kill an enemy, load your game, and kill it again. Also, it allows getting rid of loot piles storing a table key, which wasn't really needed.

Allofich:
I hadn't realized there was already code to handle potion names. I changed the first commit to use the existing method.

PR #554, opened 2017-11-18
Implement knocking back enemies

Allofich:
Implements knocking back enemies. Should be close to classic behavior.

This may give the player a bit of advantage for now. If we implement the player's view getting thrown off when they take a heavy hit, and also correct weapon speeds (the player swings fast now), it will hopefully even out.

I think having this in somewhat alleviates the "enemy constantly pressing forward while you backpedal" feeling of combat.

One thing that is happening in this PR that I don't understand is that sometimes an enemy will resume moving toward the player while still in the hurt animation. I can't see why this happens. The state should always have been reset to "Move" before an enemy gets out of being knocked back. Any idea?

### DIFF
--- a/Assets/Prefabs/Scene/DaggerfallEnemy [Game Serializable].prefab
+++ b/Assets/Prefabs/Scene/DaggerfallEnemy [Game Serializable].prefab
@@ -217,7 +217,7 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   MeleeAttackSpeed: 1.25
-  MeleeDistance: 2
+  MeleeDistance: 2.5
 --- !u!114 &11435454
 MonoBehaviour:
   m_ObjectHideFlags: 1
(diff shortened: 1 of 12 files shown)


Allofich:
The knockback is definitely more pronounced than in classic. I kind of like it like this, it makes the enemies feel a little bit more real, but it gives advantage to a player that can hit hard and consistently by keeping the enemy out of range and unable to attack. I'm not sure what to do with it at the moment.

Allofich:
Added some adjustments to enemy attacks to give them a chance to hit the player when knockback is in.

This includes changing the enemy range back from 2 to 2.5. Interkarma you adjusted it earlier, but if knockback is in I think we need this adjustment.

Interkarma:
Agree the range should be increased again now. Thank you for doing this.

Apologies for delays with merging changes. Have been on a heavier work streak for the last few weeks that should wrap up today. I'll be easing back into my hobby stuff before end of this week. :)

Allofich:
That's fine, whenever you've got the time and inclination for it.

PR #557, opened 2017-11-24
Prevent exception in SmoothFollower

Allofich:
For quite a while now, probably ever since the SmoothFollower stuff was added, I've had a problem with an exception occurring on the line

`float maxPossibleDistanceByMotorVelocity = controller.velocity.magnitude * 2.0f * Time.deltaTime;`

in PlayerMotor.cs when I try to start the game from the editor. It seems like it always happens the first time I try to run the game from the editor after getting the latest update to the master branch.

Has anyone else had the same issue? This check should keep the exception from happening.

BTW, on the topic of fixing issues with the code...

I've noticed that when loading a game, "OnMapPixelChanged" happens twice. Nystul, in TalkManager.cs, AssembleTopicLists() is called 3 times when you load a game, twice from the OnMapPixelChanged event, then once for OnLoadEvent. I know it's a work in progress, but I'm letting you know in case you didn't already. It should only need to be called once, right?

### DIFF
--- a/Assets/Scripts/Game/PlayerMotor.cs
+++ b/Assets/Scripts/Game/PlayerMotor.cs
@@ -511,7 +511,7 @@ namespace DaggerfallWorkshop.Game
                 }
             }
 
-            if (smoothFollower != null)
+            if (smoothFollower != null && controller != null)
             {
                 float distanceMoved = Vector3.Distance(smoothFollowerPrevWorldPos, smoothFollower.position);        // Assuming the follower is a child of this motor transform we can get the distance travelled.
                 float maxPossibleDistanceByMotorVelocity = controller.velocity.magnitude * 2.0f * Time.deltaTime;   // Theoretically the max distance the motor can carry the player with a generous margin.
--- a/Assets/Scripts/Game/Utility/TravelTimeCalculator.cs
+++ b/Assets/Scripts/Game/Utility/TravelTimeCalculator.cs
@@ -150,10 +150,10 @@ namespace DaggerfallWorkshop.Game.Utility
 
         public int CalculateTripCost(int travelTimeInMinutes, bool sleepModeInn, bool hasShip, bool travelShip)
         {
-            int travelTimeInDays = (travelTimeInMinutes + 59) / 60;
+            int travelTimeInHours = (travelTimeInMinutes + 59) / 60;
             int cost = 0;
             if (sleepModeInn)
-                cost = 5 * ((travelTimeInDays - pixelsTraveledOnOcean) / 24) + 5;
+                cost = 5 * ((travelTimeInHours - pixelsTraveledOnOcean) / 24) + 5;
             if ((pixelsTraveledOnOcean > 0) && !hasShip && travelShip)
                 cost += 25 * (pixelsTraveledOnOcean / 24 + 1);
             return cost;


Allofich:
Added on a fix for an incorrectly named variable from an earlier PR.

Nystul-the-Magician:
well i just react on the event, not sure what i could do better there ;)

Interkarma:
Cheers. I don't have this issue, it's likely a script execution order problem, which can vary a little between systems. Sometimes unintended problems are harder to catch because of this, and your solution is correct. Thank you for the catch. :)

> I've noticed that when loading a game, "OnMapPixelChanged" happens twice. 

Mostly likely fires once when world is reinitialised, and again when player is moved back to save position in world. The thing with event-based flow is that both cases are technically correct. It's not up to the event to decide which one should or shouldn't be raised, the event only cares the map pixel coordinates have changed. It's up to the event consumer to not perform any hefty processing on this event, or do a quick check (e.g. is game loading? is world in init state?) before doing work. If there's only a few milliseconds in it anyway, and no other problems as a result, there's probably no real advantage to changing behaviour. This is probably something best looked into as part of an optimisation pass down the road.

Interkarma:
> well I just react to the event, maybe it is sent several times?

Yep, and very normal for events to fire multiple times depending on what they are. I wouldn't worry about it unless it has measurably negative performance outcomes.

The event itself doesn't know or care what you're doing with it - it's just saying "something happened". As an event consumer, the trick is to decide "is this event appropriate for my needs?" and "should I do X on this event or find another time/way of doing X?". I generally see this as more of an optimisation question.

Nystul-the-Magician:
yeah exactly my thoughts. ;)
will take a look if I do any heavy processing but as far as I can remember it should not be too much

Allofich:
I haven't noticed a performance hit from it, but overall the player repositioning code seems like maybe it could use a tune up. When I load a game I often get the wrong music playing for a split-second, etc. I'm just speculating but maybe there's an intermediate re-positioning of the player that could be optimized out or something, which could be related to OnMapPixelChanged being sent twice when loading a game.

Also one place where I have noticed a slight performance hit is when leaving the talk window back to the 3d world, there's a slight hiccup in the framerate.

Interkarma:
Agree it could use a check-up to optimise at some point. Not too worried at the moment though.

I've also noticed the performance hit when exiting talk window (feels like about 300-ish ms on my system). I'm confident Nystul will fix, but I'm available to help if needed. :)

Nystul-the-Magician:
will take a look ;)

Nystul-the-Magician:
it might come from Resources.UnloadUnusedAssets() being called inside UserInterfaceWindow.CloseWindow().

Interkarma:
Agreed, it feels like garbage collection stall to me also.

Best way to diagnose is run with profiler open and check memory allocations and garbage collections. There's probably just one class of allocation being dumped on pop causing the stall. Once identified, the resource causing stall can be changed to use a different allocation setup (e.g. pooling) to avoid GC problems.

Whether this is with the talk window or some other UI element (there's a been a few changes elsewhere in the UI around the time this started) is yet to be seen. I've just noticed it most with talk window - but haven't had time to dive in with profiler so it could be elsewhere. Happy to let this one rest for now. :)

Allofich:
This might be a bit more of an issue than I thought. See this video of the latest build, by @numidium I assume, from around 2:43. The pause after leaving dialogue is really long and makes him think the game has frozen.

Nystul-the-Magician:
can you post the link to the video please ;)

Allofich:
Whoops, I had intended to and forgot.
https://www.youtube.com/watch?v=a7VYlLUHmoM&t=1150s

Allofich:
Just a note: When he then talks to another NPC, the pause afterwards is not nearly as long, and is more like what I've seen when I've been testing.

Nystul-the-Magician:
it looks like this happens after the window has closed. So maybe it is outside the talkwindow code?

Interkarma:
I personally haven't experienced this since I mentioned it back in November.

If someone can reproduce reliably, first port of call will be to comment out "Resources.UnloadUnusedAssets();" in UserInterfaceWindow.cs. This dumps textures, etc. that are no longer used when UI windows closed. This has been in since almost day one to help keep UI memory trim - but might be having a collection problem post talk window allocations.

Next step is to open up Profiler and check GC behaviour around time talk window closes. It should be possible to narrow it down from there.

When I get a little time, I'll try to reproduce again and dig a little deeper. I've spent a fair amount of time in the talk window lately though, and it hasn't happened again. So not sure yet, sorry guys.

Allofich:
I've never had as long a pause as in the video. I mentioned numidium's name so maybe he can provide a little insight. He did navigate through a number of options in the talk window before closing it. I'm going to try to generally recreate those circumstances:
1- Opening dialogue for first time to an NPC outside walls of a large city.
2- Click through several talk options
and see what happens...

Nothing. I'll try again with a build rather than through the editor... still nothing. So, I'm unable to reproduce it.

Allofich:
It could be mod-related. I don't use any mods, but in Galactic Chicken's videos he sometimes was talking about mod settings causing long load times, and in this video as well there were some mods active.

Nystul-the-Magician:
maybe it is really the Resources.UnloadUnusedAssets() in conjunction with large textures from mods. maybe those get unloaded when such pause happens

numidium:
I tried reproducing this a few times. The game did hang for a short moment sometimes when exiting the talk menu but not for quite as long as it did in my video. I came across something that may be of interest in my output_log.txt file:
`
IndexOutOfRangeException: Array index is out of range.
  at DaggerfallWorkshop.Game.TalkManager.expandRandomTextRecord (Int32 recordIndex) [0x00000] in <filename unknown>:0 
  at DaggerfallWorkshop.Game.TalkManager.GetOrganizationInfo (DaggerfallWorkshop.Game.ListItem listItem) [0x00000] in <filename unknown>:0 
  at DaggerfallWorkshop.Game.TalkManager.GetAnswerText (DaggerfallWorkshop.Game.ListItem listItem) [0x00000] in <filename unknown>:0 
  at DaggerfallWorkshop.Game.UserInterface.DaggerfallTalkWindow.SelectTopicFromTopicList (Int32 index) [0x00000] in <filename unknown>:0 
  at DaggerfallWorkshop.Game.UserInterface.DaggerfallTalkWindow.ListboxTopic_OnUseSelectedItem () [0x00000] in <filename unknown>:0 
  at DaggerfallWorkshop.Game.UserInterface.ListBox.RaiseOnUseItemEvent () [0x00000] in <filename unknown>:0 
  at DaggerfallWorkshop.Game.UserInterface.ListBox.UseSelectedItem () [0x00000] in <filename unknown>:0 
  at DaggerfallWorkshop.Game.UserInterface.ListBox.MouseDoubleClick (Vector2 clickPosition) [0x00000] in <filename unknown>:0 
  at DaggerfallWorkshop.Game.UserInterface.BaseScreenComponent.Update () [0x00000] in <filename unknown>:0 
  at DaggerfallWorkshop.Game.UserInterface.ListBox.Update () [0x00000] in <filename unknown>:0 
  at DaggerfallWorkshop.Game.UserInterface.Panel.Update () [0x00000] in <filename unknown>:0 
  at DaggerfallWorkshop.Game.UserInterface.Panel.Update () [0x00000] in <filename unknown>:0 
  at DaggerfallWorkshop.Game.UserInterface.Panel.Update () [0x00000] in <filename unknown>:0 
  at DaggerfallWorkshop.Game.UserInterface.UserInterfaceWindow.Update () [0x00000] in <filename unknown>:0 
  at DaggerfallWorkshop.Game.UserInterfaceWindows.DaggerfallBaseWindow.Update () [0x00000] in <filename unknown>:0 
  at DaggerfallWorkshop.Game.UserInterfaceWindows.DaggerfallPopupWindow.Update () [0x00000] in <filename unknown>:0 
  at DaggerfallWorkshop.Game.UserInterface.DaggerfallTalkWindow.Update () [0x00000] in <filename unknown>:0 
  at DaggerfallWorkshop.Game.DaggerfallUI.Update () [0x00000] in <filename unknown>:0 `

I did a bunch of scrolling around and picking dialogue options from the "tell me about..." menu before seeing this. I'll post steps to reproduce if I can get it to happen reliably.

numidium:
Turns out that error listed in output_log.txt happens in an edge case in my code where you ask about the Benevolence of Mara. I'll have to fix that later but it's unrelated to the freezing issue.

By the way, here's a list of mods I had installed just in case that's what caused the freeze:
Birds of Daggerfall 1.0.1
Distant Land 2.2.0
Enhanced Sky 2.0.5
Real Grass 2.0
Vibrant Wind 0.4

PR #558, opened 2017-11-25
Implement classic weather selection

Allofich:
This re-introduces weather changes, which are currently disabled, using something close to the way classic does it, and imports weather from classic saves.

- In classic weathers are stored in a 6-byte array, which can be found in SaveVars, with the weather for the player's current climate the one that is used. DF Unity just tracks one weather.
- Thunder (which can be chosen in classic but has no visual or audio effects) and snow have their enums reversed in classic and Unity. This PR switches them when importing from classic.
- Classic changes weather when the date changes, or when the player crosses climate zones. Because of the game's accelerated time versus the speed of real time, I feel this looks OK, even though weathers only change on at a minimum of 24-hour increments. Also, since the changes happen at midnight game-time, the abruptness of them should be less than if it happened during the day. Still, it would be nice to have the possibility for transitions at any time of day as an improvement in the future. The abrupt changes when crossing climate zones look bad and I haven't tried to recreate them.
- In addition to the 7 weather types, if the 0x80 bit is set, I think rain and snow can have fog added to them. However, fog doesn't seem to be very noticeable in DF Unity currently, at least it doesn't create the same type of thick fog as classic. These "+fog" weather types aren't implemented yet. The 0x80 bit in classic has a chance of being set whenever rain or snow is chosen.
- The "MountainWoods" climate gets mountain weather chances in Unity, but woodland chances in classic. I don't know if we should change this, but for this "classic-like" implementation, the woodland chances are used.

At the moment the array of weathers isn't stored in DF Unity saves, as it's not necessary. It might not be necessary to have it in PlayerWeather, either, but for simplicity of importing from classic saves, it's there now. Maybe we can remove it.

We could replace this with a better selection method later, including with more weather transitions, but maybe we can keep this as an option for classic behavior. Also it will allow people to experience and test the weathers again, which are disabled now.

### DIFF
--- a/Assets/Scripts/API/Save/SaveVars.cs
+++ b/Assets/Scripts/API/Save/SaveVars.cs
@@ -39,7 +39,7 @@ namespace DaggerfallConnect.Save
         const int typeOfCrimeCommittedOffset = 0x3A3;
         const int inDungeonWaterOffset = 0x3A6;
         const int breathRemainingOffset = 0x3AB;
-        const int weatherTypesOffset = 0x3B7;
+        const int climateWeathersOffset = 0x3B7;
         const int weaponDrawnOffset = 0x3BF;
         const int gameTimeOffset = 0x3C9;
         const int maleOrFemaleClothingOffset = 0x3D1; // 6 for male clothing or 12 for female clothing, matching player gender. Not bothering to read right now.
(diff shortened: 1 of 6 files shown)


Allofich:
Added in a couple minor fixes.

1) The duplicate "climateWeathers" in the savevars overwrites the first, so that is the one that is effectively used in classic when loading a game. This PR does the same now.

2) Rain and snow weathers with the 0x80 "fog" flag are imported from classic as regular rain and snow by canceling out the 0x80 flag for now, until those weathers are in DF Unity. I suppose we'll need new weather enums for them, and there's no thick fog effect at the moment, right?

Nystul-the-Magician:
fog could be easily done by setting rendersettings fog settings. Just make sure to restore clear settings once sky is clear (This is important so that e.g. far terrain mod does not break).
If you give me some time I can work it out ;)

Interkarma:
This is great, looking forward to having weather back soon.

PR #559, opened 2017-11-26
Implement magic weapon FPS graphics

Allofich:


Interkarma:
That's a simple and clever way to get the magic weapon displaying for now. Nice work!

PR #560, opened 2017-11-26
Implement weapon and shield equip/unequip delay

Allofich:


Interkarma:
Good work. You're absolutely nailing this stuff.

PR #563, opened 2017-12-06
Implement regional building answers

Allofich:
Turns out there's keys in the mapTable records for which types of buildings exist in each location.

In classic asking for armorer is bugged and always returns 0, but there was a gap in the flags and it seems to match for armorer, so this PR uses it.

The question text is still unimplemented.

### DIFF
--- a/Assets/Scripts/API/DFRegion.cs
+++ b/Assets/Scripts/API/DFRegion.cs
@@ -240,8 +240,8 @@ namespace DaggerfallConnect
             /// <summary>Whether or not this location is discovered.</summary>
             public Boolean Discovered;
 
-            /// <summary>Last 4 bytes appear to be unused.</summary>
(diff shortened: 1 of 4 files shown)


Nystul-the-Magician:
I can take a look at the question text - should be not that hard - just have to find some time to do it ;)

Allofich:
The question text looks to be the same as for local buildings, so I just re-used that.

Nystul-the-Magician:
awesome job ;)

Interkarma:
You guys rock my world. :)

I'm being smashed by work in my last couple of weeks before leave. Will get these all merged soon, and should have my life back to normal by xmas. Thanks for your patience with me.

Allofich:
That's what I figured. :) Whenever you've got time.

PR #564, opened 2017-12-09
Implement animal sounds

Allofich:
Implements animal sounds. Range and frequency should match or be close to matching classic.

One thing though, is that I don't know exactly what's going on with the billboard batching. It looked like I needed to (or it at least be simplest to) remove animal flats from billboard batching so I could add audioSources to them, but maybe they can still be batched? If not, we could remove the animal batch code.

### DIFF
--- a/Assets/Scripts/DaggerfallUnityEnums.cs
+++ b/Assets/Scripts/DaggerfallUnityEnums.cs
@@ -439,11 +439,12 @@ namespace DaggerfallWorkshop
     /// </summary>
     public enum AudioPresets
     {
-        None,                   // No changes to AudioSource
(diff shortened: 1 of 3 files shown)


Interkarma:
It's possible to drop the audio source with its own empty gameobject at scene generation time when the animal flat is rolled into batch. I do something similar with light objects in towns - i.e. the billboard object is batched but the light source is placed with its own gameobject at the same position in world.

I think in this case, it's not a big deal either way. There's fairly few animals in towns and those billboards are cheap to render. I'm happy to run with it this way, and if it becomes a problem (highly unlikely) we can follow the design I used on lights. :)

PR #567, opened 2017-12-14
Implement basics of enemy ranged attacks

Allofich:
Here's the basics of enemy ranged attacks. The timings and ranges used in the AI decisions should be the same or close to classic.

This still needs work but the basics are in and I think it's already functional enough for testing or playing around.

### DIFF
--- a/Assets/Scripts/Game/EnemyAttack.cs
+++ b/Assets/Scripts/Game/EnemyAttack.cs
@@ -31,7 +31,8 @@ namespace DaggerfallWorkshop.Game
         DaggerfallMobileUnit mobile;
         DaggerfallEntityBehaviour entityBehaviour;
         float meleeTimer = 0;
-        bool isAttacking;
+        bool isMeleeAttackingPreHitFrame;
+        bool isShootingPreHitFrame;
 
         void Start()
         {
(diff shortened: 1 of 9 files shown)


Allofich:
Added some fixes, and now hit and parry sounds are played on the target of an attack rather than on the FPSWeapon or the attacking enemy. This is how classic works and fits in to bow mechanics, since it needs to play those sounds over on the target. Classic also plays miss sounds on the target, but I haven't changed this because it might make more sense for them to play on the FPSWeapon or attacking enemy.

To play sounds on the player I'm sending messages to PlayerFootsteps, because that's how these kinds of "play sound on the player" events have been handled so far. I assume this is OK as an intermittent solution? How should sounds on the player be handled?

One issue is that it looks like enemies do LOS checks from their feet. The guards in Wayrest Castle can't see you if you're on the other side of one of the fountains in the entrance.

Interkarma:
Thank you! Looking forward to playing with this.

Enemy sight checks right now should be happening from the centre of capsule, which about mid-way in their body. But that's probably enough to block sight for any objects at least waist-height. The origin of check point could be raised a bit higher nearer to the head, that would make sense I think. Just need to account for overall capsule height when raising origin point.

I'll review and merge this one a bit later. Have run out of time for today. :)

Allofich:
Some more stuff's in now. Sorry, it should maybe be multiple commits, but when i do that I often find something in an earlier commit I want or need to change, and it's a pain doing rebases and handling merge conflicts and whatnot, and nor do I want to make a mess of "fixing problem in above commit" type commits.

Total list of changes:
1) Enemy classes that have a range attack animation shoot at the player now. Timing, range at which attack decision is made and damage should match classic or be very close.
2) Accounting for enemies who have both rangedattack1 and rangedattack2, since for them rangedattack1 is a spell animation rather than the bow and arrow. These enemies will correctly use the bow and arrow animation.
3) Hit, parry and arrow sound effects are played, and on the target, as in classic, rather than on the FPSWeapon/an attacking enemy.
4) Arrows are added to the inventories of the player or enemies when hit by bow attacks, as in classic.
5) The player loses an arrow for each bow shot, and at 0 they get "You have no arrows." message and the FPS bow does not appear ("weapons sheathed" mode is forced), same as classic.
6) Arrows can be found in loot now, in groups of 1 to 20, as in classic.

Additional work remains and I'll keep working on this stuff as I can. The main thing that needs to be done is to create an actual projectile rather than doing an instant hit. It's a little difficult to see, but classic uses a projectile, for enemy arrows at least and maybe for the player as well. I'm happy for the PR to be merged as it is, though.

PR #568, opened 2017-12-19
TextureReplacement: fix MOVE00I0.IMG

TheLacus:


TheLacus:
Waiting confirmation from Nystul, but it's a small change. As i said on forum i moved the choice between portraitImgName and facesImgName above so i can check it only once, since cif and rci share the same import method.

PR #573, opened 2018-01-02
Implement stealth basics

Allofich:
Implements stealth basics, as well as general tweaking of enemy detection AI.

- The basics of the stealth check in classic are recreated (check every in-game minute, tally player skill, detection formula). There's a bit more that's involved but we need to implement the sneak key first.
- Enemy FOV matched to classic. Vision and hearing range (I don't think enemies have "hearing" in classic) matched to range of stealth check
- Hearing is only active when the player has been detected by that enemy, since otherwise it would interfere with stealth and not allow sneaking up on enemies. The stealth check in classic basically doubles as "hearing" simulation. Since I've increased the hearing range I also force a "false" hearing check when statics are between the enemy and the player, to try to reduce enemies walking against walls because they hear the player behind it.

Right now enemy AI behavior in general is a mixture of classic and DF Unity behavior. I'm trying to first get in classic behavior (this, the ranged attacks) while preserving the improvements of DF Unity. IMO things are working pretty well with this PR. Later maybe we can have a toggle between classic behavior and enhanced AI.

This PR should also open the way to implementing backstabbing.

### DIFF
--- a/Assets/Prefabs/Scene/DaggerfallEnemy [Standalone].prefab
+++ b/Assets/Prefabs/Scene/DaggerfallEnemy [Standalone].prefab
@@ -261,7 +261,6 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   OpenDoorDistance: 2
-  GiveUpTime: 4
 --- !u!114 &11470552
 MonoBehaviour:
   m_ObjectHideFlags: 1
@@ -368,9 +367,9 @@ MonoBehaviour:
   m_Script: {fileID: 11500000, guid: 9fd446a55d098ca4f826c3c9bc93563a, type: 3}
   m_Name: 
   m_EditorClassIdentifier: 
-  SightRadius: 16
-  HearingRadius: 8
-  FieldOfView: 120
+  SightRadius: 25
+  HearingRadius: 25
+  FieldOfView: 180
 --- !u!143 &14323946
 CharacterController:
   m_ObjectHideFlags: 1
--- a/Assets/Scripts/Game/EnemyMotor.cs
+++ b/Assets/Scripts/Game/EnemyMotor.cs
@@ -25,7 +25,6 @@ namespace DaggerfallWorkshop.Game
     public class EnemyMotor : MonoBehaviour
     {
         public float OpenDoorDistance = 2f;         // Maximum distance to open door
-        public float GiveUpTime = 4f;               // Time in seconds enemy will give up if target is unreachable
         public const float AttackSpeedDivisor = 3f;       // How much to slow down during attack animations
 
         EnemySenses senses;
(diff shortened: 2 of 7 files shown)


Interkarma:
This sounds like a great start, thank you!

Regarding the hearing loss, are enemies still able to open doors when player in range? "Hearing" was my solution to reproduce the door opening mechanic from classic, as sight rays are blocked by environment.

Maybe when sneak is fully implemented, we can cut hearing when player is both sneaking and not detected. This would mean player can still be detected with hearing when blundering around outside a room occupied by an enemy. Not sure what to do about them walking against walls however, they definitely need an upgrade to their little brains sometime. :)

Allofich:
> Regarding the hearing loss, are enemies still able to open doors when player in range?

They are, I didn't change that.

One thing I realized is that we may want enemies "walking against walls", meaning walking towards the player even though they can't see him. I was thinking of situations in Daggerfall where you are in an interior and enemies come looking for you (somewhat), like when assassins or city guards spawn. I don't think master branch handled this either though, because the hearing range was quite short. And of course actual pathfinding would be best, but who knows when we'll have that. I'm happy for now with the PR the way its handling things

> Maybe when sneak is fully implemented, we can cut hearing when player is both sneaking and not detected. This would mean player can still be detected with hearing when blundering around outside a room occupied by an enemy. 

Maybe so. There's a few more things to recreate the classic stealth AI, nothing complicated, and then maybe we can go from there.

Allofich:
Implemented stealth key, including quieter hoof sound when on a horse. It doesn't affect chances of enemy detection yet.

Allofich:
Stealth key affects enemy detection now.

Allofich:
Removed stealth tally when bashing doors.

This exists in classic (although it should have also included bashing magic doors), and classic also does some kind of forced stealth check with nearby monsters when bashing or opening a door, but as far as I can tell its meaningless since the stealth check is constantly running for non-alert enemies anyway, and there's no reason to tally the stealth skill then, either.

Interkarma:
This looks excellent! Thank you for all your hard work on this. I'll be rolling out new builds within the next week to make sure everyone can get their hands on it soon. :)

ajrb:
I'm curious why you used the name *MovingLessThanHalfSpeed instead of just Stealth? Is there another function that causes player to move less than half speed like stealth?

Allofich:
I used the name "isSneaking" at first, but you can get the same effect as the "sneak" key by crouching, as what's actually checked is whether you're moving less than half your walk speed. Also, you can use the sneak key to move slowly while on a horse (making a different hoof sound), and I wasn't sure if this is supposed to represent "sneaking" on a horse or just riding it slowly. So I chose the somewhat unwieldy but I thought technically more accurate name.

Allofich:
Actually I just doublechecked and crouching does _not_ slow you down enough to get the stealth effect. I think I was under the impression it did when I named the variable, or maybe it was just the horse thing. Anyway maybe it should be renamed to isSneaking after all.

ajrb:
I think it might be less confusing, but not sure I'm any kind of authority since I didn't realise that DF had a separate stealth key/function instead of the std crouch == stealth that most games use. :-)

PR #575, opened 2018-01-06
TextureReplacement: Simplification of Image size methods (including XmlManager), other improvements 

TheLacus:
* I simplified XmlManager as it was getting out of hand. Now TextureReplacement has all the methods that deals with size of UI images, which i refactored in the process.
There are a few lines which only differs for a method name, sorry for that.
* MipMaps is enabled for .XXX textures and disabled for img, cif and rci. Is this ok or there is something more specific i should consider?
* Size of MiscFlats atlas is 4096 instead of 2048 when asset import is enabled.
* Minor cleanup and optimization.
* Allow imported emission map to be tinted with window color.

### DIFF
--- a/Assets/Scripts/Game/UserInterface/FacePicker.cs
+++ b/Assets/Scripts/Game/UserInterface/FacePicker.cs
@@ -104,9 +104,9 @@ namespace DaggerfallWorkshop.Game.UserInterface
             if (currentFaceTexture != null)
             {
                 if (raceGender == Genders.Male)
-                    facePanel.Size = TextureReplacement.GetSizeFromTexture(currentFaceTexture, raceTemplate.PaperDollHeadsMale, faceIndex);
+                    facePanel.Size = TextureReplacement.GetSize(currentFaceTexture, raceTemplate.PaperDollHeadsMale, faceIndex);
                 else if (raceGender == Genders.Female)
-                    facePanel.Size = TextureReplacement.GetSizeFromTexture(currentFaceTexture, raceTemplate.PaperDollHeadsFemale, faceIndex);
+                    facePanel.Size = TextureReplacement.GetSize(currentFaceTexture, raceTemplate.PaperDollHeadsFemale, faceIndex);
 
                 facePanel.BackgroundTexture = currentFaceTexture;
             }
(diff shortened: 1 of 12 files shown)


TheLacus:
Interkarma, i noticed UI components are assigned their size in update(). The base code used to look like this:
`Size = new Vector2(CrosshairTexture.width * CrosshairScale, CrosshairTexture.height * CrosshairScale);`

But in order to allow imported texture there is also a check for reading size from xml, for which i wrote a method:
`Size = TextureReplacement.GetSize(CrosshairTexture, defaultCrosshairFilename, CrosshairScale, true);`

I'm not sure why this is done in update since i believe there is no need to to it more than once, can we move it in constructor?

Nystul-the-Magician:
I think it is necessary for windowed mode (or whenever window size changes) - I remember me implementing automap with doing exactly what you suggest and I unintentionally broke window behaviour for window resizing completely

Interkarma:
Nystul is correct, this is for dynamically sizing frames. This can even happen in editor mode (e.g. maximise play window or resize editor).

For most components, size calculations all happen in BaseScreenComponent. If texture is loaded from Setup() as designed, the dynamic sizing should take the injected texture into account. Not much to do here provided texture is set correctly at setup time, everything else should take care of itself.

It gets bit more complicated for some elements (mainly HUD) where the components aren't parented directly to the native panel, rather they "float" on top and get scaled and positioned depending on user's display. I've done this to allow for higher resolution elements to overlay the native UI. A good example is the quest debugger HUD which runs at 2x resolution compared to native. It also allows some HUD elements (e.g. vital signs) to account for different ratio widescreens, something classic never had to deal with.

The crosshair is likely a candidate for re-parenting to native panel and removing dynamic scaling from its own component implementation. This will solve the specific example you have above, but at the cost of using higher-resolution crosshairs (as they now exist inside the native 320x200 panel area).

I agree completely we don't want to be reading from XML every frame, but I don't have a good general-purpose solution for this right now. One possible solution is to run size calcs per frame, but only read from XML when a new texture is injected (e.g. texture reference changed from property). The crosshair is likely a good testbed for this as its a very simple component to experiment with.

TheLacus:
Ah, i was sure there was a catch. Thank you for your answers, i'll separate xml reading from size calculations and assignment and see how it goes then.

Interkarma:
I love your constant improvement of this whole system. Thank you, and well done. 👍

PR #578, opened 2018-01-17
Calculate fake sky Y position with Y-shearing

afritz1:
With this method, the fake horizon now moves vertically at the same speed as the real horizon, so they should always appear "glued" together. Although having a 2D background against a 3D foreground might still make some people sick, it's about as good as it's going to get without stretching somehow (see the backgrounds in the original Doom).

Another idea I had for improving the fake sky was to have its scale be a function of field of view, because as it stands, the fake horizon has the same size on-screen regardless of what the field of view is.


Interkarma:
This is definitely a good incremental improvement! Thank you for taking the time to add this. 👍

PR #579, opened 2018-01-19
TextureReplacement: add option for compressed textures

TheLacus:
Imported textures use DXT5 if requested by user settings.


Interkarma:
Right on! You put this together very quickly. Looking forward to seeing how this can reduce memory usage for folks with larger texture mods.

PR #581, opened 2018-01-20
Allow aquatic creatures to swim around in water

Allofich:
This lets aquatic creatures move around freely while in dungeon water.

Classic has odd behavior where aquatic creatures only move horizontally. (I think they might be able to move vertically by slight amounts, not sure) I think creatures being able to move around freely like the player is the expected behavior, though.

There's one known issue. If you lead an aquatic creature to a (not overly steep) slope leading out of the water, it will follow you and get stuck above the water, unable to move. I think this happens because even when it's near the water surface it can still move horizontally, and when it does this the sloping floor pushes it up out of the water.

It's probably a pretty rare case, so if there's a cheap and simple solution for it, that would be good. Similarly we need to stop enemies somehow from walking off of ledges like they do now.

Also, Interkarma, I'd like to play with the new water effects, but how do I enable them?

### DIFF
--- a/Assets/Scripts/Game/EnemyMotor.cs
+++ b/Assets/Scripts/Game/EnemyMotor.cs
@@ -37,6 +37,7 @@ namespace DaggerfallWorkshop.Game
         float giveUpTimer;                          // Timer before enemy gives up
         bool isHostile;                             // Is enemy hostile to player
         bool flies;                                 // The enemy can fly
+        bool swims;                                 // The enemy can swim
         int enemyLayerMask;                         // Layer mask for Enemies to optimize collision checks
 
         bool isAttackFollowsPlayerSet;              // For setting if the enemy will follow the player or not during an attack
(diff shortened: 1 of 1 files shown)


Interkarma:
That's awesome! I'm not sure I noticed aquatic creatures had trouble moving around in the first place. Shows how observant I am sometimes...

For water plane, start the game and drag DungeonWater prefab into scene view. Position and scale as needed. I'll add this to scene layout builders in next few days so it's automatically scaled and positioned as required. Right now just focusing on rendering effects. It should fall into place fairly quickly from here. You've done all the hard work. :)

Nystul-the-Magician:
played around a bit, noticed the mirrorreflection script - although seems currently not to be used - at least when I add the water prefab.
Might spend some time once water implementation is finished - so that reflection mod might inject some behaviour for reflections of water - so that floor and water reflection could potentially share the reflection texture some day.
but do your thing for the moment. since mod is optional water must work without it anyway ;)

Allofich:
> I'm not sure I noticed aquatic creatures had trouble moving around in the first place.

At the moment (in master) they just walk around on the floor like a land-based character, and if they start suspended in the air/water, they will drop down to the floor like a land character.

From your Twitter pictures I guess you were testing in that part of Daggerfall Castle with the shallow water, so that might be why you didn't notice. :)

PR #582, opened 2018-01-22
Fix some reversed enemy walking animations

Allofich:
Enemy mobiles in the walking animation were reversed for the up-right, right and down-right animations. They showed as the left-looking versions.


Interkarma:
Thank you for continuing to look at this. :)

PR #583, opened 2018-01-22
Implement enemy turning

Allofich:
This implements enemy turning, modeled after classic.

In master branch right now enemies don't have to turn, they instantly move in the direction they want to move. With this PR they turn and then move.

### DIFF
--- a/Assets/Scripts/Game/EnemyMotor.cs
+++ b/Assets/Scripts/Game/EnemyMotor.cs
@@ -275,21 +275,23 @@ namespace DaggerfallWorkshop.Game
                 targetPos.y -= deltaHeight;
             }
 
-            // Get direction & distance and face target.
-            // If attacking, randomly do not do so so player has a chance to see
-            // attack animations other than those directly facing the player
+            // Get direction & distance.
             var direction = targetPos - transform.position;
             float distance = direction.magnitude;
 
-            if (mobile.IsPlayingOneShot() && !isAttackFollowsPlayerSet)
+            // If attacking, randomly follow player with attack.
+            if (mobile.Summary.EnemyState == MobileStates.PrimaryAttack)
             {
-                attackFollowsPlayer = (Random.Range(0f, 1f) > 0.5f);
-                isAttackFollowsPlayerSet = true;
+                if (!isAttackFollowsPlayerSet)
+                {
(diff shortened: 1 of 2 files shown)


Allofich:
One thing about this PR, is that using the slower classic update timing for the turning creates some jerkiness in enemy movement. It looks quite similar to the original game, so people who like the original game might like it, but a smoothed out version might look better. Maybe this PR should wait until after the upcoming release.

Interkarma:
I'll trust you and just merge. You rarely (if ever) get this stuff wrong. :) I'll let you know if I feel changes cause any problems.

Interkarma:
I've noticed that it's fairly easy to run behind an enemy now and stand still - they completely lose track of the player. This imp is totally has no idea where I went lol.

imp-confused

I also feel the turning is perhaps a bit slow. It's trivial to strafe around enemies now, and they feel very tank-like in their turning. Just a bit of tuning needed, I think. :)

Allofich:
There's definitely some tuning/fixing to be done here, I'm just not sure at the moment what it is.

I made some comparison videos.
[ComparisonVideos.zip](https://github.com/Interkarma/daggerfall-unity/files/1691572/ComparisonVideos.zip)
EnemyTurnClassic/EnemyTurnDFUnity - This is the kind of testing I was doing to get the turn rate right. If you compare these videos you'll see the enemy's turn rate after being hit is about the same.
CirclingEnemyClassic/CirclingEnemyDFUnity - Here is where more of the issues are. I tried circling an imp with a quick character in both classic and DF Unity. It's harder to do this in classic because the movement isn't as smooth, and it seems more difficult to maintain a tight circle while strafing and turning at the same time. Also it's easier to get "caught" on the enemy's collision. In DF Unity the imp definitely can't keep up as well, but I don't know if this is because the imp is turning more slowly or because the player was moving faster/in a tighter circle. Also at the end of the video the imp has lost track of me, which is something that should be fixed. If it hadn't been able to keep up in classic, though, it might also have lost track. I'll try to get around to tuning this stuff at some point.

Interkarma:
Awesome, thank for looking into it. Yeah, I definitely cut in tight circles around him, something I could never do in classic. I do genuinely appreciate the work on this and the classic-like feel you're adding. I just needs a little tuning as you say. :)

PR #584, opened 2018-01-22
Enable player interaction with shelves

ajrb:
Shop shelves have loot containers and initiate trade. (random items for now)
Libraries are not shops - reverted my previous 'correction' and corrected existing code instead
Library shelves are marked as bookshelves and implemented reading (temples & guilds too)
Moved shelf stocking to building interior layout to avoid restock when empty. 
Added member message for guilds and temple bookshelves

TODO: Serialization

### DIFF
--- a/Assets/Scripts/DaggerfallUnityEnums.cs
+++ b/Assets/Scripts/DaggerfallUnityEnums.cs
@@ -536,6 +536,7 @@ namespace DaggerfallWorkshop
         RandomTreasure,
         CorpseMarker,
         DroppedLoot,
+        Shelves
     }
 
     /// <summary>
--- a/Assets/Scripts/DaggerfallUnityEnums.cs
+++ b/Assets/Scripts/DaggerfallUnityEnums.cs
@@ -536,7 +536,6 @@ namespace DaggerfallWorkshop
         RandomTreasure,
         CorpseMarker,
         DroppedLoot,
-        Geometry,
     }
 
     /// <summary>
--- a/Assets/Scripts/Game/EnemyMotor.cs
+++ b/Assets/Scripts/Game/EnemyMotor.cs
@@ -275,21 +275,23 @@ namespace DaggerfallWorkshop.Game
                 targetPos.y -= deltaHeight;
             }
 
-            // Get direction & distance and face target.
-            // If attacking, randomly do not do so so player has a chance to see
-            // attack animations other than those directly facing the player
+            // Get direction & distance.
             var direction = targetPos - transform.position;
             float distance = direction.magnitude;
 
-            if (mobile.IsPlayingOneShot() && !isAttackFollowsPlayerSet)
+            // If attacking, randomly follow player with attack.
+            if (mobile.Summary.EnemyState == MobileStates.PrimaryAttack)
             {
-                attackFollowsPlayer = (Random.Range(0f, 1f) > 0.5f);
-                isAttackFollowsPlayerSet = true;
+                if (!isAttackFollowsPlayerSet)
+                {
+                    attackFollowsPlayer = (Random.Range(0f, 1f) > 0.5f);
+                    isAttackFollowsPlayerSet = true;
+                }
             }
-            else if (!mobile.IsPlayingOneShot())
+            else
                 isAttackFollowsPlayerSet = false;
 
-            if (!mobile.IsPlayingOneShot() || !attackFollowsPlayer)
+            if (attackFollowsPlayer)
                 transform.forward = direction.normalized;
 
             // Bow attack for enemies that have the appropriate animation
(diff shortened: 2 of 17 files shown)


ajrb:
Looks like there's a conflict with the new geometry container type you've added. Shelves need to be their own type, and I was planning other types for other fixed geometry containers like boxes and wardrobes etc. In buildings these sometimes allow you to look and steal the contents (they're private and just looking is a crime which is BS IMO) Not started this yet because I'm finding it hard to see the pattern of which are containers and which are not in classic. (maybe Allofich can illuminate me?)

Anyway, I'll need convincing that a single generic geometry type of container is a better approach. :-)

ajrb:
Allofich, are you still okay to do the stocking code port from original? It will need to go into the DaggerfallLoot.StockItems() method.

Interkarma:
>Anyway, I'll need convincing that a single generic geometry type of container is a better approach. Assuming I've not misunderstood what it's for. Probably have havn't I? :-)

The LootContainerTypes enum was originally meant to be a broad category to inform the save/load manager what general type of container it's dealing with. Right now it just informs SerializableLootContainer whether to inject a billboard at load or not, or remove container once empty. It wasn't originally intended to be a super-granular thing.

We can either add a new sub-type for geometry-based containers (which can be as granular as required) or keep building out LootContainerTypes. I'm flexible either way here. :)

If we build out LootContainerTypes, we'll just need to rephrase the if statements at SerializeableLootContainer.cs line 87 and GameObjectHelper.cs line 722 to follow correct behaviour for billboard types.

Interkarma:
I think I lean more towards a granular sub-type at the moment. It means we can spec this out with as much detail as needed, for example StoreShelves vs DungeonShelves vs HouseShelves.

The loot container itself really just wants to know whether it's dealing with a treasure pile, corpse marker, or geometry for scene management reasons. It doesn't ultimately care what the player is doing with the container contents. It just want to *contain things*, man.

ajrb:
I wasn't really considering scene management, but still not convinced we need two type variables.

Looking into what you're saying, it seems that the if statements can be simplified to inclusive of RandomTreasure and DroppedLoot for the GameObjectHelper.RemoveLootContainer() method and the billboard logic conditional on the TextureRecord/TextureArchive being set in the loot container.

Is that problematic for some reason I can't see because I tried thinking of meaningful names for two type variables and I couldn't. To me it would make more sense to have one type and change the conditionals.

Your call though, this is not something I care enough about to die on a hill for. ;-) Either will be fine, but now you have my pov on it.

Interkarma:
> Your call though, this is not something I care enough about to die on a hill for. ;-) Either will be fine, but now you have my pov on it.

Haha! Yeah, it's not a big deal either way mate. That's fine, go ahead and populate LootContainerTypes as needed. :)

I'll rework my commit now to remove that conflict and fix up the conditionals.

Allofich:
> @Allofich Are you still okay to do the stocking code port from original?

I am!

> I'm finding it hard to see the pattern of which are containers and which are not in classic. (maybe Allofich can illuminate me?)

I've mostly got it figured out.
In classic it's part of the same function called when you click on 3d statics (the same one that gives you store hours if you click a building in info mode). Here's pseudocode.
```
If (RayCastHitDistances <= 128) // Daggerfall distance units. Multiply by GlobalScale in Meshreader.cs
{
      building = GetBuildingOfObject(hit3dObject);      
      if (building != null)
      {
           byte Arch3DObjectGroup = hit3dObject.Arch3DObjectID >> 7; // If ObjectID used for loading from ARCH3D.BSA is 41002, get 410
           byte Arch3DObjectGroupIndex = hit3dObject.Arch3DObjectID & ff; // If ObjectID used for loading from ARCH3D.BSA is 41002, get 2
           if Arch3DObjectGroup  == 418 ||
           Arch3DObjectGroup == 410  && (IsHouseContainerGroupIndex(Arch3DObjectGroupIndex))
                {
                            Do stuff for opening containers in houses
                }
      }
}

bool IsHouseContainerGroupIndex(byte index)
{
        return index == 3 || index == 4 || index == 7 || index == 8 || index == 50 || index == 51 || index >= 32 && index <= 38;
}

```

There's another part to the condition, but I'm not sure what it's doing. If I figure it out and it's relevant, I'll send in a PR or post something.

ajrb:
That's so awesome, and means I can implement classic behaviour nice and easily. Amazing!

Allofich:
Whoops, what was I thinking. >> 7 won't give you the right value.

Updated the post.

ajrb:
Refactoring the PlayerActivate class code for handling loot containers FYI.

Also implementing house containers.

ajrb:
Well crap, this went wrong when I synced my fork with main. Really screwed up this PR. Closing and will create a new one. Git is a git sometimes!

Interkarma:
> I've mostly got it figured out.
In classic it's part of the same function called when you click on 3d statics (the same one that gives you store hours if you click a building in info mode). Here's pseudocode.

Awesome job at this stuff, don't know what else to say other than thank you. :)

PR #588, opened 2018-01-24
Refine temporary shop shelf stocking to be more sensible for next build

ajrb:
Since you're making a final 0.4 build in next day or so, I though the shelf stocking should be refined to be vaguely sensible so I threw together a quick algorithm to tide it over until Allofich implements the real one from classic. (@Allofich didn't want to pressure ya mate)

Should probly have done this from the start, was me being lazy.

### DIFF
--- a/Assets/Scripts/Game/Items/ItemBuilder.cs
+++ b/Assets/Scripts/Game/Items/ItemBuilder.cs
@@ -302,6 +302,20 @@ namespace DaggerfallWorkshop.Game.Items
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


Interkarma:
Awesome, cheers! I'm about to head to bed, but it looks simple enough to merge on just a read. Will check it out properly tomorrow when I can.

ajrb:
Note that general stores and pawn shops stock horse & cart but the purchasing of these special items needs a bit of work. Will do that soon.

PR #589, opened 2018-01-24
Fix dying from fall damage when fast traveling

Allofich:
In master branch it's possible to die from fast traveling if you hit the fast travel key while you were in the air (jumping for example) and the destination is lower elevation than the place you traveled from.

This fix works by clearing fall damage when teleporting. We might end up needing a different fix, because the player is also teleported to the correct position when loading a saved game, and you should still get fall damage if you saved mid-fall. You don't get any fall damage in master branch either, though, so this isn't a regression.

Or, if we can change the loading code to not use a teleport, maybe we can continue to use this fix.

### DIFF
--- a/Assets/Scripts/Game/PlayerEnterExit.cs
+++ b/Assets/Scripts/Game/PlayerEnterExit.cs
@@ -846,8 +846,8 @@ namespace DaggerfallWorkshop.Game
             Ray ray = new Ray(transform.position, Vector3.down);
             if (Physics.Raycast(ray, out hit, controller.height * 2f))
             {
-                // Cancel movement so player doesn't take falling damage if they transitioned into a dungeon while jumping
-                GameManager.Instance.PlayerMotor.CancelMovement = true;
+                // Clear falling damage so player doesn't take damage if they transitioned into a dungeon while jumping
+                GameManager.Instance.PlayerMotor.ClearFallingDamage();
                 // Position player at hit position plus just over half controller height up
                 Vector3 pos = hit.point;
                 pos.y += controller.height / 2f + 0.25f;
--- a/Assets/Scripts/Terrain/StreamingWorld.cs
+++ b/Assets/Scripts/Terrain/StreamingWorld.cs
@@ -924,6 +924,8 @@ namespace DaggerfallWorkshop
             autoRepositionOffset = repositionOffset;
             autoRepositionMethod = autoReposition;
             InitWorld();
+            // Clear falling damage so player doesn't take damage if they were in the air before a transition
+            GameManager.Instance.PlayerMotor.ClearFallingDamage();
             RaiseOnTeleportToCoordinatesEvent(worldPos);
         }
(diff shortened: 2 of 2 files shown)


Allofich:
Also this replaces an earlier fix I did for the same problem of dying when loading dungeon interiors while jumping, using the clear fall damage function, which is more to the point, instead of canceling movement.

ajrb:
That bug almost seems amusing enough to leave in. haha

Allofich:
I noticed it because of Galactic Chicken's video
https://www.youtube.com/watch?v=sbgUUC1fiC0
at 44:00. That could really suck if you hadn't saved your game recently and got killed that way :)

Another issue from that video to maybe look at is the "You have no arrows." message sticking around (around 45:00). It's not supposed to stick around that long. I may look at that tomorrow.

ajrb:
Ah, hadn't quite got that far in the vid. Gotta hand it to him, he finds a lot of issues for us. Invaluable having someone playing so much and documenting it.

Allofich:
> I assumed the message stuck around because he was repeatedly trying to activate the bow.

That's probably what it was. I haven't tried to reproduce it yet. If that's the case it's not a problem, it's working as intended.

> Gotta hand it to him, he finds a lot of issues for us. Invaluable having someone playing so much and documenting it.

Yeah, also I enjoyed how he got all excited about the dog barking, since animal noises is something I added recently. I don't usually see a person's live reaction to something I add to the project, so that was fun. :)

ajrb:
> Yeah, also I enjoyed his reaction to the dog barking, since animal noises is something I added recently. I don't usually see a person's live reaction to something I add to the project, so that was fun. :)

Yes, a lot of the work you do is back end stuff so with this and the water must be nice to see reaction to your work.

Watched the rest of the vid now and I think the no arrows may be a problem, when it happened earlier in the vid it seemed due to trying to get the bow out but later is just stuck on.

Interkarma:
> That's probably what it was. I haven't tried to reproduce it yet. If that's the case it's not a problem, it's working as intended.

I can reproduce most easily by starting a level 1 character and equipping the freebie bow they receive from initial loadout. The message hangs around until player unequips the bow or finds some arrows. NBD though, it's very minor and can be fixed anytime.

> Yeah, also I enjoyed his reaction to the dog barking, since animal noises is something I added recently. I don't usually see a person's live reaction to something I add to the project, so that was fun. :)

You don't get enough credit for the work you do. :) I try to highlight your work on Twitter. but it's awesome you finally got to see some direct visual feedback. My next blog post will be all about you guys and the work you've all done over the last few months. The only reason I haven't done this already is the good stuff keeps rolling in. A nice problem to have. :)

> Ah, hadn't quite got that far in the vid yet. Gotta hand it to him, he finds a lot of issues for us. Invaluable having someone playing so much and documenting it.

Very true, and I don't give him enough focus either. I'll try to do better.

PR #594, opened 2018-01-31
Implement classic stock for private property

Allofich:
Also has minor cleanup to store shelf stock code.


Interkarma:
I've decided you must be some kind of wizard. :) I like how you've matched the interior furniture types by building type. Is this something else you've worked out from classic disassembly?

ajrb:
Needing the model Id means it will need to be stored with the container since I'm moving stocking back to first activation. Probably toss it in textureRecord field or something.

Allofich:
You mean used the building type as the index? Yes that's how the original is doing it. Doing it the same way allowed keeping the code fairly compact.

Interkarma:
Yep. I assumed it must have come from classic by how specifically it was constructed. Nice work!

PR #601, opened 2018-02-04
Tavern room rental system fully implemented

ajrb:
It works really nicely, hope I've not missed anything.

Hope you don't mind using the serialization object to store these things in player entity. Nothing needs transforming.

### DIFF
--- a/Assets/Scripts/Game/Entities/PlayerEntity.cs
+++ b/Assets/Scripts/Game/Entities/PlayerEntity.cs
@@ -22,6 +22,7 @@ using DaggerfallWorkshop.Game.Utility;
 using DaggerfallWorkshop.Utility;
 using UnityEngine;
 using System;
+using DaggerfallWorkshop.Game.Serialization;
 
 namespace DaggerfallWorkshop.Game.Entity
 {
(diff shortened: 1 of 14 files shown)


Interkarma:
That was fast! Great stuff. Will get this merged when I can around work today.

ajrb:
I've been really enjoying programming DFU for last couple of weeks, it's all I've wanted to be doing when I have free time. (and even when I don't)  I'm rather binary when it comes to motivation - either full on like this, or off like over xmas. :-)

I wanted there to be a way for players to have persistent dropped loot before having 100k gold to blow on a boat... now there is.

Interkarma:
And I've enjoyed everything you've created. :) I'm much the same with motivation as well. I can either find the headspace or I can't. Here's looking forward to more awesome collaboration in the future.

Interkarma:
This is great! 👍 

Interesting thing - character might not wake up on top of a rest marker inside all taverns. For example Daggerfall/The Rusty Ogre Lodge has a single tavern called The White Goblin. Whenever I rest here, player just wakes up on the same spot they initiated rest. The tavern appears to have rest markers as normal, the code all looks right, and it works everywhere else I've tested so far. Not sure what's going on in the above tavern.

Let me know if you get the same thing.

ajrb:
That's interesting, I will investigate it this evening. Thanks for the report.

PR #606, opened 2018-02-08
Fix enemies losing track of player at close range

Allofich:
This should fix enemies losing track of the player at close range, and also mitigates the problem of them turning too slow, not by making them turn faster but by having them turn more often to keep the player in sight. This behavior should also be a little closer to classic. It may need more tuning but it seemed to work fairly well in testing.


Allofich:
Also put in some comments for where spell casting can go to mimic classic combat decision making, when that is supported.

Interkarma:
Thank you for taking another pass at this! 👍

PR #607, opened 2018-02-08
Implement classic tavern room costs

Allofich:
This implements the following, matched to classic:

1) Same cost formula, including free for Heart's Day
2) Commented out code for free lodging for knights, once we have guilds implemented
3) Different messages depending on how well player bargained
4) Limit on renting days into the future includes days already rented.

Regarding 3, @ajrb, this causes code duplication with the trade window. Classic uses the same function for the display of the Yes/No bartering window. Since you wrote both the trade and tavern windows, maybe you can look at combining them. I started to play with doing so but it looked like it could get messy and I think you'd have a better idea of how to do it.

Btw there's a bug/oversight here in classic. The way the bartering message works is that you get one depending on how low the price after bargaining is compared to the original cost, but it seems classic doesn't set the "before bargaining" cost for the tavern room dialog, so what message you get depends on what the original cost value of your last trade negotiation was (or is 0, I assume, if you're starting off of a freshly loaded game). For this PR I properly set the cost for the tavern dialog.

### DIFF
--- a/Assets/Scripts/API/Save/CharacterRecord.cs
+++ b/Assets/Scripts/API/Save/CharacterRecord.cs
@@ -79,6 +79,7 @@ namespace DaggerfallConnect.Save
             doc.startingLevelUpSkillSum = parsedData.startingLevelUpSkillSum;
             doc.minMetalToHit = parsedData.minMetalToHit;
             doc.armorValues = parsedData.armorValues;
+            doc.lastTimePlayerAteOrDrankAtTavern = parsedData.lastTimePlayerAteOrDrankAtTavern;
             doc.lastTimePlayerBoughtTraining = parsedData.lastTimePlayerBoughtTraining;
             doc.timeForThievesGuildLetter = parsedData.timeForThievesGuildLetter;
             doc.timeForDarkBrotherhoodLetter = parsedData.timeForDarkBrotherhoodLetter;
@@ -171,7 +172,8 @@ namespace DaggerfallConnect.Save
             reader.BaseStream.Position = 0x1fd;
             parsedData.timeStamp = reader.ReadUInt32();
 
-            reader.BaseStream.Position = 0x209;
+            reader.BaseStream.Position = 0x205;
+            parsedData.lastTimePlayerAteOrDrankAtTavern = reader.ReadUInt32();
             parsedData.lastTimePlayerBoughtTraining = reader.ReadUInt32();
 
             reader.BaseStream.Position = 0x211;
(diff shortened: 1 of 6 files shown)


Allofich:
Hmm, actually, I think I may do this differently.

It looks like classic resets the original cost value to 0 when the inventory menu is opened. At any rate, it's likely it will be 0 and you'll get the "worst" bargaining message when you ask for a room.

The thing is, if we set it "correctly," because of the way it works you are likely to get a "bargained well" message if you ask for a single day, which is probably common. The messages are kind of silly for this, like "You have bargained with me until the sun is low in the sky" etc., for a 7 gold piece room. I think maybe just enforcing the "bad" bargain, which is a straightforward "-- is my final offer," may actually be closest to classic.

ajrb:
I'm surprised that rooms are a flat 7gp / day, as I expected that tavern quality and player mercantile skill & personality stat would affect the price - like in trading, and as the messages imply haggling.

Allofich:
It does involve bargaining, take another look :)

Also I changed it to give the more sensible-sounding message. There may or may not be a need to merge in the display functions now.

ajrb:
Oh, you reused the existing function (as you should) - didn't see that. Happy to go with whatever messages you think are most appropriate. I will let you know if I see any that seem inconsistent.

Allofich:
@ajrb I think I'll go ahead and implement the food and drinks menu, including import of relevant data from classic, etc. Have you already done it locally or want to do it yourself?

Allofich:
I couldn't wait and went ahead and did it. I hope it won't cause any merge conflicts or anything for you.

ajrb:
That's fine, it was one of the next things I was planning to do but I hadn't got to it yet. Your implementation looks much better than mine would have been so that's serendipitous.
I will look at combining duplicate code and saving stuff into DFU savefiles once this is merged into master.

ajrb:
Oh, and it's very conscientious to import the minor states like last time ate or drank from classic saves. I've not done that for loot piles or rented rooms. I figure that you need the important stuff like character & items & quests as well as discovered buildings etc. But stuff like rented rooms has a far too small return on investment of time IMO... still, if you want feel free to add those as well.

Allofich:
I'll see about adding import of rented rooms in a while. I don't know if we'll ever get there but personally I'd like to eventually get all data from classic saves imported. But yeah, functionality is the most important.

Interkarma:
Awesome work, cheers! 👍

PR #609, opened 2018-02-10
Spelling changes for consistency and faction dialogue bug fix

numidium:


Nystul-the-Magician:
not sure why the spelling is an issue, especially us is standard anyway - see names like "color" - would like to see this reverted to be honest

Interkarma:
I agree with Nystul. This kind of spelling change doesn't add anything of value and could potentially break mod scripts as they compile at runtime with no awareness of the change in compiled base.

I'm from a Commonwealth country and generally use US spelling in code and my native spelling in comments. Even Unity itself uses a mixture of Commonwealth (e.g. MonoBehaviour) and US (e.g. Color) spelling. I don't feel it's necessary to modify the code of other just based on regional preference.

Could I please ask you to revert the spelling changes and just submit bug fix? Thank you numidium. :)

Assets/Scripts/Utility/TextProvider.cs
```diff
@@ -267,6 +267,9 @@ public virtual TextFile.Token[] GetRandomTokens(int id)
             // Select a random token stream
             int index = UnityEngine.Random.Range(0, tokenStreams.Count);
 
+            // Select the next to last item from the array if the length of the last one is zero
+            index = (tokenStreams[index].Length == 0 ? index - 1 : index);
```
Interkarma: I'm not overly familiar with faction dialogue. Could you give me an example of the text IDs where this is an issue that needed a fix?

As this change would be universal to all random tokens, are there any cases elsewhere where it's possible for this to end up as -1 (e.g. index 0 also has a length of 0).

numidium: The error would occur for record ID 871 (860 + 11 (B. of Mara info)). The last tokenStreams array item would contain a blank array. Therefore, expandRandomTextRecord errors out because it tries to return tokens[0].text which doesn't exist.

It's possible in theory that a text record could have a blank first token but that would cause an out-of-range exception even without my fix. So far I've only seen a blank last item in this one case with faction dialogue. As long as there's not some strange malformed set of tokens buried deep in TEXT.RSC that has a blank first item then we should be ok.

Nystul-the-Magician:
btw, you should add again your name in the contributors list numidium, it belongs there because you implemented the organization info stuff ;)

PR #610, opened 2018-02-10
Implement climbing

Allofich:
This implements climbing. It's a mix of stuff I was able to figure out from classic (skill-checking formulas) and trial and error to get something that worked decently. I'm sure this still needs work but I think it's working well enough as a first iteration.

### DIFF
--- a/Assets/Scripts/Game/FakeLevitate.cs
+++ b/Assets/Scripts/Game/FakeLevitate.cs
@@ -26,6 +26,7 @@ namespace DaggerfallWorkshop.Game
         PlayerMotor playerMotor;
         Camera playerCamera;
         float moveSpeed = 4.0f;
+        CollisionFlags collisionFlags = 0;
 
         public bool IsLevitating
         {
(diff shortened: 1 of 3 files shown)


Allofich:
For some reason collision flags retrieved from the Move function are working while out of water, but not while in the water. I can't figure that one out right now.

ajrb:
I had a thought about climbing that we should change it to be activated by a key (or a mode that's toggled by one) instead of the automatic climbing in classic which was a pain in the ass if you brushed against some geometry and got stuck 'climbing'. Lets see if it works better in DFU first I guess.

Allofich:
I'm not sure if I've ever had that problem in classic. One situation I can imagine it maybe happening is if you back into or walk sideways into a wall and continue pressing against it for a second or two because you're distracted by combat or something. You can actually climb in classic while moving sideways or backwards into the surface to climb on. In this PR you have to use the forward key to climb, although you can still look 360 degrees while you're climbing. Hopefully this will prevent that problem.

One thing about adding keys, is that we support changing keys through classic's configuration screen, but if we added anything beyond that it would not be configurable there.

ajrb:
It may be only the clumsy players among us that encounter that problem in classic! I know it happens to me a lot when walking up stairs in buildings. Looking forward to trying this out. Climbing is another of the signature aspects of daggerfall.

Interkarma:
This looks awesome. I'll merge this and have a play with it soon. Great work! :)

PR #611, opened 2018-02-10
Save breath and last ate into savedata

ajrb:
Also moved strings for later localisation


Interkarma:
It's a bummer how Daggerfall spreads text over so many places. Makes localisation very difficult to do properly.

PR #612, opened 2018-02-11
Merge quest dispensing into guild service code

ajrb:
Finish merging quest dispensing into guild service code now that we're onto 0.5. 
Should appear completely unchanged to players.
Disabled old code, but leaving in for now so can switch back if issues found.

One thing that that is changed is that in PlayerActivate lines 860-880 the trigger of quest resource behaviour and quest listener check are both now done before the quest dispensing is activated, rather than after. This doesn't seem to affect anything as far as I can see, but I may have missed something so please let me know if I need to preserve the ordering for any reason.


ajrb:
@Interkarma No rush to get this in, it can wait until you feel like taking a look as I have nothing else further I want to do on this immediately.

Interkarma:
> One thing that that is changed is that in PlayerActivate lines 860-880 the trigger of quest resource behaviour and quest listener check are both now done before the quest dispensing is activated, rather than after. This doesn't seem to affect anything as far as I can see, but I may have missed something so please let me know if I need to preserve the ordering for any reason.

I can't think of anything this would break right now. If everything appears to work correctly on your end, it's probably all good. I'll you know if I can find any problems. :)

PR #613, opened 2018-02-11
Add treasure marker containers

ajrb:
Using table entry 'O'

### DIFF
--- a/Assets/Scripts/Game/Items/LootTables.cs
+++ b/Assets/Scripts/Game/Items/LootTables.cs
@@ -123,7 +123,7 @@ namespace DaggerfallWorkshop.Game.Items
                 DaggerfallLoot.GenerateItems(lootTableKeys[locationIndex], loot.Items);
 
                 // Randomly add map
-                char key = lootTableKeys[locationIndex][0];
+                char key = lootTableKeys[locationIndex]['O'];
                 int alphabetIndex = key - 64;
 
                 if (alphabetIndex >= 10 && alphabetIndex <= 15) // between keys J and 0
--- a/Assets/Scripts/Game/Items/LootTables.cs
+++ b/Assets/Scripts/Game/Items/LootTables.cs
@@ -123,7 +123,7 @@ namespace DaggerfallWorkshop.Game.Items
                 DaggerfallLoot.GenerateItems(lootTableKeys[locationIndex], loot.Items);
 
                 // Randomly add map
-                char key = lootTableKeys[locationIndex]['O'];
+                char key = lootTableKeys[locationIndex][0];
                 int alphabetIndex = key - 64;
 
                 if (alphabetIndex >= 10 && alphabetIndex <= 15) // between keys J and 0
--- a/Assets/Scripts/Game/Items/LootTables.cs
+++ b/Assets/Scripts/Game/Items/LootTables.cs
@@ -126,7 +126,7 @@ namespace DaggerfallWorkshop.Game.Items
                 char key = lootTableKeys[locationIndex][0];
                 int alphabetIndex = key - 64;
 
-                if (alphabetIndex >= 10 && alphabetIndex <= 15) // between keys J and 0
+                if (alphabetIndex >= 10 && alphabetIndex <= 15) // between keys J and O
                 {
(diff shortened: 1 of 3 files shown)


Allofich:
I've been testing with both the save you mentioned on the forums and another character, and for the two buildings I tested (thieves guild and a regular house), it seems like its using loot table entry 'K', not 'O'. Could you change it over to that? I'm not sure about the potion and recipe chances, either.

Allofich:
I think I may know what is going on. See line 1451 in RDBLayout.cs. Dungeon type "crypt" = ID 0 gets loot table K. The houses I checked were in a "city" type location, also ID 0. In that case the map, potion and recipe chances should be as seen a little lower in that file, from line 1482, if the loot table key is within the J to O region.

ajrb:
Regular house? I had thought these were only present in TG and DB buildings. Not that it's an issue, as I implemented it to work anywhere the DF data have the treasure markers

Allofich:
If a treasure pile in a "TownHamlet" (1) location is loot table key N, then that would be strong evidence that this is correct.

ajrb:
Oh so you mean that the location type indexes the same loot table keys that dungeons do? (ln 1451 you pointed me to)

Allofich:
Seems to. I'm trying to find a loot pile in a house in a TownHamlet location right now.

Allofich:
The noble Hero of Daggerfall is currently barging into everyone's house uninvited, running into their attic, cursing under his breath when nothing is there, then moving on to the next. :)

ajrb:
I hope that some exist because that sounds awful, if slightly funny!

Might be quicker to figure out how to identify the markers in the datafiles and do a binary search... if you don't find one soon.

Allofich:
I finally found one, and it is not using K (no religious item), and the items match what is possible with N. I'll try the same house again with another character but I'm pretty confident this is correct now.

ajrb:
Cool well I've nearly finished coding it, I think. Hopefully I've understood what your theory is. You can check the code when I add to PR, and if I've got wrong end of the stick then you'll have to explain again.

Please add the house location here for posterity. Next time there may be guards! lol

Allofich:
The house is in Blackidge, Daggerfall.
fall_000

I checked it two more times and the contents still match with N, so I'm comfortable saying the theory is correct.

ajrb:
There you go, can you check the code and maybe try it out? I will revert if it's wrong. Have to go for a bit but will be back online later.

Assets/Scripts/Game/Items/LootTables.cs
```diff
@@ -93,6 +93,52 @@ public static LootChanceMatrix GetMatrix(string key)
...
+            // Get loot table key
+            if (locationIndex < lootTableKeys.Length)
+            {
+                DaggerfallLoot.GenerateItems(lootTableKeys[locationIndex], loot.Items);
+
+                // Randomly add map
+                char key = lootTableKeys[locationIndex][0];
+                int alphabetIndex = key - 64;
+
+                if (alphabetIndex >= 10 && alphabetIndex <= 15) // between keys J and 0
```
Allofich: This 0 should be an O, can you fix this typo please? :)

Allofich: The typo is from me, when I added this stuff in before.

Allofich:
I tried it out, and it seems to be working well!

Btw if you want to change up the items that can be spawned, I'm OK with that, but I would like the selection from classic to still be accessible for purists like me :). It seems like these selections of treasure were made with the dungeon types in mind, rather than the city location types. The religious items, for example, make sense for a crypt dungeon, but don't really make sense for the large settlements. Or just leave it as it is, that's fine, too.

Assets/Scripts/Game/Items/LootTables.cs
ajrb: I thought you meant the 0 in the index, changing that to 'O' broke any inerior with loot. whoops, reverted that crap. Done the correct change to the comment.  (couldn't see until I scrolled right)

ajrb:
I'm happy for these to act like classic.

Allofich:
Sounds good to me.

Interkarma:
Nice one, thank you for adding this!

PR #615, opened 2018-02-14
Implement language skills

ajrb:


Allofich:
I'd like to also implement language skills in a way that's the same as classic (don't know when yet). Is that all right?

I would add a setting for this. One question, though, is what to make the default option. So far we've mostly been having classic behavior as the default option. However, I'd feel kind of like a jerk putting ajrb's stuff that he just worked on aside as the "alternate" option. I think default options carry an air of being "recommended" even if you don't explicitly say so.

Interkarma if you could give your opinion about the direction to take for this, I'd appreciate it.

I've also been thinking about if we could have an overall "classic mode". There are lots of little nitpicky things in DF Unity already where behavior is a little different from classic and I'd like to have the option to have classic behavior across-the-board (just with bug/oversight fixes).  For example, disabling horse jumping, or restoring that dungeon song that didn't fit in with the others (personally I like that song). 

And then, on the other hand, I'm not sure that I want 100% classic behavior, either. For example, I've left it as a TODO and not gotten around to it, but I found out the monster list classic uses for graveyards a while ago. It's kind of weird, though, using lots of orcs, and I think I prefer the makeshift undead-heavy one you made, so I've been leaving that on the back burner.

Do you have an opinion about this kind of thing? Are you opposed to having a big list of settings for each little thing like this (the graveyard monster list, for example)? If we have a wide-reaching "classic mode" that covers most of these, that would involve sticking "if classicmode = true" or something in various spots around the code, would that be fine with you? This could cover changes like adding in the "You found -- gold pieces", which I assume you've intentionally left out.

I understand the changes ajrb made like expanding the types of monsters affected by languages because, I think most people would agree, the language skills aren't very useful in classic. I like to imagine that maybe we can eventually let the player convince enemies to become their followers or something. But also I'd like to always have classic behavior available if possible, and there are probably some fans of the original who feel the same way, although I haven't seen many people saying that on Twitter or the forums, etc. The people who are posting anyway seem to be happy with the DF Unity versions.

Well, anyway, any response is appreciated.

ajrb:
I'm not opposed to a big config switch for classic behaviour, but is it worth the effort & complexity? Probably worthwhile pulling all this together and starting a discussion in core dev forum. One upside to having a switch like this covering general behaviour is that I could go a little further correcting stuff like language skills knowing players can switch it off if they don't want it.

Allofich:
> I'm not opposed to a big config switch for classic behaviour, but is it worth the effort & complexity?

I think it could probably work, but that's because DF Unity has stayed fairly close to classic, so the changes could be mainly within the same framework. For large-scale changes, like for example totally replacing the way that the wandering NPCs in towns are spawned or something, that probably wouldn't be so feasible. But Interkarma would probably know better as he knows a lot more about how the project is structured and designed.

Interkarma:
I really think that some deviation from classic is perfectly OK. Especially when the change makes something more useful, more comfortable, or more sensible. 

I think it's important to separate elements from classic that are the result of a bug or oversight (e.g. skewed textures, floating objects, poor implementation, and bad game data in general) from elements that are part of the intended design, even if they're a bit quirky.

Using horse jumping as an example - was that an quirky design decision or a limitation of the motor implementation? In classic, you can levitate with a horse just fine but weirdly can't jump with one. Horses are much better known for jumping than levitating... but not with a cart, agree that should go.

For a ground-up implementation rather than a source port, I feel Daggerfall Unity is already very close to classic in overall feel and gameplay, something I confess you have a large part in realising. I'm outright not opposed to matching some of these nitpicky changes, but feel it could easily get out of hand and these restorations should be very selective based on value. What's the value of the item being restored vs. the cost in time to recreate something that could potentially just be a bug in classic? Is the value of this item equal to causing potential instability or difficulty of maintenance in future?

As long as that value exists, I'm not opposed to this option. But my feeling is that some might see this as a vector to demand reimplementation of straight-up buggy behaviours because they have a nostalgic connection to one particular bug or another. This is not a conversation I wish to be having across several points of social media. I have better things to do than defend against changes that add little value and expand maintenance complexity - so please make sure the value is actually there. :)

I describe Daggerfall Unity as "Classic Daggerfall Plus". I don't think anyone is expecting a 1:1 remake right down to the tacks. The important thing is to capture all the good stuff from Daggerfall (which is most of it) and limit some of the frustrations. I think we're achieving "Classic Daggerfall Plus" right across the board so far. And if people just want "Classic Daggerfall" exactly to the core, they always have DosBox.

ajrb:
> but not with a cart, agree that should go
Well, that's waiting for PlayerMotor rewrite, only left it like that since you wanted to minimise changes to the class. :-)

You make very good points, it could really open a can of worms. I agree with your approach, but I also want to ensure the more purist like Allofich are happy as well. I think it's good that we have different views, I probably would have gone a bit too far without being reigned in.

Allofich:
I think we're in agreement that trying to recreate bugs isn't the best use of our time, so I guess it's just a question of design. Well, maybe leave the question of doing a widespread "classic mode" off the table for now. If I have some time I may see how cleanly a classic implementation fits in alongside what ajrb has done. Maybe ajrb can then pick bits from it for his implementation if he wants them. ajrb, if you want to play with values, and Interkarma is fine with it, go ahead. If I do a classic implementation I will probably use the original values for that mode.

ajrb:
Currently it's like classic from what I know, except I added luck to the formula and made success a bit more likely. Would be interested if classic does use lang skills for other monsters, I did see a comment saying it did but there's hardly any info. Honestly, I doubt even the most experienced player will even notice it's different since these skills were so useless.

I still think that the language progression needs adjusting from classic, either make the number of uses less than 15, or allow them to be trained past 50%, or maybe even a combination of both. As they are even if you build your char around them you're going to hit 50% skill and that will be it pretty much. I think it would be justified to allow training to 90% since it's an easier skill to aquire through study rather than needing experience. I'd also like to drop the skill uses required from 15 down to 5.

Interkarma:
> Are people asking for bugs to be recreated?

Yeah. I had someone on reddit just the other day asking for a "classic mode" (exactly your phrasing) in regards to restoring buggy/exploitable container refill from classic.

https://www.reddit.com/r/daggerfallunity/comments/7va0nw/does_dfunity_still_retain_the_exploits_from_the/

The problem is that "classic mode" can mean different things to different people. It's only a matter of time before someone wants a particular bug to be reimplemented (e.g. poor container persistence) and I'm left defending why improving the container persistence was more valuable than implementing it the classic (broken) way. Or why implementing and maintaining it both ways only makes code more complex for no real gameplay benefit.

Having this option in without being extremely thoughtful about it and without communicating it very clearly, means that we're going to get those demands from all directions. As the frontman of the project, I'll be the one dealing with this across reddit, twitter, forums, email, etc. :)

> My wanting classic behavior is also actually about keeping things from getting out of hand. Since this is an open-source project, it gives a kind of "end goal". Not a real end, of course, but something to aim for, to keep things focused and avoid feature creep.

I can see this and I think it's a good approach when handled correctly. Wherever I've encountered a conflict between quality of life features and changing classic gameplay, I always revert to classic. Examples are action mode selection and spell cast/recast. This also stops the project from getting out of hand in the other direction. It can be a fine line to walk sometimes.

Allofich:
> I still think that the language progression needs adjusting from classic, either make the number of uses less than 15, or allow them to be trained past 50%, or maybe even a combination of both.

I've never tried a languages character so I can't really comment on the effects of that gameplay-wise. But if it makes languages more viable it's probably a good change. I think you should leave comments when you intentionally change something from classic. That's what I've been doing, anyway, when I correct something that seems like a bug or oversight.

ajrb:
Yep will do that, should we put a tag into these comments to mark them as behaviour changes so we can easily find them later?
e.g.
// BCHG: I upgraded classic behaviour

Interkarma:
Interestingly, a user just posted a bug report about not being able to pickpocket rats or endlessly pickpocket town NPCs.

http://forums.dfworkshop.net/viewtopic.php?f=24&t=871

Both of these changes were intentional refinements from what I can see (and I agree completely with your implementation Allofich). However, this user's perception is this change constitutes a bug in Daggerfall Unity.

As it stands, your implementation of pickpocketing is of high value. Adding an option to reproduce buggy or nonsensical behaviour like pickpocketing rats is not. You could spend that time better elsewhere adding new gameplay features. I think this is a timely example of why we need to be very careful about offering a sweeping "classic mode" option.

ajrb:
Rats with pockets? Suspention of disbelief instantly broken! lol

Allofich:
> Yep will do that, should we put a tag into these comments to mark them as behaviour changes so we can easily find them later?

That might be a good idea. In that case I should go through and mark my changes as well. There aren't that many of them.

> Interestingly, a user just posted a bug report about not being able to pickpocket rats or endlessly pickpocket town NPCs.

What a weird coincidence. Honestly, I wouldn't mind enabling pickpocketing of non-human creatures as an option, I just wouldn't choose to use it, and I didn't consider adding an option to be a priority. If someone is asking for it, though, I think it does rise in priority a bit. Not that we're the slaves of the users or being paid for this or anything.

So, yeah, normally I would respond by saying why I changed it but that I wouldn't mind adding an option for it, since it would be a simple change and obviously I can relate to wanting original behavior. But maybe you would prefer I not give that answer Interkarma?

Interkarma:
For the rat thing, I trust in you however you want this to work and I'll stand by whatever you decide. This feature is yours, and I know you've done your homework.

There's really no telling if pickpocketing rats was actually intentional or just yet another bug after fixing something else. I know which way makes more *sense* however.

This is also a pretty good example of how these discussions burn my time as well. I could be coding (or sleeping, it's 1am here), but instead I'm having a conversation about whether rats have pockets across both git and forums. It's just not a good use of my energy. :)

PR #616, opened 2018-02-14
Asset injection: Import sound files from asset bundles

TheLacus:
* Added a generic method to ModManager class to seek and import an asset of given type and name from all mods.
* Added support for sound import from mods. It was already possible to import them from disk, now they are also loaded from assetbundles with load order, like we discussed on forum for textures.

I did the same with videos but i prefer to test them a bit more before pushing.


Interkarma:
This just keeps getting better. Thank you. :)

PR #617, opened 2018-02-15
Shop & building schedules

ajrb:
Also reinstated tally crime guild calls and disabled the function instead.. much easier to enable in one place later.

### DIFF
--- a/Assets/Scripts/Game/Entities/PlayerEntity.cs
+++ b/Assets/Scripts/Game/Entities/PlayerEntity.cs
@@ -527,6 +527,10 @@ namespace DaggerfallWorkshop.Game.Entity
         /// </summary>
         public void TallyCrimeGuildRequirements(bool thievingCrime, byte amount)
         {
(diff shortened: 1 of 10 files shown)


ajrb:
The guild halls are open from 11:00-23:00 as it says on http://en.uesp.net/wiki/Daggerfall:Mages_Guild, however when I was testing in classic it never seemed to shut. Can anyone confirm?

Interkarma:
For some guilds, a certain rank means the guild hall is always open to them. Lower-ranked characters have to wait outside in the snow with the other plebs.

This seems to be different per guild however. Check the pages for each of the joinable factions and the rank entitlements are listed for each guild.

http://en.uesp.net/wiki/Daggerfall:Factions#Guild_Advancement

Fighters for example get 24/7 access after rank 6.

http://en.uesp.net/wiki/Daggerfall:Fighters_Guild

ajrb:
That makes sense now, thanks.

PR #618, opened 2018-02-16
House purchasing, ship models & player housing

ajrb:
Sorry, far too excited about this so I've just handled loading inside a bank by using the no houses for sale message. I'm happy to switch from using BuildingDirectory to datafiles, but not convinced it's worth it for this edge case. Turned out the radius was easy to get once I had model id. Didn't even need Allofich for the price multiplier.

TODO:
* Only listing buildings for sale - will need to add other houses in due course.
* Selling houses. (should be easy since the name is changed by overriding in PlayerGPS)

Anyway, this is so you can give it a spin when you need a break from magic! This was a while in the works as I gradually learned but it all came together today really nicely. Such a good feeling!

Oh, and you may want to check my changes to Place for preventing quests spawning in player houses.

### DIFF
--- a/Assets/Scripts/Game/Questing/Place.cs
+++ b/Assets/Scripts/Game/Questing/Place.cs
@@ -1,4 +1,4 @@
-﻿// Project:         Daggerfall Tools For Unity
+// Project:         Daggerfall Tools For Unity
 // Copyright:       Copyright (C) 2009-2018 Daggerfall Workshop
 // Web Site:        http://www.dfworkshop.net
 // License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
@@ -19,6 +19,7 @@ using DaggerfallConnect;
 using DaggerfallConnect.Arena2;
 using DaggerfallWorkshop.Game.UserInterfaceWindows;
 using FullSerializer;
+using DaggerfallWorkshop.Game.Banking;
 
 namespace DaggerfallWorkshop.Game.Questing
 {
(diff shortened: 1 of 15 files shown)


Interkarma:
This looks awesome, can't wait to play through it and test out.

As it's a fairly large PR crossing a few different systems, I'll put into a branch on my end and merge branch once I'm done testing. If you have any changes on this one, just send them to the 'housePurchasing' branch on my end. Will get it master soon.

Cheers mate! :)

Interkarma:
Well, this is rather amazing. Loving every part of it so far.

I'll let you know any feedback I have before the weekend is out and get everything merged to master. I can tell how much work and love you've put into this. It's also been a pleasure watching you ladder upwards towards this through all the prerequisite steps like world serialisation. Kudos on delivering such a great feature! :)

Assets/Scripts/Game/Questing/Place.cs
```diff
@@ -851,6 +854,10 @@ SiteDetails[] CollectQuestSitesOfBuildingType(DFLocation location, DFLocation.Bu
                         // Match building against required type
                         if (buildingSummary[i].BuildingType == buildingType || wildcardFound)
                         {
+                            // Building must not be a player owned house
+                            if (DaggerfallBankManager.IsHouseOwned(buildingSummary[i].buildingKey))
```
Interkarma: I think this is a good and perfectly direct way to filter out owned houses from quest selection. There's always that small chance an active quest has already selected this building before player purchases house, but that would be an uncommon edge case and should not affect the operation of most quests, as the markers remain valid for spawns anyway. I'm happy with this approach. 👍

Interkarma:
No problems with this so far. Will merge over to master tomorrow.

One small item of feedback. Would it be possible to increase lighting on spinning house in bank UI a little bit? The object just feels a touch dim to me at the moment.

ajrb:
Yep, I'd not noticed in my excitement. Done now, and carefully selected intensity so it doesn't overlight desert houses. Also, notice the shadows... took me ages to implement that one line of code! haha

PR #619, opened 2018-02-17
House selling and lighting 3d models

ajrb:
Also enabled mesh replacement, but still needs testing by a moddy person.
Added location name into savefile in case people forget where they bought houses.


Interkarma:
Awesome, thank you for the update!

PR #622, opened 2018-02-18
Added mage guild quests to table, mostly commented out

ajrb:
Also a correct for dupe TG training skill


Interkarma:
Awesome. And thanks for taking the time to play-test those quests. :)

PR #623, opened 2018-02-18
Climbing Control Changes

LorrMaster:
Made changes to the PlayerMotor script to give the player more control while climbing. However, these changes deviate from the original game's climbing mechanics.

Basically you can now let go of the movement controls in order to stop while climbing, press the key to move backwards in order to climb down, and cancel climbing by pressing the jump key.

### DIFF
--- a/Assets/Scripts/Game/PlayerMotor.cs
+++ b/Assets/Scripts/Game/PlayerMotor.cs
@@ -248,7 +248,7 @@ namespace DaggerfallWorkshop.Game
 
             // Climbing
             uint gameMinutes = DaggerfallUnity.Instance.WorldTime.DaggerfallDateTime.ToClassicDaggerfallTime();
-            if (!InputManager.Instance.HasAction(InputManager.Actions.MoveForwards)
+            if (InputManager.Instance.HasAction(InputManager.Actions.Jump)
                 || (collisionFlags & CollisionFlags.Sides) == 0
                 || failedClimbingCheck
                 || fakeLevitate.IsLevitating
@@ -495,7 +495,10 @@ namespace DaggerfallWorkshop.Game
 
         private void ClimbMovement()
         {
-            controller.Move(Vector3.up * Time.deltaTime);
+            if (InputManager.Instance.HasAction(InputManager.Actions.MoveForwards))
+                controller.Move(Vector3.up * Time.deltaTime);
+            if (InputManager.Instance.HasAction(InputManager.Actions.MoveBackwards))
+                controller.Move(-Vector3.up * Time.deltaTime);
             if (climbingContinueTimer <= (systemTimerUpdatesPerSecond * 15))
                 climbingContinueTimer += Time.deltaTime;
             else


Interkarma:
My preference is for classic-like controls by default and changes to climbing mechanics to become an option or a mod.

There are some upgrades planned for PlayerMotor later in 0.5 lifecycle that will better enable mods to plug in replacement scripts for different movement types. I feel this would be a better time to change climbing mechanics.

Alternatively, I'd be happy to accept this as an option with appropriate setting in INI for "sticky climbing" (or whatever you think is most descriptive).

Allofich, as the one implementing climbing initially, do you have any feedback here?

Thanks guys.

Allofich:
As I'm sure you know my "ideal" is that very-close to classic behavior be the base (wherever we can get it, and with bugs fixed), and be always available to use.

This particular mod is small enough that it may not cause any problems to just stick in an .ini setting for it. Even with 1 "mod" though, it may start to look a little messy and it would be something anyone would need to account for if going back and changing the basics of the climbing code. This won't necessarily happen, but if down the road 2 or 3 other people pop up with different changes they want to add to climbing, and 2 or more of these mods change the same piece of code in different ways, we could have a real mess on our hands.

> There are some upgrades planned for PlayerMotor later in 0.5 lifecycle that will better enable mods to plug in replacement scripts for different movement types.

That would be very cool! It might really help for gameplay modding.

I suppose an optimistic approach here is to accept the PR as an .ini setting (LorrMaster would need to change it to a setting I guess) with the idea of later switching it over to a new system.

LorrMaster:
Okay, that's fine. I'm just trying to get familliar with the code right now and figured that giving the player more movment options while climbing would make sense as a QOL change.

ajrb:
It may be better to hold fire until:
1) Allofich has finished classic climbing implementation
2) PlayerMotor has been re-written

Interkarma:
I haven't started work on this yet, but here's a general map of what I want to do with PlayerMotor in 0.5.

1. *Change PlayerMotor to a state machine.* Each type of movement (walking, riding, climbing, levitating, swimming, etc.) has it's own state script with full control over input and movement. Execution is transferred to active state during Update().
2. *PlayerMotor can accept replacement state scripts.* This would allow a mod to use different movement styles by creating new scripts to replace default behaviours. It will be up to the mod creator to maintain script compatibility with game.
3. *State rules will control which transitions are valid.* Generally you can move from one state to another, but rule support will maintain valid transitions.
4. *General improvements.* This item is to prevent player from becoming stuck on single-sided polys (e.g. in caves), prevent falling when running down steep slops in dungeons, and reduce or eliminate needing to crouch through doors near slopes (e.g. Sentinel). Fixing these problems is deeper than I first thought. Daggerfall's world is not designed for a rigid player capsule. I might have to build a new controller for this and will put it off until late in 0.5 or possibly defer this final item until later. In any case, the above 3 will happen first.

LorrMaster, your riding and climbing changes would be a perfect fit for item 2 above. You'll have complete freedom to make this work how you wish without needing to modify base code. You can just copy the default script, make your changes, and package to a mod. Then anyone who prefers your motor changes over classic can just grab the mod package.

Interkarma:
Also LorrMaster, I'm happy for this one to go in as an option. I can help you setup the option if any help is needed. This one is only a few lines of code so no major changes needed.

It might be worth waiting until Allofich has finished his work on climbing first, as the flow might change once he's done.

How does that sound? :)

LorrMaster:
Thanks! I think I'll just wait for the PlayerMotor script rewrite before doing anything else with movement.

In the meantime I'll do other stuff, like maybe try and make parts of the code easier to mod. I also know the Creation Kit's magic system pretty well, so I might be able to help with that once you guys start implementing it.

Interkarma:
No worries! I'll close off this one for now then.

PR #624, opened 2018-02-19
Implemented interaction mode icon for HUD 

ajrb:
This is something I thought would compliment the minimal HUD of DFU and show the interaction mode at a glance for players. Intended to avoid accidental pickpocketing etc.

Art took much longer than the code, but is placeholder only since I really suck at drawing. I hope I managed to make them good enough to demonstrate the idea and get it into the codebase. Then we can petition the forums to create some nice icons for us.

Chuck this on a branch and give it a go when you get a chance. It can be disabled in settings just like crosshair.


ajrb:
I tried to ensure they can easily be modded, including being replaced by different sized icons. I think standard ones should be plain and monochrome, but some people will want max bling mod I'm sure.

Interkarma:
This is a nice idea, I'll play with this and give you any feedback.

Another option for the icons is to snip out the interaction icons Daggerfall uses for these states in the large UI. Could do this at runtime with current UI, or just edit out to individual icons with transparency like you have now. File is MAIN01I0.IMG.

main01i0

ajrb:
I thought you would prefer a more minimal look, but could give that a go if you like. (darn those couple of hours cursing at paint and my ineptitude... lol)

Interkarma:
I like your icons and the minimal look. :)

It's just likely users might miss the original icons since these were already present in the game and fit the overall style. But considering the whole thing is an option, I don't really mind if you want to skip icons for now.

ajrb:
Planning to add an option for classic icons tonight.

Interkarma:
Awesome! I really do like the clean minimal look by the way. It's really unobtrusive and gets out of the way. I don't even really notice the icon after a while unless I want to look at it. I personally think you made the right choice.

It will be interesting to see with classic icons as well, though. I'm sure many people will appreciate that option.

PR #625, opened 2018-02-20
Add classic interaction mode icons as option

ajrb:


Interkarma:
Cheers! Will get this merged to master shortly.

Interkarma:
Hazelnut, would you be happy to make classic icons the default setting? A quick sample on Twitter shows a definite preference for classic. And after more playtesting the classic icons are easier to parse over a greater range of lighting conditions. It's easy for the minimal icons to become lost in gray cobblestone in towns, for example.

I'll leave final decision to you though as this one is your feature. :)

ajrb:
Quite  low sample size of people who have not tried it in game yet. I'd like to leave it as it is and poll forumites in a few weeks if that's ok. I think people are more likely to bother to switch to classic to see them which is why I went for this default.

Interkarma:
That's fine mate. Final choice will be yours, as mentioned.

PR #627, opened 2018-02-22
Add ability to get "work" from merchants

numidium:
NPCs can now give answers about "work." Added a generic window for quest offering that can be shared among all the NPC social groups whenever questing for those groups is implemented.

I added a merchant quest table with one quest that seems to work. A table for nobility is included as an example of extending functionality of the new questing window but those quests are not handled as of yet.

### DIFF
--- a/Assets/Scripts/Game/PlayerEnterExit.cs
+++ b/Assets/Scripts/Game/PlayerEnterExit.cs
@@ -4,7 +4,7 @@
 // License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
 // Source Code:     https://github.com/Interkarma/daggerfall-unity
 // Original Author: Gavin Clayton (interkarma@dfworkshop.net)
-// Contributors:    
+// Contributors: Numidium
 // 
 // Notes:
 //
@@ -1032,6 +1032,12 @@ namespace DaggerfallWorkshop.Game
                         holidayTextPrimed = true;
                     }
                     holidayTextLocation = location;
+
+                    // Generate a list of merchant NPCs to get work from
+                    if (!TalkManager.Instance.MerchantQuestorsAreSet(location.LocationIndex)) 
+                    {
+                        TalkManager.Instance.SetMerchantQuestors(location.LocationIndex);
+                    }
                 }
             }
         }
--- a/Assets/Scripts/Game/TalkManager.cs
+++ b/Assets/Scripts/Game/TalkManager.cs
@@ -22,6 +22,7 @@ using DaggerfallWorkshop.Game.UserInterface;
 using DaggerfallWorkshop.Game.Serialization;
 using DaggerfallWorkshop.Game.Entity;
 using DaggerfallWorkshop.Game.Questing;
+using System.Linq;
 
 namespace DaggerfallWorkshop.Game
 {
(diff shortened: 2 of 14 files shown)


Interkarma:
This looks like good work, thank you! That wasn't an easy job to take on, and from a quick read of the code I think you've hit all the right points.

Will test this in a branch and push to master shortly.

Nystul-the-Magician:
wow great you picked up where I had to stop because of my limited knowledge
talk functionality gets more and more complete.

ajrb:
This is something related to what I'm working on at the moment, which is quest dispensing.  Great to see the talk manager parts, because I didn't want to get into that. I have a few queries/observations:

1) Is the parsing of data really neccessary in TalkManager? It already does this in GetBuildingList() method so maybe these could be combined?

2) I would like the duplicated conditionals that offer quests moved into a single location in TalkManager.TalkToStaticNPC() rather than having lots of them in each place with talk option, unless there's a reason for this I cannot see?

3) I view the quest offer window class as a placeholder until the real quest dispensing code is done and appropriate quests can be offered using it. You've implemented it as a UI window class even though it's not really, but that doesn't matter if you see it as just a temporary front end for testing this 'get work' code?

Interkarma:
> Is the parsing of block data really neccessary in TalkManager? It already does this in GetBuildingList() method so maybe these could be combined?

Agree these could be combined. I'm likely to accept PR as it is once I'm done with testing, but I'd like to see more work on it to optimise. :)

> I would like the duplicated conditionals that offer quests moved into a single location in TalkManager.TalkToStaticNPC() rather than having lots of them in each place with talk option, unless there's a reason for this I cannot see?

numidium, are you able to assist Hazelnut with this one?

> but that doesn't matter if you see it as just a temporary front end for testing this 'get work' code?

That would be my interpretation. Your work on guilds, services, and quest dispensing is the end goal, and this implementation can be removed later just as you did for my placeholder quest dispensing from Fighters and Mages guilds.

numidium:
I’ll see what I can do regarding the parsing code. 

The duplicated if statements are left over from when I thought it would be handled differently, so yes I should be able to combine them into TalkToNPC. 

And yes the new “window” is temporary. Hazelnut, feel free to delete it once you’re finished with the quest dispenser. :)

Interkarma:
> I’ll see what I can do regarding the parsing code.

Cheers! And if there's no way to combine sensibly (i.e. it just makes everything harder to understand or more difficult, or breaks something else) then please leave as-is for now.

I'm testing now and will have any feedback for you soon.

Interkarma:
One small bug I've found is asking about work after loading an interior save results in a null exception, as exterior doesn't exist yet (suppress world).

Just needs a null check in TalkManager.WorkAvailable on merchantRecords or to init merchantRecords as an empty dict at define time.

Interkarma:
I'm very happy with this! 👍 It's a great starting point and can be built on over time to refine.

I will merge to master soon. Just a few more items of feedback.

1. Will need to implement serialization of available work at some point, at least for last city entered.
2. The OnEnterLocationRect doesn't fire when loading a game inside a city (this one is on me), which means work isn't generated until player leaves area and returns. I'll look at raising this even when loading a game or exiting an interior save for first time.
3. Have the infamous BLANK for questor in journal text, as questor may not be set. Not sure if on my side or yours though. I'll check this in more detail soon.

None of the above are blocking issues however. Great work mate! :)

PR #628, opened 2018-02-23
Clear rented rooms for new games

ajrb:
Sorry about the multiple commits - I updated my fork after pushing the fix to it which causes Github to do this. I'm avoiding this most of the time now, but I forgot this morning.


ajrb:
Also fixed an issue with loading old pre-bank deed saves

PR #629, opened 2018-02-23
Asset injection: Improvements for movies and images

TheLacus:
* Import movies from loose files
* Import Img, Cif and Rci files from mods. I had to merge seeking and import in a single method to avoid iterating mods and locations twice, but i don't think we lost much.

### DIFF
--- a/Assets/Scripts/Game/UserInterface/HUDCrosshair.cs
+++ b/Assets/Scripts/Game/UserInterface/HUDCrosshair.cs
@@ -10,7 +10,6 @@
 //
 
 using UnityEngine;
-using DaggerfallWorkshop.Utility.AssetInjection;
  
 namespace DaggerfallWorkshop.Game.UserInterface
 {
@@ -47,12 +46,7 @@ namespace DaggerfallWorkshop.Game.UserInterface
 
         void LoadAssets()
         {
-            if (TextureReplacement.CustomTextureExist(defaultCrosshairFilename))
-                CrosshairTexture = TextureReplacement.LoadCustomTexture(defaultCrosshairFilename);
-            else
-                CrosshairTexture = Resources.Load<Texture2D>(defaultCrosshairFilename);
-
-            crosshairSize = TextureReplacement.GetSize(CrosshairTexture, defaultCrosshairFilename, true);
+            CrosshairTexture = DaggerfallUI.GetTextureFromResources(defaultCrosshairFilename, out crosshairSize);
         }
     }
 }
--- a/Assets/Scripts/Game/DaggerfallUI.cs
+++ b/Assets/Scripts/Game/DaggerfallUI.cs
@@ -896,6 +896,27 @@ namespace DaggerfallWorkshop.Game
             return texture;
         }
 
+        /// <summary>
+        /// Get a texture from resources with modding support.
+        /// </summary>
+        public static Texture2D GetTextureFromResources(string name)
+        {
+            Texture2D tex;
+            return TextureReplacement.TryImportTexture(name, out tex) ? tex : Resources.Load<Texture2D>(name);
+        }
+
+        /// <summary>
+        /// Get a texture from resources with modding support.
+        /// Size is read from xml or set as texture size.
+        /// </summary>
+        public static Texture2D GetTextureFromResources(string name, out Vector2 size)
+        {
+            Texture2D tex = GetTextureFromResources(name);
+            if (!TextureReplacement.TryGetSize(name, out size))
+                size = new Vector2(tex.width, tex.height);
+            return tex;
+        }
+
         public static DaggerfallMessageBox MessageBox(string message, bool wrapText = false, IMacroContextProvider mds = null)
         {
             DaggerfallMessageBox messageBox = new DaggerfallMessageBox(Instance.uiManager, Instance.uiManager.TopWindow, wrapText);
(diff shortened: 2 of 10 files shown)


TheLacus:
@Interkarma ImageReader.GetSubTexture() doesn't work correctly with imported images due to rect being based on default resolution. 
screenshot 2018-02-23 21 45 16
A workaround is to scale it to account size difference
`weaponsAndArmorNotSelected = ImageReader.GetSubTexture(baseTexture, ScaledRect(weaponsAndArmorRect, baseTextureScale));`
`weaponsAndArmorSelected = ImageReader.GetSubTexture(goldTexture, ScaledRect(weaponsAndArmorRect, goldTextureScale));`
but it looks a bit messy. Do you think is possible to just avoid this dependency (ie use relative coordinates)? The annoying part is that rect is used for subtexture(which is based on res) but also for button placement (which is fixed for native panel).

Assets/Scripts/Game/UserInterface/HUDCrosshair.cs
```diff
@@ -47,12 +46,7 @@ public override void Update()
 
         void LoadAssets()
         {
-            if (TextureReplacement.CustomTextureExist(defaultCrosshairFilename))
-                CrosshairTexture = TextureReplacement.LoadCustomTexture(defaultCrosshairFilename);
-            else
-                CrosshairTexture = Resources.Load<Texture2D>(defaultCrosshairFilename);
-
-            crosshairSize = TextureReplacement.GetSize(CrosshairTexture, defaultCrosshairFilename, true);
+            CrosshairTexture = DaggerfallUI.GetTextureFromResources(defaultCrosshairFilename, out crosshairSize);
```
ajrb: I just copied this last week for loading images for the icon UI component. Should I update it to use this method then?

TheLacus: Hi, yes. GetSize() is now only intended for Img and CifRci (gets original texture size for panel layout if xml is not found), GetTextureFromResources is a simple helper method which set size as resolution if there is no xml. The biggest difference is that now it also loads textures from .dfmods

Interkarma:
> ImageReader.GetSubTexture() doesn't work correctly with imported images due to rect being based on default resolution.

Agree this is limiting and confusing. And I would have only had classic assets in mind at the time. I'm assuming the overload you're most interested in is this one, as it uses Texture2D for input:

`public static Texture2D GetSubTexture(Texture2D texture, Rect subRect)`

I likely wrote this prior to Graphics.CopyTexture() being available in Unity (or I just wasn't aware of it at the time). Rather than convert back to Color32 and hand off, it's probably more efficient to use Unity's native Graphics.CopyTexture() to do the work. Even it's doing something similar under the hood, it's probably still more efficient.

I'll try something now that might help and get back to you in a couple of hours. :)

Interkarma:
I've made some changes that should at least be a step along the road to helping.

First, I've enabled GetSubTexture() to use the more efficient Graphics.CopyTexture() when available. This will speed up process across the board, but especially for larger replacement textures. It falls back to old memory buffer method if CopyTexture() not available.

Second, GetSubTexture() can use accept relative coordinates by setting useRelativeCoords=true. These are converted back to pixel coordinates relative to the input texture of GetSubTexture().

Third, I've added a helper ImageReader.ConvertToRelativeSubRect() that can convert classic pixel coordinates into a classic source texture to relative coordinates.

Now we can start tackling the really hard part of the problem:

> The annoying part is that rect is used for subtexture(which is based on res) but also for button placement (which is fixed for native panel).

It's difficult to build the UIs by hand and think in relative coordinates. I'm hoping to strike a compromise if possible.

As the classic pixel coordinates sent to GetSubTexture() are hard-coded anyway, these calls can now use ConvertToRelativeSubRect() with the classic width and height to get relative dimensions.

Then the actual call to GetSubRect() can succeed using replacement texture and relative coords. Placement in classic panel should be as before.

I don't think we're going to be able to avoid updating all those existing GetSubTexture() calls in some way. All we can do is try to make it as easy as possible for everyone.

I'm interested in hearing feedback from you guys where we go from here. Just remember not to make too much work for ourselves. :)

TheLacus:
Thank you :) I'll start with a small note: CopyTexture doesn't work with texture compression
```
Graphics.CopyTexture with a region will not copy readable texture data for compressed formats (source texture format 12)
Graphics.CopyTexture can only copy between same texture format groups (d3d11 base formats: src=76 dst=27)
```
Even if we preserve format
`Texture2D newTexture = new Texture2D((int)subRect.width, (int)subRect.height, AssetInjection.TextureReplacement.TextureFormat, false);`
we still have this issue
`For block-based formats (e.g. DXT, ETC), the region size and coordinates must be a multiple of compression block size (4 pixels for DXT).`

TheLacus:
My first idea was to scale rect for the ratio of real size and default size, which is just a slightly different way to do the same thing.

If we aim at making the least possible changes we could do something like this
```
// fullscreen
public static Texture2D GetSubTexture(Texture2D texture, Rect subRect)
{
        return GetSubTexture(texture, subRect, new Vector2(320, 200));
}

// popup
public static Texture2D GetSubTexture(Texture2D texture, Rect subRect, Vector2 virtualSize)
{
       return GetSubTexture(texture, ConvertToRelativeSubRect(subRect, (int)virtualSize.x, (int)virtualSize.y), true);
}
```
Basically we assume rect is based on native panel unless window size is given. We'd still need to fix all popups buttons though.

Interkarma:
Thank you for the heads up. I clearly didn't read that properly earlier. I will ensure GetSubTexture() falls back to memory copy method when handed a compressed texture. I'll pop this fallback in today.

That's an interesting solution. Happy for you to try this and see how well it rolls.

PR #631, opened 2018-02-24
Minimize repetition in work dialogue code

numidium:
I merged SetMerchantQuestors into GetBuildingList and made it part of its loop. Also updated TalkToStaticNPC so it checks to see if the NPC is a potential questor.


Interkarma:
Thank you for all your hard work on this! 👍

PR #632, opened 2018-02-24
Settings window: "Theme" settings

TheLacus:
Added a few settings which where not exposed on GUI
screenshot 2018-02-24 17 58 37

InteractionModeIcon was not present in SaveSettings(), i assumed this wasn't intended but let me know :)


ajrb:
Nope, probably not intentional. I'll take a look, and since we have that slider, can I introduce other options? Thinking none, minimal, icons, classic.

TheLacus:
Sure. Slider is based on `InteractionModeIconModes`  enum, just add new values. 
I created this enum because i need an index for each allowed string, but it's up to you to choose how to manage these options, settings window will just follow.

PR #633, opened 2018-02-25
Add GetSubTexture() overload for relative subrects

TheLacus:
This follows #629. Native panel size is only used in inventory window, so i wrote a single method that accepts a DFSize argument. A few classic textures were off by one pixel (which is visible at classic resolution) so relative rect is used only when necessary. Tell me if you have any suggestions or requests :)


TheLacus:
screenshot 2018-02-25 15 41 09

PR #636, opened 2018-02-26
Guilds - initial implementation

ajrb:
I could use another pair of eyes to test what I've done so far if you fancy giving it a spin.

Fighters, Mages, and all Temples & Templars should work. Resting is allowed in all once you join, and rank update checks are every day - for testing purposes. Basically rest 24 hours, save, edit rep/skills, load and speak to a guild person to get promotion / demotion.


Interkarma:
This looks like really find work based on just a read. Going to merge this to a branch now as I'm really looking forward to taking it for a test drive! :)

PR #641, opened 2018-03-03
Match disease infliction by enemy type to classic

Allofich:
@ajrb Here is the PR we discussed. Interkarma, could you let ajrb take a look before merging? I didn't change much of what he had implemented.

I also added some todos for various bits to add into the damage code.

### DIFF
--- a/Assets/Scripts/Game/EnemyAttack.cs
+++ b/Assets/Scripts/Game/EnemyAttack.cs
@@ -195,16 +195,7 @@ namespace DaggerfallWorkshop.Game
             playerEntity.TallySkill(DFCareer.Skills.Dodging, 1);
 
             if (damage > 0)
-            {
                 GameManager.Instance.PlayerObject.SendMessage("RemoveHealth", damage);
-
-                if (!playerEntity.Disease.IsDiseased())
-                {
-                    Diseases disease = FormulaHelper.CalculateChanceOfDisease(entity);
-                    if (disease != Diseases.None)
-                        playerEntity.Disease = new DaggerfallDisease(disease);
-                }
-            }
 
             return damage;
         }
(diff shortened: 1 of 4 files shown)


ajrb:
Looks great, thanks for this. Happy for this to be merged ASAP.

Allofich:
Great. I found a small mistake, though. I want to fix that first.

ajrb:
Okay, well I think Interkarma is asleep so you have time. :)

I'm trying to figure out how best to allow formulas to be redefined by mods. Easy way of supplying a new implementation of everything will be too restrictive since it would only allow one mod to change formula.

Allofich:
Fixed. For enemies with multiple hit attempts per attack (including werewolves and wereboars), there is a chance of disease for each of these hits that has damage > 0. I was accidentally checking against `damage`, which stores the damage total, rather than the individual hits. (BTW this "check lycanthropy per hit" behavior is why I wanted to put the call from here, as in classic, rather than checking the damage result in EnemyAttack).

Allofich:
It can be in a separate PR, but I'm also looking at the fatigue drain for nymphs/lamias. There really seems to be a lot of slip-ups and inconsistencies in the damage code (I've added a few notes in my latest commit above and changed the silver bonus on skeletal warriors to affect werebeasts instead).

Also different monsters cause different possible diseases. I'll try to add that in later. Again it can be a different PR.

Allofich:
Added the other things in I talked about. Diseases are picked by enemy type and nymph/lamia fatigue damage is implemented in what seemed the most sensible way to work out the inconsistencies between classic and DF Chronicles.

Also I borrowed a message used in Arena for when the player loses all fatigue in water (as would probably happen if they were drained by a lamia) since Daggerfall doesn't have any special message for this situation.

Allofich:
Implemented the saving throw function. Right now it just works for diseases but it's basically the same for all types of saving throws.

Interkarma:
This is great, guys. :)

PR #642, opened 2018-03-05
added demo quest showing off rumor mill in conjunction with dialog link

Nystul-the-Magician:
example quest to demonstrate rumor mill and dialog link:
get information after (and only after) hearing the correct rumor - this is extending vanilla daggerfall's behaviour - once an npc talked about the rumor the linked dialog option will be available in the tell me about section - rumor mill gives one of two possible answers (and you will have to try quite often, since these 2 answers only get added to the list of rumors) - only one of the two answers is revealing the person name in tell me about section (after getting the correct answer you still have to switch dialog option and then it will show up in the tell me about section) you can then ask peasants for the correct place where the person can be found (todo: seems like the quest person does not show up, not sure why, but this is not related to talk mechanics)


Nystul-the-Magician:
will add the other demo quests first as well and then reissue the pr

PR #647, opened 2018-03-08
Initial version of manging quest tables

ajrb:
Just Fighters and Mages so far.


Interkarma:
This looks like great progress, thank you.

PR #648, opened 2018-03-09
 Add reputation change calls to quest end for guild quests

ajrb:


Interkarma:
This keeps getting better! Thank you for all your work on this. :)

PR #649, opened 2018-03-10
Progress on diseases

Allofich:
Implements:
- Classic diseases' stats (which attributes they affect, etc.)
- Diseases that endlessly lower stats vs. those that don't. When a non-endlessly-lowering disease finishes, it sets its number of days left to 254 (0xFE) in classic. Doing it the same way, which should aid in importing diseases from classic saves.
- Length of disease and damage values gotten like in classic.
- Diseases lower stats in the same manner as classic and can kill the player.

Diseases are disabled now until a way to cure them is implemented.

I haven't figured out how classic does the incubation period yet. Also there is no "you may not survive" message for traveling. Length of disease period will need to be saved and loaded. Poisons aren't implemented yet. Neither are imported from classic saves yet.

### DIFF
--- a/Assets/Scripts/API/Save/SaveTree.cs
+++ b/Assets/Scripts/API/Save/SaveTree.cs
@@ -382,6 +382,7 @@ namespace DaggerfallConnect.Save
         NPCFlat = 0x08,
         Spell = 0x09,
         GuildMembership = 0x0a,
+        DiseaseOrPoison = 0x0b,
         QBNData = 0x0e,
         QuestTree = 0x10,                           // Fixsave calls this "quest tree".
         EnemyMobile = 0x12,
--- a/Assets/Scripts/Game/DaggerfallUI.cs
+++ b/Assets/Scripts/Game/DaggerfallUI.cs
@@ -1043,7 +1043,7 @@ namespace DaggerfallWorkshop.Game
         public DaggerfallMessageBox CreateHealthStatusBox(IUserInterfaceWindow previous = null)
         {
             DaggerfallMessageBox healthBox = new DaggerfallMessageBox(uiManager, previous);
-            healthBox.SetTextTokens(GameManager.Instance.PlayerEntity.Disease.GetMessageId());    // TODO: Various diseases are in msgs 100-117
+            healthBox.SetTextTokens(GameManager.Instance.PlayerEntity.Disease.GetMessageId());
             healthBox.ClickAnywhereToClose = true;
             return healthBox;
         }
--- a/Assets/Scripts/Game/Entities/DaggerfallDisease.cs
+++ b/Assets/Scripts/Game/Entities/DaggerfallDisease.cs
@@ -15,15 +15,83 @@ namespace DaggerfallWorkshop.Game.Entity
 {
     public class DaggerfallDisease
     {
+        public struct DiseaseData
+        {
+            // Affected stats
+            public byte STR;
+            public byte INT;
+            public byte WIL;
+            public byte AGI;
+            public byte END;
+            public byte PER;
+            public byte SPD;
+            public byte LUC;
+            public byte HEA;
+            public byte FAT;
+            public byte SPL;
+            public byte minDamage;
+            public byte maxDamage;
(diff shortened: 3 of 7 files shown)


Allofich:
Added:
- Status popup doesn't show disease until it's done incubation (which is I think how classic works)
- Warning popup if player tries to fast travel with an incubating disease (also should be same as classic)

ajrb:
I'll add the cure service with stubs for cost and the actual curing.. then you can fill those in when you're ready.

Allofich:
Sounds good.

ajrb:
Actually I need a bit of info about what happens when curing because I have no way of testing it in classic as I don't have any chars with diseases to use. If you could tell me what the price is based on, and if it differs with each temple, or with membership. I already have the special cheaper cures for Arkay membership ready to plug in, so you can ignore that.

Allofich:
I'll take a look at that stuff soon and let you know.

I'll update this PR pretty soon. As I mentioned in the other PR I think we need to restore damaged stats when curing a disease.

Interkarma:
The incoming magic & effect system is being designed for diseases, poisons, advantages/disadvantages, spell effects, etc. Basically any programmatic short or long-running change to player entity that can be applied/removed should be able to fit inside effect system.

Part of that design is the ability to apply/remove an effect without affecting base stats. There's also some tolerance being built in for "misbehaving" effects (such as bugs in my scripts or in effects added by mods) to ensure they can be removed easily without any permanent change to stats. I'll be talking about this more in next blog post.

I'm not sure if you want to hold off on business-end of diseases modifying any stats for a little bit longer until the effect system is further developed. You should be able to dovetail your work on diseases with effect system at some point, although it's bound to go through some iteration before settling down.

But if you're on a roll for now, I'm happy for diseases to be developed independently and we can look at migrating to effect system later if needed. I'm happy just to have you guys helping with this stuff. Thank you. :)

Assets/Scripts/Game/Entities/DaggerfallStats.cs
```diff
@@ -46,14 +46,14 @@ public partial class DaggerfallStats
...
 
-        public int LiveStrength { get { return GetLiveStatValue(DFCareer.Stats.Strength); } }
-        public int LiveIntelligence { get { return GetLiveStatValue(DFCareer.Stats.Intelligence); } }
-        public int LiveWillpower { get { return GetLiveStatValue(DFCareer.Stats.Willpower); } }
-        public int LiveAgility { get { return GetLiveStatValue(DFCareer.Stats.Agility); } }
-        public int LiveEndurance { get { return GetLiveStatValue(DFCareer.Stats.Endurance); } }
-        public int LivePersonality { get { return GetLiveStatValue(DFCareer.Stats.Personality); } }
-        public int LiveSpeed { get { return GetLiveStatValue(DFCareer.Stats.Speed); } }
-        public int LiveLuck { get { return GetLiveStatValue(DFCareer.Stats.Luck); } }
+        public int LiveStrength { get { return GetLiveStatValue(DFCareer.Stats.Strength); } set { mods[0] += value - GetLiveStatValue(DFCareer.Stats.Strength); } }
```
Interkarma: Please don't make any changes to "mods" array at this time. These way these values are intended to work isn't what you might expect. They are going to be flattened back to zero with every tick of effect system. Just leave them as-is for now, thank you!

Interkarma: To give you more detail, I purposefully want to insulate any system from making changes here without going to through the effect system. There will be some protections built into effects to keep them in line and prevent changes "escaping" from an effect back onto live stats.

Allofich:
Ok, for now I'll just print stat changes to the Debug log or something and not actually change stats.

Allofich:
Updated.

Stat changes are removed (just comments).
Total damage is tallied, with the idea that the damage caused by the disease can be removed when the disease is cured.
Put in a check for daysOfSymptoms = 0 in ApplyDiseaseEffects(), which should catch any saved games made since diseases were added but before this PR is merged.

PR #650, opened 2018-03-10
JH classic quest fixes - first pass complete

ajrb:
This is the work that Jay_H has been doing to fix classic quests. First pass complete for all the guild quests.

### DIFF
--- a/Assets/StreamingAssets/Quests/L0B00Y02.txt
+++ b/Assets/StreamingAssets/Quests/L0B00Y02.txt
@@ -4,6 +4,7 @@
 -- Questor: Brotherhood
 -- Repute: 0
 -- QuestId: 2
+-- Edited for Daggerfall Unity by Jay_H
 Messages: 16
 Quest: L0B00Y02
 -- Message panels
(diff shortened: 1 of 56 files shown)


Nystul-the-Magician:
one question about N0C00Y10.txt:
why was _thief_ renamed into _enemy_ ? - this obfuscates things a bit since in the first place you are asked to track down a book thief by the quest giver

Assets/StreamingAssets/Quests/L0B00Y02.txt
```diff
@@ -79,8 +79,8 @@ QuestorPostfailure:  [1009]
 
 QuestLogEntry:  [1010]
 %qdt:
- _qgiver_ of the Dark Brotherhood
- in __qgiver_, ___qgiver_
+ _qgiver_ of the Dark
+ Brotherhood in ___qgiver_
  has me paying a visit on =assassin_,
```
Allofich: Minor question: From what I know of the quest scripting, it looks like the original text mentions the city and region where you got the quest from here, right? Was replaced with just one or the other? I don't see the reason for the change.

Allofich:
Added a question, but overall it looks like a lot of solid work went into this! It must have taken a lot of time and dedication to look at all these quests for bugs.

Assets/StreamingAssets/Quests/L0B00Y02.txt
Interkarma: I'm not sure, but Jay might have just made that change for everything to fit on a single line. The full location+region text might have been too many characters wide in some cases.

PR #651, opened 2018-03-10
Add donation and cure disease guild services

ajrb:
Cure is very skeleton for now, Allofich and I will flesh out soon.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallGuildServicePopupWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallGuildServicePopupWindow.cs
@@ -576,15 +576,11 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
                 {
                     // Deduct gold, and apply blessing if member
                     playerEntity.DeductGoldAmount(amount);
-                    int factionId = guild.GetFactionId();
-                    if (guild.IsMember() && guild.GetType() == typeof(Temple))
-                        ((Temple)guild).Blessing(playerEntity, amount);     // TODO what happens when blessing is applied in classic? a message?
-                    else
-                        factionId = (int)Temple.GetDivine(buildingFactionId);
+                    int factionId = (int)Temple.GetDivine(buildingFactionId);
 
                     // Change reputation
                     int rep = Math.Abs(playerEntity.FactionData.GetReputation(factionId));
-                    if (UnityEngine.Random.Range(1, 100) <= (2 * amount / rep + 1))
+                    if (UnityEngine.Random.Range(1, 101) <= (2 * amount / rep + 1))
                         playerEntity.FactionData.ChangeReputation(factionId, 1, true);
 
                     // Show thanks message
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallGuildServicePopupWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallGuildServicePopupWindow.cs
@@ -23,7 +23,7 @@ using DaggerfallWorkshop.Game.Guilds;
 
 namespace DaggerfallWorkshop.Game.UserInterfaceWindows
 {
-    public class DaggerfallGuildServicePopupWindow : DaggerfallPopupWindow
+    public class DaggerfallGuildServicePopupWindow : DaggerfallPopupWindow, IMacroContextProvider
     {
         #region UI Rects
 
@@ -59,6 +59,8 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
         const int TrainSkillId = 5221;
         const int notEnoughGoldId = 454;
         const int insufficientRankId = 3100;
+        const int tooGenerousId = 702;
+        const int donationThanksId = 703;
 
         Texture2D baseTexture;
         PlayerEntity playerEntity;
@@ -238,6 +240,14 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
 
                 case GuildServices.Training:
                     TrainingService();
+                    break;
+
+                case GuildServices.Donate:
+                    DonationService();
+                    break;
+
+                case GuildServices.CureDisease:
+                    CureDiseaseService();
                     break;
 
                 case GuildServices.BuyMagicItems:   // TODO: switch items depending on npcService?
@@ -473,7 +483,6 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
                 DaggerfallMessageBox messageBox = new DaggerfallMessageBox(uiManager, uiManager.TopWindow);
                 messageBox.SetTextTokens(tokens);
                 messageBox.ClickAnywhereToClose = true;
-                messageBox.ParentPanel.BackgroundColor = Color.clear;
                 uiManager.PushWindow(messageBox);
             }
             else
@@ -488,7 +497,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             }
         }
 
-        public void ConfirmTraining_OnButtonClick(DaggerfallMessageBox sender, DaggerfallMessageBox.MessageBoxButtons messageBoxButton)
+        private void ConfirmTraining_OnButtonClick(DaggerfallMessageBox sender, DaggerfallMessageBox.MessageBoxButtons messageBoxButton)
         {
             CloseWindow();
             if (messageBoxButton == DaggerfallMessageBox.MessageBoxButtons.Yes)
@@ -509,7 +518,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             }
         }
 
-        public void TrainingSkill_OnItemPicked(int index, string skillName)
+        private void TrainingSkill_OnItemPicked(int index, string skillName)
         {
             CloseWindow();
             List<DFCareer.Skills> trainingSkills = GetTrainingSkills();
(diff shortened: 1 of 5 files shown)


Allofich:
Curing a disease should (I assume) also restore any lowered stats. I think disease records in classic store a tally of how many points they have deducted from the player, which is probably used when a cure is applied to undo the effects. I'll incorporate that into https://github.com/Interkarma/daggerfall-unity/pull/649.

ajrb:
Yes, it should restore the stats but I think we should wait for the effect system Interkarma is building to be ready before implementing disease effects. There's still a lot of mechanics around incubation & curing that can be finished.

Allofich:
For the reputation gain from donation in classic, it works like this:

```FactionRecord templeFaction = GetFactionFromBuildingType();

    if (ValueThatPlayerInputted >= 1)
    {
        if (GetPlayerHasEnoughGold(ValueThatPlayerInputted))
        {
            SubtractPlayerGold(ValueThatPlayerInputted);
            int chanceForReputationGain = 2 * ValueThatPlayerInputted / (abs(*(TempleFaction.PlayerReputation)) + 1);
            if (Random(1, 100) <= chanceForReputationGain ) // random inclusive of min and max
                ++*TempleFaction.PlayerReputation;
            DisplayTEXTRSCMessage(703); // thank you for your donation
        }
        else
        {
            DisplayTEXTRSCMessage(702); // too generous
        }
    }
```

The more gold that is donated the better the chance, and the higher the reputation (or lower if negative) the lower the chance of a change.

ajrb:
Hmm, my testing of rep gain was that you gain 1 point for each donation no matter the size of it. I shall add this logic then.

Allofich:
I'm pretty sure this is for reputation gain. I haven't found how the blessing works, yet. It makes sense, too, as you could otherwise just donate 1 gold piece over and over to get maximum reputation, right? But I haven't tested in-game.

> What does "(abs(*(" mean?

Typo, sorry. I'll fix it.

ajrb:
Yeah, I was pretty surprised at the results of my testing but I only tried a few different amounts and then saved and checked with a save editor. For anything chance based trial and error is useless at revealing the truth... thankfully we have you on teh team!

Allofich:
Still, doing repeated donations of 1 gp for example, it definitely shouldn't be increasing every time. I'll try to test it.

ajrb:
No need - I trust your analysis, and I was trying to say that basically my testing was pretty worthless. :)

Allofich:
I just tested anyway, and it seems to hold up. I got reputation gains from donating the default of 1000 and nothing when I donated 1.

By the way I was trying at a Temple of Kynareth (Infinite Ascencion of Kynareth in Daggerfall Region) and occasionally I get a mistaken message with "BLANK" in it, like:
"Certainly BLANK accepts all donations..."

Two other messages also randomly show up, both of which correctly say "Kynareth". I think BLANK shows up as the default in classic when it fails to get the correct name for some reason. I wonder what shows up with this PR?

ajrb:
Yes, I have a TODO comment in about that. Need to handle the macro but is within the quest system so I need to discuss with Gav first. Was holding off for a bit, since theres plenty to do on guilds yet! :)

Allofich:
Looks like the problem message uses the `%gdd` macro, which it looks like you are handling in this PR, so it will probably work fine with this PR.

Allofich:
More info on the `%gdd` macro (sorry I know you are focused on other things right now).

I don't think this is just a typo of `%god`. But it seems to be kind of broken. Unlike the `%god` macro, this one depends on the player's temple affiliation. I was getting BLANK because I was donating while not belonging to a temple. But after I joined it became "God of Logic". That's supposed to be Julianos, not Kynareth.

"God of Logic" is among a list of different deity descriptions in FALL.EXE. I don't see "Goddess of Air" in there, and this would need an "a" attached to it to look right for this message (which might make it look incorrect elsewhere), but it's another puzzle we could work on later.

Allofich:
Regarding blessings from donations, I'm not sure those are even in the game. I see reference to it on the UESP but I haven't found anything on it in the binary and I just gave all my player's cash away in donations in classic and didn't get any indication of a stat blessing.

ajrb:
Could have been unimplemented, I know I've never recieved a blessing in all my (limited) testing.

Allofich:
> Could have been unimplemented, I know I've never recieved a blessing in all my (limited) testing. So got into compendium (where most uesp info comes from) but not the game.

I tried giving a huge amount of gold (the first time I thought I gave a lot but actually my character didn't have much and I just didn't notice), but still no noticeable blessing. I'm back to doubting it's actually in the game, but I'll keep an eye out for it.

> Thanks for the gdd info, guess it's god description maybe?
That was my guess, too.

> No messages in text.rsc that I can find
The messages are in FALL.EXE. Do a search for "God of Logic" and you'll find them. They aren't all for the deities with temples, though. Assuming Kynareth would be "Goddess of Air", it's missing from the list.

Allofich:
They're also preceded by the deity names, and looks like the order matches up. "Kynareth" isn't in there at all, but other deities from the lore are.

Raen.Arkay.Zenithar.Notorgo.Mara.Shandar.Akatosh.Julianos.The Bandit God.Dibella.Jephre.Vigryl.Q'Olwen.Stendarr.Ephen.

God of Agriculture.God of Birth and Death.God of Work and Commerce.Messenger God.Goddess of Love.God of War.God of Time.God of Logic.God of Thieves.Goddess of Beauty.the Storyteller.God of the Sea.(wisdom?).God of Mercy.God of the Wild.

ajrb:
Have added GodDesc macro handler.

ajrb:
Added the rep gain chance

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallGuildServicePopupWindow.cs
```diff
@@ -574,14 +574,18 @@ private void DonationMsgBox_OnGotUserInput(DaggerfallInputMessageBox sender, str
...
                     int factionId = guild.GetFactionId();
                     if (guild.IsMember() && guild.GetType() == typeof(Temple))
                         ((Temple)guild).Blessing(playerEntity, amount);     // TODO what happens when blessing is applied in classic? a message?
                     else
                         factionId = (int)Temple.GetDivine(buildingFactionId);
-                    playerEntity.FactionData.ChangeReputation(factionId, 1, true);
+
+                    // Change reputation
+                    int rep = Math.Abs(playerEntity.FactionData.GetReputation(factionId));
+                    if (UnityEngine.Random.Range(1, 100) <= (2 * amount / rep + 1))
```
Allofich: You should add 1 to the maximum to get it the same as classic, because UnityEngine.Random.Range is non-inclusive of the maximum.

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallGuildServicePopupWindow.cs
```diff
@@ -574,14 +574,18 @@ private void DonationMsgBox_OnGotUserInput(DaggerfallInputMessageBox sender, str
             {
                 if (playerEntity.GetGoldAmount() > amount)
                 {
-                    // Deduct gold, change reputation and apply blessing
+                    // Deduct gold, and apply blessing if member
                     playerEntity.DeductGoldAmount(amount);
```
Allofich: Should the blessing be only for members? It seems to me like anyone who donates should be able to get it. I suppose restricting it gives a bit more uniqueness to joining a particular temple, but allowing it for all temples gives a reason to donate if you're not a member (I'm not aware of any benefit to raising your reputation with a temple other than for gaining ranks as a member). Also, the later games aren't a very strong source of reasons for Daggerfall IMO because game design and lore changed a good bit, but that said, in Morrowind and Oblivion you can get blessings without being a member of a religious faction.

Allofich:
> Have added GodDesc macro handler.

Did you forget to add it to the PR?

ajrb:
> >   Have added GodDesc macro handler.
> Did you forget to add it to the PR?
No it'll be part of the next PR

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallGuildServicePopupWindow.cs
ajrb: I don't believe that's the case. From the unity documentation:

Returns a random float number between and min [inclusive] and max [inclusive]

ajrb: WHy on earth is the int version an exclusive max in Unity? I had only read the float version documentation. Thanks for pointing that out.

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallGuildServicePopupWindow.cs
ajrb: I think you're right now I think about it - not quite sure why I had it in my head blessings were only given to members.

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallGuildServicePopupWindow.cs
Allofich: Maybe they expect the most common use of it to be something like UnityEngine.Random.Range(array[0], array.length).
The "int" example on https://docs.unity3d.com/ScriptReference/Random.Range.html looks like it may be the same way. I'm guessing (have not actually looked into it) that if "SceneCount" was 3, you would want to give "LoadScene" either 0, 1, or 2.

PR #652, opened 2018-03-12
Set player death sound by race and gender

Allofich:
This sets the death sound by race and gender.

Background:
There are 3 pain sounds for each race/gender combination in the sound files. Classic only uses one of these, the second in the Male Wood Elf set, for the player regardless of their race or gender in the original game.

By my judgment the sounds generally seem to sound like
1. Hurt/Exertion
2. Hurt/Exertion more
3. Dying

The "hurt" sounds might also be interpreted as exertion while attacking. For example the Breton Female "Hurt 1" is the same as the Nymph's attack. But generally across the board I think they sound more like being hurt than doing an attack. I think the 3rd one is pretty unambiguous across the board as a "dying" sound, so that's the only one I'm using for this so far. I'm thinking maybe the other two could be used as pain sounds for something like > 50% health and < 50% health.

I initially made this as a setting, and I still have that version locally, but this is pretty much going into the realm of bug fix I think, since you shouldn't get a male voice for a female character. If people are nostalgic for the voice used by classic, we could have the setting, but I think maybe we could just wait to see if anybody wants it first.

### DIFF
--- a/Assets/Scripts/Game/PlayerDeath.cs
+++ b/Assets/Scripts/Game/PlayerDeath.cs
@@ -31,8 +31,6 @@ namespace DaggerfallWorkshop.Game
         public float FadeDuration = 2f;
         public float TimeBeforeReset = 3;
 
-        public SoundClips PlayerDeathSound = SoundClips.PlayerDeath;
-
         StartGameBehaviour startGameBehaviour;
         DaggerfallEntityBehaviour entityBehaviour;
         PlayerEntity playerEntity;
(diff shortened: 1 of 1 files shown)


ajrb:
Sounds good (ho ho) :-D
Makes me wonder what it would be like having exertion sounds when swinging weapons & making attacks though.

Assets/Scripts/Game/PlayerDeath.cs
```diff
@@ -162,8 +160,66 @@ private void PlayerEntity_OnDeath(DaggerfallEntity entity)
...
+            // it may have been meant for when the player dies.
+            SoundClips sound = SoundClips.None;
+            PlayerEntity playerEntity = GameManager.Instance.PlayerEntity;
+            switch (playerEntity.RaceTemplate.ID)
+            {
+                case (int)Races.Breton:
+                    if (playerEntity.Gender == Genders.Male)
+                        sound = SoundClips.BretonMalePain3;
+                else
+                        sound = SoundClips.BretonFemalePain3;
```
ajrb: Would be more compact written in a method call like this:

```c#
sound = GetRaceGenderPain3Sound((Races) playerEntity.RaceTemplate.ID)
```

```c#
SoundClip GetRaceGenderPain3Sound(Races race)
{
    switch ()
    {
        case Races.Breton:
            return (playerEntity.Gender == Genders.Male) ? SoundClips.BretonMalePain3 : SoundClips.BretonFemalePain3;
        case Races.Redguard:
             return (playerEntity.Gender == Genders.Male) ? SoundClips.RedguardMalePain3 : SoundClips.RedguardFemalePain3;
        // .... etc ....
    }
}
```

Just a suggestion, functionally it's fine.

Allofich: Sure, I'll rewrite that way.

Allofich:
> Makes me wonder what it would be like having exertion sounds when swinging weapons & making attacks though.

Yeah, I might experiment with that. It may be a bit subjective whether they sound more like being hurt or attacking, and it might not cleanly match across all combinations. Like, for one race/gender combination it may sound OK as an attacking sound, but then with another it sounds more like a being-hurt sound. 

I feel fairly confident about the third sound, though.

Allofich:
Updated to use your suggestion for the formatting.

When I have some time I'll experiment with using the other sounds as attack sounds and/or hurt sounds. Maybe make a setting for them.

PR #653, opened 2018-03-13
Knightly orders done

ajrb:


Interkarma:
This looks good. Lots of work in there. 👍

PR #654, opened 2018-03-14
Quest Packs implementation

ajrb:
Implemented the functionality to load new quests in DFU. It's still only for guild quests for now, but it will load lists and quests from mods as well as StreamingAssets/QuestPacks for people who may not want to have to build, maintain and release a full modpack just to add quests.

Also fixed error in the classic quest list and added a DFU quest list with the CUSTOM01 quest in it. I'm not sure if you wanted to ship DFU with some non-classic quests in, but this is where they should go if you do.


Assets/Game/Addons/ModSupport/ModManager.cs
```diff
@@ -370,6 +370,19 @@ public static string GetAssetName(string assetPath)
             return assetPath.Substring(startIndex).ToLower();
         }
 
+        /// <summary>
+        /// Utility method for gettings lines of text from a text asset
+        /// </summary>
+        public static List<string> GetTextAssetLines(TextAsset asset)
+        {
+            StringReader reader = new StringReader(asset.text);
```
TheLacus: Just a very small note: StringReader implements IDisposable so you probably want a using block.

ajrb: Another instance where I prove I am still transitioning from Java to C#. Great to have stuff like this pointed out - thank you. :-)

Interkarma:
This is awesome thank you!

And thanks for fixing the schema bug in Quests-Classic. I just came across that one this morning and noticed you'd already fixed. Nice one. :)

Interkarma:
Also, I'm happy for CUSTOM01 to be left out. It was just a sample quest to show off quest writing to help community get started. Jay's tutorials do a much better job of that now.

ajrb:
So I should remove the DFU specific quest list then? You don't want to ship with extra quests then?

Interkarma:
Please keep the list, I was really just saying I'm happy to follow your decisions here. I like the fact CUSTOM01 is in for now. :)

PR #655, opened 2018-03-17
Implement classic cure disease cost (WIP)

Allofich:
Implements classic cost for cure diseases.

Some additional work needs to be done. @ajrb, I need your help to figure out some stuff. The main thing, which I think should be fixed before this is merged, is displaying the correct cost amount in the dialogue. Temple healers use the same "bartering dialogue" as shopkeepers. However right now the amount given in the dialogue doesn't reflect the actual amount that curing the disease costs. These messages have a `%a` macro that is used in classic, but in DF Unity it looks like this macro is unimplemented. Why do the shopkeepers show the correct value? We need it for this dialogue, too, and the cure calculation will need to be figured out before the message shows, not after as it is now.

Otherwise, this doesn't have to be in right now, but the cure service gives a discount based on rank. It wasn't obvious to me where to get the player's rank in the temple.

### DIFF
--- a/Assets/Scripts/API/Save/CharacterRecord.cs
+++ b/Assets/Scripts/API/Save/CharacterRecord.cs
@@ -79,6 +79,7 @@ namespace DaggerfallConnect.Save
             doc.startingLevelUpSkillSum = parsedData.startingLevelUpSkillSum;
             doc.minMetalToHit = parsedData.minMetalToHit;
             doc.armorValues = parsedData.armorValues;
+            doc.timeToBecomeVampireOrWerebeast = parsedData.timeToBecomeVampireOrWerebeast;
             doc.lastTimePlayerAteOrDrankAtTavern = parsedData.lastTimePlayerAteOrDrankAtTavern;
             doc.lastTimePlayerBoughtTraining = parsedData.lastTimePlayerBoughtTraining;
             doc.timeForThievesGuildLetter = parsedData.timeForThievesGuildLetter;
--- a/Assets/Scripts/Game/Entities/PlayerEntity.cs
+++ b/Assets/Scripts/Game/Entities/PlayerEntity.cs
@@ -73,6 +73,7 @@ namespace DaggerfallWorkshop.Game.Entity
         protected int thievesGuildRequirementTally = 0;
         protected int darkBrotherhoodRequirementTally = 0;
 
+        protected uint timeToBecomeVampireOrWerebeast = 0;
         protected uint lastTimePlayerAteOrDrankAtTavern = 0;
 
         protected DaggerfallDisease disease = new DaggerfallDisease();
@@ -131,6 +132,7 @@ namespace DaggerfallWorkshop.Game.Entity
         public uint TimeForDarkBrotherhoodLetter { get { return timeForDarkBrotherhoodLetter; } set { timeForDarkBrotherhoodLetter = value; } }
         public int ThievesGuildRequirementTally { get { return thievesGuildRequirementTally; } set { thievesGuildRequirementTally = value; } }
         public int DarkBrotherhoodRequirementTally { get { return darkBrotherhoodRequirementTally; } set { darkBrotherhoodRequirementTally = value; } }
+        public uint TimeToBecomeVampireOrWerebeast { get { return timeToBecomeVampireOrWerebeast; } set { timeToBecomeVampireOrWerebeast = value; } }
         public uint LastTimePlayerAteOrDrankAtTavern { get { return lastTimePlayerAteOrDrankAtTavern; } set { lastTimePlayerAteOrDrankAtTavern = value; } }
         public float CarriedWeight { get { return Items.GetWeight() + ((float)goldPieces / DaggerfallBankManager.gold1kg); } }
         public float WagonWeight { get { return WagonItems.GetWeight(); } }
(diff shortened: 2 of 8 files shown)


ajrb:
Great, I'll work on this and sort out the issues you raised. Not sure how I get this WIP into my fork but google should be able to tell me I hope.

I noticed that you have stuff to handle multiple diseases, I'd implemented based on only being able to have 1 disease at a time, which seemed simpler. I don't see that being able to get several diseases at the same time adds much to the gameplay, but if classic does it then I guess you would like to have that in DFU as well? If so, I guess I need to expand the player disease data. Let me know if it's important to you.

Allofich:
> I don't see that being able to get several diseases at the same time adds much to the gameplay, but if classic does it then I guess you would like to have that in DFU as well? If so, I guess I need to expand the player disease data.

I think we should have it in, yes. It doesn't have to be high priority, though, so it can go on the back burner.
In classic "incubating" lycanthropy/vampirism count as 1 disease for the cost calculation. I think that they do not have a disease record at this point, they just use the time values in the player record that are dedicated to them. When the player has actually become a werebeast or vampire, then these do have disease records, but the temple can't cure them at that point.

Allofich:
Whenever I want to test someone's branch I do it from the Git Bash prompt like this

1. make a test branch to pull into
2. checkout test branch
3. git pull `nameOfGitRepository` `wipBranchName`

and it works. It's probably not necessary but I have always given my test branch the same name as the WIP branch just to be safe. I get the  `nameOfGitRepository`  from the "Clone or download" button on that person's repository.

Allofich:
I should probably do some more research on the multiple diseases thing first, so don't bother yourself with it now.

Allofich:
> It's not much work, and am happy to do it if you want.

I think you may have edited this in? (Or I just missed it before) If it's easy to do and, possibly (but probably not) undo later, then yes, please go ahead. I haven't actually been able to test this part in-game yet.

ajrb:
Yeah, will leave multiple diseases for now. One thing I am not sure about is whether the cost vs. trade price after haggling check for indexing correct message is done before applying rank mod and Arkay benefit. I'll assume it's done before to match normal shopping and if I am wrong I will move it.

ajrb:
Also I don't see how the rank modifier calc is correct. For rank 0 it multiplies the price by 256! Is that right? So members at rank 0 (novice) in a temple have same cost as a non-member?

ajrb:
The rank multiplier is exactly the same as the Arkay rank reduction - is this correct? So Arkay simply applies the rank cost reduction twice?

ajrb:
Right, all done. Have submitted a PR to your fork repo branch. Accept and then this PR should auto-update I think. Love this teamwork stuff.

Allofich:
Thanks, merged. I need to tweak a few things, though.

Allofich:
Updated. A few things:

1. I put it back to using bit shifts and integers, as otherwise the final result is slightly off from classic. It makes it easier to check against in-game behavior this way and I figure we might as well get the same values as classic does.
2. The way you had it set, ajrb, the override ReducedCureCost was not being called, the base one from the Guild class was. I've just redone it to avoid calling out to any other classes. Personally I find this a lot simpler and easier to deal with then having to follow execution across different classes with overrides and guild managers and all this stuff, but if you want to rework it to match with the other similar functions you can. It might be a good idea to check that those other similar functions, like ReducedIdentifyCost, are also actually going to the override version of the function and not the one in the base Guild class.

ajrb:
The reason for the guild implementations to have the logic for benefits is so new guilds (from mods) can easily provide the same benefit by overriding the methods. Hardcoding it to Arkay temple is easier to follow in the code, but means that benefit isn't available to mods. 

I'm pretty sure that the overrides are being called correctly, how did you test this?

Allofich:
> I'm pretty sure that the overrides are being called correctly, how did you test this?

I wasn't getting the expected result, so I put in a Debug.Log statement and it indicated that the base Guild class was being called.

I'll double-check to be sure.

Allofich:
> The reason for the guild implementations to have the logic for benefits is so new guilds (from mods) can easily provide the same benefit by overriding the methods.

Okay, that's a good reason. If you figure out the issue (sorry this is a hobby project after all, and I don't really feel like hunting that down right now...) and post here what to do, I can go in and edit that commit. Or if it's kind of complicated maybe another PR.

Allofich:
For now I'll go ahead and change that commit to not remove the function, but I think I'll leave in the bit shifts for now for the reason I said above.

ajrb:
Yep bit shifts are just fine. Leave them in.

Are you sure you are a member of the temple you're asking for a cure from? If not it will simply call the guild verson.

ajrb:
GuildManager returns a NonMemberGuild instance when you're not a member. So if you joined Kynareth and ask for cures in Arkay temple, the basic guild class will handle it. The override member benefit method doesn't get called. This is as designed.

Allofich:
> GuildManager returns a NonMemberGuild instance when you're not a member. So if you joined Kynareth and ask for cures in Arkay temple, the basic guild class will handle it. The override member benefit method doesn't get called. This is as designed.

Ah, whoops, well then it's probably working fine then. Let me try with a higher rank than 1 in Arkay and see if it gives me the correct discount.

Allofich:
Yes it does work, sorry for the misunderstanding.

So this should be ready for merging, then.

Allofich:
Oh, and in that case I guess I don't need that "rank != -1" check. Let me remove it.

Allofich:
Removed it.

ajrb:
I forget sometimes that object oriented programming isn't always obvious to people used to pure procedural programming, probably since I've been doing OOP at work for over 15 years. I should have explained that better and sooner, sorry. The advantage OO gives for the harder to follow code path it results in, (this case allowing mods to adjust cure pricing if they offer cure service) isn't always obvious either.

Anyway looks good now. One question about vamp/lycan diseases where you added a new variable to player entity - is this handled differently enough to the normal diseases that it should be done outside of the daggerfallDisease class do you think?

ajrb:
makes sense, lets see how it pans out. They're quite involved with essentially changing the player character class.

Interkarma:
Thank you guys for all your work on this 👍

PR #656, opened 2018-03-18
Bugfix: save factionId with quest

ajrb:
This was a pretty bad and stupid oversight on my part! This fix stores the faction id so that the rep is changed correctly when the quest finishes. Currently it only changes the faction rep if you don't save and reload between taking the quest and completion which is pretty rare.

Would you please create a new build and remove the #102 build once this bugfix is merged and validated. I consider it that serious. (as in it breaks the whole guild system since you can't get promotions)


ajrb:
Fixed some quests that require the guild name of the qgiver, this is probably not complete.

Interkarma:
No problems, will issue new builds today and remove 102 as requested. It's OK to have problems like these and fix them, it's all part of the live build process.

PR #657, opened 2018-03-18
Fix enemies without ranged attacks not pursuing

Allofich:
At some point I seem to have accidentally made it so that all enemies hold their ground at range, not just those with ranged attacks. This fixes it.

### DIFF
--- a/Assets/Scripts/Game/EnemyMotor.cs
+++ b/Assets/Scripts/Game/EnemyMotor.cs
@@ -355,7 +355,10 @@ namespace DaggerfallWorkshop.Game
             if (!senses.TargetIsWithinYawAngle(5.625f))
             {
                 TurnToTarget(direction.normalized);
-                return;
+                // Returning here is closer to classic behavior but disabling
+                // it provides an easy trick to make enemies harder to outmaneuver,
+                // since the enemy will resume moving. May revisit this later.
+                //return;
             }
 
             var motion = transform.forward * moveSpeed;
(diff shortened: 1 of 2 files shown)


Allofich:
Fixed some more issues.
- Pursuing when at long range was unnecessarily being slowed down
- Enemies were not properly pursuing the last spot they saw the player when they lost LOS

I'm pretty sure these were both problems that I accidentally introduced while working on the AI some time ago.

Finally, while working on these fixes I experimented with taking out a return statement so that AI would still move while turning to face the player while pursuing them. Coincidentally I found this helps with the issue of being able to outmaneuver them too easily. It also has the benefits of giving them slightly less predictable behavior and giving more opportunity to see the walking animations from the side.

As a downside it can cause them to walk into walls or off of ledges a bit more, but overall I found the effect more positive than negative, so I thought I'd tack it on for people to play with. And the issue of walking off ledges is something that needs to be addressed regardless.

I still am trying to recreate the original AI as closely as possible but maybe little enhancements like this can later be grouped together as an improved AI mode.

Interkarma, if you're planning on releasing another build, maybe give the AI a spin first. If you don't like it you can revert the second commit. Or if you haven't merged the PR by that point I could remove it.

ajrb:
Great, the holding ground was quite exploitable with a bow :-)

Allofich:
Yeah I watched a video of GalacticChicken's some time ago and noticed the enemies he shot at with the bow were just standing there. And I got one of those "Aw man, I bet something I did caused that." feelings. :)

Interkarma:
Thank you for continuing to refine this. 👍

PR #658, opened 2018-03-18
Texture replacement: Import billboard textures from mods

TheLacus:
Textures for billboards can now also be imported from mods, in addition to loose files. This include xml files with optional scale and uv.

I was hesitant to do this at first because there is a possible issue of unwanted mixing of resources when a mod includes a texture with an xml, and another one in a lower position includes the same texture but without the xml. The same may happens for emission maps. If this became a problem we could fix this at a later time loading everything only from the last mod that include the first main texture frame. 

What's left now are textures for 3d meshes. I think it would be a pity not to benefit of Unity materials, so my idea is to load them directly. It's a bit of an hassle for the most simple texture packs with only the albedo, but i believe is the wisest idead in the long run. Textures from loose files would still be imported individually and set on the (default or imported) material. What do you think?


Interkarma:
This is great! Agree it would be nice for model to benefit from Unity materials.

It's always possible for someone to contrive a weird game-breaking mix of mods, but that's much the case for any other game. The best quality mods will always float to the top without these possible issues.

PR #660, opened 2018-03-20
Lay groundwork for random enemy encounters

Allofich:
This adds the groundwork for random enemy encounters implemented like classic. Right now there is are no functional changes.

Adds:
- All enemy encounter lists from classic
- Classic function for choosing enemies (not being used for anything yet, and not tested)
- Incomplete classic function for spawning random enemy encounters

Interkarma, I hope maybe you can give some advice for filling out the remaining parts. These are the remaining issues:
- The main issue: How to spawn a nearby enemy? I know this is implemented for quests, but is it tightly wound up in that, or is there a convenient method available for spawning an enemy nearby, with the only parameter being the enemy type?
A couple of minor issues:
- Need to set and store a variable to disable monster spawning in some circumstances(fine to just add it to the PlayerEntity?)
- Need to somehow check whether in the rest menu. Classic has a global variable that stores the current popup window mode and just checks that from within this function. I wonder if we have a similar convenient way to handle it all just from within this one function.

Also this function, like the regional price adjustments and climate changes, should perhaps not be in the PlayerEntity. I've just been sticking everything in there from classic that is involved with updating the game world when the in-game time (minutes) changes.

### DIFF
--- a/Assets/Scripts/Game/Entities/PlayerEntity.cs
+++ b/Assets/Scripts/Game/Entities/PlayerEntity.cs
@@ -299,9 +299,70 @@ namespace DaggerfallWorkshop.Game.Entity
                 }
             }
 
+            for (uint l = 0; l < (gameMinutes - lastGameMinutes); ++l)
+                IntermittentMonsterSpawn(l + lastGameMinutes + 1);
+
             lastGameMinutes = gameMinutes;
 
             //HandleStartingCrimeGuildQuests(Entity as PlayerEntity);
+        }
+
+        public void IntermittentMonsterSpawn(uint Minutes)
+        {
+            //TODO: if (InOutsideWater)
+            //          return;
+            bool timeForSpawn = ((Minutes / 12) % 12) == 0;
+
+            //TODO: Monster spawns are prevented when time is advanced for: 1) Traveling 2) Loitering 3) Turning into vampire 4) In prison
+            bool preventMonsterSpawns = false;
+
+            if (!preventMonsterSpawns && timeForSpawn)
+            {
+                if (!GameManager.Instance.PlayerEnterExit.IsPlayerInside)
+                {
+                    uint timeOfDay = Minutes % 1440; // 1440 minutes in a day
+                    if (GameManager.Instance.PlayerGPS.IsPlayerInLocationRect)
+                    {
+                        if (timeOfDay < 360 || timeOfDay > 1080) // night
+                        {
+                            if (UnityEngine.Random.Range(0, 24) == 0)
+                            {
+                                MobileTypes enemy = RandomEncounters.ChooseRandomEnemy(false);
+                                // TODO: Spawn enemy
+                                //SpawnEnemyNearPlayerInLocation(enemy);
+                            }
+                        }
+                    }
+                    else
+                    {
+                        if (timeOfDay >= 360 && timeOfDay <= 1080) // day
+                        {
+                            if (UnityEngine.Random.Range(0, 36) != 0)
+                                return;
+                        }
+                        else // night
+                            if (UnityEngine.Random.Range(0, 24) != 0)
+                            return;
+                        MobileTypes enemy = RandomEncounters.ChooseRandomEnemy(false);
+                        // TODO: Spawn enemy
+                        //SpawnEnemyNearPlayerInWilderness(enemy);
+                    }
+                } // in interior
+                else
+                {
+                    if (GameManager.Instance.PlayerEnterExit.IsPlayerInsideDungeon)
+                    {
+                        // TODO: If in rest menu
+                        if (UnityEngine.Random.Range(0, 36) == 0)
+                        {
+                            // TODO: Not sure how enemy type is chosen here.
+                            MobileTypes enemy = RandomEncounters.ChooseRandomEnemy(false);
+                            // TODO: Spawn enemy
+                            //SpawnEnemyNearPlayerInLocation(enemy);
+                        }
+                    }
+                }
+            }
         }
 
         /// <summary>
--- a/Assets/Scripts/Game/Entities/PlayerEntity.cs
+++ b/Assets/Scripts/Game/Entities/PlayerEntity.cs
@@ -34,6 +34,8 @@ namespace DaggerfallWorkshop.Game.Entity
         #region Fields
 
         bool godMode = false;
+        bool preventEnemySpawns = false;
+        bool isResting = false;
 
         const int testPlayerLevel = 1;
         const string testPlayerName = "Nameless";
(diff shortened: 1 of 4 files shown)


ajrb:
Ooh, this is exciting! Random encounters will be awesome.

I agree that this doesn't belong in PlayerEntity, hopefully Interkarma will have a good suggestion for a place for something that needs to run as gime time changes.

Checking the rest menu, is that the only one it checks - not travel or transport etc? Anyway, UserInterfaceManager.ContainsWindow() should allow for checking whether a UI window is up or not.

Interkarma:
This is a great start, thank you!

> The main issue: How to spawn a nearby enemy? I know this is implemented for quests, but is it tightly wound up in that, or is there a convenient method available for spawning an enemy nearby, with the only parameter being the enemy type?

No problems. I'll split this out from CreateFoe and allow a generic overload for just enemy type. Will get to this as soon as I can.

> Need to set and store a variable to disable monster spawning in some circumstances(fine to just add it to the PlayerEntity?)

That's fine. I understand your concerns around things like regional price adjustments. Maybe this can be moved into a world state class at some point to separate from player object. I'm happy with this in PlayerEntity for now to keep things moving. The player is the centre of the world in this game, after all.

>  Need to somehow check whether in the rest menu.

If you just need if rest window is active, you can use DaggerfallUI.Instance.UIManager.TopWindow property to get a reference to current window at top of stack. From there should be easy to check if class type is DaggerfallRestWindow.

`if (DaggerfallUI.Instance.UIManager.TopWindow.GetType() == typeof(DaggerfallRestWindow))`

Could probably still use "is" in this case as well, even though we're not checking for inheritance.

`if (DaggerfallUI.Instance.UIManager.TopWindow is DaggerfallRestWindow)`

Hope that helps!

Interkarma:
For rest window, another option might be to create an event you can hook that ticks every hour. This means you can just listen for event and trigger encounter check then. Would that be easier?

ajrb:
Can random encounters happen when you fast travel as well, or would that be a new feature?

Allofich:
They don't. They're disabled because otherwise (I assume) you could end up with a ton of enemies spawned when you appeared at the new location, because of all the in-game minutes that would have passed between the start and end of the fast travel.

Random encounters or other events while traveling, like in Fallout (which I assume you like :) ) would be cool, but that's something to look at later IMO. BTW there is an unused old menu in the Daggerfall graphics for fast travel that including options for stopping at inns, dungeons, etc. that are encountered while traveling. It might be interesting if you could uncover dungeons along the path of a fast travel. Another enhancement I've thought of is instead of the game warning you when you have an incubating disease about the danger of traveling, that the player could be dropped out of fast travel at a location along their travel path at the point that the disease symptoms begin to appear.

Right now I'm just looking at recreating the classic behavior, though.

ajrb:
How ever did you guess I liked Fallout? ;-)

Yeah classic behaviour first. I just wasn't sure if classic did encounters when fast travelling. I suspect they intended to because of the reckless option for travel. I like the idea of dropping player out of fast travel if they encountered anything (enemies, disease symptoms etc) and would be a reason to see bits of the world that are not locations like towns.

Allofich:
Yeah the cautious vs. reckless travel option feels like it wasn't developed enough. I think you can easily exploit it by just recklessly traveling to a location near your destination, then using cautious for the short remainder (although I purposely avoid using exploits like this so haven't actually tried it) to get the benefits of healing and arriving at day without the time penalty. Encounters mid-travel seems like a natural way to expand on this. Maybe if you were cautious traveling you could spot enemies on the way and decide whether to fight or avoid them. Always fun to imagine this kind of stuff, and so much easier than actually creating it.

Fallout was a great game, definitely :)

Allofich:
> I purposely avoid using exploits like this so haven't actually tried it

Actually I lied. Now that I think about it I used this trick a number of times to avoid arriving at Daggerfall at night and make sure I didn't exceed quest time limits.

ajrb:
I was thinking cautious would avoid encounters and reckless would have a chance of them stopping your journey for a fight / hold up. I think having to click to decide whether to avoid or not would be fine as long as encounters were limited in number so you never had to click several times to simply get where you're going.

Allofich:
Yeah, maybe so. At any rate it might not work very well with the default random encounters because I think you would encounter a ton of enemies on any long trip so you would constantly be getting stopped.

Allofich:
Updated

- Random enemy creation is prevented for the time raised by fast travel
- For dungeons, random enemies only appear when in rest mode
- Added a fix for the "lastGameMinutes" in the playerEntity. I found that when loading a Unity save, the playerEntity was allowed to run Update() after the worldtime was updated to the saved game world time but before the playerEntity's lastGameMinutes value was updated to the saved game world time. This caused an infinite loop when the game being loaded had an earlier world time than the game being loaded from. I fixed this by temporarily halting player update during the loading process. I also added a check to make sure the infinite loop situation won't happen. Loading from classic did not have this problem.

- I did not prevent enemy encounters during loitering, as classic does. I don't know why classic does this. Maybe the idea is enemies don't approach the alert player? At any rate I've left this out for now.

All that's left is to place in the enemy creation functions.

Interkarma:
> Just to let you know, this isn't only for when resting, encounters can happen when you're awake. As I understand it an enemy has a chance to appear every minute of a 12-minute period that begins every 144 minutes of game time.

My apologies, I obviously wasn't paying attention this morning when I posted. :) Will make sure this works as required.

Assets/Scripts/Game/Entities/PlayerEntity.cs
```diff
@@ -192,6 +201,9 @@ public static int GetRemainingHours(RoomRental_v1 room)
 
         public override void Update(DaggerfallEntityBehaviour sender)
         {
+            if (preventUpdate)
```
ajrb: Is there anything stopping this from working using the existing mechanism? In other words change this to:

``if (SaveLoadManager.Instance.LoadInProgress)``

Allofich: Yes that should work. Good I can remove that bool then.

Allofich:
Removed the superfluous bool I had added to stop updating. There was already functionality there as ajrb pointed out.

Also removed the check that lastGameMinutes is <= gameMinutes. This should never happen now and if  it does that's a problem that needs to be fixed.

Interkarma:
Thank you for the all the hard work Allofich! I'll get this merged by the weekend and add that spawning method. Looking forward to seeing all this in action when ready.

Interkarma:
@Allofich have added a generic foe spawner now in recent merge. Let me know if any problems using.

PR #662, opened 2018-03-22
Handle exterior NPCs and add Witch Coven popup with quests

ajrb:
Also:
-Added social quests (merchants, nobles etc)
-Expanded utility of dumplocblocks command, the command can now take multiple block names.
-Added another shelf id


ajrb:
RmbBlockFlatObjectRecord & RmbBlockPeopleRecord are now identical structs, so we should discuss maybe merging them. I wasn't sure how you would want to handle this.

ajrb:
I found that the merchant questor stuff that was added by Numidium has been (mostly) broken by the update I made to nameseed. I have a fix ready but I need buildingKey at line 1575 of TalkManager and was not fully confident of the best way to ensure I get it in here. Interkarma, could you assist here and give me a line of code that would work correctly in here when you get a chance please?

Then it should work as it did originally, but now working with the quest systems I've built.

numidium:
@ajrb - My apologies, I'm currently unable to PM you on the forums. My account there is brand new and it won't let me send PMs until I've made more posts.

It's been a couple of weeks since I've looked at the code in depth but my plan was to have a separate dictionary for nobility questors. AFAIK, nobility is the only other social group besides merchants that offer quests instead of opening a dialogue window at random. Since they aren't mentioned by NPCs when you ask about work, we can wait until the player enters a castle to generate their dictionary.

When I was first working on this I thought we could populate nobles by adding another case in the inner loop of SetMerchantQuestors but the nobility is usually found in dungeon blocks and I believe those are structured differently from the exterior blocks we fetch merchant People records from.

Extending the work system to nobility should just be a matter of making a function that populates a dictionary from dungeon (castle) block data instead of exterior block data.

Interkarma:
> I have a fix ready but I need buildingKey at line 1575 of TalkManager and was not fully confident of the best way to ensure I get it in here. Interkarma, could you assist here and give me a line of code that would work correctly in here when you get a chance please?

Line 1575 of TalkManager is an inner loop of the following:

`foreach (BuildingSummary buildingSummary in buildingsInBlock) {}`

So you should be able to get the current buildingKey using:

`buildingSummary.buildingKey`

Alternatively, you can use the static MakeBuildingKey() method to construct the key using blockX, blockY, recordIndex of building. At line 1575 this should yield the same as buildingSummary.buildingKey:

`int buildingKey = BuildingDirectory.MakeBuildingKey(x, y, j);`

My apologies if I've misunderstood and not been helpful. Just let me know and I'll try to help further. :)

ajrb:
@Interkarma - thanks, sorry didn't look up far enough last night. I got as far as realising MakeBuildingKey wasn't going to work.

@numidium - no problem, here is fine. That should be fine with my changes when you get to it. What about the peasant/commoner quests 'A*.QRC' though? How do they get started, do you know?

Allofich:
I've been looking at the regional "conditions" (famine, plague, etc.) of http://forums.dfworkshop.net/viewtopic.php?f=19&t=890, and 
 I think part of this data may be related to nobles offering quests, just to let you know. It may be quite a while though before I figure that part out, though.

ajrb:
The code had a big issue where it was looking at people records for every building in a block, for every building in a block. Found it because the building keys I was getting were not what they should have been. Anyway after spending several hours I think I have it all fixed up and working.

Switched to store building summary for merchant questors, but I think just the building name could be stored, and that way buildings that don't resolve (I assume merchant npc's in unnamed buildings) can be filtered out as well. @numidium is there any reason why more building info is being stored along with the npc data? I can't see any need, but maybe you have plans that require it?

ajrb:
@Allofich - thanks, I think that stuff can be plugged in once the framework is there. I also think we need to look at how merchants are selected in classic for being questors, in this code each merchant in a location has a 25% chance of offering a quest and I'm not sure that matches classic.

numidium:
@ajrb 

>What about the peasant/commoner quests 'A*.QRC' though? How do they get started, do you know?

I'm looking at the quests whose file names start with "A" on UESP and they appear to all be offered by merchants (shopkeepers/innkeepers). I don't know of any peasant/commoner quests.

>is there any reason why more building info is being stored along with the npc data? I can't see any need, but maybe you have plans that require it?

I was only using the building info to fetch the building name. If we're storing that by other means then there's no need.

ajrb:
Thanks, yes looks like UESP has those in the merchants list. Maybe they never go around to implementing a peasant quest offer system and just allocated them to merchants. UESP also says "You can be assigned these quests by any NPCs in guildhalls, temples, taverns, etc. However, the chances of getting them from merchants and innkeepers is much higher."

No idea how accurate that is. I might need to experiment with classic a bit. Did your research ever turn up a quest from a non-merchant npc?

numidium:
Nope, I've never seen quests offered from merchants who weren't shopkeepers or innkeepers come to think of it. I think there are NPCs who belong to the merchant social group in those other places the article mentioned so we might have to specify _only_ shopkeepers/innkeepers in SetMerchantQuestors instead of just the broader merchant social group.

Interkarma:
Thanks for all your work on this! 👍

JayH2971:
Occasionally, some non-merchant NPCs will offer merchant quests. Occasionally you'll get a rumor telling you that the questgiver for the Mages Guild is offering work, for example. If you then use the "talk" instead of the "get quest" option, you'll get a merchant quest rather than a guild one. I recently read a comment by someone (don't remember where) who said that the Mages Guild questor asked to be taken to a distant city, and because of that quest the questor has never returned to the guild hall.

Br'itsa, the gem maid, will infrequently offer merchant quests in classic.

Very rarely, random patrons in taverns can offer a quest.

ajrb:
Thanks Jay, that's really useful info which matches my current theory that daggerfall assigns both merchant and commoner/peasant quests to npcs of either social group simply because most commoners are not in named buildings, so asking about work cannot easily point you in their direction. There are less commoners in named buildings than merchants so that is why they less commonly give out the quests. 

UESP implies that the merchant and commoner quests A* & K* were essentially combined due to this and/or lack of time. I plan to update the work mechanism to identify merchants, commoners and nobles (except ones in the 3 dungeon palaces in the game) and assign work. We can see how that feels, it may need to be combined because it's harder to get the commoner quests, but we can easily do that later.

PR #663, opened 2018-03-25
Modify Assets/Resources/Songs/sunnyday.mid , it was too damn long.

heilkitty:
Assets/Resources/Songs/sunnyday.mid had empty MIDI events at the end, so its length was 23:58 total, most of it just silence. Removed them in Muse, now it's 0:56, perfectly reasonable.


Interkarma:
Cheers, it's nice someone was able to fix that. 👍

PR #664, opened 2018-03-26
Spawn random encounters

Allofich:
Random encounters spawn now. Thanks for the FoeSpawner, Interkarma.

I don't know if adding a FoeSpawner to the "options" part of DaggerfallUnity prefab is correct style-wise, I just followed the example of existing uses of InstantiatePrefab.

In game it works fine. While in the wilderness in classic enemies actually spawn from a little farther away, but currently at that distance enemies don't react to the player in DF Unity nor does the "There are enemies nearby" message appear, so for now I just have it using the default FoeSpawner distances.

### DIFF
--- a/Assets/Scripts/Game/Entities/PlayerEntity.cs
+++ b/Assets/Scripts/Game/Entities/PlayerEntity.cs
@@ -347,11 +347,7 @@ namespace DaggerfallWorkshop.Game.Entity
                         if (timeOfDay < 360 || timeOfDay > 1080) // night
                         {
                             if (UnityEngine.Random.Range(0, 24) == 0)
-                            {
-                                MobileTypes enemy = RandomEncounters.ChooseRandomEnemy(false);
-                                // TODO: Spawn enemy
-                                //SpawnEnemyNearPlayerInLocation(enemy);
-                            }
+                                GameObjectHelper.CreateFoeSpawner(RandomEncounters.ChooseRandomEnemy(false), 1);
                         }
                     }
                     else
@@ -364,9 +360,7 @@ namespace DaggerfallWorkshop.Game.Entity
                         else // night
                             if (UnityEngine.Random.Range(0, 24) != 0)
                             return;
-                        MobileTypes enemy = RandomEncounters.ChooseRandomEnemy(false);
-                        // TODO: Spawn enemy
-                        //SpawnEnemyNearPlayerInWilderness(enemy);
+                        GameObjectHelper.CreateFoeSpawner(RandomEncounters.ChooseRandomEnemy(false), 1);
                     }
                 } // in interior
                 else
@@ -376,12 +370,8 @@ namespace DaggerfallWorkshop.Game.Entity
                         if (isResting)
                         {
                             if (UnityEngine.Random.Range(0, 36) == 0)
-                            {
                                 // TODO: Not sure how enemy type is chosen here.
-                                MobileTypes enemy = RandomEncounters.ChooseRandomEnemy(false);
-                                // TODO: Spawn enemy
-                                //SpawnEnemyNearPlayerInLocation(enemy);
-                            }
+                                GameObjectHelper.CreateFoeSpawner(RandomEncounters.ChooseRandomEnemy(false), 1);
                         }
                     }
                 }
--- a/Assets/Scripts/Utility/RandomEncounters.cs
+++ b/Assets/Scripts/Utility/RandomEncounters.cs
@@ -1346,7 +1346,7 @@ namespace DaggerfallWorkshop.Utility
             if (chooseUnderWaterEnemy)
                 encounterTableIndex = 19;
             else if (playerEnterExit.IsPlayerInsideDungeon)
-                encounterTableIndex = ((int)playerGPS.CurrentLocationType);
+                encounterTableIndex = ((int)playerEnterExit.Dungeon.Summary.DungeonType);
             else if (playerEnterExit.IsPlayerInsideBuilding)
             {
                 DFLocation.BuildingTypes buildingType = playerEnterExit.BuildingType;


Allofich:
> In game it works fine.

Well actually it's a little rough around the edges. This will probably need continued work. It's playable, though.

Interkarma:
That was quick, thank you! :)

No need to place the prefab object in options, my apologies for not making this easier. I've just added GameObjectHelper.CreateFoeSpawner() that should help you do this in a single call when needed. You can remove the option settings in DaggerfallUnity singleton class.

Also, would you like an option to disable line of sight check for placement? It's probably a good idea to keep this enabled for live spawns but resting spawns can use any area around the player with no concern about pop-in.

Allofich:
OK, thanks. I'm using the new function you added.

> Also, would you like an option to disable line of sight check for placement? It's probably a good idea to keep this enabled for live spawns but resting spawns can use any area around the player with no concern about pop-in.

Yes, I think that would be good to have, thanks.

Interkarma:
No worries, will get that option in soon!

Interkarma:
@Allofich I'm getting a few reports of people experiencing random spawns inside rented tavern rooms or their own homes, or in guild halls. Would you be happy for me to disable intermittant spawns just to get out a fresh build without this behaviour? I don't remember ever getting intermittent spawns while inside buildings in classic (only in dungeons and outside at night).

The frequency of random spawns in dungeons is also very high compared to classic. Even if you've matched the percentage chance, the check is running inside the Update() loop which means it's happening every frame. For end users with high frame rates, this will result in more spawns than users with low frame rates. This is something we'll need to fix, as random checks like this shouldn't be frame-rate dependent. Rather, they should run once per fixed amount of time (e.g. once every game second or whatever is appropriate).

Just let me know if any help needed. I'm more than happy to help tighten this up with you. And I still need to give you an option to spawn without a LOS check. :)

Allofich:
Just go ahead and disable spawns for now. There are a few related fixes I was planning anyway. Enemies in classic spawn farther away from the player than they are in DF Unity right now. But I think that right now in DF Unity if they spawn that far away they cannot see the player because of their vision is too short, so that needs to be changed as well. And also DF Unity has a shorter distance check for "enemies are nearby" than classic and that may need to be changed as well. This should help with guards, though, giving them more of that "they show up from the distance somewhere and are chasing me" feeling than just suddenly appearing by the player.

Also what I said above about waking the player up mid-hour might not be correct, as I'm pretty sure I've sometimes woken up in classic with more than one enemy spawned. I'll investigate closer into how classic handles that.

Interkarma:
> A spawn can only happen when ((Minutes / 12) % 12) == 0

This isn't happening in dungeons though. Step through with debugger or place some Debug statements to see what I mean. Here's the general flow right now.

1. Rest steps game time forward by 60 minutes.
2. The following loop iterates once per game minute passed (so 60x, one per minute of rest), calling IntermittentEnemySpawn() each time.

`for (uint l = 0; l < (gameMinutes - lastGameMinutes); ++l)`

3. In IntermittentEnemySpawn(), it only does the timeForSpawn check in first part of conditional block. Otherwise it flows straight into the else statement where dungeon spawning is done.

4. This does your roll of `UnityEngine.Random.Range(0, 36) == 0` but it's not sensitive to any time limits. So this chance is being hit up to 60x times per hour of rest. The player is being woken up by anywhere from 1-60 enemies not because they're not being woken up. It's because all 1-60 enemies are being placed in a single rest tick.

> Just go ahead and disable spawns for now

Cheers. I'll just disable building spawns for now so player isn't harassed just by going in and out of buildings around town. And I'll change how the dungeon spawns are done so they respect the same timeForSpawn check. I'll also add a boolean break so the time catch up will stop iterating once something spawns.

Interkarma:
Stepping through this again, I was wrong earlier about the upper limit to number of enemies per rest tick. It was being limited by the timeForSpawn 12-minute check. But with 60 hits per rest tick, this still allowed for up to 5 spawns per tick in dungeons with an unlucky roll.

I'm also not sure why people are getting spawns in towns while going in and out of buildings. This might be related to something in the back-end like a location rect check failing causing code to flow through to wilderness logic. I'll look into this more later.

I've restructured it a little to make flow easier for me to read with a bit less nesting, but have otherwise left it exactly as you wrote it. The only major change is to break when an enemy is selected so player should wake up rather than allow multiple spawns during that update cycle. I'll test and post this soon.

Interkarma:
Looks like there are two reasons player isn't being woken up by spawns.

The first is the loop mentioned above, so multiple enemies can be queued during one rest tick. The second is the queuing system the spawner uses. The enemy is queued but not in world until a good spawn position is found. This might take multiple cycles during which more enemies can spawn. This can keep going if player is very unlucky.

I've added a BreakRestForEnemy() method to DaggerfallUI that will notify rest UI if it should break for an enemy for any reason. This will only trigger if rest UI is open and a random enemy spawns.

Interkarma:
I've also added an option to disable line of sight check in FoeSpawner. This allows us to place foes at a greater distance than before during rest. You're right that they need to be able to see the player from a bit further afield, we can tune this.

Interkarma:
I made some refinements to how rest is broken and removed my hacky method to forcibly break rest from outside UI.

The new method considers FoeSpawner to be an enemy in potentia for purposes of preventing rest or fast travel, and for breaking rest in progress. This makes the break automatic as part of normal enemy detection and fixes a bug where player was able to immediately re-initiate rest while enemy spawn was pending but not yet released by foe spawner.

The other nice side effect of this is that if player is in a weird spot where nothing can spawn (e.g. a very small room) this gives the illusion enemy spawned somewhere outside the room - because spawn will release the moment a ray seeks a suitable spot (e.g. player exits small room).

The end result feels much nicer with only a single spawn and more interesting placement under previously impossible spawn conditions. Enemies still spawn frequently enough to harry player from continuous rest, but hopefully not so frequently they can still heal most times.

PR #665, opened 2018-03-26
Start implementation of regional conditions

Allofich:
This is the start of implementing regional conditions. Only famine is handled now, and there are no in-game effects, only debug.log messages about the famine phases. As with randomEncounters, this is being handled in PlayerEntity for now but could be moved later. This is all matched to classic behavior.


Interkarma:
Always a pleasure watching you work this stuff out and get it into the game. :)

PR #666, opened 2018-03-26
Fix bug with resting in houses causing exception cant find guild for …

ajrb:
…faction id 0

Thanks for finding this Allofich, and apologies if it got in your way or caused you any grief. When I changed the guild manager design while doing temples I missed that statement.


Allofich:
No trouble, it's just something I noticed while testing.

PR #667, opened 2018-03-27
 Generalise the social quest work allocation system to cover commoners & nobles

ajrb:


Interkarma:
This is looking good!

PR #669, opened 2018-03-29
 Add TG + DB, make non-member quests availiable when rep less than 0 

ajrb:
3 days ago I thought I would knock out the remaining two guilds that evening... yeah, there was a lot of extra stuff to do than I expected so that didn't happen. 

Here it is now though - all types of quests except for daedra are now availiable in game!


Interkarma:
Wow that's a lot of work! Great stuff, looking forward to having these guilds in next round of builds.

PR #671, opened 2018-03-30
TextureReplacement: some size fixes

TheLacus:
Fixed a few windows reported on forums. Also removed a workaround for paperdoll background which is not necessary anymore since we can cut imported textures now.


Interkarma:
Thank you for fixing these!

PR #672, opened 2018-03-30
Work on diseases

Allofich:
Does the following:

1 - Clean up, simplify disease code.
2 - Import diseases from classic, all data identified.
3 - Support for multiple diseases, as in classic. Can show multiple diseases in status window, also as in classic.
4 - Classic's "incubation" figured out. Diseases just go into effect as soon as the date changes. Put in an approximation of this for DF Unity. It's kind of mechanical for diseases to always take effect exactly at midnight, so made it a random value of 1 to 1440 minutes (full day), effectively the same  min and max values it could be in classic.
5 - Matched disease enums to classic


Allofich:
I realized the incubation thing I added is pretty pointless, because the application of disease effects happens based on the date changing anyway, so it was just giving a chance for an extra 24-hour buffer. I took it out and behavior is the same as classic now.

ajrb:
I think you should swap our names round on DaggerfallDisease, you're the author now I think, :-)

Looks good at a first pass, and matches classic wheras I was guessing with a lot of what I did. Hope you found my code easy to change and it wouldn't have been easier doing from scratch.

Interkarma:
Great work, thank for this!

PR #674, opened 2018-03-31
Fix quests

ajrb:
Have gone through all guild quests and removed all changes Jay_H made to work around unimplemented quest system features. These will all start working once those features are finished.

Still a few work arounds in the TG + DB joining quests to allow players to join for now.

Updated all notes in quest list, and removed all the errors that were caused by testing using command line instead of a questor.

Fix commoner assigned quests by finding people of region faction id and allowing faction id 0 on static npc.


Interkarma:
Thank you Hazelnut!

PR #676, opened 2018-04-01
Fix loading issue with active commoner quests

ajrb:
Quest system is not designed for commoners who have no faction in DF data. We should discuss if the faction should be set to same as I use for work npc data, but I decided to leave it alone and simply allow faction id of 0 for questors.

Also fixed some guild quest macros by adding an external MCP to quest (tried keeping it external, but some messages are generated internally only - e.g. journal entries) so that any %abc macros not handled by quest MCP can be handled by an appropriate context MCP such as guild.


Interkarma:
Thank you for the fix. I honestly don't believe I considered these NPC quests at all when building this out, at that's a nice workaround. 

I appreciate your work on handling those macros. You seem to have a good solution for handling the internal macros or passing off to MCP. Will let you know if any problems pop up during playtesting.

PR #680, opened 2018-04-06
Add Camera Recoil to DFU

MeteoricDragon:
Adds head recoil (screen stagger) when taking damage.  "Camera Recoil" Toggle on the first page of settings.  Default is ON.  Camera sway is **increased per point of damage taken**. Formula is in GetRotationVector() in HeadRecoiler.cs.


Interkarma:
Thank you. :) Sending to a branch to test now.

PR #681, opened 2018-04-07
Add "already defaulted" bank behavior

Allofich:
Just a note, I started to add behavior for reading classic's bank records because it wasn't in what I thought was the "usual" spot, in ReadRecords() in SaveTree.cs. It seems like that should be consistent. But at least it meant the work of converting the read data to DF Unity was already done.

Maybe Interkarma though the same thing when he made this bug report?
http://forums.dfworkshop.net/viewtopic.php?f=24&t=788
Or maybe he made that before the code to read the data was added? Anyway the boolean that was being skipped is now read and used, so that report can be closed.

### DIFF
--- a/Assets/Scripts/Game/DaggerfallBankManager.cs
+++ b/Assets/Scripts/Game/DaggerfallBankManager.cs
@@ -34,7 +34,7 @@ namespace DaggerfallWorkshop.Game.Banking
         NOT_PORT_TOWN           = 0285,
         ALREADY_OWN_HOUSE       = 0286,
         NO_HOUSES_FOR_SALE      = 0287,
-        ALREADY_DEFAULTED       = 0288,    // not used in game?
+        ALREADY_DEFAULTED       = 0288,
         ALREADY_HAVE_LOAN       = 0289,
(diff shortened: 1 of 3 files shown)


Allofich:
I didn't describe this well, it reads in the value from classic and shows the corresponding message, but doesn't actually set the flag itself. I assume this is should be set when the loan payback due date has passed without the loan amount being paid, but I want to test classic first.

Allofich:
Classic doesn't ever seem to ever set the flag, although the message will show if you hack your saved game. Seems like an oversight, or maybe they thought being barred from ever loaning again after a default was too severe? There is also bank assassin spawning code, but I haven't been able to get this to trigger in classic so far, in which case the loan due date would be pointless. I'll keep looking at that.

Allofich:
From the bug report:

> Notably right now this includes credit notes and loans in classic save data.

These are both imported already, unrelated to this PR.

Interkarma:
> These are both imported already, unrelated to this PR.

Thanks for letting me know! I will mark that old bug report as resolved then. :)

PR #682, opened 2018-04-07
Progress on conditions

Allofich:
Changes to faction power are disabled until power can be raised in some way.


Allofich:
Removed the debug message for failure to find a match in `FindFactionByTypeAndRegion()` because it would spam the log whenever the conditions update was called and witches factions were searched for in all regions.

PR #684, opened 2018-04-08
Implement normalization of reputations

Allofich:
Implements gradual return of reputations to 0, matched to classic.

Also some cleanup, comments, etc.

I removed the unused stuff for legal reputation in PersistentFactionData.cs, trying to keep the mess of all this WIP stuff under control a bit, since legal reputation all being done in the playerEntity's regionData array at the moment. If you want to refactor this stuff later, Interkarma, to make it all more tidy or fit some code organization you have in mind, that's OK with me.

I'm looking into guard spawning, btw, which ties into legal reputation work.


Interkarma:
No problems, happy for that to be removed and done differently. 👍

PR #685, opened 2018-04-08
Implement blood splash for townspeople and basic guard spawning

Allofich:
This adds blood splashes for killing townspeople and basic guard spawning (only for killing townspeople for now).

I think that guard spawning behavior in classic involves LOS checking from townspeople, and nearby "townsperson guards" and regular townspeople (1/3 chance) are the first choices for spawning guards. The fallback behavior after that is what this PR uses. I may need some help from you Interkarma to do that. Basically I would want to be able to iterate through the townspeople, doing an LOS check from each of them to the player, and also to be able to use the location of a townsperson as a spawn point (not sure if we actually want to do that for regular townspeople, it may have been intended to simulate them "clearing the streets" but it may be more immersive to not do it that way, but for guards as townspeople at least we'll want the ability to spawn from their position).


Interkarma:
Looking forward to playing with this! I'm wrapping up for the evening, but will get this in tomorrow sometime. :)

PR #688, opened 2018-04-14
Modsettings: Improvements

TheLacus:
Commit descriptions explain in details, this is a short list of changes:
* Local values and local presets are serialized as json files like mod presets, and all implements the same interface. They are created and applied to an instance of ModSettingsData.
* Sections can have an optional description. Section description is shown below their name while individual keys description are still shown as tooltips.
* Sections can have a "IsAdvanced" flag for options which should not be changed by most users but for which is still worth not being hardcoded.
* Not strictly related to mod settings, but i've hadded a ui class which easily draws multiple textboxes in a rect. I'm using it for tuples, as a corrispective of EditorGUI.MultiFloatField.


Interkarma:
Awesome! Thanks again for continuing work on this.

PR #689, opened 2018-04-14
More classic data identified + Save Explorer improvements

Allofich:
Recently I found that FALL.EXE contained a collection of strings that matched to the object types, although it's incomplete. This updates the `RecordTypes` in `SaveTree.cs` (I believe these are referred to as "Objects" in classic), noting the names from classic and also information on them I have reversed.

I also improved Save Explorer a little.
- Do not start with the faction drop-down open. This was an annoyance as I had to close it every time.
- Identify all "Container" types, not just the wagon.

And of course the work on identifying record types carries over to Save Explorer.

### DIFF
--- a/Assets/Scripts/Game/Entities/PlayerEntity.cs
+++ b/Assets/Scripts/Game/Entities/PlayerEntity.cs
@@ -575,7 +575,7 @@ namespace DaggerfallWorkshop.Game.Entity
                         newItem.TrappedSoulType = MobileTypes.None;
                 }
                 // Add to local inventory or wagon
-                if (containerRecord.IsWagon)
+                if (containerRecord.RecordRoot.SpriteIndex == 4) // Wagon containers have sprite index 4
                     wagonItems.AddItem(newItem);
                 else
                     items.AddItem(newItem);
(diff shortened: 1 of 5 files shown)


Assets/Scripts/Game/Entities/PlayerEntity.cs
```diff
@@ -575,7 +575,7 @@ public void AssignItems(SaveTree saveTree)
                         newItem.TrappedSoulType = MobileTypes.None;
                 }
                 // Add to local inventory or wagon
-                if (containerRecord.IsWagon)
+                if (containerRecord.RecordRoot.SpriteIndex == 4) // Wagon containers have sprite index 4
```
Interkarma: The property IsWagon also performs the (recordRoot.SpriteIndex == 4) check. Is there a reason you removed that property?

I would rather this being performed in the ContainerRecordData and exposed as IsWagon, rather than having the PlayerEntity check for sprite index directly.

Interkarma:
Great work mate! It's amazing how you keep working this stuff out.

> Do not start with the faction drop-down open. This was an annoyance as I had to close it every time.

My apologies for that. I think I made that happen when working on faction reading and forgot to set it back.

Assets/Scripts/Game/Entities/PlayerEntity.cs
Allofich: No big reason. First I took out the wagon specific checks in favor of identifying all container types, and then when I saw there was that one spot using the wagon check I figured since it was just one spot it would be OK to just have it directly do the check itself, as I'm not sure that we'll ever need to do the check elsewhere. But I'll go ahead and stick it in again.

PR #693, opened 2018-04-17
Minor fixups for classic record reading

Allofich:
Added back wagon check to ContainerRecord.

Revised some comments.


Interkarma:
Thanks for putting wagon check back. You're right it probably wasn't needed as only used in one place after you changed the container checks. But still nice to have that abstraction. Will get all these PRs merged within next 24 hours.

Allofich:
I had accidentally modified some spacing in one of the files but it's fixed now.

PR #694, opened 2018-04-17
More work on guards and crime

Allofich:
- Added crime enum and variable to track player's crime to playerEntity, and import this value from classic saves.
- Set the crime variable for the easy cases (murder, pickpocketing, vagrancy). The others will take a little more work to get right.
- Guards spawn further away (max distance should be equal to classic). Enemy sight increased (also FOV, which hadn't been properly updated before. I think I only changed the prefab for the Standalone Enemy and not the Serializable one). I believe the FOV matches classic and the sight matches the maximum distance an enemy can exist in classic (in outdoors environments) before being despawned, which may or may not be the same as sight range. Regardless, pursuing by guards or other monsters works much better.

A couple notes:
- Using the foe spawner, when multiple enemies spawn at once they always seem to be at the same distance. Is this intended? This isn't necessarily a bad thing but I think when multiple guards spawn in classic they each get their own random distance. I tried creating multiple 1 enemy FoeSpawners in a loop but the result was the same.
- Classic only stores a single "type of crime committed" variable for the player. We should probably sum together all crimes instead, but that can come later. We'd have to figure out how to adjust the court screen to handle multiple crimes, though.

### DIFF
--- a/Assets/Scripts/Game/EnemySenses.cs
+++ b/Assets/Scripts/Game/EnemySenses.cs
@@ -25,7 +25,7 @@ namespace DaggerfallWorkshop.Game
     {
         public static readonly Vector3 ResetPlayerPos = new Vector3(float.MaxValue, float.MaxValue, float.MaxValue);
 
-        public float SightRadius = 25f;         // Range of enemy sight
+        public float SightRadius = 4096 * MeshReader.GlobalScale;         // Range of enemy sight
         public float HearingRadius = 25f;       // Range of enemy hearing
         public float FieldOfView = 180f;        // Enemy field of view
 
--- a/Assets/Scripts/Game/Entities/PlayerEntity.cs
+++ b/Assets/Scripts/Game/Entities/PlayerEntity.cs
@@ -1271,6 +1271,21 @@ namespace DaggerfallWorkshop.Game.Entity
             Treason = 14,
         }
 
+        // Values after index 0 are from FALL.EXE. It does not seem to have a valid value for the last crime "Treason," so just using half of "High Treason" value here.
+        short[] reputationLossPerCrime = { 0x00, 0x0A, 0x05, 0x0A, 0x08, 0x14, 0x0A, 0x02, 0x01, 0x02, 0x02, 0x4B, 0x02, 0x08, 0x24 };
+
+        public void LowerRepForCrime()
+        {
+            int regionIndex = GameManager.Instance.PlayerGPS.CurrentRegionIndex;
+
+            // Lower legal reputation
+            regionData[regionIndex].LegalRep -= reputationLossPerCrime[(int)crimeCommitted];
+
+            // TODO: Faction reputation adjustments
+            //FactionFile.FactionData peopleFaction;
+            //FactionData.FindFactionByTypeAndRegion(15, regionIndex + 1, out peopleFaction);
+        }
+
         #endregion
 
         #region Event Handlers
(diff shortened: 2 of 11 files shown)


Allofich:
> Using the foe spawner, when multiple enemies spawn at once they always seem to be at the same distance.

Hm, now when I tried it they spawned at different distances, so maybe it's fine.

Allofich:
There seems to be a problem with the FoeSpawners created in playerEntity.SpawnCityGuards() persisting between loading a game. Try killing a townsperson then loading a saved game, and guards will appear in the saved game, even if the location is different. It's not caused by this PR, it's in master branch as well. Any ideas?

Allofich:
Just to be clear, I'm fine with this to be merged. The things I mentioned above are just a few issues I noticed with the FoeSpawner that can perhaps get some look later.

Interkarma:
Thanks for the catch. I'll see what I can do about those two issues today.

And good job on this. Looking forward to reviewing.

Allofich:
Some fixes and additions:

- Added surrender dialogue when hit by guard (same trigger as classic). Surrendering isn't functional yet.
- Surrender dialogue boolean reset when no guards nearby (same as classic. Actually for classic the check is for any enemies, not guards, so this is a minor improvement)
- Fixed maximum guards check. I forgot the player was also a `DaggerfallEntityBehaviour` so the previous check also counted the player. Also the check is for guards only now, which is different from classic (which just checks for number of any enemies) but the sensible/probably intended behavior.

The crime mechanics are one thing where definitely lots of changes could be made for more realistic behavior. I'm not going to get bogged down in that, though, and looking to recreate classic behavior, just with minor refinements like above (I really doubt, for example, that non-guards are supposed to count against the guard spawn limit. They probably just didn't have time to code for "guards + other enemies" situations.).

Assets/Scripts/Game/EnemySenses.cs
```diff
@@ -25,7 +25,7 @@ public class EnemySenses : MonoBehaviour
     {
         public static readonly Vector3 ResetPlayerPos = new Vector3(float.MaxValue, float.MaxValue, float.MaxValue);
 
-        public float SightRadius = 25f;         // Range of enemy sight
+        public float SightRadius = 4096 * MeshReader.GlobalScale;         // Range of enemy sight
```
Interkarma: Wow! That's the radius of a whole city RMB block. Had no idea it was that large. Good job finding that. :)

Assets/Scripts/Game/Entities/PlayerEntity.cs
```diff
@@ -432,8 +444,8 @@ public bool IntermittentEnemySpawn(uint Minutes)
 
         public void SpawnCityGuards(bool forceSpawn)
         {
-            // Don't spawn if more than 10 enemies are already in the area
-            if (UnityEngine.Object.FindObjectsOfType<DaggerfallEntityBehaviour>().Length > 10)
+            // Don't spawn if more than 10 guards are already in the area
+            if (GameManager.Instance.HowManyEnemiesOfType(MobileTypes.Knight_CityWatch) > 10)
```
Interkarma: This is a good fix. I've been looking into adding a serializable entity type to mobile NPCs as well so they can be saved/loaded and affected by spells.

Interkarma:
This is really amazing, thank you for the work! At this rate, I can really see the bones of crime & punishment coming together.

PR #697, opened 2018-04-21
Crimes modify faction reputations

Allofich:
Crimes now modify player's reputation with "people" factions in the same manner as classic.

Also identified the parts of the faction records in `SAVEVARS.DAT` that were unidentified. If we wanted to it's probably possible to create the parent-child relationships of factions just by reading in and using the 3 pointers at the end of each faction data record from `SAVEVARS.DAT`, rather than creating them by reading `FACTION.TXT`, but of course what we have already works.

### DIFF
--- a/Assets/Scripts/API/FactionFile.cs
+++ b/Assets/Scripts/API/FactionFile.cs
@@ -175,6 +175,10 @@ namespace DaggerfallConnect.Arena2
             public int vam;
             public int rank;
             public int randomValue;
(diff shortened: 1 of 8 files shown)


Allofich:
Added work on surrendering to guards.
- Whether or not you can surrender or you are killed, depending on reputation (matched to classic)
- Open court screen (non-functional)

Interkarma:
This is getting exciting, keep it up! :)

PR #698, opened 2018-04-22
Croucherfullbackup

MeteoricDragon:
Hey Interkarma.  Here is smooth croucher code, I have tested the code and haven't seen the player get stuck in any doorways, and it works well with HeadBobber and CameraRecoil.

This code makes the player smoothly (and quickly) crouches or stands up.  There is also a function that uses a Spherecast to check if the player has room to stand up.  If the player tries standing up when there's no room, the code waits until the player has moved out of the area and then stands him up.

When crouching, the code first lowers the camera, then crouches the playerController.  When standing, the opposite order happens.

Most of the code is in croucher.cs.   PlayerMotor is mostly untouched, I kept in mind what you said about PlayerMotor being due for a rewrite.  I felt that maybe if I reallocated crouching code from playerMotor into a separate class (croucher.cs) and include a variable for that class in PlayerMotor to access its functions, it would make the rewrite slightly easier as well as keep PlayerMotor less touched.

### DIFF
--- a/Assets/Croucher.cs
+++ b/Assets/Croucher.cs
@@ -34,6 +34,7 @@ namespace DaggerfallWorkShop.Game
         private float camCrouchLevel;
         private float camStandLevel;
         private float crouchTimer;
+        private bool bStandController;
 
         private const float timerMax = 0.1f;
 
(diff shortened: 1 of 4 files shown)


Interkarma:
Excellent! Looks good thank you for the work. Will do another PR merge shortly. :)

PR #699, opened 2018-04-23
Make it easier to max out volume

MeteoricDragon:
Makes it easier to max out the volume bar on sound and music

I want to add more fixes to this branch before it gets merged to master, but I wanted to create it to discuss some potential fixes to making all the sounds scale with the audio volume setting.

PlayerFootsteps uses an AudioSource called customAudioSource and a DaggerfallAudioSource called dfAudioSource.  The former is used for default footsteps and the latter is used for other footsteps. 

TransportManager does something similar for keeping the horse whinny sound separate from riding sounds.

In addition to horse and footstep sounds, it seems like dungeon ambient sounds and enemy sounds aren't being affected.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallPauseOptionsWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallPauseOptionsWindow.cs
@@ -112,6 +112,9 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
         #region Event Handlers
         private void SoundBar_OnMouseClick(BaseScreenComponent sender, Vector2 position)
         {
+            // make it easier to max out volume
+            if ((position.x / barMaxLength) > 0.99f)
+                position.x = barMaxLength;
             // resize panel to where user clicked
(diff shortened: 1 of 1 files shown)


Interkarma:
Thanks! I'm happy to help with the other sound adjustments if you like. Just let me know how I can help. :)

MeteoricDragon:
If you know where the sounds are that need adjustment, feel free to go ahead and change them.  
I tried putting these two lines in the enemy fixed update method but didn't' seem to help with enemy sounds:

```
If (GameManager.IsPaused)
    dfAudioSource.AudioSource.Volume = DaggerfallUnity.Settings.SoundVolume;
```

Interkarma:
The spell sounds call PlayOneShot(), which does the scaling automatically. The volume scale for spell effects seemed to be working OK last night when I generated builds? Is this not working on your end?

Having read this though, I just spotted a bug where volume scale was applied twice when calling one of the overloaded methods (once by overload and once by base method). I didn't spot this when reviewing code, and have fixed now.

MeteoricDragon:
I haven't tested it yet, I was just remembering it.

PR #701, opened 2018-04-23
Asset injection: Import terrain textures from mods

TheLacus:
* Terrain textures can be shipped in a single dfmod file.
* Removed some old methods for loose files import which are not needed anymore.

I'm using Shader.PropertyToID() because methods for access to shader properties have an int overload, which is faster than the string one (i believe the string method just call PropertyToID everytime).  I wrote a static class to holds all these values like the post processing stack does. What do you think of moving it to MaterialReader and use it project wide?


Interkarma:
I think that sounds like a great idea mate.

TheLacus:
Something like [this](https://github.com/TheLacus/daggerfall-unity/commit/e67e814607de43604e06d852a87cdfaaa94aeb69)? Tell me if you want it organized differently.

Interkarma:
That looks great! :)

PR #702, opened 2018-04-24
Framework for background questions

MeteoricDragon:
Lays some framework only.  Creates a class for background questions that are supposed to be able to keep track of which classes it is applicable to in the 12-question questionnaire for the character creation.  Has been tested somewhat

**Also not sure if this is the appropriate folder for the files**

### DIFF
--- a/Assets/BackgroundQuestion.cs
+++ b/Assets/BackgroundQuestion.cs
@@ -40,16 +40,39 @@ namespace DaggerfallWorkshop.Game
         private int questionID;
         private string question;
 
-        private Dictionary<char, string> answerDictionary;
+        private Dictionary<char, string> answerTexts;
+        private Dictionary<char, Action> answerResults;
         private char[] answerKeys = { 'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h' };
+        private char selectedAnswerIndex;
 
+        #region Properties
         public int QuestionID  {  get { return questionID; } }
         public CareerCheckList Careers { get { return careers; } }
         public string Question   {  get { return question; } }
         /// <summary>
         /// Char refers to the character indexing the answer, string is the text answer for that character to display
         /// </summary>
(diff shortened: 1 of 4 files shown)


deepfighter:
As I have not much to contribute on the coding side of this project I just want to get the attention on the corresponding forum thread and my post where I gathered some information regarding the BIOG files which may be helpful for the coding: [](http://forums.dfworkshop.net/viewtopic.php?p=11616#p11616)

MeteoricDragon:
I've discovered that because the biography files in classic are divided by file for each class, the enum property I have in background question is kind of superfluous.   I'll comment it out for now.

MeteoricDragon:
I believe that this is about as far as I can reasonably go.  I'll leave it up to @Interkarma whether this will be useful or not as a prefab.

Interkarma:
Hey mate, I'm sorry for not responding yet. I was waiting to see where you went with this and trying to figure out internally whether I can fit this into my future plans or not.

At this time, I'm planning to use the BIOG question files directly rather than convert to another format. This should allow them to be easily translated as now by anyone with a text editor. It would also allow already translated BIOG files to be used.

For this sort of work when tackling a new system, I usually follow a "tools first" approach. I would here first write a parser inside the DaggerfallConnect namespace to handle those BIOG files, then start implementing classic UI process, and then let that inform the appropriate data structures for questions and answers while building out UI and linking back to game systems.

I certainly don't want to throw out any of your work out-of-hand, but not sure yet where I'd fit this into the pipeline without any of the prerequisite framework around the BIOG files. I'm happy to leave this open if it's a subject you'd like to continue working on? Otherwise, I can just close and refer back when time comes for these class questions to implemented to see if anything can be used.

MeteoricDragon:
Yes, I understand.  I saw your BIOG parser just before your comment here and that's part of what led me to think you've already got this part of DFU covered.  If there's a place you think the files can connect in the future. let me know and I'll see if I can figure out a useful transition.

PR #703, opened 2018-04-25
Progress court screen

Allofich:
Note that there seems to be an issue with the MIDI playback in DF Unity and the song used for the court screen (Song_11) where the instrument at the beginning of the song doesn't play. At first I thought the song wasn't playing at all but you wait a few seconds you'll hear music.


Interkarma:
Excitement! :) Yeah, some of those song didn't convert from HMI very well. If someone has a better version of that song, we can replace it.

PR #706, opened 2018-04-27
Minor adjustments to enemies

Allofich:
- Change silver damage bonus back to Skeletal Warrior.

Recently someone working on OpenTESArena found that original Arena gives silver weapon damage bonuses against undead, so I no longer think this was a mistake. However, it might be a mistake or oversight that it only applies to the Skeletal Warrior and not other undead types. Changing that might affect gameplay too much, though, if it made undead enemies too easy, so I just put it back on the Skeletal Warrior like in classic.

- Make enemies stop when turning while pursuing again (not needed because of the faster turning now). This will remove the odd-looking behavior of dashing off to the side when you approach enemies from behind, etc.

About the turning, though, I think it may be too fast at the moment. Specifically, if an enemy goes from facing sideways to looking towards you the turn looks instantaneous. I think it should be slow enough that you see all the in-between animation frames.

### DIFF
--- a/Assets/Scripts/Game/EnemyMotor.cs
+++ b/Assets/Scripts/Game/EnemyMotor.cs
@@ -391,7 +391,7 @@ namespace DaggerfallWorkshop.Game
 
         private void TurnToTarget(Vector3 targetDirection)
         {
-            const float turnSpeed = 30f;
+            const float turnSpeed = 20f;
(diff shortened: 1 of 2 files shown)


Allofich:
Just did a quick try with the turn rate down to 20 from 30 and it seems good. I can see the turning frames and it doesn't seem too easy to get behind them.

PR #707, opened 2018-04-27
Refinements to court screen/arresting

Allofich:
Moved the message down and removed the background dimming. The court screen is already a dark screen and you can't see it clearly if it is dimmed.

Of course as I'm sure you know Interkarma there is no dimming for any pop-up messages in classic. I've been generally leaving it in place in DF Unity because I assume you put it in because you like it, but I don't know if you have anything decided about it. For this case at least I think it looks better without.

### DIFF
--- a/Assets/Scripts/Game/EnemyAttack.cs
+++ b/Assets/Scripts/Game/EnemyAttack.cs
@@ -35,6 +35,7 @@ namespace DaggerfallWorkshop.Game
         float meleeTimer = 0;
         bool isMeleeAttackingPreHitFrame;
         bool isShootingPreHitFrame;
+        int damage = 0;
 
         void Start()
         {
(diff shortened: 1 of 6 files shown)


Allofich:
Updated. It's now possible to plead Guilty and the fine will be applied and the prison sentence time passed, and reputations will be adjusted. No prison screen or attempting Not Guilty yet. I corrected what seemed to be a couple oversights in the way classic handles reputation modifications.

Unfortunately, there is a fairly long delay when leaving the court screen and the player is being relocated to the area entrance. I just mimicked other code in the project for this part. Is there a better way to do it?

Interkarma:
> Unfortunately, there is a fairly long delay when leaving the court screen and the player is being relocated to the area entrance. I just mimicked other code in the project for this part. Is there a better way to do it?

Code looks right on a read. Delay shouldn't be any more than say the time taken to fast travel somewhere. Will merge and take a closer look when I can.

PR #708, opened 2018-04-29
Fix loading saves outside losing guild membership data

ajrb:
Pretty stupid mistake on my part. :(


Interkarma:
All good, thanks for fixing mate! :)

PR #710, opened 2018-04-29
Headnod

MeteoricDragon:
This upgrade to HeadBobber.cs adds a head nodding mechanic to the headbobbing movement.  A very subtle nod.  

@Interkarma  I was wondering if it would be acceptable to implement a three-stage toggle into the Daggerfall pause menu for those who prefer the original version.

First stage: box unchecked, Head bobbing turned off.
Second Stage: box lower half checked.  Head bobbing movement as it exists in latest build is turned on.
Third Stage: box completely checked.  Head bobbing movement includes the new nodding movement for increased detail.

Cycling these stages: Click the box to cycle between them much like the box is toggled between the two stages right now :)

the head bobbing option in the pre-launch menu would need to be changed to a slider.

### DIFF
--- a/Assets/Scripts/Game/Player/HeadBobber.cs
+++ b/Assets/Scripts/Game/Player/HeadBobber.cs
@@ -14,6 +14,7 @@ namespace DaggerfallWorkshop.Game
     }
 
     [RequireComponent(typeof(CharacterController))]
+    [RequireComponent(typeof(PlayerFootsteps))]
     public class HeadBobber : MonoBehaviour
     {
         private BobbingStyle bobStyle = BobbingStyle.Walking;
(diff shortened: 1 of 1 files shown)


Interkarma:
Sure thing, shouldn't be too hard to knock together a 3 state button, or just use a slider like some of the other settings. Happy with whatever you go with here.

MeteoricDragon:
Do you think that the meaning of the different States of the button would be unclear based on what I described? Without some sort of tool tip on the pause window

MeteoricDragon:
That was actually my second idea.

By the way, here is the code in action:

https://www.youtube.com/watch?v=BWsJSWkeAOA&feature=youtu.be

Allofich:
> user can simply toggle headbob on/off in pause menu like now and set the *headbob *style from advanced settings at startup

I would prefer this. The head movement is quite good but it's possible I may try to recreate the original game's head movement (for both head bobbing and swaying on being hit), so I'd prefer that modified behaviors not be too tightly integrated into the base game. For example if the box became a 3-check version, I might have to figure out how to work around that if I wanted to try to recreate original movement. Or, if someone else wanted to make their own heab-bobbing behavior, they may have the same issue. I think being able to pick what type of "head bobbing" you want, and just turn that on and off, is the best.

PR #711, opened 2018-04-30
Progress court screen

Allofich:
Player can now choose "Not Guilty" and try to debate or lie.
Time is now passed for prison sentence (still no prison screen).
Flags set for execution/banishment. (No effect yet. They are supposed to cause guards to spawn when player enters a settlement in that region from then on)
Messages positioned correctly.

Unfortunately, the pause at the end when returning the player from the court screen seems to be even longer now (perhaps because of raising time for the prison sentence). People will probably think the game has frozen so this needs to be fixed. Any ideas?

### DIFF
--- a/Assets/Scripts/Game/EnemyAttack.cs
+++ b/Assets/Scripts/Game/EnemyAttack.cs
@@ -189,7 +189,7 @@ namespace DaggerfallWorkshop.Game
                 if (entity.MobileEnemy.ID == (int)MobileTypes.Knight_CityWatch)
                 {
                     // If hit by a guard, lower reputation and show the surrender dialogue
-                    if (!playerEntity.HaveShownSurrenderToGuardsDialogue)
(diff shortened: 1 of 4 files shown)


Interkarma:
I tried a guilty plea today after merging and didn't notice any unusual delay (took less than 2 seconds to respawn outside of town). Are you running with any mods at all?

Allofich:
Turns out I just thought it was being delayed greatly because I hadn't set the message boxes to be closed by clicking the mouse button.

I added that property and also put in manual calls to Update() to eliminate a minor delay between message boxes that could happen. (I don't really understand the Update() frequency of Unity objects but it seems this DaggerfallCourtWindow doesn't update so frequently by itself?) Also prevented enemy encounters from spawning for the time spend in prison or being released from prison.

Interkarma:
Not sure about spawns lingering. Could be FoeSpawner, yeah. I'll have a solid crime-spree later and see what I find. :)

PR #712, opened 2018-04-30
Talkwindow implementation: questor is added to talk options + fix

Nystul-the-Magician:
questor of current quest is added to list of persons in "where is"->"person" section of talk window
answers not implemented right now
fix: quest related talk topics now get correctly removed if quest is neglected


Interkarma:
Nice one! Thank you for the work. :)

Nystul-the-Magician:
@TheExceptionist: you can give it a try, so i guess you are familiar with the codebase now ;)
I need to add other quest npc as talk entry as well - you might want to wait for that
another task that is open where a helping hand is even more needed is the generic rumors stuff - some reverse engineering might be required here though. We would have to find out how vanilla df generates rumors and when...

TheExceptionist:
Alright, I'll try to figure it out!

PR #713, opened 2018-04-30
Fix enemies shooting bow too much

Allofich:
My earlier fix in https://github.com/Interkarma/daggerfall-unity/pull/657 included removing a limiter on the AI decision code that timed it to classic's slower update rate, but doing so made bow attacks way too rapid. This fixes it by putting it back for the ranged attack (while also avoiding the earlier problem that was the reason I removed it).


Interkarma:
Awesome! :) Thanks for fix. I was absolutely *annihilated* by knights in the Daggerfall treasury while testing something unrelated the other day haha.

PR #715, opened 2018-04-30
Fixed a talk window bug

TheExceptionist:
I noticed a dialogue bug when switching from the "Where is -> Person" to "Where is -> Location". The root cause the variable "inListboxTopicContentUpdate" remaining true after the player asked "Where is -> Person". I rearranged when the variable is set to true to solve this issue.

Relatively small fix, but I'm still getting used to the codebase :)

### DIFF
--- a/Assets/Scripts/Game/TalkManager.cs
+++ b/Assets/Scripts/Game/TalkManager.cs
@@ -174,6 +174,8 @@ namespace DaggerfallWorkshop.Game
         int currentKeySubjectBuildingKey = -1;
         int reactionToPlayer = 0;
 
+
+        //The lists that contain the topics to select
         List<ListItem> listTopicTellMeAbout;
         List<ListItem> listTopicLocation;
         List<ListItem> listTopicPerson;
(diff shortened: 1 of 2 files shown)


Interkarma:
Thanks for the fix! As change is relative minor I'll just merge straight to master. I'm not as familiar with the talk interface as some others, so will rely on @Nystul-the-Magician to let me know if any subtle problems with this change I'm not aware of.

Nystul-the-Magician:
I just trust you here. Funny side-node: could not remember the use of inListboxTopicContentUpdate at all - will try to make my brain remember tomorrow ;)

Nystul-the-Magician:
i took another look - to me this additional line might be dangerous since it deactivates a mechanism i implemented to resolve a gui click related infinite loop - see the long comment in function SelectTopicFromTopicList() before first use of inListboxTopicContentUpdate

setting this variable will likely break the mechanism i fear.

I am still not getting what the bug you mentioned is - if i get rid of the statement nothing changes when switching between "Where is -> Person" to "Where is -> Location" does work as intended.

Could you explain more what the bug is?

Interkarma:
Hi @TheExceptionist, are you able to discuss this with @Nystul-the-Magician please. :)

Please let me know the outcome so we can decide on status of this change. Thank you all!

Nystul-the-Magician:
i will undo the changes since it might cause problems, currently rewriting some topic list handling anyway

PR #716, opened 2018-05-04
Add TG and DB bribing judge, and fixes

Allofich:
- Adds Thieves Guild and Dark Brotherhood bribing judge.
- Fixes enemy-clearing method I added last time (it should have been destroying the game objects)
- Addressed a couple oversights in classic. The player should always be healed and get their reputation partly repaired if they are found not guilty now.

### DIFF
--- a/Assets/Scripts/Game/Entities/EnemyEntity.cs
+++ b/Assets/Scripts/Game/Entities/EnemyEntity.cs
@@ -107,6 +107,9 @@ namespace DaggerfallWorkshop.Game.Entity
                 return;
             }
 
+            if (careerIndex == (int)MobileTypes.Knight_CityWatch)
+                level += UnityEngine.Random.Range(3, 7);
+
             this.mobileEnemy = mobileEnemy;
             this.entityType = entityType;
             name = career.Name;
--- a/Assets/Scripts/Game/GameManager.cs
+++ b/Assets/Scripts/Game/GameManager.cs
@@ -602,13 +602,13 @@ namespace DaggerfallWorkshop.Game
             {
                 DaggerfallEntityBehaviour entityBehaviour = entityBehaviours[i];
                 if (entityBehaviour.EntityType == EntityTypes.EnemyMonster || entityBehaviour.EntityType == EntityTypes.EnemyClass)
-                    Destroy(entityBehaviour);
+                    Destroy(entityBehaviour.gameObject);
             }
 
             // Also check for enemy spawners that might emit an enemy
             FoeSpawner[] spawners = FindObjectsOfType<FoeSpawner>();
             for (int i = 0; i < spawners.Length; i++)
-                Destroy(spawners[i]);
+                Destroy(spawners[i].gameObject);
         }
 
         /// <summary>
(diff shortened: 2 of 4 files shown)


Allofich:
City guards are 3 to 6 levels above the player in classic. That's implemented now.

Interkarma:
Thanks for all your work on this, it's amazing to have this coming together so far ahead of schedule.

>The player should always be healed and get their reputation partly repaired if they are found not guilty now.

I was found not guilty the other day and died immediately after respawning outside of town. Sounds like this will be fixed now, good work.

Allofich:
> I was found not guilty the other day and died immediately after respawning outside of town. Sounds like this will be fixed now, good work.

Well you are supposed to be sent out with only 1 health for banishment (and probably put in a dungeon with 1 health for execution, which doesn't seem to have made it into the final game), but I wonder why you died. Fall damage? Lingering damage from a hit by a guard? Or maybe it was an arrow hit from a spawned random encounter (random encounters were spawning immediately after being released from court/prison but that should be fixed now). Let me know if it happens again or if you think you know why it happened.

Interkarma:
In think it was remaining damage from a guard. I died the moment i respawned.

Allofich:
OK, I may have to revisit how that works. I had one point where multiple guards were attacking me and although the first guard's hit was nullified in order to show the surrender dialogue, I think another guard's hit got in, so damage from a guard seems likely.

Assets/Scripts/Game/GameManager.cs
```diff
@@ -602,13 +602,13 @@ public void ClearEnemies()
             {
                 DaggerfallEntityBehaviour entityBehaviour = entityBehaviours[i];
                 if (entityBehaviour.EntityType == EntityTypes.EnemyMonster || entityBehaviour.EntityType == EntityTypes.EnemyClass)
-                    Destroy(entityBehaviour);
+                    Destroy(entityBehaviour.gameObject);
```
Interkarma: Nice catch. I make this mistake fair bit as well.

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallCourtWindow.cs
```diff
@@ -140,9 +142,51 @@ public override void Update()
                     fine = playerGold;
                 }
 
-                // TODO: Chance to free player if in Dark Brotherhood or Thieves Guild
+                DaggerfallMessageBox messageBox;
+                if ( crimeType == 4 || crimeType == 3) // Assault or murder
+                {
+                    // If player is a member of the Dark Brotherhood, they may be rescued for a violent crime
```
Interkarma: I'm not sure I knew this could happen. Very nice work rolling this in too.

Assets/Scripts/Game/Entities/EnemyEntity.cs
```diff
@@ -107,6 +107,9 @@ public void SetEnemyCareer(MobileEnemy mobileEnemy, EntityTypes entityType)
                 return;
             }
 
+            if (careerIndex == (int)MobileTypes.Knight_CityWatch)
+                level += UnityEngine.Random.Range(3, 7);
+
             this.mobileEnemy = mobileEnemy;
```
Allofich: Careless me, I just realized I applied the additional levels here _after_ health is rolled (although before skill levels are calculated). I'll double-check whether the guards in classic get a health boost from their additional levels.

PR #717, opened 2018-05-04
Preparation for assigning enemy spell lists

Allofich:
Enemy spell lists and a setup for assigning spells to enemies, same method as classic. Nothing is actually assigned yet.

The spell lists are the same as in the executable (although they are each terminated by `0xFF` there), just arrays of bytes that are the spell IDs.

### DIFF
--- a/Assets/Scripts/DaggerfallUnityStructs.cs
+++ b/Assets/Scripts/DaggerfallUnityStructs.cs
@@ -206,6 +206,7 @@ namespace DaggerfallWorkshop
         public string LootTableKey;                 // Key to use when generating loot
         public int HitFrame;                        // Frame of attack animation at which hit on target is attempted
         public int Weight;                          // Weight of this enemy. Affects chance of being knocked back by a hit.
+        public bool CastsMagic;                      // Whether this enemy casts magic. Only used for enemy classes.
     }
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


Allofich:
Added some more, for assigning enemy spell stats. Note that enemies don't follow the same rules as the player for number of spell points. Right now `MaxMagicka` is set to follow the player's rules. I think you have particular ways you want entity values to work, so I haven't touched that, but to get the values of classic we need to be able to set enemy entities to have an arbitrary value for maximum spell points.

Allofich:
Added a `CastsMagic` property to `mobileEnemy`, since it doesn't seem like `HasRangedAttack1` and `HasRangedAttack2` are enough to determine which enemies use magic. I haven't been able to find this data in classic so this is just based on whether the enemies have magic animation frames.

Interkarma:
Great work! 👍 

Just so you have an idea of my timeline - as soon as I've built out majority of spell effects (I'm hoping to have done around 90% before end of May) then I'll work on implementing standard spells list. I'm likely to export SPELLS.STD to its own JSON file or something though.

Are these monster spell IDs just referencing a spell in SPELLS.STD? Should be very easy to link up if that's the case.

Allofich:
> Are these monster spell IDs just referencing a spell in SPELLS.STD? Should be very easy to link up if that's the case.

Yes, each is a reference to the 79th byte of each 89-byte spell record. `spellRecord.index` in `DaggerfallSpellReader.cs`.

PR #718, opened 2018-05-04
Talkwindow - progress with "Where Is"->"Person" talk topics, "Where Is" -> "Location" named Residences

Nystul-the-Magician:
- fixed quest N0C00Y10 (mages guild non-member missing book quest (my test quest))
- topic list generation strategy/policy update (only generated when talkwindow is opened (exception: if dialog link reveals new topics then it is immediately updated)
- questor dictionary is now only generated once (since it takes long and caused a pause which was noticable), we need to decide if we want to generate this on first talk window open after a location change/savegame load/etc.  or if we want to create it immediately on location change/savegame load/etc.
- asking for quest people outside now works (compass direction or map reveal)
- asking for quest people inside will point one to the building and compass direction (this is wrong when player is already in the same building and I will work on a different approach for such case - see forum post and tell me your opinions: http://forums.dfworkshop.net/viewtopic.php?f=23&t=672&p=11888#p11887)
- savegame support - there was a problem with loading conversation data regarding quest resources, quest data is restored first now and then the conversation data is loaded and data is relinked to loaded quests
- questor is hidden from topic list if it is the dialog partner
- quest place resources which are regional houses will now show up in list "Where Is" -> "Location" if pc learned about them through dialog or quest
- quest place houses are now revealed on map correctly (revisited after Interkarma's feedback - much cleaner now)
- update to ExteriorAutomap regarding quest named residence resolving after Interkarma's feedback


Interkarma:
This is excellent progress Nystul!

> we need to decide if we want to generate this on first talk window open after a location change/savegame load/etc. or if we want to create it immediately on location change/savegame load/etc.

I'm a bit torn here. On one hand, loaded new locations is already a bit heavy and every ms saved is good. On the other hand, players are likely to be more forgiving of load times when fast travelling than delays when opening UIs, which should feel almost instant at all times.

I'd probably lean towards generate when changing locations. Would you be able to wrap the function in a couple of stopwatch timers and give me some idea of how many ms it takes to generate this list?

Nystul-the-Magician:
yeah will put a timer on it and tell you. It feeled like something about half a second to me.
would also tend to go for generating it on location change

Assets/Scripts/Game/DaggerfallExteriorAutomap.cs
```diff
@@ -762,9 +763,34 @@ private void createBuildingNameplates(DFLocation location)
...
-                                // only create nameplates for buildings different than "Residence"
-                                if (discoveredBuilding.displayName != "Residence")
+                                
+                                if (!RMBLayout.IsResidence(buildingSummary.BuildingType)) // guilds, shops, taverns, palace (general case)
+                                {
                                     newBuildingNameplate.name = discoveredBuilding.displayName;
+                                }
+                                else if (RMBLayout.IsResidence(buildingSummary.BuildingType)) // residence handling
+                                {
+                                    // check if residence is involved in active quest (this might be replaced by function call - for now I implemented this by myself
```
Interkarma: I'll roll this into a nice dictionary lookup in QuestMachine that gets refreshed when quests are compiled or removed. That way the automap population only need to a do a single lookup against the dict. This should work great for now though! :)

Assets/StreamingAssets/Quests/N0C00Y10.txt
```diff
@@ -108,12 +108,6 @@ Message:  1025
 <ce>                   who takes things without asking,
 <ce>                       if you know what I mean.
 
-RumorsPostfailure:  [1006]
```
Interkarma: Bonus fix! 👍

Assets/Scripts/Game/DaggerfallExteriorAutomap.cs
```diff
@@ -762,9 +763,34 @@ private void createBuildingNameplates(DFLocation location)
                             PlayerGPS.DiscoveredBuilding discoveredBuilding;
                             if (GameManager.Instance.PlayerGPS.GetDiscoveredBuilding(buildingSummary.buildingKey, out discoveredBuilding))
                             {
-                                // only create nameplates for buildings different than "Residence"
-                                if (discoveredBuilding.displayName != "Residence")
+                                
+                                if (!RMBLayout.IsResidence(buildingSummary.BuildingType)) // guilds, shops, taverns, palace (general case)
```
Interkarma: @Nystul-the-Magician Nystul, I updated post in forums. Please feel free to no longer filter residences from automap. I know that I've been adamant about keeping map free of non-functional clutter, but tying the automap in with all the other game systems that can modify nameplates (especially since mods can do this) is going to be much too hard bordering on impossible. The automap shouldn't have to worry about this.

I think the best solution is is just for me to admit that it's best to do things like classic here and just display nameplate as set by discovery system. If that nameplate is a non-useful residence then so be. This is how classic works anyway, and I doubt many people will complain. It will also allow users to potentially edit any nameplates on automap as well.

I can admit when my decisions are wrong and just make things more difficult for you guys. This is one of those cases. :)

Nystul-the-Magician: what do you think about this:
one way to go would be to filter indeed by name "Residence" but get this string from recently introduced ConversationText.txt file. Translators would then be cautious to translate this correctly but it would work.
where does the building name creation / discovery mechanism get the string "Residence" from? maybe best thing would to make a common mechanism that all systems use for deriving string "Residence" and automap could then filter it. Of course this mechanism must support translations

PR #719, opened 2018-05-05
Height changer

MeteoricDragon:
Changes:  

- Rename Croucher to PlayerHeightChanger

- Migrate code in PlayerMotor that adjusts the player height when mounting/dismounting horse to PlayerHeightChanger

- Fix bug where if mounting horse while crouched, would lock player into crouch speed, even on horse.

- Delete deprecated commented code

- Horse Mount/dismount now smoothly & quickly transitions like crouching.


Interkarma:
There's a commit in there from myself. Don't forget to sync your fork up before sending the PR over to me. There shouldn't be any of my "minor spell fixes" commits in there at all. :)

I like the idea of combining the various height modifiers to a single state class. A very sensible improvement!

MeteoricDragon:
I think I successfully fixed it. @Interkarma , Please look at it and let me know if there are any other problems with the commits.

Interkarma:
That looks much better, cheers mate! I'll test and merge today sometime.

PR #723, opened 2018-05-08
Climbing motor

MeteoricDragon:
Like the other two PRs, this one Isolates climbing motor commands.

### DIFF
--- a/Assets/Prefabs/Player/PlayerAdvanced.prefab
+++ b/Assets/Prefabs/Player/PlayerAdvanced.prefab
@@ -35,6 +35,7 @@ GameObject:
   - component: {fileID: 114189154217185308}
   - component: {fileID: 114126215008192106}
   - component: {fileID: 114232921892998986}
+  - component: {fileID: 114607307154016044}
   m_Layer: 0
   m_Name: PlayerAdvanced
   m_TagString: Player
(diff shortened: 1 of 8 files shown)


Interkarma:
Nice! You totally get one of the core concepts for new player motor, which is to split the various movement modes into their own handler class and route based on a state machine of some kind. Making a start on decoupling all the movement states is a pretty solid lead on the problem, thank you. :)

I'll merge this PR as soon as I get time to test properly.

MeteoricDragon:
Thanks, I completed migrating the sliding section of code in playerMotor to a FrictionMotor class, but I don't know where to test the sliding portion of the migration.  Where can the player test sliding behavior?  I launched the code and ran around and things seem ok.  but I don't know how the sliding stuff behaves.  If you know a place i can test it, i'll test it.

Interkarma:
The sliding stuff is from FPSWalkerAdvanced which PlayerMotor was originally based on before some heavy upgrades. Sliding has never really been in the game, and some of the slopes player is expected to walk up (like the steep ramp in Privateer's Hold) are extremely steep. So the slope limit is already so high it precludes most chances to slide.

I'm very happy to have this split out though, thank you that! :)

MeteoricDragon:
You're welcome.  So should I conclude that it won't be necessary to test the sliding code?

Interkarma:
Yeah, probably not necessary. We don't have a baseline to compare it to. As long as nothing is changed overall and player can still walk up and down steep ramps like the one in PH, I'm happy. :)

PR #724, opened 2018-05-08
Talkwindow - fixes and improvements

Nystul-the-Magician:
fixed bug with remote place resources (http://forums.dfworkshop.net/viewtopic.php?f=24&t=997, http://forums.dfworkshop.net/viewtopic.php?f=24&t=996)

Location->General subgroup is now hidden if no entries are available

topic list creation is no longer done while accepting quest

hard strings are now loaded from ConversationText.txt


Interkarma:
Thank you for the rapid fix @Nystul-the-Magician. You rock!

PR #725, opened 2018-05-09
Fix wrong faction IDs

Allofich:
Fixes http://forums.dfworkshop.net/viewtopic.php?f=24&t=1000. I got the faction IDs reversed.


Allofich:
Please also check https://github.com/Interkarma/daggerfall-unity/pull/726 before deciding whether to merge this.

PR #727, opened 2018-05-09
Spell costs code refinement

ifkopifko:
It will spit out correct costs once offsetGold values are corrected. That is planned for the next pull request.

### DIFF
--- a/Assets/Scripts/Game/Formulas/FormulaHelper.cs
+++ b/Assets/Scripts/Game/Formulas/FormulaHelper.cs
@@ -1187,16 +1187,12 @@ namespace DaggerfallWorkshop.Game.Formulas
             int skillValue = GameManager.Instance.PlayerEntity.Skills.GetLiveSkillValue((DFCareer.Skills)effectTemplate.Properties.MagicSkill);
 
             // Duration costs
-            int durationMinGoldCost = 0, durationMinSpellPointCost = 0;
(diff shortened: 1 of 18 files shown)


Interkarma:
Welcome, and thank you for the refinements! 👍 

I'll leave this PR open until you have parity with the current output. Just add your offsetGold value corrections to this PR and I'll be able to review the whole thing in one hit when ready.

Interkarma:
Thank you! I'll merge this now and let you know if any troubles. Thank for the help with this, not just here but with the formulas as well. :)

Interkarma:
Just a heads up, the new formula is no longer in parity with classic. It seems to have problems with Magnitude-only effects (e.g. Drain Endurance) and Duration+Chance effects (e.g. Paralyze). My old formula was also able to handle these cases with classic parity.

My testing process is to load up the same character with 50 in all magic skills in both classic and DFU, then spin up all kinds of wild combinations in spellmaker in both versions and to make sure they come out the same.

ifkopifko:
Will look into it and come up with a fix.

Interkarma:
> Will look into it and come up with a fix.

It's all good, working OK now. I edited my earlier comment. Not sure what I did wrong in my earlier tests. I'm pretty much running on fume tonight, so mistake is likely mine. I'll try again when less tired and let you know if any more problems. :)

PR #728, opened 2018-05-09
Modsettings window: Improvements

TheLacus:


TheLacus:
I don't remember if this was already discussed, and i can't find it in translation forum, but do you intend to use text tables for these non-classic windows? I can start moving some strings if you want.

Interkarma:
Yes please, I'm planning to use text tables to store all text hard coded game text the player needs to see. Error messages the player doesn't see (debug text, exceptions, etc.) can stay as is. It's not as sophisticated as a real localisation setup, but we're also fighting classic data on this front. I think as long as the translation folk can edit the text somewhere they can make do.

Just a heads up, I need to make a small upgrade to table class to support string literals better as commas will currently break schema. It might be necessary to rework these tables slightly in future if you do start using them.

PR #729, opened 2018-05-10
Use a few more faction enums

Allofich:
Very minor PR. Just fixes a spelling error and uses the faction enums in a few more spots.

I started to go through the code looking for anywhere that faction IDs were used as numbers, but I wasn't sure of an effective way to search for that, and I came across usage of faction IDs defined within guild files, such as the factionID defined within the `ThievesGuild` class. I just left that alone, but I'm using the enums for the code I write.

### DIFF
--- a/Assets/Scripts/API/FactionFile.cs
+++ b/Assets/Scripts/API/FactionFile.cs
@@ -196,7 +196,7 @@ namespace DaggerfallConnect.Arena2
 
             Random_Noble = 242,
 
-            Teahers_of_Zen = 243,
+            Teachers_of_Zen = 243,
 
             Court_of_Balfiera = 244,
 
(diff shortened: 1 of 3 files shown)


Interkarma:
That's fine. :) Perhaps this is something @ajrb will be interested in later when he has more time. I'll start using the enums moving forwards when I need to reference a faction.

ajrb:
Oh, when did these get added? For some reason I had assumed that you'd decided that the faction entries should be read from FACTION.TXT and not hardcoded. I suppose it's unlikely to change, but I was avoiding this to try to stay within the existing design.

If you confirm these are sticking around I will move all my code to use the enum names rather than the id numbers.

ajrb:
Great, regret not bringing it up before then. Good job Allofich.

Allofich:
For me the most useful reason for having these is that it is a numerically sorted list of factionIDs. So if you want to find out what a certain faction ID refers to, it's easy to scan this list for it. Not sure how much that particular way of using it will matter from here on out, though.

PR #730, opened 2018-05-10
Guard improvements

Allofich:
~1) Problem with player taking damage from guards if they surrendered while being hit nearly simultaneously by 2 or more guards should be fixed.~ (Removed, still seems to need work)
2) Implemented guards spawning for player with a low legal reputation or who has been banished. Behavior is the same as classic. (Actual spawning behavior of guards is still different, still need to implement LOS-check-based spawning)

### DIFF
--- a/Assets/Scripts/Game/Entities/PlayerEntity.cs
+++ b/Assets/Scripts/Game/Entities/PlayerEntity.cs
@@ -342,7 +342,7 @@ namespace DaggerfallWorkshop.Game.Entity
 
             // TODO: Right now enemy spawns are only prevented when time has been raised for
             // fast travel. They should later be prevented when time has been raised for
-            // 1) Turning into vampire 2) Serving prison sentence
+            // turning into vampire
             // Classic also prevents enemy spawns during loitering,
             // but this seems counterintuitive so it's not implemented in DF Unity for now
             if (!preventEnemySpawns)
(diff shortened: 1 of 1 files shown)


Allofich:
Oh man, don't try to loiter somewhere you've been banished. :)

image


I think this should be solved by resolving enemies not waking up the player, though, so I don't think it's a problem with this PR specifically.

Also the player won't die for some reason. This isn't specific to guards or this PR, when I get to 0 health the player doesn't die.

Interkarma:
> I think this should be solved by resolving enemies not waking up the player, though, so I don't think it's a problem with this PR specifically.

I really need to get on that.

> Also the player won't die for some reason. This isn't specific to guards or this PR, at 0 health the player doesn't die.

That's new. Do you mean player doesn't die under any circumstances?

Allofich:
> Do you mean player doesn't die under any circumstances?

 I wouldn't die no matter how long enemies attacked me. I assumed it was guards at first but the same thing happened with non-guards.

I see you updated the master branch. I'll try again with the very latest code.

Allofich:
It's possible I messed something up when I modified EnemyAttack recently working on the surrender implementation.

Allofich:
Hmm, I don't die even from fall damage.

And a new issue. No breath meter when underwater (I was going to try drowning).

Allofich:
If you don't have these maybe my setup is somehow messed up locally.

Interkarma:
Hmm sounds funky. I just did a couple of quick tests then and was able to die ok from guards and monsters. Didn't think to try drowning. If you can, clone a fresh copy of code from master and let me know if problem continues. I can do more testing in morning, just shutting down for night here.

Allofich:
Seems to be fine now, but I'm not sure why. Here's the order of what I tried.
1. Tried starting from a classic save rather than a Unity save, same problems. Tried Unity save one more time.
2. Tried re-importing all assets. Same problems.
3. Tried "Build And Run" command (rather than running from the editor), problems gone.
4. Now problems are gone when running from the editor.

Normally I skip the `DaggerfallUnityStartup` scene and start straight from `DaggerfallUnityGame`, but the built .exe ran from the startup scene. Maybe that had something to do with it.

Interkarma:
I'm happy you were able to get this sorted. Not sure what was going wrong there. But you're right, probably build or startup UI popped something back into alignment.

PR #731, opened 2018-05-10
Talkwindow - progress and fixes

Nystul-the-Magician:
- rumors get removed correctly on quest decline now (one part of the bug report in http://forums.dfworkshop.net/viewtopic.php?f=24&t=1002&p=12092#p12092)
exception handling for the reported exception as well - TODO: still have to find out how "tell me about" entries could stay on the list after quest has finished
- npc reject message is now a random msg from text.rsc like in classic when first approaching npc, second time is "you get no response"
- question's answers take into account reputation with the npc now (similar to greeting msg)
- implemented guild member greetings, added macro handlers for %fnpc, %fpc, %fpa and %hnr
- work/questor list is now saved with conversation data
- fixed merchant missing book quest


Interkarma:
Thank so you so much for all your hard work on this. :)

PR #733, opened 2018-05-11
%hnr "Ma'am" instead of "Madam", work/merchant quest offers accept closes npc submenu

Nystul-the-Magician:
info clicking won't reveal quest place resources that are available for chat but were not revealed by npc
player now must get them revealed to show on the map


Interkarma:
I really appreciate all your work on this Nystul. Thank you!

ajrb:
I have a minor issue with the change to close the popup window made to DaggerfallQuestOfferWindow class. I spent a while making sure it matched classic behaviour, why does it need to be closed to update dialog link?

Interkarma:
I understand it was to work around a code execution order issue. This solves the problem in the short term and then can potentially be addressed a better way. Here's the forum post. I'll let you and Nystul work it out from here. :)

http://forums.dfworkshop.net/viewtopic.php?f=23&t=672&start=150#p12110

Nystul-the-Magician:
ajrb, classic always closes this submenu - at least for me!

PR #734, opened 2018-05-11
thieves guild and dark brotherhood guild buildings are revealed again 

Nystul-the-Magician:
now filtering by factionId of building to show them when player is in guild and thus have them marked on the map
map marking also happens now directly when joining guild in same town - you don't have to reenter the city/town

only shows the tg, dbh guild nameplate when player has joined them
due to the new filtering this had to be handled specially

### DIFF
--- a/Assets/Scripts/Game/DaggerfallExteriorAutomap.cs
+++ b/Assets/Scripts/Game/DaggerfallExteriorAutomap.cs
@@ -27,6 +27,7 @@ using DaggerfallWorkshop.Game.Entity;
 using DaggerfallWorkshop.Game.Serialization;
 using DaggerfallWorkshop.Game.Utility;
 using DaggerfallWorkshop.Game.Questing;
+using DaggerfallWorkshop.Game.Guilds;
 using Wenzil.Console;
 
 namespace DaggerfallWorkshop.Game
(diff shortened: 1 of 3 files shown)


Nystul-the-Magician:
will take a look - thanks for letting me know

ajrb:
Not so sure about having the specific conditional in automap class though. Sure there's a good reason I have missed, but before it just needed the discovery of a building to be set then it would override the natural name 'residence' - this is also done for player housing. Why does that no longer work from the building name code that's shared, but specific conditionals in map?

Nystul-the-Magician:
the conditional was always there, before it was the check for string "Residence" - which is suboptimal because translations can easily break this mechanism - even if we translate the comparison string as well this is error-prone. So I went for method calls and enum checking instead of string comparison.
I get your point with the player housing. Maybe best thing would be to introduce a flag in discoveryData called nameIsOverridden or something - lets see if I can improve this

Nystul-the-Magician:
working on it - now the check would look like this:
if (!RMBLayout.IsResidence(buildingSummary.BuildingType) || discoveredBuilding.isOverrideName)
only thing to know is that player has to leave and enter town when loading an old savegames to make the new mechanism work with old savegames. no problem with new though

Nystul-the-Magician:
new pull request: https://github.com/Interkarma/daggerfall-unity/pull/741

PR #735, opened 2018-05-11
Document and read classic effect flags

Allofich:
Just information, no functionality. It might be useful since you're building out effects.

### DIFF
--- a/Assets/Scripts/API/Save/CharacterRecord.cs
+++ b/Assets/Scripts/API/Save/CharacterRecord.cs
@@ -149,8 +149,8 @@ namespace DaggerfallConnect.Save
             parsedData.magicEffects1 = reader.ReadByte();
             parsedData.magicEffects2 = reader.ReadByte();
             parsedData.magicEffects3 = reader.ReadByte();
+            parsedData.magicEffects4 = reader.ReadByte();
 
-            reader.BaseStream.Position = 0x8D;
             parsedData.currentSpellPoints = reader.ReadInt16();
             parsedData.maxSpellPoints = reader.ReadInt16();
 
@@ -315,9 +315,10 @@ namespace DaggerfallConnect.Save
             public Byte level;
             public PlayerReflexes reflexes;
             public UInt32 physicalGold;
-            public Byte magicEffects1; // x1 = paralyzed, x4 = invisible, x8 = levitating, x20 = lock, x40 = open
-            public Byte magicEffects2; // x1 = silenced, x10 = chameleon, x20 = shade, x40 = slowfall
-            public Byte magicEffects3; // x1 = jump, x8 = waterbreathing,, x10 = waterwalking
+            public Byte magicEffects1; // x1 = paralyzed, x4 = invisible, x8 = levitating, x10 = light, x20 = lock, x40 = open, x80 = regenerating
+            public Byte magicEffects2; // x1 = silenced, x2 = spell absorption, x4 = spell reflection, x8 = spell resistance, x10 = chameleon, x20 = shade, x40 = slowfall, x80 = climbing
+            public Byte magicEffects3; // x1 = jumping, x2 = free action, x4 = lycanthropy, x8 = water breathing, x10 = water walking, x20 = diminution (not implemented), x40 = shield, x80 = detect
+            public Byte magicEffects4; // x1 = darkness, x2 = tongues, x4 = intensify fire (not implemented), x8 = diminish fire (not implemented)
             public Int16 currentSpellPoints;
             public Int16 maxSpellPoints;
             public Int16 reputationCommoners;
--- a/Assets/Scripts/API/Save/CharacterRecord.cs
+++ b/Assets/Scripts/API/Save/CharacterRecord.cs
@@ -125,9 +125,8 @@ namespace DaggerfallConnect.Save
             parsedData.baseHealth = reader.ReadInt16();
 
             reader.BaseStream.Position = 0x60;
-            parsedData.timePlayerBecameWerebeast = reader.ReadUInt32();
-
-            reader.BaseStream.Position = 0x6c;
+            parsedData.lastTimeUrgeToHuntInnocentSatisfied = reader.ReadUInt32();
+            parsedData.timeAfterWhichShieldEffectWillEnd = reader.ReadUInt32();
             parsedData.unknownLycanthropy = reader.ReadInt16();
 
             reader.BaseStream.Position = 0x74;
@@ -180,6 +179,7 @@ namespace DaggerfallConnect.Save
             reader.BaseStream.Position = 0x211;
             parsedData.timeForThievesGuildLetter = reader.ReadUInt32();
             parsedData.timeForDarkBrotherhoodLetter = reader.ReadUInt32();
+            parsedData.shieldEffectAmount = reader.ReadUInt32();
 
             reader.BaseStream.Position = 0x21f;
             parsedData.darkBrotherhoodRequirementTally = reader.ReadByte();
@@ -305,7 +305,8 @@ namespace DaggerfallConnect.Save
             public UInt32 skillsRaisedThisLevel2; // Flags for skills 32 through 34.
             public Int32 startingLevelUpSkillSum; // The starting total of all the primary skills, the two top major skills and the top minor skill
             public Int16 baseHealth;
-            public UInt32 timePlayerBecameWerebeast; // Needs confirming
+            public UInt32 lastTimeUrgeToHuntInnocentSatisfied;
+            public UInt32 timeAfterWhichShieldEffectWillEnd;
             public Int16 unknownLycanthropy; // Lycanthropy stage? Set when inflicted with lycanthropy.
             public UInt32 playerHouse; // Building ID of player's house. 0 if player doesn't own a house.
             public UInt32 playerShip; // Probably same type of data as above, for player's ship. 0 if player doesn't own a ship.
@@ -351,6 +352,7 @@ namespace DaggerfallConnect.Save
             public UInt32 lastTimePlayerBoughtTraining;
             public UInt32 timeForThievesGuildLetter;
             public UInt32 timeForDarkBrotherhoodLetter;
+            public UInt32 shieldEffectAmount;
             public Byte vampireClan;
             public Byte effectStrength; // Used for Open and Shade effects at least.
             public Byte darkBrotherhoodRequirementTally;


Interkarma:
It's amazing how you just keep working this stuff out! Incredible. It's really interesting information to see how this is stored by classic save data, and it gives some nice insight into how the classic effects were built.

Allofich:
Interkarma, together with finding out how these flags work I've come across the functions for the various magic effects. Since you're actively working on it, it seems best if I supply what I figure out to you and you implement it into the way you are making the system. What's a good way to do it? For example, here is the pseudocode for the paralyze effect:
```
bool Effect_Paralysis(spellObject Spell, int AdditionalValue, entity Target)
{
       if (Target.Race == 4) // High Elf (Also probably makes grizzly bear (monsterID 4) immune)
              return false;
       if (Target == Player && Player.ImmunityFlags & 1) // Not sure why this is limited to the player
              return false;
       if (Target.MagicEffects1 & 1) // Paralysis.
              return false;
       if (Target.MagicEffects2 & 0x80) // Climbing.
              return false;
       
       int chance = AdditionalValue + Spell.Byte86
       if (DFRandom(1, 100) > chance)
             return false;

       Target.MagicEffects1 |= 1;
       if (Target == Player)
             TopOfScreenMessage("You are paralyzed.");
}
```

I'm not sure where the "additional value" comes from yet. Also I checked `DaggerfallSpellReader.cs` but it looks like the 86th byte is not read yet, it's just skipped over.

Allofich:
Effect_Damage

```
void Effect_Damage (spellObject Spell, int Value, entity Target, char *a4@<edi>, int a5@<esi>)
{
  int subType = spellObject[2 * Value + 1];
  if ( subType == 0 ) // Damage health
  {
         if ( Target != Player || !GodMode )
         {
                ApplyDamage(Target, spellObject[2 * Value + 80]);
         }
  }
  else if ( subType == 1 ) // Damage fatigue
  {
    short resultFatigue = Target.CurrentFatigue - spellObject[2 * Value + 80];
    if ( resultFatigue < 0 )
         resultFatigue = 0;
    Target.CurrentFatigue = resultFatigue;
  }
  else if ( subType == 2 ) // Damage spell points
  {
    Target.CurrentSpellPoints -= spellObject[2 * Value + 80];
    if (Target.CurrentSpellPoints < 0 )
      Target.CurrentSpellPoints = 0;
  }
}
```
"Value" as I called it above is probably the number of the spell's effect (first effect, second effect or third effect).

BTW if you don't know already this is a useful resource: http://en.uesp.net/wiki/Daggerfall:SPELLS.STD_indices.

Allofich:
> The way my effect system works is different enough that we may not have much use for these. 

For effect_damage, I can see that, but for paralysis, etc. we want things like the high elf immunity, right? I think it's good to reference to this to make sure the effects are covering all the needed behavior.

Allofich:
> I'm across the SPELLS.STD info, I even link to it from my effect system status.

Ah, sorry. I haven't been following the spell progress really because it took all my time and attention to do my own stuff.

Interkarma:
Paralysis resistance will be just another effect, not something added into the paralysis effect itself (which I've already completed). I'm building something much larger and more powerful than classic's effect system, with mods in mind from day one, something classic never had to worry about.  :)

But I agree it's useful to have these insights from the classic effects as well.

Interkarma:
By any chance have you come across how spell reflection in this manner? That would be an awesome effect to know more about, as I know almost nothing about it right now. It's a bit harder to divine internal behaviour just from testing as well.

Allofich:
> By any chance have you come across how spell reflection works in this manner?

In these effect functions, spell absorption, reflection and resistance all just set the appropriate flag. That is:

`Player.MagicEffects2 |= 2` for spell absorption
`Player.MagicEffects2 |= 4` for spell reflection, and
`Player.MagicEffects2 |= 8` for spell resistance

These flags are then checked as needed later. I have found a check for the flag in the function where spells hit, but I need some time to figure it out.

In the meantime, though, I also found a kind of interesting place where the flag is in the function for enemies casting spells at their opponents, like follows:
```
if ( Enemy.CurrentSpellPoints ) >= 0 )
  {
    if ( Target.MagicEffects2 & 4 ) // spell reflection
    {
      int chance = (Enemy.Intelligence) / 2;
      if ( DFRandom(1, 100) < chance )
        Enemy.CurrentSpellPoints = 0;
    }
  }
  else
  {
    Enemy.CurrentSpellPoints = 0;
  }
```

At first I didn't realize what this was and thought maybe I had misidentified something, but looks like it is a bit of "AI" to stop enemies from casting any more offensive spells once they realize their opponent has spell reflection on.

I'll also look at the resistance and absorption effects.

Interkarma:
This is fascinating... Thank you for looking into that.

Allofich:
Ok, it looks something like this:

```
  // First try spell absorption check (TODO), if failed, try reflection + resistance
  if (Target.MagicEffects2 & 4) // spell reflection
  {
    int chance;
    CheckForAntiMagicSpells(Target, 21, &chance);
    if (DFRandom(1, 100) <= chance)
    {
      Target = spellBaseRecord[47]; // guessing a bit here, but I think a pointer to the caster exists from the 47th byte of the spell's base record
      TopOfScreenMessage("Spell was reflected.");
    }
  }
  
  if (Target.MagicEffects2 & 8) // No spell resistance effect. Hit target with spell and try saving throw
	return SavingThrow(...)
  
  // Otherwise, target has a spell resistance effect
  CheckForAntiMagicSpells(Target, 22, &chance);
  if ( DFRandom(1, 100) > chance )
    return SavingThrow(...);
  TopOfScreenMessage("Spell was resisted.");
  
  return false;
  
```

```
  bool CheckForAntiMagicSpells(Entity Target, int ID, int &chance)
  {
	// Iterate over Target's subrecords in a loop
    if (subrecord.Type == 9) // A spellObject record
    {
      int spellData = i[71]; // The spell record at the end of the 71-byte base record
      for (int i = 0; i < 3; ++i)
      {
        if ( (spellData + 2 * i) == ID )
        {
          chance = (spellData[i + 86]);
          return true;
        }
      }
    }
	return false;
  }  
```

`SavingThrow` is already implement by the way (I think in FormulaHelper), but for diseases. I'll try to look into what is needed to use it for spells, but the basics are the same I think. I'll try to look at spell absorption later.

It looks like reflected spells in classic just switch around the target, the projectile doesn't bounce back. Maybe that could be an idea for an alternate behavior.

Allofich:
So while I assume you know this, although the Spell Reflection effect within a spell just sets the magiceffects flag, the spell it belongs to attaches itself as a subrecord to the player. The flag just seems to be used to decide whether to bother searching through the subrecords when the entity is hit by a spell.

Allofich:
Here's the whole function when a spell hits (includes the above reflection and resistance stuff), except the calls to the saving throw function. The saving throw function that is in FormulaHelper is enough for diseases but it needs to be expanded to also handle spells. Also it looks to me like the spell absorption career abilities in classic won't function unless the magic effect flag for spell absorption is set. Maybe it gets permanently set somewhere for those characters but if not, those abilities might not work in classic.

```
int OnSpellHit (spellObject Spell, entity Target) // returns how much of the spell's magnitude is applied to the target
{
  if (Spell.SpellRecord.rangeType == 0) // Self-targeting I think
	return 100;
  
  if (Spell.Caster == Target && Spell.SpellRecord.rangeType == 3) // not sure which one this is
	return 100;
	
  if (Target.MagicEffects2 & 2 // spell absorption effect
	  || Target.Career.SpellAbsorptionFlags & 4 // spell absorption always
	  || Target.Career.SpellAbsorptionFlags & 2 && !IsInLight() // spell absorption in darkness and not in light
	  || Target.Career.SpellAbsorptionFlags & 1 && IsInLight() // spell absorption in light adn in light
  {
	if (Target.MagicEffects2 & 2) // spell absorption effect
	{
		int chance;
		if (CheckForAntiMagicSpells(Target, 20, &chance) // true if spell with absorption effect found
		{
			if (DFRandom(1, 100) > chance)
				goto ReflectionAndResitanceChecks; // absorption failed
		}
		else // Bug? I think this is supposed to be the check for spell absorption using career flags, and should be outside of the magiceffects flag check.
		{
			chance = Target.Level + 25;
			if (DFRandom(1, 100) > chance)
				goto ReflectionAndResitanceChecks; // absorption failed
		}
	}
	
	int spellCost = CalculateCastingCost(Spell, Target) // The target gets how much the spell would cost if THEY cast it. I thought maybe this was a bug, since characters with poor magic skills will get more spell points, but considering the below, this reduces opportunities for absorption by those characters since they won't have room for the spell points, so maybe it's intended.
	if (Target.CurrentMagicka + spellCost <= Target.MaxMagicka) // Target has to have room for the spell points or no absorption happens
	{
		Target.CurrentMagicka += spellCost;
		TopOfScreenMessage("Spell was absorped."); // Should be "absorbed."
		return 0; // 0% of spell hits target
	}
  }
  ReflectionAndResistanceChecks:
  if (Target.MagicEffects2 & 4) // spell reflection
  {
    int chance;
    CheckForAntiMagicSpells(Target, 21, &chance);
    if (DFRandom(1, 100) <= chance)
    {
      *caster = spellBaseRecord[47];
      TopOfScreenMessage("Spell was reflected.");
    }
  }
  
  if (Target.MagicEffects2 & 8) // no spell resistance effect
	return SavingThrow(...)
  
  // Otherwise, has a spell resitance effect
  
  CheckForAntiMagicSpells(Target, 22, &chance);
  if ( DFRandom(1, 100) > chance )
    return SavingThrow(...);
  TopOfScreenMessage("Spell was resisted.");
  
  return 0;
  
```

```
bool IsInLight()
{
  if ((GameMinutes % 1440) >= 360 && (GameMinutes % 1440) <= 1080 )
    return Player.IsOutside;
  else
    return false;
}

```

Allofich:
I put up a PR that prepares SavingThrow(). With that, the calls to SavingThrow() above should be
```
SavingThrow(Spell.SpellRecord[6], EffectFlags[Spell.SpellRecord[6]], Target,
5 * (Target.Level - Caster.Level);
```

`EffectFlags` is [0x08, 0x10, 0x04, 0x20, 0x02]. These are the same as the flags in `DFCareer.EffectFlags`, i.e. Fire, Frost, Poison, Shock, Magic. The last parameter is a modifier to the saving throw chance.

Allofich:
By the way I probably should have named the `OnSpellHit()` function something like `ApplyTargetSpellDefenses()` now that understand it better. I don't know for sure but I assume the final spell magnitude applied to the target is the spell magnitude * the result from this function.

Interkarma:
So conceptually, what's the numerical difference between "resistance" and "critical weakness" to an element? I'm assuming "immunity" is just that.

I also understand that classic can have issues when stacking "critical weakness" values (e.g. stacking critical weakness to disease) that will actually result in the player becoming immune instead.

What I need to do is tease out the intended behaviour so that I can re-implement in a cleaner more robust way in my spell system.

Allofich:
> I'm sorry, I hadn't made it to a computer when I wrote this, was working remotely off my phone.

No need to apologize!

> I also understand that classic can have issues when stacking "critical weakness" values (e.g. stacking critical weakness to disease) that will actually result in the player becoming immune instead.

Yes, I think there may be some sort of problem there, but I'm not sure what it is exactly. When I was working on diseases I was trying to catch one in classic, and I gave my character critical weakness to disease and a big negative modifier in the biography disease resist mod, but he seemed to instead become immune (gave myself huge health and let disease-causing enemies hit me repeatedly, but nothing). I eventually was able to get some diseases after I removed any kind of weaknesses or resistances from the character.

Allofich:
I found how the parameters of a particular instance of a spell casting are calculated from that spell's data.

Pseudocode:
```
CalculateSpellInstanceValues(spellRecord spellRecord, int casterLevel, int percentOfSpell)
{
  if ( spellRecord[72] >= 200 ) // if icon is >= 200? Not sure what this is
    casterLevel = 8;
  for ( int i = 0; i < 3; ++i )
  {
    if ( spellRecord.effects[i] != 0xFF ) // effect exists, not 0xFF
    {
      if ( spellRecord.effects[i].durationBase == 0xFF ) // no base duration
      {
        spellRecord.effects[i].calculatedDuration = 0xFF; // no calculated duration
      }
      else if ( SpellDefinitionData[12 * spellRecord.effects[i].index] & 1 ) // See SpellDefinitionData below
      {
        (spellRecord.effects[i].calculatedDuration) = percentOfSpell         // percentOfSpell may be 100 or may have been lowered by the target's resistance
                                    * ((spellRecord.effects[i].durationMod * (casterLevel / spellRecord.effects[i].perLevel))
                                     + spellRecord.effects[i].durationBase)
                                    / 100;
      }
      else
      {
        spellRecord.effects[i].calculatedDuration = 0;
      }
      if ( SpellDefinitionData[12 * spellRecord.effects[i].index] & 2 )
        spellRecord.effects[i].calculatedChance = spellRecord.effects[i].chanceBase
                                + spellRecord.effects[i].chanceMod * (casterLevel / spellRecord.effects[i].chancePerLevel);
      if ( SpellDefinitionData[12 * spellRecord.effects[i].index] & 4 )
      {
        int calculatedPerLevel1 = caster.Level / (spellRecord + spellRecord.effects[i].magnitudePerLevel);
        int calculatedPerLevel2 = DFRandom(spellRecord.effects[i].magnitudeLevelBase, spellRecord.effects[i].magnitudeLevelHigh) * calculatedPerLevel1;
        spellRecord.effects[i].calculatedDamage = (calculatedPerLevel2 + DFRandom(spellRecord.effects[i].magnitudeBaseLow, spellRecord.effects[i].magnitudeBaseHigh)) * percentOfSpell / 100;
        spellRecord.effects[i].magnitudeLevelBase = 0; // Not sure if there's any reason for this
      }
    }
  }
}
```

I noted it in a comment above, but `percentOfSpell` is the result of the `ApplyTargetSpellDefenses()` or whatever it would be called above. 100 for a fully non-resisted spell.

It doesn't matter for reading spells from `SPELLS.STD`, but for spell records in save games, the calculated duration, chance and damage for the effects in a spell are at, looking from the start of the spell data at the 71st byte:
WORD 74-75 (all of these are inclusive) = calculatedDuration1
WORD 76-77 = calculatedDuration2
WORD 78-79 = calculatedDuration3
BYTE 80 = calculatedDamage1
BYTE 81 = calculatedDamage2
BYTE 82 = calculatedDamage3
BYTE 86 = calculatedChance1
BYTE 87 = calculatedChance2
BYTE 88 = calculatedChance3

`SpellDefinitionData` is found at 0x1B3F4B in `FALL.EXE`. It is an array of 12-byte records, for the various effect types. I only know the meaning of the first byte, as used in the above function.

```
  03 00 00 00 00 00 00 00 00 00 00 00
  05 05 05 00 00 00 00 00 00 00 00 00
  01 00 00 00 00 00 00 00 00 00 00 00
  02 02 02 00 00 00 00 00 00 00 00 00
  04 04 04 00 00 00 00 00 00 00 00 00
  02 00 00 00 00 00 00 00 00 00 00 00
  02 02 02 00 00 00 00 00 00 00 00 00
  04 04 04 04 04 04 04 04 00 00 00 00
  03 03 03 03 03 00 00 00 00 00 00 00
  05 05 05 05 05 05 05 05 00 00 00 00
  04 04 04 04 04 04 04 04 04 04 00 00
  04 04 04 04 04 04 04 04 04 04 00 00
  03 00 00 00 00 00 00 00 00 00 00 00
  01 01 00 00 00 00 00 00 00 00 00 00
  01 00 00 00 00 00 00 00 00 00 00 00
  01 00 00 00 00 00 00 00 00 00 00 00
  02 00 00 00 00 00 00 00 00 00 00 00
  02 00 00 00 00 00 00 00 00 00 00 00
  05 00 00 00 00 00 00 00 00 00 00 00
  03 00 00 00 00 00 00 00 00 00 00 00
  03 00 00 00 00 00 00 00 00 00 00 00
  03 00 00 00 00 00 00 00 00 00 00 00
  03 00 00 00 00 00 00 00 00 00 00 00
  01 01 00 00 00 00 00 00 00 00 00 00
  01 01 00 00 00 00 00 00 00 00 00 00
  01 00 00 00 00 00 00 00 00 00 00 00
  01 00 00 00 00 00 00 00 00 00 00 00
  01 00 00 00 00 00 00 00 00 00 00 00
  01 00 00 00 00 00 00 00 00 00 00 00
  01 01 01 01 01 01 01 00 00 00 00 00
  01 00 00 00 00 00 00 00 00 00 00 00
  01 00 00 00 00 00 00 00 00 00 00 00
  01 00 00 00 00 00 00 00 00 00 00 00
  02 02 02 02 00 00 00 00 00 00 00 00
  03 00 00 00 00 00 00 00 00 00 00 00
  05 00 00 00 00 00 00 00 00 00 00 00
  05 00 00 00 00 00 00 00 00 00 00 00
  01 00 00 00 00 00 00 00 00 00 00 00
  03 00 00 00 00 00 00 00 00 00 00 00
  01 01 01 00 00 00 00 00 00 00 00 00
  02 00 00 00 00 00 00 00 00 00 00 00
  01 00 00 00 00 00 00 00 00 00 00 00
  03 00 00 00 00 00 00 00 00 00 00 00
  00 00 00 00 00 00 00 00 00 00 00 00
  03 00 00 00 00 00 00 00 00 00 00 00
  05 00 00 00 00 00 00 00 00 00 00 00
  05 00 00 00 00 00 00 00 00 00 00 00
  05 00 00 00 00 00 00 00 00 00 00 00
  05 00 00 00 00 00 00 00 00 00 00 00
  05 00 00 00 00 00 00 00 00 00 00 00
  05 00 00 00 00 00 00 00 00 00 00 00
  00 00 00 00 00 00 00 00 00 00 00 00
  80 80 80 00 00 00 00 00 00 00 00 00
  00 00 00 00 00 00 00 00 00 00 00 00
  00 00 00 00 00 00 00 00 00 00 00 00
  00 00 00 00 00 00 00 00 00 00 00 00
  00 00 00 00 00 00 00 00 00 00 00 00
  00 00 00 00 00 00 00 00 00 00 00 00
  81 81 81 81 81 81 81 81 00 00 00 00
  82 83 84 85 86 00 00 00 00 00 00 00
  87 88 89 8A 8B 8C 8D 8E 00 00 00 00
  00 00 00 00 00 00 00 00 00 00 00 00
  8F 90 91 92 93 94 95 96 97 98 00 00
  00 00 00 00 00 00 00 00 00 00 00 00
  99 00 00 00 00 00 00 00 00 00 00 00
  9A 00 00 00 00 00 00 00 00 00 00 00
  00 00 00 00 00 00 00 00 00 00 00 00
  00 00 00 00 00 00 00 00 00 00 00 00
  00 00 00 00 00 00 00 00 00 00 00 00
  9B 00 00 00 00 00 00 00 00 00 00 00
  9C 00 00 00 00 00 00 00 00 00 00 00
  9D 00 00 00 00 00 00 00 00 00 00 00
  9E 00 00 00 00 00 00 00 00 00 00 00
  9F 00 00 00 00 00 00 00 00 00 00 00
  A0 00 00 00 00 00 00 00 00 00 00 00
  A1 00 00 00 00 00 00 00 00 00 00 00
  A2 00 00 00 00 00 00 00 00 00 00 00
  A3 00 00 00 00 00 00 00 00 00 00 00
  A4 00 00 00 00 00 00 00 00 00 00 00
  A5 00 00 00 00 00 00 00 00 00 00 00
  A6 A6 A6 A6 A6 A6 A6 00 00 00 00 00
  A7 00 00 00 00 00 00 00 00 00 00 00
  A8 00 00 00 00 00 00 00 00 00 00 00
  A9 00 00 00 00 00 00 00 00 00 00 00
  00 00 00 00 00 00 00 00 00 00 00 00
  00 00 00 00 00 00 00 00 00 00 00 00
  AA 00 00 00 00 00 00 00 00 00 00 00
  00 00 00 00 00 00 00 00 00 00 00 00
  AB 00 00 00 00 00 00 00 00 00 00 00
  AC 00 00 00 00 00 00 00 00 00 00 00
  AD AE AF 00 00 00 00 00 00 00 00 00
  00 00 00 00 00 00 00 00 00 00 00 00
  B0 00 00 00 00 00 00 00 00 00 00 00
  00 00 00 00 00 00 00 00 00 00 00 00
  00 00 00 00 00 00 00 00 00 00 00 00
  B1 00 00 00 00 00 00 00 00 00 00 00
  00 00 00 00 00 00 00 00 00 00 00 00
  00 00 00 00 00 00 00 00 00 00 00 00
  00 00 00 00 00 00 00 00 00 00 00 00
  00 00 00 00 00 00 00 00 00 00 00 00
  00 00 00 00 00 00 00 00 00 00 00 00
  00 00 00 00 00 00 00 00 00 00 00 00
```

Interkarma:
This is amazing! Thank you for putting that together and documenting so thoroughly here. This kind of detail will be invaluable for setting up the final touches on effects down the road.

PR #737, opened 2018-05-12
 renamed automap classes/files, palace mark on map fix, organizationinfo impr., implemented spymaster

Nystul-the-Magician:
renamed DaggerfallAutomap and DaggerfallExternalAutomap into Automap and resp. ExternalAutomap
this might create some problems with mods, but since I really wanted to rename these 2 files/classes since half a year now I just did it. They are now more in line to all the other files in the folder (exception is the DaggerfallBankManager), palaces are now correctly revealed, progress with TalkManager


Interkarma:
Provided any mods use GameManager like they're supposed to, this shouldn't be too much of a problem. Do you know any mods that might access these game objects under the old name?

Nystul-the-Magician:
the thing is the scripts name and class names changed as well. I could think of the archeologist mod with this locator device...

ajrb:
No effect on Arch mod that I can see, I just checked.

Nystul-the-Magician:
that's good to hear

PR #738, opened 2018-05-12
Mark incomplete quests as broken

KirillSmirnov:
Several quests in vanilla DaggerFall are incomplete: only .QRC or
.QBN file is present and other is missing. Mark these quests as
broken explicitly.

Signed-off-by: Kirill K. Smirnov <kirill.k.smirnov@gmail.com>


Allofich:
These quests seem to be unused variants of existing quests. You can convert them by substituting in other quests that match for them. You can see that they match because relevant keywords are in the .QBN files if you examine them, and they appear in places that make sense in the converted file.

A0C00Y04-> Use A0C0XY04.QRC
N0C00Y01-> Use N0B40Y07.QRC
R0C40Y23-> Use R0C4XY23.QRC

KirillSmirnov:
@Allofich  I've diffed the QBN parts these pairs of quests and got:
- N0C00Y01 differs greatly from N0B40Y07, and the quest flow seems meaningless and incomplete to me.
- R0C40Y23 is almost identical to R0C4XY23 before the commit 7848b1682c2624845d83665f40b785968ef3c98a.
- A0C0XY04 is an improved version of A0C00Y04. They do differ but not so much.

So I don't think that these incomplete quests should be improved: either we get a duplicate or a completely new quest.

KirillSmirnov:
Indeed

ajrb:
These quests don't get given out unless they're in a quest list. Guess Gav should take the call whether to delete them at this point or not.

Interkarma:
Sounds quite reasonable to just delete them then. Cheers!

Allofich:
Just a suggestion. Since these seem to be incomplete quests, maybe remove the "FAILED" next to them, since they can't be "passed" anyway? Otherwise we will never have a list that is empty of "FAILED".

Interkarma:
Agree with Allofich, best just to remove those quest line items from table.

Allofich:
I think it could be OK to still document them as incomplete in classic, I just meant remove the "FAILED" line that makes it look like they should be supported. But removing the lines completely seems OK, too.

KirillSmirnov:
I suggest to keep these lines in table to maintain the one-to-one correspondence 'line entry' <-> 'DF quest file'. This way it is much easier to track a quest migration progress.There are still a lot of quests to migrate.

Interkarma:
>I think it could be OK to still document them as being incomplete in classic, I just meant remove the "FAILED" word next to them that makes it look like they should be supported. But removing the lines completely seems OK, too.

>I suggest to keep these lines in table to maintain the one-to-one correspondence 'line entry' <-> 'DF quest file'. This way it is much easier to track a quest migration progress.There are still a lot of quests to migrate.

That sounds fine guys. Maybe just change the word "FAILED" to something else indicating these quests aren't valid, compared to valid quests that just don't work yet. I'll be happy to merge then. :)

KirillSmirnov:
Done

PR #739, opened 2018-05-12
Identify more classic data, prepare SavingThrow() to accept spells

Allofich:
With this, SavingThrow() is matched to classic. As soon as there is a way to store and access resist flags and resist amounts on an actor, it can be used.


Assets/Scripts/Game/Formulas/FormulaHelper.cs
```diff
@@ -842,7 +871,12 @@ static int SavingThrow(DFCareer.EffectFlags effectFlags, DaggerfallEntity target
             if (toleranceFlags == DFCareer.Tolerance.LowTolerance)
                 savingThrow = 75;
 
-            savingThrow += biographyMod;
+            savingThrow += biographyMod + modifier;
```
Interkarma: This actually answers my question in the other commit. I'm sorry, I hadn't looked through the code when I posted that. This looks most useful thank you!

Assets/Scripts/Game/Formulas/FormulaHelper.cs
```diff
@@ -818,15 +818,44 @@ static int GetBonusOrPenaltyByEnemyType(DaggerfallEntity attacker, EnemyEntity A
             return damage;
         }
 
-        static int SavingThrow(DFCareer.EffectFlags effectFlags, DaggerfallEntity target)
+        static int SavingThrow(int effectType, DFCareer.EffectFlags effectFlags, DaggerfallEntity target, int modifier)
```
Interkarma: If you can, would it be OK to make a small change here to rename effectType to elementType? We'll need to remap my usage of elements back to a classic enum or something to pass into here for type safety eventually.

Allofich:
It's renamed to elementType now.

> We'll need to remap my usage of elements back to a classic enum or something to pass into here for type safety eventually.

Sure, I don't mind if you go ahead and make such changes if you need to.

PR #742, opened 2018-05-13
implemented SpyMaster pop-up message box like in classic, quest residence discovery behavior changed

Nystul-the-Magician:


Interkarma:
I'm amazed we're finally at the point where even more obscure features like the spy master are being implemented. Very exciting! I'll merge when I can.

Nystul-the-Magician:
spymaster was quite easy to do (unless I have overseen some features):
-) always knows answers
-) only gives anyinfo messages instead of rumor messages when a quest topic is asked about under "tell me about" - basically all %hnt2 macros become %hnt macros :)

Interkarma:
Still impressed! :)

Nystul-the-Magician:
thank you ;)

PR #745, opened 2018-05-14
Player motor changes

MeteoricDragon:
Migrates code responsible for enabling input to move the player along the ground to PlayerGroundMotor. Also migrates the code enabling moving platforms to move the player to the same new class. 

Migrates a couple Height variables unused by PlayerMotor to PlayerHeightChanger.

Sorts PlayerMotor into Regions.

Adjusts access modifiers of classes I've created and adds XML documentation to several methods.

### DIFF
--- a/Assets/Prefabs/Player/PlayerAdvanced.prefab
+++ b/Assets/Prefabs/Player/PlayerAdvanced.prefab
@@ -38,6 +38,7 @@ GameObject:
   - component: {fileID: 114607307154016044}
   - component: {fileID: 114549470222698842}
   - component: {fileID: 114101796857139080}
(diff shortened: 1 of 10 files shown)


Interkarma:
Thank you for continuing to clean up PlayerMotor. There's probably not much more splitting out you can now? :)

I'll get this tested and merged when I can.

Interkarma:
I've found one issue with this in testing. Mounting horse or cart results in player sinking into ground. Doesn't look like riding height is correct after being moved. 

When you have a fix, please don't forget to send to my branch "player-motor-changes" instead of master, as that's where I'm test everything. Otherwise seems OK so far! :)

image

PR #746, opened 2018-05-14
tg and dbh guild halls discovered when loading a interior savegame and getting promoted

Nystul-the-Magician:
tg and dbh guild halls are now also correctly discovered when loading a interior savegame and getting promotion in interior and then leaving building

had to introduce new event in StreamingWorld class:  OnAvailableLocationGameObject so that GameManager.Instance.StreamingWorld.GetCurrentBuildingDirectory() will return the building directory and not null
initially I though existing OnCreateLocationGameObject(dfLocation) event would work but it turned out that it is fired too early (inside the function creating the gameobject, but gameobject is assigned outside afterward)
not sure if OnCreateLocationGameObject should do what OnAvailableLocationGameObject does now but I didn't want to break existing stuff.

### DIFF
--- a/Assets/Scripts/Game/Guilds/DarkBrotherhood.cs
+++ b/Assets/Scripts/Game/Guilds/DarkBrotherhood.cs
@@ -82,6 +82,14 @@ namespace DaggerfallWorkshop.Game.Guilds
         {
             // Register for location entry events so can auto discover guild houses.
             PlayerGPS.OnEnterLocationRect += PlayerGPS_OnEnterLocationRect;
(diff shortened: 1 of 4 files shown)


TheLacus:
This is somewhat similar to what i asked [here](http://forums.dfworkshop.net/viewtopic.php?p=12050#p12050).

Interkarma:
> This is somewhat similar to what i asked here.

Sorry for the delay on that. The event is easy, but I wanted to see if I could address some of the startup order problems as well. I'll try to make time to get to this today or tomorrow.

PR #747, opened 2018-05-14
Fix RideHeight not initializing.

MeteoricDragon:


Interkarma:
Everything looking good with this change. I'll merge to master and let you know if I come across anything else. Thank you for the work!

PR #748, opened 2018-05-14
Classic quests janitorial

KirillSmirnov:


Assets/StreamingAssets/Tables/QuestList-Classic.txt
```diff
@@ -157,7 +164,7 @@ A0C00Y00, Commoners, N, 0,
 A0C00Y07, Commoners, N, 0,
 A0C00Y08, Commoners, N, 0, Passed. questgiver name is incorrect (almost like a codename)
 A0C00Y10, Commoners, N, 0,
-A0C00Y11, Commoners, N, 0,
+-A0C00Y11, Commoners, N, 0, FAILED. The _healer_ does not respond. Need to handle 'clicked anNPC and at least anAmount gold otherwise do aTask'.
```
Interkarma: This action variant should be fairly easy to add. I'll see if I can knock it together today or tomorrow.

Interkarma: I've implemented this conditional action. It appears to work as expected on A0C00Y11 when speaking with healer for talisman with correct popup fired based on player's gold count.
There are a couple of other issues with that quest still. I can't seem to find NPCs by asking around town yet, and I feel time limit is much too tight considering the travel involved. But those are separate issues to the quest action which should be working now.

PR #752, opened 2018-05-16
Player motor

MeteoricDragon:
So, with these changes, I'm reasonably sure I'm done with PlayerMotor.  At least on a massive scale.

### DIFF
--- a/Assets/Scripts/Game/Player/AcrobatMotor.cs
+++ b/Assets/Scripts/Game/Player/AcrobatMotor.cs
@@ -55,7 +55,7 @@ namespace DaggerfallWorkshop.Game
         /// Jump! But only if the jump button has been released and player has been grounded for a given number of frames
         /// </summary>
         /// <param name="moveDirection"></param>
-        public void DoJump(ref Vector3 moveDirection)
+        public void HandleJumpInput(ref Vector3 moveDirection)
         {
             // Cancel jump if player is paralyzed
             if (GameManager.Instance.PlayerEntity.IsParalyzed)
--- a/Assets/Scripts/Game/Player/HeadBobber.cs
+++ b/Assets/Scripts/Game/Player/HeadBobber.cs
@@ -5,6 +5,9 @@ using DaggerfallConnect;
 
 namespace DaggerfallWorkshop.Game
 {
+    /// <summary>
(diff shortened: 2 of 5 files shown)


Interkarma:
Nice one! Thank you for your work on all of this. The motor setup is significantly cleaner now without breaking anything. That's a tough challenge for any programmer. When the time comes to rework anything else in the player motor setup, it will be much easier thanks to you. 👍

MeteoricDragon:
Thanks.  I'm actually making some XML documentation for a few pieces right now for this code right now.  The next commit I push for this will have no functional changes. just documentation.

MeteoricDragon:
I've got a question.  As standard programming practice I've heard it's best to make classes unable to modify fields of other classes inside their own methods because it makes the classes more independent.  But I find with the PlayerMotor, it is quite difficult to make the classes I created independant of Playermotor because they do need to access and set some fields or properties of PlayerMotor.  I figure a programmer should be able to read the name of a method, and maybe the xml description and know exactly what the method does without reading the implementation of the method.   Is there any advice regarding this topic that you've gained experience with?

Interkarma:
There's a few reasons you'll see publicly accessible fields in a lot in MonoBehaviour-derived classes in Unity:
* Public fields are exposed to Inspector in Unity Editor automatically to edit component properties at design time or run time.
* The state of public fields are serialized automatically with scene data when you save your scene (public properties are not).
* Programmers can easily setup property drawers (and other fun tricks) to assign ranges and tooltips to these fields for editor.

I don't recommend changing these particular fields to properties, as you'll lose the editor integration and  links from prefabs, etc. Unity has it's own idea of "standard programming practice". :)

BTW, you can still add XML comments to these fields if you like, any documentation builder will still pick them up.

MeteoricDragon:
Ok thanks.

ajrb:
> As standard programming practice I've heard it's best to make classes unable to modify fields of other classes inside their own methods because it makes the classes more independent. 

This is standard object oriented programming practice, yes. 

OOP 101 if you like. The concept is called encapsulation and means that the amount of code that can alter a member variable is restricted in size and scope which gives less chance of bugs & unexpected side effects etc.

I had a mini-rant about C#'s lack of any protection levels between protected and internal. I suspect maybe assemblies are supposed to provide this delineation, but not done the research to know.

PR #753, opened 2018-05-16
 Add guildjoin and guildrank commands for guild testing 

ajrb:
* Add guildjoin and guildrank commands for guild testing
* Remove temp testing features from guilds (i.e. no free resting and rank changes only every 28 days)
* Use faction id enum where appropriate
* Fix gdd macro for donations when not temple member


Interkarma:
It's always good to see another Hazelnut PR in the queue. Will get this merged soon.

PR #754, opened 2018-05-17
Finish Fortify Effects

MeteoricDragon:
Implements the remaining 5 fortify effects, first 3 originally authored by @Interkarma, Last 5 continued by the same example set by him.  This one was actually pretty easy, just compare what you did and continue the pattern. Tested and verified spells are available now in spellmaker.


Interkarma:
Thank you! 👍 

Just so you know, I don't normally accepts PRs into systems I'm in the middle of building out. But as you correctly surmised, this is just a matter of setting the correct properties in the inherited effect classes. I'll quickly double-check later and merge.

MeteoricDragon:
You're welcome.  I'll keep that in mind when I look at systems you're building in the future.  When you find time to work on the transfer attributes system, I'd gladly assist you the same way in completing them as I did for Fortify attributes if it is as simple as Fortify was.

Interkarma:
Thanks again for knocking these out. I'll take you up on the offer if you like on the Transfer effects. I'll be getting to those soon once I'm happy with the teleport effect.

PR #756, opened 2018-05-18
Optionally allow TextBox to receive negative or floating point numbers

TheLacus:
TextBox numeric mode only allows positive numbers (I assume this is what's required by classic) but mod settings have keys which allow user input for negative numbers and decimals (specifically _TupleFloatKey_ and _TupleIntKey_). 

So i added an enum property to texbox to safely receive also minus and decimal mark when required. Default option is _Natural_, which is current behaviour. Interkarma are you ok with this? If you prefer, i can replace entirely _isNumeric_ flag with a _None_ option.


Interkarma:
This is a very positive and useful change. Thank you!

PR #757, opened 2018-05-18
 Add new quest action to check skill level 

ajrb:
Been sitting on this code for 2 months... only needed a few fixes and testing.


Assets/Scripts/Game/Questing/Actions/SkillLevel.cs
```diff
@@ -0,0 +1,94 @@
...
+    /// Triggers when player has reached specified skill level or higher.
+    /// </summary>
+    public class SkillLevel : ActionTemplate
+    {
+        DFCareer.Skills skill;
+        int minSkillValue;
+
+        public override string Pattern
+        {
+            get { return @"skill (?<skillName>\w+) (?<minSkillValue>\d+)"; }
```
Interkarma: Just a minor thing, but would you mind changing this to "when skill" so it reads more like a trigger condition?

If you want to go full throttle, you could make it "when skill (?<skillName>\w+) is at least (?<minSkillValue>\d+)" haha :)

ajrb:
Sure, I was patterning off LevelCompleted and being consistent with that. You are right that is the odd one out. Hold off on merging until I get to that.

Interkarma:
Cheers mate. It's such a minor thing, and it didn't really worry me either way. Just thought it might be clearer for a new quest author someday in future as it's a custom effect unknown in Tipton's docs. Thanks for doing that. :)

PR #759, opened 2018-05-19
status info messagebox fix, hotkey window toggling for remaining windows implemented

Nystul-the-Magician:
according to your example I implemented all remaining windows that benefit from toggle behavior, hope you find this better now ;)

furthermore I fixed status info messagebox bug, status hot-key proceeds messagebox now as implemented in vanilla daggerfall, enter key will proceed, escape key will close messagebox immediately
also status hotkey will now proceed messagebox as convenience feature

hotkey toggling implemented for remaining windows:
- character sheet
- log book
- automaps
- travel map
- rest window
- transportation window

### DIFF
--- a/Assets/Scripts/Game/DaggerfallUI.cs
+++ b/Assets/Scripts/Game/DaggerfallUI.cs
@@ -1157,10 +1157,12 @@ namespace DaggerfallWorkshop.Game
         {
             // Setup status info as the first message box.
             DaggerfallMessageBox statusBox = new DaggerfallMessageBox(Instance.uiManager, Instance.uiManager.TopWindow);
+            statusBox.ExtraProceedBinding = InputManager.Instance.GetBinding(InputManager.Actions.Status); // set proceed binding for statusBox to key binding for status window
             statusBox.SetTextTokens(22);
 
             // Setup health info as the second message box.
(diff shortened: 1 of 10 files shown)


Interkarma:
Thank you for your persistence with this Nystul! I'll merge now and let you know if any issues encountered.

I'm sure this will make more than a few people happy. Maybe even I will come around to toggle after a while. :)

Interkarma:
@Nystul-the-Magician Hey mate, found one small problem with travel map after change. This breaks exiting using ESC. Now have to press ESC twice to exit travel map, and the second press "falls through" and opens pause menu. Let me know if any help needed. :)

Otherwise very happy with it so far 👍

Nystul-the-Magician:
will take a look thanks for letting me know

Interkarma:
Sounds good to me. Thank for you looking into that. :)

Nystul-the-Magician:
do we want to make "esc" return to game when in region subwindow or to world map?

Nystul-the-Magician:
I will stick to first option since I didn't manage to get the second working anyway
(if you want I will investigate more how to only close regional window when in regional view
this is trickier to achieve than I initially thought)

pull request issued

PR #760, opened 2018-05-19
Import guild memberships from classic

Allofich:
Imports guilds from classic, including rank and time of last rank change.

The only used guild value I know of that is not imported is one that seems to be involved in guild promotions and rewards.

@ajrb I had to make a couple guild-related methods public so I could access them, I hope that's alright with you. Also I had to do some conversion work to match to the days-since-promotion that you're using. Classic goes by minutes. (As far as I have seen classic always uses minutes for game time related calculations)

### DIFF
--- a/Assets/Scripts/API/Save/DiseaseOrPoisonRecord.cs
+++ b/Assets/Scripts/API/Save/DiseaseOrPoisonRecord.cs
@@ -92,7 +92,7 @@ namespace DaggerfallConnect.Save
             // Copy base record data
             base.CopyTo(other);
 
-            // Copy item data
+            // Copy parsed data
             other.parsedData = this.parsedData;
         }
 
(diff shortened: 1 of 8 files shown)


Nystul-the-Magician:
i like this

ajrb:
Hey Allofich, looks good to me man! There's a little re-factoring I'd like to do once this is merged to re-protect some methods - if that would be ok with you? Mainly move some of the creation of memberships to be responisbiliity of GuildManager. I'm really pleased you've done this work though, mainly cos I didn't want to. :-D

ajrb:
Oh, and I used days simply to make the conditional check simpler and easier to understand. I gave no thought to classic TBH, so thanks for having my back on that one!

Interkarma:
Looks good to me, and as @ajrb is happy, I'll just merge this now. He can add his own refinements later as needed.

You guys are seriously awesome!

PR #763, opened 2018-05-20
Court screen updates

Allofich:
- Prison screen implemented
- Fix to not allow using ESC key to cancel court window or court message boxes, which would mess things up as it wasn't set up to handle it
- Fix to prevent camera recoil when leaving court screen

Also, although it's not in classic, I added a "cheat" key to speed up the day countdown by holding down ESC, which is nice to have while testing if nothing else. I don't know how hardcore people want to be about actually having to wait through the prison sentence.

Interkarma, is there a good way to change the color of text? The text on the prison screen should be different from the default color, as in this screenshot.
prisonscreen


Interkarma:
Yep, you can change colour of text by setting label.TextColor to whatever you need. Just let me know if any problems and I'm happy to help. :)

Allofich:
Yeah, I found it after that comment. For some reason I thought that was not supported. I'm adding that in plus a few fixes.

Allofich:
Color is fixed. I added them to the definitions you mentioned.

Also the screen is reset now so it won't be showing the prison screen the 2nd time onward (as it was at first with this PR)

PR #764, opened 2018-05-20
ConversationData.txt - fixes for old savegame format issues - should fix most of the problems

Nystul-the-Magician:
RestoreConversationData now checks for quest resource entries and rumors of orphaned quests (that are no longer running) and removes them

additional checks should prevent some issues I found when above mechanism would not be in place (double net mechanism)

### DIFF
--- a/Assets/Scripts/Game/TalkManager.cs
+++ b/Assets/Scripts/Game/TalkManager.cs
@@ -545,6 +545,11 @@ namespace DaggerfallWorkshop.Game
             get {  return listTopicThing; }
         }
 
+        public static string TextDatabase
+        {
+            get { return textDatabase; }
+        }
+
         public bool ConsoleCommandFlag_npcsKnowEverything
         {
             get { return consoleCommandFlag_npcsKnowEverything; }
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Game/TalkManager.cs
```diff
@@ -1922,16 +1946,27 @@ public void RestoreConversationData(SaveDataConversation data)
...
+                        if (questInfo.resourceInfo.ContainsKey(key)) // if list of quest resources contains a matching entry
                         {
                             questInfo.resourceInfo[key].questResource = item; // update (relink) it
                         }
                     }
 
                 }
             }
-
             listRumorMill = data.listRumorMill;
```
Interkarma: I've found an old save (player is on main quest) where data.listRumorMill returns null. This causes the subsequent for loop to throw an exception. It might be worth doing some null handling on this line as well. I can provide copy of save if needed. :)

Assets/Scripts/Game/TalkManager.cs
```diff
@@ -1922,16 +1946,27 @@ public void RestoreConversationData(SaveDataConversation data)
...
+            {
+                ulong questID = listRumorMill[i].questID;
+                if (GameManager.Instance.QuestMachine.GetQuest(questID) == null)
+                {
+                    Debug.Log(String.Format("save data contains orphaned rumors for quest with id {0}. Removing these rumors...", questID));
+                    listRumorMill.Remove(listRumorMill[i]);
+                }
+            }
+
             if (listRumorMill == null)
```
Interkarma: Update: Oh, you have this null check already - sorry mate. Just need to move this up above the for loop then. :)

Nystul-the-Magician: thanks for improving this ;)

PR #765, opened 2018-05-20
escape key now closes travel map window correctly

Nystul-the-Magician:
if wanted I will investigate if it is possible to only close regional window when in regional view - currently escape closes travel map window completely when in there
but this is more complex than thought at first glance


Interkarma:
Yep, this is perfect now. No other problems observed with toggle. Thank you! :)

PR #769, opened 2018-05-20
Crusher

MeteoricDragon:
Implements a functional crushing system.  Also fixes a bug that was introduced with the smooth crouching update.  When loading a game with a different crouching state than your previous, your game would be stuck in either crouching while walking full speed, or walking while crouching speed.

**There is a problem that prevents this from being Merge-ready.**  There is a SphereCollider I attached to the PlayerAdvanced components and set it as a trigger, and I call OnTriggerEnter and perform the crush logic in the PlayerCrush class.  Because of the position of the SphereCollider, many rays and projectiles are blocked if they are below a certain angle (because the trigger collider blocks them)  I am seeking advice how to work around this.  I can imagine some but was wondering the best solution.

### DIFF
--- a/Assets/Scripts/Game/PlayerCrush.cs
+++ b/Assets/Scripts/Game/PlayerCrush.cs
@@ -1,0 +1,51 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace DaggerfallWorkshop.Game
+{
+    public class PlayerCrush : MonoBehaviour
+    {
+        private PlayerHeightChanger heightChanger;
+        private PlayerMotor playerMotor;
+        private SphereCollider crushCollider;
+        private CharacterController controller;
+    
+	    void Start ()
+        {
+            heightChanger = GetComponent<PlayerHeightChanger>();
+            playerMotor = GetComponent<PlayerMotor>();
+            
+            crushCollider = GetComponent<SphereCollider>();
+            crushCollider.center += new Vector3(0, 0.45f);
+            controller = GetComponent<CharacterController>();
+	    }
+	
+	    void Update ()
+        {
+
+        }
+
+        private void OnTriggerEnter(Collider other)
+        {
+            // if we didn't collide with a crushing object, return
+            if (other.gameObject.GetComponent<MeshCollider>() == null)
+                return;
+
+            // If player is standing, crushing object forces them into a crouch, 
+            if (!playerMotor.IsCrouching && heightChanger.HeightAction != HeightChangeAction.DoCrouching)
+            {
+                Debug.Log("Force Crouch");
+                heightChanger.HeightAction = HeightChangeAction.DoCrouching;
+            }
+            // if player already crouching, then kill.
+            else if (playerMotor.IsCrouching)
+            {
+                GameManager.Instance.PlayerEntity.SetHealth(0);
+                Debug.Log("Crush!");
+            }
+
+
+        }
+    }
+}
--- a/Assets/Scripts/Game/PlayerCrush.cs
+++ b/Assets/Scripts/Game/PlayerCrush.cs
@@ -35,17 +35,13 @@ namespace DaggerfallWorkshop.Game
             // If player is standing, crushing object forces them into a crouch, 
             if (!playerMotor.IsCrouching && heightChanger.HeightAction != HeightChangeAction.DoCrouching)
             {
-                Debug.Log("Force Crouch");
                 heightChanger.HeightAction = HeightChangeAction.DoCrouching;
             }
             // if player already crouching, then kill.
             else if (playerMotor.IsCrouching)
             {
                 GameManager.Instance.PlayerEntity.SetHealth(0);
-                Debug.Log("Crush!");
             }
-
-
         }
     }
 }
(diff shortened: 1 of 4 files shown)


Interkarma:
I had this issue with the player's capsule originally and "solved" it the wrong way (by re-positioning ray origin). When I wrote the magic casting setup, I had a bit more experience and used Physics.IgnoreCollision() instead. Here's the appropriate line from DaggerfallMissile.cs so the player's fired missiles do not interact with their own collider.

    Physics.IgnoreCollision(caster.GetComponent<Collider>(), this.GetComponent<Collider>());

I still need to return to PlayerActivate and WeaponManager to refactor those ray and sphere casts to ignore player's own capsule. This is something I'd like to address, please.

But in this case, do you really need another trigger collider? Could you not use the player's capsule instead?

MeteoricDragon:
Perhaps I could?  I'll need to check it out.

Interkarma:
Thank you mate, it would be preferable if you could. Things can get a bit screwy to manage with multiple colliders on same object. It can work, but it introduces more complexity I'd like to avoid if possible.

MeteoricDragon:
@interkarma I don't think the playerController can be used to detect the crushing, I tried using OnCollisionEnter and the method wouldn't trigger when placed inside PlayerCrush.cs.  
The only thing I can think of to try is instead of using a trigger, inside an update call a spherecast upwards and see if it detects any MeshColliders inside the player. I'll try it and see if it works.

Interkarma:
No worries. I merged some changes this morning that should help with this. If you can do the following:
1. Ensure up to date with master
2. Add your SphereCollider trigger like before, except on own object underneath SmoothFollower. Size and position as required.
3. Set this object to use Player layer in Inspector (note: be sure to set layer, don't change tag).
4. Attach your crusher to this new object rather than root of player. You can still get the other references you need via GameManager in the usual way.

Any rays from PlayerActivate should automatically ignore the new trigger now due to layering. If the same thing happens to weapon sphere casts, I can add layer masking there as well.

Just let me know if any help needed. :)

MeteoricDragon:
I just saw your message now.  I already worked a solution using SphereCast to detect any MeshColliders within the player.  It seems to work perfectly.  If the player is standing, it forces him/her to crouch, if he/she is crouching, gets the hit distance from the origin.  next frame if hit distance is lesser than the previous frame, and If grounded, Crush!

Interkarma:
I'm not sure how I feel about a spherecast every frame. Probably OK in this context as it has limited range and Unity should optimise internally. I'll try it in a branch and see how it goes.

As you're no longer using a SphereCollider, could you please revert those other changes to PlayerAdvanced.prefab. You only want to merge the effective changes, not the experimental stuff.

Assets/Scripts/Game/PlayerCrush.cs
```diff
@@ -0,0 +1,58 @@
...
+            if (!playerMotor.IsCrouching)
+                distance = (controller.height / 2f) - (controller.height * 0.1f);
+            else
+                distance = (controller.height / 2f);
+
+            Ray ray = new Ray(controller.transform.position, Vector3.up);
+            RaycastHit hit = new RaycastHit();
+            if (Physics.SphereCast(ray, controller.radius * 0.85f, out hit, distance))
+            {
+                if (hit.collider.GetComponent<MeshCollider>())
```
Interkarma: Curious why grabbing MeshCollider<>() here. Is this to test if player is being crushed under actual scene geometry compared to say an enemy walking off a ledge above and landing on player's head?

MeteoricDragon: LOL. 🤣 earlier when I was using the CapsuleColider as a form of detection, I was getting forced to crouch when I was walking over dead enemy corpses and loot piles.  And I kept it in the code when i transferred and modified it for the update method. I suppose it's still possible when an enemy dies on some ledge, if you walk up to the ledge you might trigger a collision.

MeteoricDragon:
Ok, Had to remove the SphereCollider, then close and restart Unity before the changes would appear in GithubDesktop. but now it's done and pushed to the branch.  Hope this works well.  Do you have a performance benchmark that informs you?  I might invest in one of them.

Assets/Scripts/Game/PlayerCrush.cs
Interkarma: Yep, this is a reasonable solution to the problem of dealing with non-level geometry like enemies and loot piles as they don't carry mesh colliders. Well done.

My other concern is what happens if player manages to crouch and crawl in somewhere very tight (e.g. on top of a shelf very close to the ceiling). They would potentially be both grounded and within range to trigger a crush. That's something we'd have to test. I'll stick it to a branch and we'll go from there.

Interkarma:
Unity has a decent profiler built in (Windows menu > Profiler). I don't think in this case it will be much of a problem, but will let you know.

Assets/Scripts/Game/PlayerCrush.cs
Interkarma: This works very well so far! There's a spot in Mantellan Crux where player can crawl under a low shelf slightly lower than crouching head height., But with the tolerances you've put in place it all just works as expected. The player literally has to stand under a lift and let it lower all the way down onto them, which is perfect.

Also did some profiling, no real impact to physics system. I think Unity perform these kinds of tests very efficiently and continuously under the hood. So no problems with that either.

I'll play with it a bit more and try to break it. Will merge to master soon. Good work! 👍

Interkarma:
I'm happy with this, no problems encountered. Will go ahead and merge with master now.

PR #772, opened 2018-05-21
Fix for commoner quest handout closing HUD

ajrb:
Needs to be reviewed and given green light by Nystul before merge.


Nystul-the-Magician:
looks good to me ;) thx for fixing

Interkarma:
Thank you! 👍

PR #775, opened 2018-05-22
Add ability to init quest scripts when new game started

ajrb:


Interkarma:
This should keep the quest guys on the forums happy. :)

PR #776, opened 2018-05-23
Add ability to get work inside castles

numidium:
This is a continuation of my work on the "work" functionality in TalkManager.

I used a separate dictionary because 1. The block structure for castles/dungeons is different than towns and 2. the process of looking for work in a castle is simpler. There's no "work" dialogue in classic for castle questors so an NPC list doesn't need to be pre-populated.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallQuestPopupWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallQuestPopupWindow.cs
@@ -92,7 +92,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
 
                 // Show refuse message
                 sender.CloseWindow();
-                ShowQuestPopupMessage(offeredQuest, (int)QuestMachine.QuestMessages.RefuseQuest, false);
+                ShowQuestPopupMessage(offeredQuest, (int)QuestMachine.QuestMessages.RefuseQuest);
             }
         }
 
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallQuestPopupWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallQuestPopupWindow.cs
@@ -92,7 +92,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
 
                 // Show refuse message
                 sender.CloseWindow();
-                ShowQuestPopupMessage(offeredQuest, (int)QuestMachine.QuestMessages.RefuseQuest);
+                ShowQuestPopupMessage(offeredQuest, (int)QuestMachine.QuestMessages.RefuseQuest, false);
             }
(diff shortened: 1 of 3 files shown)


Nystul-the-Magician:
Thanks for adding this. Great work!

Interkarma:
Cheers! Looking forward to getting this merged. Just let me know if you'd like to look at saving state first, or if that's something you'd like to tackle later. I'm happy to leave the PR open if you want to keep adding to it for now.

numidium:
Adding the dict to the save state is a good idea. I might as well add that to this PR. I’ll get on that after work.

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallQuestPopupWindow.cs
```diff
@@ -92,7 +92,7 @@ protected virtual void OfferQuest_OnButtonClick(DaggerfallMessageBox sender, Dag
 
                 // Show refuse message
                 sender.CloseWindow();
-                ShowQuestPopupMessage(offeredQuest, (int)QuestMachine.QuestMessages.RefuseQuest, false);
+                ShowQuestPopupMessage(offeredQuest, (int)QuestMachine.QuestMessages.RefuseQuest);
```
ajrb: Can you explain why this false has been removed please?

ajrb:
Added a question to the code diff.

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallQuestPopupWindow.cs
numidium: The window was staying open after the user clicks “no” to an offered quest. I removed this and the CloseWindow on Setup so that the quest giver window closes itself properly regardless of whether the quest offer is shown after clicking “talk” or upon clicking the NPC.

ajrb: You've not used the latest version of this class for your work, so I would appreciate it if you could update to the latest version and test before you remove this false param.

numidium: “You've not used the latest version of this class...”

Silly me. Thank you for catching that.

numidium: “You've not used the latest version of this class...”

Silly me. Thank you for catching that.

Interkarma: Good catch Hazelnut. I missed that one on the read.

numidium:
Hopefully the issues have been addressed now.

Interkarma:
Thank you @numidium! I'll merge this to master now so it will be in next round of builds.

PR #779, opened 2018-05-26
Mostly finish up region and faction conditions

Allofich:
I finally got through this massive function for handling the background events that set region flags for things like wars, etc., along with changes to faction power caused by this. This is also the basis for non-quest rumors and messages on the town sign boards, which are created by these events (I haven't implemented that).

This follows classic very closely, plus fixing the usual bugs/oversights.

There's one part that seems to be an oversight that I may visit again later. There's a section of data that I found was a list of the bordering regions of all the game's regions (including the "Bjoulsae River" region that isn't in the final game). It's used rather oddly in connection with the other regions that regions will go to war with, because the order doesn't match the region order from FACTION.TXT, so regions will go war with the border regions of what seem like unrelated regions. My best guess of what is supposed to happen is that regions go to war with their bordering regions, but due to some mix-up during development or something the index numbers for the data didn't match the final region numbers. I may revisit it later.

As usual the function is in the PlayerEntity. I'm fine with it being moved out to some other class if desired.

### DIFF
--- a/Assets/Scripts/API/FactionFile.cs
+++ b/Assets/Scripts/API/FactionFile.cs
@@ -652,8 +652,8 @@ namespace DaggerfallConnect.Arena2
             public int maxf;
             public int vam;
             public int rank;
-            public int randomValue;
-            public int randomPowerBonus;
+            public uint rulerNameSeed;
+            public int rulerPowerBonus;
             public int ptrToNextFactionAtSameHierarchyLevel;
             public int ptrToFirstChildFaction;
             public int ptrToParentFaction;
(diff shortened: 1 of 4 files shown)


Interkarma:
That's a crazy amount of work! Thank you for all of that. I'll read through and merge when I can.

It's such a small thing in the game to have that external sense of things happening elsewhere in the world. It never seemed to impact the player directly, but it adds so much flavour to hear about these events going on. I always felt in Daggerfall the game was bigger than myself, and this feature was part of creating that feeling.

Allofich:
I agree it's nice to have, it feels like a (very simple) turn-based strategy game running in the background. It would be fun (and perhaps useful for debugging or modding) if there was some way to visually see what's going on, although I'm not intending to try such a thing. It also does have a very small game impact. The power of regional factions vs. the merchants faction affects how prices tend to change, and the power of factions affects how much money they give you as reward for quests (when it's not explicitly defined in the quest file). But I'm not sure that there is much beyond that, besides rumors and sign-post messages.

I haven't looked at all at how rumors are being handled in DF Unity, but that might be the next thing to do here. I know @Nystul-the-Magician has been wanting to know where the rumors like those for new rulers come from, and this is the answer. (The "new ruler" stuff starts at line 1391 in PlayerEntity.cs.) I don't completely understand classic's rumors but I think we need to be able to supply at least the TEXT.RSC textID for the rumor and up to 2 factions (in this case just the faction that has had its ruler changed, but two for rumors about wars, alliances, etc.) to create the rumors.

PR #781, opened 2018-05-28
Finish Stendarr avoid death perk

ajrb:


Allofich:
Looks like that information on UESP came verbatim from the Daggerfall Chronicles. I don't know for 100% sure but I don't think this is perk is actually in classic. I haven't found any code in the damage or kill functions that intervenes and saves the player if they're going to die. I'm OK with this anyway, though, as to me the Daggerfall Chronicles seems closer to the intended experience than what was actually released, anyway.

One thing that it definitely seems worth adding from DF Chronicles are the "Racial Starting Statistics" from page 22. I don't know how that didn't get in the final game, when it's the other Elder Scrolls titles, including Arena.

ajrb:
Thanks Allofich.

PR #786, opened 2018-05-29
 implemented "Where am I?" question/answers,  special npc "where is" answer in same location

Nystul-the-Magician:
- implemented "Where am I?" question/answers 
- trimming castle names read from text.rsc when last character is '.' (change to Assets/Scripts/Utility/BuildingNames.cs - should be ok, but pls take a look - I did this because I think it is an improvement since palace name would often break sentences with this '.', discovery system discovers castles without problems after this change - nameplates no longer contain the '.' character as well)
- special npc "where is" answer about location/person topic if in location:
 -- for building asked about that is the same building where the npc talk partner is in: npc always knows he is in the very same building and gives special answer now
 -- for quest person "where is" answer if npcs knows about person there is also a specific answer if npc is in same building as person

### DIFF
--- a/Assets/Scripts/Utility/BuildingNames.cs
+++ b/Assets/Scripts/Utility/BuildingNames.cs
@@ -145,6 +145,8 @@ namespace DaggerfallWorkshop.Utility
                                 break;
                             }
                         }
+                        if (a[a.Length - 1] == '.')
+                            a = a.Remove(a.Length - 1); // remove character '.' from castle text record entry if it is last character
                     }
                     else
                     {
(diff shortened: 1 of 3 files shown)


Assets/Scripts/Utility/BuildingNames.cs
```diff
@@ -145,6 +145,8 @@ public static string GetName(int seed, DFLocation.BuildingTypes type, int factio
                                 break;
                             }
                         }
+                        if (a[a.Length - 1] == '.')
```
Interkarma: Could you instead use `a = a.TrimEnd(".")` here? That might look a little neater and be friendlier to string class. Result will be same string if does not in fullstop.

Interkarma:
This is really exciting! Looking forward to playing with this one.

Assets/Scripts/Utility/BuildingNames.cs
Nystul-the-Magician: sure ;)

PR #788, opened 2018-05-30
TextureReplacement: read scale from xml (MobilePersonBillboard)

TheLacus:
Allow custom scale for wandering npcs as discussed on forums.  Instead of using raycast as previosly said, i left position of parent gameobject with Motor as is, and moved child gameobject with billboard meshrender via transform. This doesn't affect small variations so behaviour with original graphics is unaltered.

### DIFF
--- a/Assets/Scripts/Game/Utility/PopulationManager.cs
+++ b/Assets/Scripts/Game/Utility/PopulationManager.cs
@@ -181,6 +181,11 @@ namespace DaggerfallWorkshop.Game.Utility
                     poolItem.scheduleEnable = false;
                     poolItem.npc.RandomiseNPC(GetEntityRace());
                     poolItem.npc.Motor.InitMotor();
+
+                    // Adjust billboard position for actual size
+                    Vector2 size = poolItem.npc.Billboard.GetBillboardSize();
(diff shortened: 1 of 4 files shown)


Interkarma:
It's awesome how clean these injections look now you've matured all the systems to support it.

Assets/Scripts/Game/Utility/PopulationManager.cs
```diff
@@ -200,6 +205,7 @@ void UpdateMobiles()
                     poolItem.npc.Motor.gameObject.SetActive(false);
                     poolItem.active = false;
                     poolItem.scheduleEnable = false;
+                    poolItem.npc.Billboard.transform.localPosition = Vector3.zero;
```
Interkarma: This has been found to cause an exception after fast travelling around a while (it triggered right on arrival at a new town). I'll add a quick null check here, but wanted to let you know in case some other action is also needed. :)

TheLacus: Thank you. I believe this happens if poolItem is marked for recycle before npc info and texture are set, which is also when position is changed. If this does't happens it shouldn't be needed to reset to parent position, so a null check should be enough.

PR #789, opened 2018-06-06
Extend start quest to load from quest packs

ajrb:
I've changed the start quest action to pick up quests in the following order

1) Any classic quests in /Quests folder.
2) Any init quests registered by quest list.
3) Any quests in same folder as an init quest.
4) Any guild quests registered by quest list.
5) Any social quests registered by quest list.

### DIFF
--- a/Assets/Scripts/Game/Questing/Actions/StartQuest.cs
+++ b/Assets/Scripts/Game/Questing/Actions/StartQuest.cs
@@ -1,4 +1,4 @@
-﻿// Project:         Daggerfall Tools For Unity
+// Project:         Daggerfall Tools For Unity
 // Copyright:       Copyright (C) 2009-2018 Daggerfall Workshop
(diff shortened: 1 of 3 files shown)


ajrb:
Wondering why the log line for parsing quest has two LFCRs at the beginning - makes it look like a blank log line in Unity console.

 `LogFormat("\r\n\r\nParsing quest {0}", questName);`

Interkarma:
> Wondering why the log line for parsing quest has two LFCRs at the beginning - makes it look like a blank log line in Unity console.

Nothing to wonder about, the intention was quite literally to put in white space to make the thing more readable. :)

Easy enough to change in console output (I can just TrimStart() the Debug.Log() copy of the output). As long as the white space remains in quest_log, I'm happy.

Interkarma:
Thank you for this! Looks fine on a read. I'll test also before releasing next builds. From there, the quest authors are most likely to pick up any problems.

PR #790, opened 2018-06-06
Refactor CalculateAttackDamage and make more formula overridable

ajrb:
New overridable formula:
    CalculateAttackDamage
    CalculateSuccessfulHit
    CalculateStruckBodyPart
    CalculateBackstabChance
    CalculateBackstabDamage


ajrb:
Since testing these changes is rather subjective (seemed fine to me and no errors) I would request that these code changes are reviewed in more detail than normal as there are no unit tests to catch any mistakes.

Interkarma:
Thank you again, I appreciate the assist with this. 👍 

Looks OK on a read. Will test and let you know if any problems I can find. As the same code is being called by default setup, it should be OK.

This is something @Allofich can probably give feedback from testing as well. He's very familiar with subjective feel of the formulas during gameplay.

PR #791, opened 2018-06-08
Near death experience backup

MeteoricDragon:
@Interkarma I haven't put the settings in yet.  I was wondering if you can playtest this in a separate branch and share your opinion on what parts should be toggle-able.

There's a flash, a flicker, and a slow strobe.  Perhaps a single checkbox for both flicker and strobe.  Or a checkbox for each?


Interkarma:
Happy to test. Are you still looking at creating this as a mod?

JayH2971:
I could test this if it were a mod. I can't build DFU branches from git.

MeteoricDragon:
@interkarma This code contains no directional hit indicators.  I was thinking you were referring to the directional indicators as the mod-appropriate content after our conversation in the forum.  I can learn how to change all of it to a mod if you wish.  My desire to implement the the flicker and slow strobe is because I felt while playing that warning the player about his/her health when its low makes the game more accessible and less frustrating to modern gamers.  For those immersed in the game could lose attention to their health.  And if this were a mod for the game, they may say "Why wasn't this in the game to begin with?" The flicker, and slow strobe would be toggleable pre-launch options. The single flash occurs anytime the player is hit and shouldn't have a toggle.

Interkarma:
I apologise for the confusion, I thought we were discussing the whole feature. It's fine though, I can see how this would help people new to the game so let's proceed. :)

I'm happy for a classic-like red flash to be on by default and without a toggle, provided it's very quick and basic like classic. Something like this has been in the game for years with my PlayerDamage component up until very recently anyway, so people should already be familiar with the idea by now. Everything on top of that is optional and a bonus to play.

I'll test and give you feedback soon. Cheers mate.

MeteoricDragon:
by the way, there is a bug that I know of.  When the player is killed, I can't seem to get any flash or flicker to occur.  Also, when first loading. the fade-in from black seems not perfect, as if it gets interrupted.

Interkarma:
I have played with this now and I can see the work you've put into it. However, a classic-like red flash when player damaged has been in the game for years now. It was only disabled briefly during the UI rendering refactor in last few experimental builds. It also works when player dies (flashes red then fades to black) with no additional changes required to HUD classes.

I'd still be happy to import the other features as an option for players, but I'd prefer something more light-weight. This implementation feels very heavy for such a simple feature (e.g. ShowPlayerDamage gets the job done in about a dozen or so lines of code). If you can clean it up to a single minimal component with no changes to other parts of game, I would be much happier.

But considering how much you wish to add to this, and the ideas you have for future work (directional damage indicators for example), I feel this really would be better suited as a mod package. The classic gameplay implementation (red flash) is already in place and this is just extending that in a preferential way. When building a mod, you have all the freedom you like to create this in a format yourself and others will enjoy without being constrained by the core.

I hope that makes sense and does not discourage you. I value the work you've added to the game so far. This one just needs a little more work before I can consider merging with core. Otherwise it would make an excellent mod you can put all the bells and whistles into as you wish. :)

MeteoricDragon:
Ok.  I will happily make that work.  I have a few more ideas to try.  I have so far deleted the FlickerFlash class (the one that makes the screen flash red) in favor of the original ShowPlayerDamage class.

PR #792, opened 2018-06-10
Load RMB blocks like classic, add "not a port town" popup

Allofich:
Changes RMB block loading to work in the same manner as classic. This shortens the code and removes the need for special cases for "CUST" blocks.

Along while doing this I identified 2 more bytes of the exterior data, including the one used for the message you get when you try to buy a ship in a location that is not by water.

### DIFF
--- a/Assets/Scripts/API/DFLocation.cs
+++ b/Assets/Scripts/API/DFLocation.cs
@@ -547,6 +547,15 @@ namespace DaggerfallConnect
             /// <summary>Unknown.</summary>
             public Byte[] Unknown2;
 
+            /// <summary>Unknown.</summary>
+            public Byte Unknown3;
+
+            /// <summary>ASCII value for first letter in .RMB file name.</summary>
+            public Byte Letter1ForRMBName;
+
+            /// <summary>If non-zero, ships can be purchased at banks here. Also has an unknown use.</summary>
+            public Byte PortTownAndUnknown;
+
             /// <summary>Only first Width*Height elements will have any meaning.</summary>
             public Byte[] BlockIndex;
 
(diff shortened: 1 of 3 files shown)


Interkarma:
This code goes way back to around 2002 when I wrote Daggerfall Cartographer. I believe there are a few locations where the incorrect block is selected, which this sounds like it could get resolved by this PR. As long as selection is equally or more correct, happy to have this one incoming. :)

I'll test in a branch and merge soon. Cheers!

Interkarma:
Yes! This seems to have fixed my problem with a desert-styled RMB block being selected for the middle of Wayrest/Chestertower. This seems to be correct now, but I just need to pass a few other systems as this one is fairly fundamental to world layout. Looking good though!

Interkarma:
I need to revert this I'm sorry. It breaks some locations (e.g. Direnni Tower). This needs more testing.

Allofich:
I found the fix for it, but I'm kind of confused how to open a pull request for it, because the merge/revert stuff of the previous PR is interfering.

The problem was that the value used for the index of the array of 2nd letters of the .RMB filename needed to all be done as if it's being done in the al register, so bits going beyond the scope of a byte should be chopped off.

See
https://github.com/Allofich/daggerfall-unity/commit/4cd100faeab58b1118f5ff708c3f38bede7ddb16

Tested with:
Direnni Tower
Scourg Barrow
Shedungent (These 3 custom dungeons were all loading wrong. Seems the problems happened with `CUST` blocks for unique dungeons)
Wayrest
Sentinel
Daggerfall
Privateer's Hold
Random dungeon exterior
Random town
Random temple town
Witches' coven in Daggerfall region
.. and they were all OK with the above fix.

Allofich:
This also fixes an additional "0" I had had to add to the end of the 2nd letter of the .RMB name. I had wondered about that because the original game data didn't seem to have a 0 there. I've removed it and also updated the comment to use <remarks> for the supplemental information.

Allofich:
OK, I figured out how to a PR for this. I just had to revert the revert. :) I'll open a new PR, then.

PR #793, opened 2018-06-11
Make death of player show red

MeteoricDragon:
Edits FadeBehavior.FadeHUDToBlack() to have fadeStartColor = a blood red color and fade that to black instead of clear to black.  I tested it in the game and it looks better than the inconsistent behavior that occurs that when you get hit and no flash of any sort occurs and it just fades to black.  Try it out.

### DIFF
--- a/Assets/Scripts/Game/FadeBehaviour.cs
+++ b/Assets/Scripts/Game/FadeBehaviour.cs
@@ -64,10 +64,11 @@ namespace DaggerfallWorkshop.Game
             if (fadeTargetPanel == null || !allowFade)
                 return;
 
-            fadeStartColor = Color.clear;
+            fadeStartColor = new Color(0.39f, 0, 0, 0.4f);
+            //fadeStartColor = Color.clear;
             fadeEndColor = Color.black;
             this.fadeDuration = fadeDuration;
-            fadeTargetPanel.BackgroundColor = Color.clear;
+            //fadeTargetPanel.BackgroundColor = Color.clear;
             fadeInProgress = true;
         }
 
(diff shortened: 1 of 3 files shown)


Assets/Scripts/Game/FadeBehaviour.cs
```diff
@@ -64,10 +64,11 @@ public void FadeHUDToBlack(float fadeDuration = 0.5f)
             if (fadeTargetPanel == null || !allowFade)
                 return;
 
-            fadeStartColor = Color.clear;
+            fadeStartColor = new Color(0.39f, 0, 0, 0.4f);
```
Interkarma: I'd like this change not to be made, thank you. The entire purpose of FadeHUDToBlack() is to fade from clear to black. This changes the purpose of the method and hardcodes a different start colour. It is used in many more places than just when player dies.

MeteoricDragon: I have an alternative solution, then.  But by the way, visual studio's "Find all references" when used on FadeHUDToBlack() yields only one occurrence in PlayerDeath.  But regardless it's best to keep this method unchanged in case other code uses it.

Regardless, The better solution is that I can create another method FadeHUDBloodToBlack() in FadeBehavior.cs and have PlayerDeath call that method instead.

Assets/Scripts/Game/PlayerDeath.cs
```diff
@@ -158,7 +158,7 @@ private void PlayerEntity_OnDeath(DaggerfallEntity entity)
             startCameraHeight = mainCamera.transform.localPosition.y;
             targetCameraHeight = playerController.height - (playerController.height * 1.25f);
             currentCameraHeight = startCameraHeight;
-            DaggerfallUI.Instance.FadeBehaviour.FadeHUDToBlack(FadeDuration);
+            DaggerfallUI.Instance.FadeBehaviour.FadeHUDBloodToBlack(FadeDuration);
```
Interkarma: That is better, thank you. :)

PR #796, opened 2018-06-11
TextureReplacement: Fix issue with emission maps on billboards

TheLacus:


Interkarma:
Thank you for this! :)

PR #797, opened 2018-06-12
Fix .RMB name resolving

Allofich:
Continued from https://github.com/Interkarma/daggerfall-unity/pull/792


Interkarma:
Awesome, thank for your continuing to fix this! I was very happy to see that it fixed some of my known problem areas at last, and it made into master before I discovered any problems. I apologise for the inconvenience caused by the reversion, I'm glad you were able to sort that out.

I'll get this back into master soon.

PR #798, opened 2018-06-13
Near death experience simplified

MeteoricDragon:
This code version is greatly simplified.  It removes the flash I created in favor of the flash you have. and simplifies the code so flicker code is done mostly inside flickerBase.  The FlickerFast and FlickerSlow just use different settings by initializing different values for the same fields inside Init().  Let me know what you think.


MeteoricDragon:
Also, This couldn't automatically merge so I went through and resolved all conflicts.  There still isn't a setting for these flickers yet.  That's because I want to get a feel for what you think before I implement it.  A mod version for this with directional indicators can come later. :)

ajrb:
Um, looks like your PR contains more than just your changes. Some kind of git mess? Just wanted to point that out in case you hadn't realised

TheLacus:
They are there due to merge. Try to rebase from upstream master :)

ajrb:
Yes I suspect it's due to not syncing your fork/branch. I feel your pain, I had allsorts of issues when I first started with github, even though I was familiar with basic cmd line git usage.

MeteoricDragon:
@ajrb @TheLacus @Interkarma  Actually, I think I have a cleaner solution with less commits and easier to follow than rebasing.  I have a .txt copy of all the files I modified.  I could just close this PR, begin a new branch from upstream master.  Open the files in the local master, replace the files with the modified ones. and just issue a new PR on the upstream master branch.  It would mean Interkarma would need to re-make the branch though.  But would be much cleaner because it would eliminate all the re-writes I did.

ajrb:
Yep, I did that a few times in the early days... labour intensive and you have to be really on the ball to ensure you don't miss anything (and test it all again) but it gives a clean PR.

Not saying there's a better way, but when I get in a mess I want to keep it simple even if it takes some effort.

Interkarma:
I'm OK with a new PR if that's the easiest solution for you. Just send it when ready and I'll delete the current branch and merge the new code to its own branch.

Thank you for continuing work on this one. :)

PR #799, opened 2018-06-13
Abstracts HealthChangeDetector

MeteoricDragon:
This commit takes health change detection code from CameraRecoiler and abstracts it to a new class "HealthChangeDetector" and makes it a component of playerAdvanced.  No Functionality changes.

This will allow easier modding for mods related to changes in Player Health, and other future functions.

Also, one additional Entity property is added to DaggerfallEntity, which allows shorthand way to get the current percentage of max health left of that entity.

This will also be used by the Near Death Experience code.  I figured it's better to make the code available for re-use rather than retype the same code in any new code that requires it.

### DIFF
--- a/Assets/Prefabs/Player/PlayerAdvanced.prefab
+++ b/Assets/Prefabs/Player/PlayerAdvanced.prefab
@@ -40,6 +40,7 @@ GameObject:
   - component: {fileID: 114101796857139080}
   - component: {fileID: 114220683863726122}
   - component: {fileID: 114033680799267548}
+  - component: {fileID: 114397629012878356}
   m_Layer: 0
   m_Name: PlayerAdvanced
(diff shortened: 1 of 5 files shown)


MeteoricDragon:
@Interkarma  the latest code doesn't allow the player to die when health reaches zero.  I can't imagine why my changes would do it.  they don't modify the health, they just read it.

Interkarma:
Update: I haven't been been able to reproduce this one yet. I've been able to die normally from monsters, guards, falling, stat drain, fatigue loss while monsters nearby, all the usual stuff. I'm not sure why this would be happening on your end either, but all seems OK here.

MeteoricDragon:
I closed Unity, reopened and now I'm dying.  I'm happy about that 🥇

PR #804, opened 2018-06-16
Implement classic reputation propagation

Allofich:
Implements reputation propagation based on reversing classic code. This is a response to http://forums.dfworkshop.net/viewtopic.php?f=19&t=910 but.

This does not yet 100% replicate classic behavior for quest rewards, just the propagation part. I used http://en.uesp.net/wiki/Daggerfall:The_Infested_House as a test case. In classic the quest reputation reward seems to be given based on the faction ID of the NPC quester, in this case faction 846 ("Questers"), the ID for the NPC in a knightly order who gives quests. DF Unity appears to instead be using the faction ID of the knightly order. This causes a different propagation and different results.

Would it be an easy fix to assign the faction ID of the NPC clicked on rather than of the faction itself? I don't know if that is how it works for all quests in classic but I would assume so for now unless there is evidence otherwise. If that change is made I think we will have behavior equivalent to classic.

@ajrb Regarding the thread discussion above.
I think the difference on UESP might be because in one case whoever wrote it accidentally compared the before + after results after doing 2 quests or something. At any rate, the reputation reward will vary depending on the faction relationships with the region and knightly order (as well as any allies or enemies they may have randomly added for that game).

### DIFF
--- a/Assets/Scripts/Game/Entities/PlayerEntity.cs
+++ b/Assets/Scripts/Game/Entities/PlayerEntity.cs
@@ -1693,7 +1693,7 @@ namespace DaggerfallWorkshop.Game.Entity
             FactionFile.FactionData peopleFaction;
             FactionData.FindFactionByTypeAndRegion(15, regionIndex + 1, out peopleFaction);
 
-            FactionData.ChangeReputation(peopleFaction.id, -(reputationLossPerCrime[(int)crimeCommitted] / 2));
+            FactionData.ChangeReputation(peopleFaction.id, -(reputationLossPerCrime[(int)crimeCommitted] / 2), true);
         }
 
         public void RaiseReputationForDoingSentence()
@@ -1705,7 +1705,7 @@ namespace DaggerfallWorkshop.Game.Entity
             FactionData.FindFactionByTypeAndRegion(15, regionIndex + 1, out peopleFaction);
 
             // Classic changes reputation here by (1 - halfOfLegalRepPlayerLostFromCrime) / 2). Probably a bug.
-            FactionData.ChangeReputation(peopleFaction.id, (halfOfLegalRepPlayerLostFromCrime - 1) / 2);
+            FactionData.ChangeReputation(peopleFaction.id, (halfOfLegalRepPlayerLostFromCrime - 1) / 2, true);
         }
 
         public bool SurrenderToCityGuards(bool voluntarySurrender)
--- a/Assets/Scripts/Game/Guilds/GuildManager.cs
+++ b/Assets/Scripts/Game/Guilds/GuildManager.cs
@@ -120,6 +120,14 @@ namespace DaggerfallWorkshop.Game.Guilds
             return memberships.ContainsKey(guildGroup);
         }
 
+        public bool GetJoinedGuildOfGuildGroup(FactionFile.GuildGroups guildGroup, out Guild value)
+        {
+            if (memberships.TryGetValue(guildGroup, out value))
+                return true;
+
+            return false;
+        }
+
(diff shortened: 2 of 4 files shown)


ajrb:
Hey @Allofich - changing the faction id given to the quest for reward should be easy, but I remember testing this in classic so let me re-check this first. Will get back to you this weekend.

Allofich:
The reputation change given at quest end (both the +5 for success and the -2 for failure) in classic is given to the factionID found at offset +2 of the QBN object (type 0x0e object).

I think the value comes from the NPC clicked on, except if the NPC has faction ID 240 ("Temple Missionaries"), in which case the factionID of the current building (i.e. the temple) is used. So the propagation code for Temple Missionaries might not ever actually be used. And it seems that if the first letter of the quest filename is "Y", (Mehrunes Dagon quest) then the factionID is set to 0.

ajrb:
The other issue about some of the UESP awards being +10, are you saying that you think this is a UESP mistake? I've not had time to try the save JH provided on forums yet - will do it later, but have to go out now to a 50th birthday thing of a friend.

Allofich:
I didn't think about trying to do a non-member quest for another temple or order than the one you already joined.

In classic, if you have joined a knightly order then go to a different knightly order, clicking "Join Guild" will do nothing but "Get Quest" will get a non-member quest. I haven't checked the reputation changes but I am pretty confident it will happen as you say, you would get a reputation bonus to the guild you had joined. Seems this is an edge case that classic doesn't handle.

I didn't quite understand your suggestion on what to do on the first reading, I'd need to give it a bit more close reading and probably look at the code. Honestly I don't think it's a priority to fix, though. I don't think there's any tangible benefit to raising your reputation with other orders, is there? It seems, though, like doing a non-member quest for an order _before_ you've joined any should give a reputation raise with that order, not just a double boost to the "Generic Knightly Order" as happens in classic. This is what the quest reward text implies, it's always something like "I'll put in a good word with my order." Do you know if the Generic Knightly Order faction reputation used for determining if you can join an order? Even if it is, though, it still seems like maybe that should be changed, to get a single bonus to the generic faction and one to the order you did a quest for, rather than two for the generic faction.

If you don't mind, I would prefer if you made any such change (after this is merged), as I'm sure you know your way around the DF Unity quest code much better than me.

ajrb:
> Anyway, if you don't mind, I would prefer if you made any such changes that you were suggesting (after this is merged), as I'm sure you know your way around the DF Unity quest code much better than me.

I don't mind at all, as long as you are happy with it. I will check JH save he posted now, and see if that validates your theory about the UESP being wrong. I had a suspicion from the start since there didn't seem to be any reason for it.

Interkarma:
This is really exciting you guys! Very happy to see the teamwork between you two (and I guess by extension all three of us with my earlier work on rep stuff in the mix). Will be great to have this in the questing mix, finally. :)

Allofich:
Hey Interkarma, I saw you tweeted out

> Thanks to Allofich and @GamerHazelnut, we finally have reputation propagation matched to classic Daggerfall. 

which is technically correct (hopefully, I haven't tested it very rigorously), but just to be sure you know, the resulting reputation changes are not the same as classic yet because of the above issue I mentioned with Hazelnut. The reputation in classic begins the propagation from the faction ID of the NPC that was clicked on to get the quest, while DF Unity is not doing this. (In the quest I tested it seemed to be using the faction ID of the knightly order itself as the quest faction, so propagation is starting from there).

Interkarma:
>  just to be sure you know, the resulting reputation changes are not the same as classic yet because of the above issue I mentioned with Hazelnut. 

I understand, but that's more of a tuning thing than anything wrong with your work or the function of propagation itself. Take a little credit for your efforts. :)

PR #805, opened 2018-06-16
Near death experience simplified

MeteoricDragon:
This version is greatly simplified and cleaner.  Also includes a setting in Advanced Settings window to toggle the effects off.

When the player is reduced to < 40% health.  they will experience a flashing red flicker effect.  If < 20% health, the player fades in and out of darkness until their health is raised above 20% threshold.


Interkarma:
Thank you for all your hard work on this one. I'll get this reviewed and merged as soon I can. :)

PR #806, opened 2018-06-17
NPCs don't stop if weapon drawn

MeteoricDragon:
Mobile Town NPCs now won't stop if your weapon is drawn.  

Also makes TODO note for other NPC stop conditions:

**Conditions to do in future:**
Player is in beast form.
Player just got caught committing a crime
Player is invisible.

### DIFF
--- a/Assets/Scripts/Game/Entities/PlayerEntity.cs
+++ b/Assets/Scripts/Game/Entities/PlayerEntity.cs
@@ -153,6 +153,8 @@ namespace DaggerfallWorkshop.Game.Entity
         public Crimes CrimeCommitted { get { return crimeCommitted; } set { crimeCommitted = value; } }
         public bool HaveShownSurrenderToGuardsDialogue { get { return haveShownSurrenderToGuardsDialogue; } set { haveShownSurrenderToGuardsDialogue = value; } }
         public bool Arrested { get { return arrested; } set { arrested = value; } }
+        public bool IsInBeastForm { get; set; }
+        public bool IsInvisible { get; set; }
 
         #endregion
 
--- a/Assets/Scripts/Game/MobilePersonMotor.cs
+++ b/Assets/Scripts/Game/MobilePersonMotor.cs
@@ -212,13 +212,19 @@ namespace DaggerfallWorkshop.Game
             bool withinIdleDistance = (distanceToPlayer < idleDistance);
             bool playerStandingStill = GameManager.Instance.PlayerMotor.IsStandingStill;
             bool sheathed = GameManager.Instance.WeaponManager.Sheathed;
+            //bool enemiesNearby = GameManager.Instance.AreEnemiesNearby();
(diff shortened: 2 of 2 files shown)


MeteoricDragon:
@Allofich do you know of a good way to set a variable that indicates that NPC's should ignore the player because of a crime that he/she just committed? Like a "public panic" mode that gets reset after a while?  PlayerEntity.CrimeCommitted != Crimes.None seems like an appropriate one but maybe not, because the crime only gets reset after Player goes to court.  The player can flee a town, come to a different town, and NPC's there would "panic" too if CrimeCommitted is used.

Allofich:
Maybe check whether there any hostile enemies nearby? `GameManager.Instance.AreEnemiesNearby()` 
If the player has been detected doing a crime then guards will have spawned so this should work.

MeteoricDragon:
@Allofich  ok, it works! thanks!  I found out that AreEnemiesNearby is an expensive method to call if it's being called on all MobilePersonMotor instances every frame.  Initially, I made all instances of MobilePersonMotor call it every frame, but now I made it so that only NPC's that are close enough, no weapon drawn, and you're not moving will call the method.

Interkarma:
>I found out that AreEnemiesNearby is an expensive method to call if it's being called on all MobilePersonMotor instances every frame.

I'm glad you caught this and found a workaround. I'm happy with this for now, as player won't be able to stand still very long with NPCs while guards are on their tail anyway. Either the NPC will start moving away or the guard will catch up to the player. Either way the expensive call is limited to a brief spurts.

There might be an even cheaper way though. I'll have a think about this and perhaps we can improve even more later. I think this will be fine for now though. :)

MeteoricDragon:
> There might be an even cheaper way though. I'll have a think about this and perhaps we can improve even more later. I think this will be fine for now though. :)

@Interkarma I have a possible solution.  If GameManager was to call `AreEnemiesNearby()` every frame and save it to a property of the same name.  And then reassign all `AreEnemiesNearby()` references to the property instead.  It would guarantee one call per frame and might be a better solution for mods needing to know if any enemies are nearby as well.

Interkarma:
> It would guarantee one call per frame and might be a better solution for mods needing to know if any enemies are nearby as well.

One call per frame is exactly what I want to avoid. :) This search was designed only to run when player opens rest or fast travel UI. Anything that searches scene for game objects is too expensive.

My preference is to build on your first idea above, and add some logic for an "alarm state" that begins when player commits a crime and ends when they leave the area. Shouldn't be too difficult, then all you have to check is a boolean flag.

Allofich:
I thought I remembered doing something like this, and I remembered. The `haveShownSurrenderToGuardsDialogue` gets reset to false whenever there are no guards nearby, using `HowManyEnemiesOfType()`. The call is made from within `Update()` in `PlayerEntity` but I didn't notice any performance hit from it. Maybe it's because it does not have a distance check in it, while `AreEnemiesNearby()` does.

PR #807, opened 2018-06-17
Potion maker & Item enchanter basic implementation

ajrb:
- Upgraded ItemListScroller component to be flexible for use beyond inventory UI window.
- Fixes column swap issue when scrolling to end of lists with odd numbers of items.
- Implemented the basic buttons, item scrolling and movement for both potion and item maker UI screens. (You can't actually create potions or enchanted items yet)
- These are now prepared for when the effect system is ready to enable their functionality.

I'm really happy how the ItemListScroller handles item display, once I upgraded it the potion screen was really easy to do.


Interkarma:
Very happy to have these underway. Looking forward to seeing them in action soon!

PR #808, opened 2018-06-18
Re-work classic DF rep propagation algorithm

ajrb:
- Simplify code, thanks to Allofich for implementing as classic and removing guesswork.
- Works from guild faction ids rather than questor npc id.
- Fixes classic bug with knightly orders where rep isn't propagated to children of generic order.
- Fixes classic bug with knightly orders where rep is doubled for generic order when performing non-member quests.
- Keeps classic behaviour of not propagating for a specific knightly order. I assume due to all affiliated nobles being aloof from such matters, but this is easily changed if we want.

### DIFF
--- a/Assets/StreamingAssets/Tables/QuestList-Classic.txt
+++ b/Assets/StreamingAssets/Tables/QuestList-Classic.txt
@@ -202,7 +202,7 @@ K0C30Y03, Merchants, N, 30,
 -P0B00L04, Vampires, M, 0,
 -P0B00L06, Vampires, M, 0,
 -P0B01L02, Vampires, M, 1,
--P0B10L07, Vampires, M, 10,\
+-P0B10L07, Vampires, M, 10,
 -P0B10L08, Vampires, M, 10, In vanilla DF P0B10L08.QBN file is missing but P0B1XL08.QBN is used instead.
 -P0B10L10, Vampires, M, 10,
 -P0B1XL08, Vampires, M, 1X, BROKEN. QBN file is a counterpart of P0B10L08.QRC
@@ -237,7 +237,7 @@ R0C20Y07, Nobility, N, 20,
 R0C20Y22, Nobility, N, 20,
 R0C30Y25, Nobility, N, 30,
 -R0C40Y23, Nobility, N, 40, BROKEN. QRC file is missing in DF+DFQFIX
--R0C4XY23, Nobility, N, 4X, Needs support for until _thing_ perform:
+-R0C4XY23, Nobility, N, 4X, Needs support for until _thing_ performed:
 R0C60Y24, Nobility, N, 60,
 
 -- Daedra Princes
--- a/Assets/StreamingAssets/Tables/QuestList-Classic.txt
+++ b/Assets/StreamingAssets/Tables/QuestList-Classic.txt
@@ -202,7 +202,7 @@ K0C30Y03, Merchants, N, 30,
 -P0B00L04, Vampires, M, 0,
 -P0B00L06, Vampires, M, 0,
 -P0B01L02, Vampires, M, 1,
--P0B10L07, Vampires, M, 10,
+-P0B10L07, Vampires, M, 10,\
 -P0B10L08, Vampires, M, 10, In vanilla DF P0B10L08.QBN file is missing but P0B1XL08.QBN is used instead.
 -P0B10L10, Vampires, M, 10,
 -P0B1XL08, Vampires, M, 1X, BROKEN. QBN file is a counterpart of P0B10L08.QRC
@@ -237,7 +237,7 @@ R0C20Y07, Nobility, N, 20,
 R0C20Y22, Nobility, N, 20,
 R0C30Y25, Nobility, N, 30,
 -R0C40Y23, Nobility, N, 40, BROKEN. QRC file is missing in DF+DFQFIX
-R0C4XY23, Nobility, N, 4X,
+-R0C4XY23, Nobility, N, 4X, Needs support for until _thing_ perform:
 R0C60Y24, Nobility, N, 60,
 
 -- Daedra Princes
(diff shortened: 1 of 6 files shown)


Interkarma:
Cheers mate! Absolutely loving the work you and @Allofich have done on this.

I'll get everything merged soon. This is my busiest couple of weeks in the year and everything is boiling over at the moment.

ajrb:
No rush, it can wait 'til you have the time to check it.

Allofich:
Hey ajrb, not really important, but

> Fixes classic bug with knightly orders where rep isn't propagated to children of generic order.

I wasn't aware of any such bug, and I thought that when I tested it classic did properly propagate this? Did your testing find otherwise?

One other thing -- how are we handling the faction ID for reputation rewards for noble quests? For example if you do a quest for King Gothryd or Queen Barenziah or someone, will their individual faction get a full reputation change as in classic?

ajrb:
Yes, when I tested it last night I got no propagation for generic order factions. I think anyway, this is just from memory - will double check later.

Good point about the social quests, is propagation supposed to happen for them - I assumed so, although looking at faction data these don't seem to have children or parents in most cases. Where they do, all affiliations will get half rep and the npc questor faction gets full rep.

Allofich:
> is propagation supposed to happen for them - I assumed so, although looking at faction data these don't seem to have children or parents in most cases.

I assume so, too. The "King Gothyrd" faction for example has Daggerfall as a parent so I believe if you do a quest for him you should get full rep for "Daggerfall" and "King Gothryd" and half-rep for all the various factions under "Daggerfall".

ajrb:
The last thing I am unsure about is the couple of long lists of rep changes for failed quests on the http://en.uesp.net/wiki/Daggerfall:Reputation page. Have you seen any evidence of such a thing in the classic code?

Allofich:
No I haven't. I wonder if maybe their comparison of reputation before and after (assuming that's how they did it) just happened to coincide with one of the regular calls to normalizing of reputations, hence all the +1 and -1.

I thought of another minor thing. In classic since propagation starts from the NPC clicked on, this also should mean that the allies and enemies considered are considered from the that NPC's faction. I think in the case of the Mages Guild it's the "Mercenary Mages" faction, which is allied to the "Master of Academia" and an enemy of "The Patricians", which I think are 2 other Mages Guild NPC factions. Maybe there was some inter-Mages Guild faction rivalries planned, but it doesn't seem like this would amount to much, as the positive effect from propagating down from the Mages Guild parent should cancel out any negatives due to enemies anyway. And outside of the inter-relationships between that group of people, it makes sense that allies/enemies of the Mages Guild itself would like you/dislike you for doing quests for them, so it may be better to do the enemy/ally thing from the parent faction, as it seems you are doing in this PR.

ajrb:
Added 3 minor bugfixes:
- Inns now free for members of knightly orders when fast travelling, as applicable
- Buy spells available to non-members of mages
- Comment out noble quest that has unimplemented syntax 'until _thing_ performed:'

Allofich:
If you're doing the free inns for knights thing, they also get free ship fast travel over water.

ajrb:
Thanks, I checked and I added that a while ago. Not sure how I missed free inns for fast travel before. Free inns is restricted to knightly orders, whereas free ship travel can be offered by a new guild in a mod.

ajrb:
> I thought of another minor thing. In classic since propagation starts from the NPC clicked on, this also should mean that the allies and enemies are considered from that NPC's faction. I think in the case of the Mages Guild it's the "Mercenary Mages" faction, which is allied to the "Master of Academia" and an enemy of "The Patricians", which I think are 2 other Mages Guild NPC factions. Maybe there was some inter-Mages Guild faction rivalries planned, but it doesn't seem like this would amount to much, as the positive effect from propagating down from the Mages Guild parent should cancel out any negatives due to enemies anyway. And outside of the inter-relationships between that group of people, it makes sense that allies/enemies of the Mages Guild itself would like you/dislike you for doing quests for them, so it may be better to do the enemy/ally thing from the parent faction, as it seems you are doing in this PR.

Thanks for the input, this would be easy enough to change by moving the ally changes before navigating up to the root faction. Having analysed the faction structure a little for guilds, I'm going to do this since it makes sense for temples & orders, is irrelavant to all other guilds except mages. Mages is an odd one, the allys are on the questor npc faction but testing for all the UESP listsed rep changes for failed quests showed that classic doesn't do anything with these ally/enemy links. They seem to be for other internal guild factions and as you say, the intention of rivalries never made it in the game.

This change makes no difference to noble/commoner/merchant factions from what I can see, since they generally don't have

ajrb:
Well I have truely debunked the long lists of rep propagation for failed MB and Noble quests - complete fiction. Noble quests do, however, give a -12/-6 for failure instead of default -2/-1 but the success values are the default 5/2... don't fail the toffs! :-)

Due to this, and the fact that noble quests affect the root parent faction (i.e. region) by full amount, I have added another condition for full amount.

Assets/StreamingAssets/Tables/QuestList-Classic.txt
```diff
@@ -237,7 +237,7 @@ R0C20Y07, Nobility, N, 20,
 R0C20Y22, Nobility, N, 20,
 R0C30Y25, Nobility, N, 30,
 -R0C40Y23, Nobility, N, 40, BROKEN. QRC file is missing in DF+DFQFIX
-R0C4XY23, Nobility, N, 4X,
+-R0C4XY23, Nobility, N, 4X, Needs support for until _thing_ performed:
```
Interkarma: This is already a supported task type (see line 235 of Task.cs). I'll add this quest to my internal tracker to take a look at later, as it doesn't sound like this quest is executing properly.

Interkarma:
Thanks again for all your work on this. 👍

Assets/StreamingAssets/Tables/QuestList-Classic.txt
ajrb: Ah, silly me, I simply checked the action classes and found none called until. Can't recall what led me to look at it now.

PR #809, opened 2018-06-18
Sorceror Magicka restrictions and mages guild benefit.

MeteoricDragon:
If the player has the NoRegenSpellPoints career property set to true:

->Spell points no longer recover when resting.
->Rest until fully healed doesn't require full magicka.
->The player can recharge spell points at mages guild by clicking any service like in classic.

@Nystul-the-Magician The TextProvider ID is needed for following messages so they can be used in TextProvider.GetRandomTokens.  Do you know the ID?

-> I see your magicka is low. You sorcerors really need to be more careful. Here, let me recharge you.
-> It is customary for the guild to recharge the magicka of sorcerors.

msgbox.SetText is used as a makeshift solution until the GetRandomTokens approach is used.

### DIFF
--- a/Assets/Scripts/Game/Entities/PlayerEntity.cs
+++ b/Assets/Scripts/Game/Entities/PlayerEntity.cs
@@ -1783,7 +1783,7 @@ namespace DaggerfallWorkshop.Game.Entity
                 DaggerfallUnity.Instance.WorldTime.Now.RaiseTime(1 * DaggerfallDateTime.SecondsPerHour);
                 int healthRecoveryRate = FormulaHelper.CalculateHealthRecoveryRate(this);
                 int fatigueRecoveryRate = FormulaHelper.CalculateFatigueRecoveryRate(MaxFatigue);
-                int spellPointRecoveryRate = FormulaHelper.CalculateSpellPointRecoveryRate(MaxMagicka);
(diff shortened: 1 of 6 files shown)


ajrb:
Hi, I can supply the right ids for those messages. Unfortunately this is not the way I intended to implement the mage guild functionality. I want all guild benefits to be controlled by the guild classes, allowing full modding etc.

Would you mind if I implement the commit "Mages Guild recharges NoRegenSpellPoints" in the way I intended (just didn't because mana wasn't in use back then) please?

This PR can be merged and then I will update it - just want to be sure you're okay with that first MD.

MeteoricDragon:
That's fine.  In fact I think I know where you want the mage guild functionality.

MeteoricDragon:
@ajrb To make your update for you easier rather than leave a mess, I've modified the PR with a new commit based on what you described in your comment above.  Feel free to make further adjustments.

ajrb:
That's exactly it, thanks. The id you need is 465, it has several variants for GetRandomTokens to select from.

ajrb:
Great, this is good to go now.

ajrb:
Actually, there is one more change required - according to UESP you have to be a member of mages guild to get recharged. See comment on code.

PR #810, opened 2018-06-21
Horse Riding sound global fix

MeteoricDragon:
Horse riding sound now scales with global volume.

@Interkarma  do you know where I can find the command that plays audio for monsters?  I found the hit sound effects in EnemySounds but am looking for the code that plays the sounds that you hear the monster randomly making every so often.  Such as squeaks for rats, screams for skeletons, grunts for giants.

### DIFF
--- a/Assets/Scripts/Game/TransportManager.cs
+++ b/Assets/Scripts/Game/TransportManager.cs
@@ -86,7 +86,7 @@ namespace DaggerfallWorkshop.Game
             ridingAudioSource.loop = false;
             ridingAudioSource.dopplerLevel = 0f;
             ridingAudioSource.spatialBlend = 0f;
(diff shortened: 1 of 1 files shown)


Interkarma:
The bark sounds (squeak, screams, grunts, etc.) are fired from EnemySounds.cs at line 196.

I can see you've done a pass over this before, but it doesn't look like global volume scale is being applied yet.

MeteoricDragon:
I see, thanks.  I don't know how I missed that.

I think I got _everything_ now.  The enemies sounds, plus now burning torches will scale volume when you close the DaggerfallPauseOptionsWindow, instead of needing to load a save game to apply changes.

Well, technically, the movies don't scale in volume.  I don't know where the volume for those would be set anyhow, or if it's necessary to scale their sound.

PR #813, opened 2018-07-01
Import discovered locations from classic saves

Allofich:
This implements reading `MAPSAVE.SAV` from classic saves and discovering the locations marked within it as discovered.

Although it works correctly, unfortunately it is extremely slow and will take around 2 or 3 minutes to load a classic save. Interkarma, I'm hoping you will know where the bottleneck is.


Allofich:
I found out the cause of the delay and was able to optimize around it. There isn't any significant delay now.

The delay was caused by calling `DaggerfallWorkshop.DaggerfallUnity.Instance.ContentReader.MapFileReader.GetLocation` for each of the examined locations. I only needed that to get the location name to pass to `DiscoverLocation()`, so I rewrote the code to only call `GetLocation()` if needed.

Interkarma:
Awesome work!

PR #814, opened 2018-07-03
Height changer adjustments and partial swim implementation

MeteoricDragon:
Go ahead and give a try.

Includes integration with swimming speed, skill usage, skill tally.

The horse can be mounted/unmounted while swimming. Slight boost to height when mounted.

Known bugs:

When loading a save game, the height is sometimes wrong but can be fixed by crouching once or twice.

Not done yet:

- No footstep change for swimming yet. 
- Will adjust headbobber to bob differently for swimming.  
- Need to adjust EnemyMotor to keep non-flying enemies from going on water tiles.

### DIFF
--- a/Assets/Scripts/Game/Player/PlayerHeightChanger.cs
+++ b/Assets/Scripts/Game/Player/PlayerHeightChanger.cs
@@ -25,6 +25,7 @@ namespace DaggerfallWorkshop.Game
     [RequireComponent(typeof(PlayerMotor))]
     [RequireComponent(typeof(CharacterController))]
     [RequireComponent(typeof(HeadBobber))]
+    [RequireComponent(typeof(LevitateMotor))]
     public class PlayerHeightChanger : MonoBehaviour
     {
         private HeightChangeAction heightAction;
@@ -44,6 +45,7 @@ namespace DaggerfallWorkshop.Game
         private PlayerMotor playerMotor;
         private CharacterController controller;
         private HeadBobber headBobber;
+        private LevitateMotor levitateMotor;
         private Camera mainCamera;
         private float controllerStandHeight = 1.78f;
         private float controllerCrouchHeight = 0.45f;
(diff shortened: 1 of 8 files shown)


Assets/Scripts/Game/Player/PlayerHeightChanger.cs
```diff
@@ -109,148 +148,300 @@ private void Update()
...
+            if (!OnWater)
+            { 
+                if (dismounting)
+                    eyeChange = -1 * eyeHeight;
+                else if (mounting)
+                    eyeChange = eyeHeight;
             }
+            controller.transform.position += new Vector3(0, heightChange / 2 + eyeChange);
+            //Debug.LogFormat("IsGrounded = {0}, \nHeight = {1}", playerMotor.IsGrounded, controller.height);
+
```
Allofich: I think you should remove this commented-out debug statement.

Allofich:
I gave it a quick test, seemed to work fine.

Just a few notes:
- The swim speed in classic is indeed the same for swimming outside and in dungeons, so no problem there.
- In classic you don't get any skill tally for swimming outdoors. I can maybe see a justification for that (deep-diving in dungeons being a different skill than just paddling around at the surface outdoors) but I think it's fine to get skill increases for outdoor swimming. I think that's the expected behavior, it's what I would expect anyway.

MeteoricDragon:
I can't seem to find the water paddling sound.  I thought it might be footstepShallow or footstepSubmerged but it sounds like a foot striking ground instead.

MeteoricDragon:
Is there a convenient way to have an EnemyTileMapIndex to prevent enemies from swimming in water?

Interkarma:
I think this is fine to merge straight to master. Cheers for all the work! 👍 

It will probably be another couple of weeks until I release new builds, so plenty of time to test and clean up. I'll let you know any feedback I can once I start playing with the changes properly.

PR #816, opened 2018-07-07
Implement illusion effect behavior and adjust stealth detection

Allofich:
This implements enemy AI behavior for when the player has the Invisibility, Shade or Chameleon effects active, in a manner similar to classic.

Spells for these, etc. are not added by this. But you can test the behavior by setting the flags `IsInvisible` `IsAShade` or `IsBlending` in `PlayerEntity`. The effects are dispelled if you do damage to an enemy, the same as in classic.

Invisibility should work identically to classic (except for the "giveUpTimer" as I explain below). For shade and chameleon, though, the behavior differs. This is because, if I understand correctly, these effects are pretty much useless in classic. This is because as soon as an enemy has detected the player during one of its constant updates, they then see the player for the remainder of their "giveUpTimer" (as it's being called in `EnemyMotor.cs`). When the timer runs out, because of how frequently AI updates happen they are almost sure to immediately detect the player again and set the timer again, allowing them to see the player for another run of the timer. Maybe classic's behavior was created with a slower AI update rate in mind? I tried a shade effect in classic and indeed it didn't seem to do anything to stop an enemy from seeing me.

So for this PR, instead Shade and Chameleon effects are rolled against every classic AI update, so they can see you for a moment, then they can't, then they can, etc. I'm pretty happy with the behavior, it gives a feeling that they are searching for you but can't see you well.

Some enemy types ignore these effects and can always see the player. This is recreated from classic.

Also the enemy detection was refactored some. and enemies can now get the player's location from the player failing a stealth roll. I think this also happens in classic, and it recreates enemies coming through doors for you, etc. as they do in that game, and generally makes them feel a little more alive since you won't always find them in their starting spots. I had previously not added this because I didn't want to get rid of DF Unity's superior pursuit behavior where AI will go to the last spot they saw the player rather than try to walk through walls, but I added a timer to account for this. If the enemy recently had LOS with the player they will not update the location from the stealth check (i.e. they will go to the last spot they saw the player and not try to walk through a wall).

### DIFF
--- a/Assets/Scripts/DaggerfallUnityStructs.cs
+++ b/Assets/Scripts/DaggerfallUnityStructs.cs
@@ -206,7 +206,8 @@ namespace DaggerfallWorkshop
         public string LootTableKey;                 // Key to use when generating loot
         public int HitFrame;                        // Frame of attack animation at which hit on target is attempted
         public int Weight;                          // Weight of this enemy. Affects chance of being knocked back by a hit.
-        public bool CastsMagic;                      // Whether this enemy casts magic. Only used for enemy classes.
+        public bool CastsMagic;                     // Whether this enemy casts magic. Only used for enemy classes.
+        public bool SeesThroughInvisibility;        // Whether this enemy sees through the shade, chameleon and invisibility effects.
     }
 
     /// <summary>
--- a/Assets/Scripts/Game/EnemyMotor.cs
+++ b/Assets/Scripts/Game/EnemyMotor.cs
@@ -178,7 +178,7 @@ namespace DaggerfallWorkshop.Game
 
         private void Move()
         {
-            // Cancel movement and animations if paralyzed, but still allow gavity to take effect
+            // Cancel movement and animations if paralyzed, but still allow gravity to take effect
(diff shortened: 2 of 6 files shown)


Interkarma:
This looks great thank you! :) I'm impressed you've managed to work those behaviours into the existing AI setup, and I'm looking forward to playing with the tuned shade/chameleon effects.

As soon as I'm clear from work commitments (a few days), I'll test, merge, and get the spell effects in so we can wire all of this up for the player in next builds.

Interkarma:
@Allofich - I'm about to start work on the front-end magic effects for these. As part of this process, I'll need to implement effects for Invisibility-Normal, Invisibility-True, and so on.

I'm collapsing the bool fields (isInvisible, isBlending, isAShade) into a single EntityVisibility enum with values like (Normal, InvisibleNormal, InvisibleTrue, etc.) In the weapon checks, I won't dispel if player is rocking one of the "true" forms, as attacks shouldn't dispel these based on effect description.

Are you OK with all this?

Allofich:
I'm OK with it!

PR #818, opened 2018-07-10
Fix Crushing bugs

MeteoricDragon:
players can wedge themselves into a slanted wall and climb up and kill themselves by triggering the overhead player crush check.  This fixes it by doing the folllowing:

Disable player crouching while climbing.
Disable Climbing while player crouching

### DIFF
--- a/Assets/Scripts/Game/Player/PlayerHeightChanger.cs
+++ b/Assets/Scripts/Game/Player/PlayerHeightChanger.cs
@@ -119,7 +119,7 @@ namespace DaggerfallWorkshop.Game
                 heightAction = HeightChangeAction.DoDismounting;
                 toggleRiding = false;
             }
-            else if (!playerMotor.IsRiding)
+            else if (!playerMotor.IsRiding && !playerMotor.IsCrouching)
             {
                 // Toggle crouching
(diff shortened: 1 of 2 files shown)


Interkarma:
Fair compromise. We'll see if anyone complains about not being able to climb while crouched and can iterate on both climbing and crushing in future.

Assets/Scripts/Game/Player/PlayerHeightChanger.cs
```diff
@@ -119,7 +119,7 @@ public void DecideHeightAction()
                 heightAction = HeightChangeAction.DoDismounting;
                 toggleRiding = false;
             }
-            else if (!playerMotor.IsRiding)
+            else if (!playerMotor.IsRiding && !playerMotor.IsCrouching)
```
Interkarma: I've noticed player can no longer stand back up after this change. Once they crouch, they're stuck crouching. This happens because the condition only allows crouch toggle when player is not already crouch. I think the intention was to prevent crouch toggle while climbing, rather than when already crouched?

I've reverted this change now so player can toggle crouch again. At this time, they will simply fall if trying to change stance while climbing.

PR #819, opened 2018-07-13
Bug fixes

ajrb:
* One of the repair messages showing "%pct[srcDataUnknown]" instead of name.
* Stack weight checking only for 1 item instead of whole stack


Interkarma:
Thank you Hazelnut!

PR #820, opened 2018-07-14
Move env log to persistent data path

lucianposton:
This will enable logs to be written even when the game dir is installed
in a system dir.


Interkarma:
Cheers, that's a decent place for it, especially now that output log and quest logs are going there as well. Almost everything is in the one place now.

PR #821, opened 2018-07-14
Import and handle expiring items

Allofich:
This does a few things:

1. Identifies and imports time-limit data from items created with the "Create Item" spell in classic.
2. Items with a time limit will be removed when the time limit is reached. Note this doesn't happen in classic (items created with "Create Item" never disappear, despite the fact that you can set duration when creating a Create Item spell) due to a mismatch in the flag set and the flag checked. (Byte 43 of the itemRecord is ORed with 0x10, but what is checked for 0x10 is byte 21 of the base record. If you manually set that classic then will remove the item when the duration runs out.)
3. Fixed a bug with unequipping broken weapons. It relied on `GetEquipSlot()`, but for weapons that can be equipped in either slot (one-handed weapons), it would return the unequipped hand, as it was apparently written specifically for when looking for a free slot to equip to, not to check the slot for an already equipped weapon. I think I didn't realize this when I wrote the weapon breaking code. This is fixed now and the fixed way is also used for unequipping expiring items made with Create Item.

### DIFF
--- a/Assets/Scripts/API/Save/SaveTree.cs
+++ b/Assets/Scripts/API/Save/SaveTree.cs
@@ -354,6 +354,7 @@ namespace DaggerfallConnect.Save
         public UInt32 RecordID;                 // Unique ID of this record. Called "mapID" and "map identifier" by Fixsave.
         public Byte QuestID;                    // Associated quest ID of this record (0 if no quest)
         public UInt32 ParentRecordID;           // ID of parent record
+        public UInt32 Time;                     // Time in game minutes. Known uses: Time for magically-created items to disappear, time for items under repair to be done being repaired.
         public UInt32 ItemObject;               // chunktcl's description: ItemObject. Active spell/spell book/permanent treasure container
         public UInt32 QuestObjectID;            // chunktcl's description: QuestObjectID
         public UInt32 NextObject;               // chunktcl's description: Link to next object in series
--- a/Assets/Scripts/API/Save/SaveTreeBaseRecord.cs
+++ b/Assets/Scripts/API/Save/SaveTreeBaseRecord.cs
@@ -248,8 +248,11 @@ namespace DaggerfallConnect.Save
             // ParentRecordID
             recordRoot.ParentRecordID = reader.ReadUInt32();
 
+            // Time
+            reader.BaseStream.Position = 43;
+            recordRoot.Time = reader.ReadUInt32();
+
             // ItemObject
-            reader.BaseStream.Position = 47;
             recordRoot.ItemObject = reader.ReadUInt32();
 
             // QuestObjectID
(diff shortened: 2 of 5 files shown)


ajrb:
Good stuff there mate. Do these temporary items have weight, or are they weightless?

Allofich:
They do have weight.

I noticed one edge case, that these items will not disappear if they dropped into a treasure pile. (Until the player tries to add them to their inventory again) I think the same thing would happen in classic.

Allofich:
As for "Create Item" itself, the menu of items is a static list of strings, at 0x1ACAFE in FALL.EXE (Leather Cuirass through Robes). In typical Daggerfall fashion, the spell doesn't work like it seems like it should. All of the armor selections seem to be the same as far as the material of the item produced, i.e. It's the same whether you pick "Leather Cuirass", "Chain Cuirass" or "Steel Cuirass," you can potentially get any of the 3 whichever you select. As far as I can tell the material selection is the same as when an item is spawned in loot or on an enemy, so you'll usually get leather for armor items, and the material of metals is dependent on the player. (Although I seem to practically always get "Steel" like the menu suggests. I don't know if that was just a coincidence. But when I hacked my character level to 99 I did get a Daedric weapon)

The description of Create Item says

```
Item lasts for _ round(s), plus _ round(s) for every _ level(s) of the caster.
```

and this is correct, with "round" meaning a minute of game time. Seems pretty useless if the items are made to actually disappear as they were supposed to, instead of lasting forever like in classic.

Interkarma:
This is cool! Very timely as well, as I was looking at starting work on Create Item shortly.

Allofich:
Yeah, although I'm not sure what the best way to implement it is.

One way is to do it the same as classic (never expire, and you get level-based metal types). That way the effect is somewhat useful, but the in-game description of the effect is inaccurate, and players who don't know this will be shooting themselves in the foot by trying to add duration that will just make the spell more expensive for no gain. Also the materials of items in the summon menu will be inaccurate.

Another way is to implement it the way it seems like it was supposed to work. They would expire like in the effect description and item materials would be created according to the menu (so, nothing beyond Steel). But then the effect will be essentially useless.

The other option is to retool it. For example, make the "rounds" of duration be game hours, or maybe even game days (probably hours would be better), rather than game minutes. Maybe make the item spawned always (or often) have a magical enchantment. Then there would be a use for creating a piece of leather armor or the robes, etc. The list of items in classic (divided into leather, chain and plate armor, and different weapon types) was probably made with different class forbidden equipment restrictions in mind, so I think it makes sense to use a list like that.

One more thing, in classic if you cast a Create Item spell and try to cast it again immediately afterwards, it won't do anything. You have to wait a while (I don't know how long, I used rest/fast travel). I found a comment on-line that the duration of the spell is how long you can't recast. If that's true, then I guess if the item expiring was working that would mean you could only have one created item at a time.

Allofich:
Since we don't have enchantments yet anyway, maybe a good middle-ground is:

1) Make the rounds of duration be for hours rather than minutes, then expire
2) Keep the original list of items, but remove "Steel" from the front of the descriptions of weapons and plate armor, and let them give a material according to player level as in classic.

Allofich:
Any way you choose to do it is fine with me!

PR #822, opened 2018-07-17
Vitals loss indicator

MeteoricDragon:
@Interkarma please review and let me know what you think of this code.

The goal is to display feedback indicators that show how much health, fatigue, or magicka was lost in an action. I think it may be useful for new players to understand what is happening to their vitals.


Interkarma:
I've played with this and I really like it! The result feels very natural and creates a nice link between all the other feedback like head rocking and health flash with the vitals decreasing. Good work. :)

It would be great to have a toggle for this, just for the classic purists, but I'm happy for it to be on by default. It's a subtle enough change that I doubt many would object, especially if they can turn it off.

I'll play with this a bit more and let you know if any other feedback.

PR #824, opened 2018-07-18
Gains in vitals bars added

MeteoricDragon:
Added bars to serve as gains indicators.  Renamed VerticalProgressIndicator to VerticalProgressSmoother, because that class is used to extend a VerticalProgress object so it can smoothly update its amount if needed.  Gains and loss bars work independently and simultaneously.

In the code you'll notice that the vitals gains bars are just VerticalProgress bars.  That's because they never need to be smoothed.  The vitals base bars needed to be changed to VerticalProgressSmoother.

Have not added a setting quite yet.  Will do that soon.


MeteoricDragon:
Ok, I think everything's done.  It's got loss and gain indicators.  and the load game function seems to update the vitals fine.  and the setting in advanced game settings is working too.

Interkarma:
Cheers! Will get this in shortly and test. 👍

PR #825, opened 2018-07-19
Identify classic magic item data

Allofich:
- Identifies/reads data from `MAGIC.DEF` that was unknown
- Organizes classic enchantments like classic, as two values, one that identifies the type of enchantment (on use, on equip, on hit, special artifact effects, etc.) and one that serves as a parameter for that type of enchantment (an spell ID for use with `SPELLS.STD`, an identifier for the particular type of special artifact effect, etc.)

Classic enchantments were being stored as an array of integers. The two parts of each effect were also being saved separately as two different integers. To keep compatibility with existing DF Unity saves they are still saved that way, and converted as necessary. I think it's all working OK. I had to add a check that the effect type is 65535 (0xFFFF) when the item power popup because those values have been read and saved as unsigned so far, but with this PR I'm reading 0xFFFF from `MAGIC.DEF` or magic items in classic saves -1 now to fit in nicely with an enum for effect types.


Interkarma:
This is great work, thank you! With magic items coming up so quickly in the pipe, this kind of PR is really valuable. 👍

Assets/Scripts/Game/UserInterfaceWindows/HardStrings.cs
```diff
@@ -251,6 +251,66 @@ public class HardStrings
         public const string cannotTravelWithEnemiesNearby = "You cannot travel with enemies nearby.";
 
         public const string powersUnknown = "Powers unknown.";
+
+        public static readonly string[] itemPowers = { "Cast when used:",
```
Interkarma: At some point, I need to help with a better way to store this text outside of HardStrings. My text tables are only a part-measure for now and have a few limitations (e.g. one text result per key, commas break table schema in text). This is fine here for now though, it can be moved to a new text setup later once that exists.

Interkarma:
This all seems to work perfectly. It really helps bridge the gap between magic item data and the effect system as well. Thank you again for your work on this, magic items have just taken a big step forwards.

PR #826, opened 2018-07-22
Setup initial class creation window

numidium:
Not much functionality here, just setting up the navigation and background for the custom class creator. This is mainly to stake my claim, as it were, on this feature since it's what I'm focusing on currently. Others are welcome to add to it, of course.

I commented out the "Custom" option on the class select since there's no functionality on the window as of yet. You can test it out by uncommenting line 62 of CreateCharClassSelect.cs.

### DIFF
--- a/Assets/Scripts/Game/UserInterface/StatsRollout.cs
+++ b/Assets/Scripts/Game/UserInterface/StatsRollout.cs
@@ -67,11 +67,13 @@ namespace DaggerfallWorkshop.Game.UserInterface
             set { SetStats(startingStats, workingStats, value); }
         }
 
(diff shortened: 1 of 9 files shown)


numidium:
The nativeImageName should be "CUST00I0.IMG" instead of "CHAR02I0.IMG." Fixing that in the next commit.

Interkarma:
Thanks for making a start on this! 👍

Interkarma:
It looks like you're making amazing progress with this, thank you! I'm going to merge it into a branch on my end matching yours so I can test. I'm looking forward to trying out what you have so far.

PR #829, opened 2018-07-28
Implement creation of magic items in loot

Allofich:
Creation of magic items in loot (treasure piles, enemy equipment, house containers). Everything should be the same as classic except for the gold value of the items, which I still need to figure out (they'll just have the value of the non-magic versions for now). The magic item stats use the data read from `MAGIC.DEF`.

Also still need to do stocking of magic items by magic item merchants. (would be a separate PR)

### DIFF
--- a/Assets/Scripts/Game/Items/ItemBuilder.cs
+++ b/Assets/Scripts/Game/Items/ItemBuilder.cs
@@ -10,9 +10,12 @@
 //
 
 using System;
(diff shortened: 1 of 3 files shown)


Interkarma:
This looks good! I'm about to make another big push on rolling out more classic effects, then I can help with rolling out magic items further. This is really exciting for the project. Cheers! 👍

PR #831, opened 2018-08-01
Fix ArgumentOutOfRangeException

lucianposton:
*Templatize BaseImageFile.Load*

- Templatize BaseImageFile.Load to encapsulate common behavior in base.
- Expose result of Load as property to clients. There are better ways to
  expose the result to clients, but it would require uncoupling a bunch
  of code not worth touching for the sake of fixing the exception in the
  next commit.

*Fix ArgumentOutOfRangeException when arena2 dir unset*

Note, it'd be better if arena2 resources were loaded after arena2 dir is
set by the player (including LoadSpellIcons specifically for this
exception), but I'll leave that as an excercise for someone more
familiar with how arena2 resources are loaded.

### DIFF
--- a/Assets/Scripts/API/BaseImageFile.cs
+++ b/Assets/Scripts/API/BaseImageFile.cs
@@ -38,6 +38,11 @@ namespace DaggerfallConnect.Arena2
         /// </summary>
         internal FileProxy managedFile = new FileProxy();
 
+        /// <summary>
+        /// Indicates whether file has been loaded.
+        /// </summary>
+        private bool isLoaded = false;
+
         #endregion
 
         #region Class Structures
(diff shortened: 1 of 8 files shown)


Interkarma:
Hello and welcome! :)

> Note, it'd be better if arena2 resources were loaded after arena2 dir is
set by the player (including LoadSpellIcons specifically for this
exception)

This should not happen based on how the the startup process operates. I've not been able to reproduce the error this is fixing in live game.

I'd like to look at the root cause in more detail. Are you able to give me a process to reproduce the ArgumentOutOfRangeException before player has set Arena2 path? For example, are you starting game from DaggerfallUnityStartup (where path is acquired for player) scene or playing DaggerfallUnityGame scene directly in editor?

Cheers!

Interkarma:
Thank you for identifying this issue and steps to reproduce. I've decided to handle in a slightly different way, but have credited you in fix as well. Thank you! 👍

PR #832, opened 2018-08-02
Swimming bug fixes

MeteoricDragon:
This one is a really important fix: Fixes bug introduced by the indicator bars.  Because the property call in DaggerfallHUD for updating the breath bar was discarded in favor of updating the breath bar in the vitals class (to be consistent with the other vitals updating in the class), the SetRemainingBreath method was overlooked, which does more than set the amount, it sets the color of the bar according to the percent left. **Which causes the breath bar not to appear.**  Now the code properly sets the color and the breath bar appears.

Also Fixes bug identified by @Allofich , namely, if you load a game from swimming in a dungeon, you continue at swim speed.  Now the code checks to make sure the player is not on a water tile before clearing the swimming flag.

Also refactors code in the HUDVitals class to make a little more sense.

update: fixes to the vitals bars prevent the bars from displaying more than is actually tallied by the playerEntity.  could be caused when the player gets struck immediately after healing.

Placed the loss indicators behind the gain indicators.  Looks better if healing immediately after taking damage.  and can see both indicators working simultaneously.

Multiple consecutive healing or damage actions will no longer reset the timer or add delay for the smoothing.  It will just update the target amount to smooth the bar to.

### DIFF
--- a/Assets/Scripts/Game/PlayerEnterExit.cs
+++ b/Assets/Scripts/Game/PlayerEnterExit.cs
@@ -308,8 +308,9 @@ namespace DaggerfallWorkshop.Game
             else
             {
                 // Clear flags when not in a dungeon
-                // commenting this out allows player to swim outside - MeteoricDragon
-                //isPlayerSwimming = false;
+                // don't clear swimming if we're outside on a water tile - MeteoricDragon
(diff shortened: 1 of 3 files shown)


Interkarma:
Thank you for continuing to work on this!

PR #833, opened 2018-08-04
Implement magic item gold value calculation

Allofich:
This PR calculates and assigns gold values to magic items in loot. Values should match classic.

This was kind of complicated because it also involved reversing the formula for how spell costs are calculated.

A problem here Interkarma is that I saw you already have code in place for calculating spell costs.

I looked at using your existing function but it didn't seem that it was going to be straightforward to do so, as it is built into your system and the data representations you are using for spells. Sorry I didn't do it that way but I didn't know how to convert over when the only data I had was a SPELLS.STD index (that's the only data specified by spell enchantments on items). Also I wanted to figure out how classic was doing things.

The stuff I added is self-contained and won't interfere with anything you've done, so it shouldn't harm anything by merging, but in FormulaHelper there will be both "classic way" and the way you've implemented for getting spell costs. Maybe you could merge over the classic method and data into your system similar to how you did for diseases, or just use it for reference if you want.

### DIFF
--- a/Assets/Scripts/Game/Formulas/FormulaHelper.cs
+++ b/Assets/Scripts/Game/Formulas/FormulaHelper.cs
@@ -19,6 +19,8 @@ using DaggerfallWorkshop.Game.MagicAndEffects;
 using System.Collections.Generic;
 using DaggerfallConnect.Arena2;
 using DaggerfallWorkshop.Game.Items;
+using DaggerfallWorkshop.Utility;
+using DaggerfallConnect.Save;
 
 namespace DaggerfallWorkshop.Game.Formulas
 {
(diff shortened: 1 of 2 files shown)


Interkarma:
It's definitely valuable to know how classic is doing things, and very important to have that in place. Thank you for doing the legwork on this! 👍 

Something we'll need to consider is what happens when a custom (modded) spell effects are used on a magic item? How do we plan to calculate the custom gold and casting cost for new effects not present in the hardcoded arrays? It will also be possible for modders to replace the built-in effects with custom versions by injecting a new effect implementation with the same key, and those replacement effects might have different costs.

Because of this design requirement, the approach I've taken is driven entirely by the effects themselves. The current cost formulas can both reproduce classic costs while allowing cost calculations for new effects - whereas the classic formulas are incapable of handling custom effects.

When I get to implementing magic items fully, the legacy magical effects array on items will be deprecated and mainly used for item import and matching back to IEntityEffect bundles.

In any case, we'll join the two halves together somehow. I'm incredibly grateful for everything you're doing here. The more information we have on Daggerfall's guts the better. :)

Allofich:
Sounds good, thanks.

PR #834, opened 2018-08-05
Class creator

numidium:
I'm happy to announce that the class creator is now in a usable state. This was a lot more work than I expected. It could probably use some refactoring in the future (hopefully fewer switch statements) but what I have currently works correctly as far as I can tell.

Edit: I totally forgot about the "Help" button. That isn't built out yet but it's not essential for usability so the class creator is still functional as promised.


Interkarma:
This is great, looking forward to checking it out!

Interkarma:
I've been playing with this and have no issues moving over to master. Good job!

I'll let you know any feedback once I've had a chance to play with it in more detail. :)

Interkarma:
Hi @numidium! Your custom class creator UI is part of latest builds and users can play with it now.

Just letting you know there's a thread on the bug reports forums flagging a couple of issues. I'll also summarise below.
https://forums.dfworkshop.net/viewtopic.php?f=24&t=1213

* Custom characters reportedly only gain 1 HP per level. I've not confirmed yet.
* Starting to add an advantage/disadvantage then pressing escape on the secondary list will add a partial entry such as "bonus to" or "resistance to" without the second part. Ideally pressing escape in secondary list would fallback to primary list, and pressing escape on primary list would pop back to main UI.
* Selecting "bonus to hit undead" will result in only "bonus to hit" being added.

Thank you for all your time and hard work on this, it's an awesome feature to have so far ahead of schedule and everyone seems really hyped to be testing it.

numidium:
Hey, @Interkarma! Thanks for bringing this to my attention. I would have been surprised if there weren't at least a couple of bugs.  I'll keep an eye on that thread.

PR #835, opened 2018-08-05
Match magic item merchant stock to classic

Allofich:
Also fixed up list of enemy soul points and added that to enemy definitions.

When magic merchants stock soul traps with souls in them, it's a random selection from 0 through 42.
As far as I can see, there is nothing stopping it from choosing the unused "Horse" or the dragonling with 0 enchantment points. Since other 0-value animals can also be chosen I added a "Horse" stub to the enemy list to allow for the soul to appear. Not sure if we should allow those or not.

However, when an item with a soul bound to it breaks, the stored enemy is supposed to spawn. I assume classic probably crashes or something if an item with a horse soul breaks, so if we allow horse souls we'll have to account for that. We should account for that anyway in event somebody imports a classic save with an item with a horse soul.


Interkarma:
Only in Daggerfall would we need to worry about the weight of horse's soul. :)

Good job!

ajrb:
I think I only ever intended this method to be temporary code. Is this complete now then?

PR #836, opened 2018-08-07
Vitals fixes

MeteoricDragon:
Overall better interaction between indicator bars and base bars.  If you heal after or before you take damage or heal damage, the bars appear where you'd expect, and won't hide each other except in special cases where you heal completely after losing, then the gain bar will cover the loss bar.  If the player heals while the gain is smoothing, the bar will make a briefer pause than initial pause before continuing to smooth again.  and same goes for loss bar when taking damage repeatedly.  Also refactored the code.


Interkarma:
Thank you for continuing to work on this and refining it. I've really enjoyed playing with this feature since it came in. There's something very satisfying about that short delay before a bar changes size, it really does make the experience more polished.

PR #837, opened 2018-08-07
Read classic rumor data

Allofich:
No functionality yet, just reads `RUMOR.DAT` when importing a classic save and stores the data in a list of rumors.


Interkarma:
Awesome work! I was always happy to have a "good enough" save import from classic, but with your work it's probably going to be 95%+ by the time we're done.

Allofich:
Yes, both types are in the file. The 9th byte is flags. What I know so far about the flags is:
0x01 = means a "rumor" (actually a public announcement) that shows on a city entrance sign, rather than spoken in dialogue
0x04 = I think means it's a quest rumor, and can show if the relevant quest is active
0x08 = I think means the rumor is spoken in dialogue

Nystul-the-Magician:
good to know, thx for letting me know ;)

PR #839, opened 2018-08-10
Fix and clean flicker

MeteoricDragon:
When setting breakpoints during the flashing and flickering, the amount of alpha can go past the intended amount because of more time passing due to taking more time to process the breakpoint.  If someone had a slowdown in their computer, this could also happen.  Their player's screen can end up permanently red.  This update fixes that.  Also, fixed some redundancy in code logic and simplified it.


Allofich:
I just had a situation where the screen ended up permanently red (with master, not this PR). I think it happened when I opened the inventory screen while being hit by an enemy.

MeteoricDragon:
@allofich hopefully this PR should fix that.

PR #841, opened 2018-08-12
Fix books from classic saves to open

Allofich:
Fixes https://forums.dfworkshop.net/viewtopic.php?f=24&t=678.

The only check classic does for whether to open the book screen when doing `USE` in the inventory on an item is whether it is in group 7, the books group. Changing DF Unity to also just check this solves this issue.

Also made various fixes to the item template enums that were mixed up and causing some issues. (For example, a book in classic can have group index 1 through 4, which I think is supposed to be for different quality books, but only one of these is "book" in master, causing the books imported from classic to fail)

The changes to the enums are all from data in the .EXE (I found where they are listed) so they should all be correct.

Also included latest ItemTemplates.txt and MagicItemTemplates.txt since these are out-of-date.

### DIFF
--- a/Assets/Scripts/API/ItemsFile.cs
+++ b/Assets/Scripts/API/ItemsFile.cs
@@ -27,7 +27,7 @@ namespace DaggerfallConnect.FallExe
         public long position;                       // Position record was read from FALL.EXE
         public Byte[] name;                         // Display name
         public Int32 baseWeightUnits;               // Base weight in 0.25kg units
-        public Int16 hitPoints;                     // Hit points
+        public UInt16 hitPoints;                    // Hit points
         public Int32 capacityOrTarget;              // Capacity of container or target of effect
         public Int32 basePrice;                     // Base price before material, mercantile, etc. modify value
         public Int16 enchantmentPoints;             // Base enchantment points before material
@@ -435,7 +435,7 @@ namespace DaggerfallConnect.FallExe
             item.position = reader.BaseStream.Position;
             item.name = reader.ReadBytes(nameLength);
             item.baseWeightUnits = reader.ReadInt32();
-            item.hitPoints = reader.ReadInt16();
+            item.hitPoints = reader.ReadUInt16();
             item.capacityOrTarget = reader.ReadInt32();
             item.basePrice = reader.ReadInt32();
             item.enchantmentPoints = reader.ReadInt16();
--- a/Assets/Scripts/Game/Items/ItemBuilder.cs
+++ b/Assets/Scripts/Game/Items/ItemBuilder.cs
@@ -248,7 +248,7 @@ namespace DaggerfallWorkshop.Game.Items
         public static DaggerfallUnityItem CreateRandomBook()
         {
             Array enumArray = DaggerfallUnity.Instance.ItemHelper.GetEnumArray(ItemGroups.Books);
-            DaggerfallUnityItem book = new DaggerfallUnityItem(ItemGroups.Books, Array.IndexOf(enumArray, Books.Book));
+            DaggerfallUnityItem book = new DaggerfallUnityItem(ItemGroups.Books, Array.IndexOf(enumArray, Books.Book0));
 
             // TODO: FIXME: I don't know what the higher order bits are for the message field. I just know that message & 0xFF encodes
             // the ID of the book. Thus the books that are randomly generated are missing information that the actual game provides.
--- a/Assets/Scripts/Game/Items/ItemEnums.cs
+++ b/Assets/Scripts/Game/Items/ItemEnums.cs
@@ -289,15 +289,15 @@ namespace DaggerfallWorkshop.Game.Items
         Champion_straps = 181,
     }
 
-    public enum Books  // modified to conform with real Daggerfall save game - IC112016
+    public enum Books
     {
-        UNKNOWN_BOOK_TYPE = 276,
-        Potion_recipe = 277,
-        Parchment = 279,
-        Book = 279,
+        Book0 = 277,
+        Book1 = 277,
+        Book2 = 277,
+        Book3 = 277,
     }
 
-    public enum Furniture // Index numbers are from ItemTemplates.txt
+    public enum Furniture
     {
         Plain_single_bed = 217,
         Fancy_single_bed = 218,
@@ -337,7 +337,7 @@ namespace DaggerfallWorkshop.Game.Items
         Bandage = 249,
         Oil = 252,
         Candle = 253,
-        Parchment = 278,
+        Parchment = 279,
     }
 
     public enum ReligiousItems  //checked
@@ -575,9 +575,8 @@ namespace DaggerfallWorkshop.Game.Items
         Spellbook = 132,
         Soul_trap = 274,
         Letter_of_credit = 275,
-        //Ruby = 0,
-        UNKNOWN_MISC_ITEM, // modified order to ensure Potion_recipe is index 4 -IC112016
-        Potion_recipe = 277,
+        Ruby = 0,
+        Potion_recipe = 278,
         Dead_Body = 281,
         House_Deed = 285,
         Ship_Deed = 286,
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
@@ -1419,16 +1419,8 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             // Handle local items
             if (item.ItemGroup == ItemGroups.Books)
             {
-                // Unreadable parchment (the one with a note graphic) is actually in UselessItems2
-                if (item.TemplateIndex == (int)Books.Book || item.TemplateIndex == (int)Books.Parchment)
-                {
-                    DaggerfallUI.Instance.BookReaderWindow.BookTarget = item;
-                    DaggerfallUI.PostMessage(DaggerfallUIMessages.dfuiOpenBookReaderWindow);
-                }
-                else if (item.TemplateIndex == (int)Books.Parchment)
-                {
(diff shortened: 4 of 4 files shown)


Allofich:
Also fixed an issue with the hitpoints of items being read as a signed short for the item templates, which made items with lots of hit points, like Lich Dust, become negative. It's read as an unsigned short now.

Assets/Resources/MagicItemTemplates.txt
```diff
@@ -1,1359 +1,3070 @@
 [
     {
-        "position": 4,
```
Interkarma: I'm curious about the change from "position" to "index" here? This value is the byte position the record is found in the source binary file (from binaryReader.BaseStream.Position at export time) - not the sequential index of the record. I've used this "position" convention elsewhere as well, as it matches the StreamReader naming.

It should be functionally fine as I'm not consuming that value elsewhere, but I feel it's a bit less clear now. I was mainly just curious why the need to change?

Assets/Resources/MagicItemTemplates.txt
```diff
@@ -1,1359 +1,3070 @@
...
-        "name": "The Masque of Clavicus Vile",
```
Interkarma: This was an intentional edit on my part. The item's full name is the "The Masque of Clavicus Vile". I don't mind it being truncated again to match classic however.

Assets/Resources/MagicItemTemplates.txt
```diff
@@ -1,1359 +1,3070 @@
...
     },
     {
-        "position": 128,
+        "index": 128,
         "name": "The Mace of Molag Bal",
         "type": "ArtifactClass1",
         "group": 3,
         "groupIndex": 11,
         "enchantments": [
-            538,
```
Interkarma: I've noticed the first effect param is condensed to just "SpecialArtifactEffect" and discards unique value read from the binary. This data is exported directly to JSON from MAGIC.DEF, and I had intended to hold exported data in its raw form.

This value will also help identify which "SpecialArtifactEffect" should be deployed. With these changes, that identifying value read from MAGIC.DEF is lost and the artifact items no longer have a unique effect value for that artifact. I feel this will make it more difficult to map these values back the implemented effect I had intended.

I would prefer the raw "enchantments" data is preserved, even if in just another form such as "legacyEnchantments". What are your thoughts?

Interkarma:
Is this a fresh export of the ItemTemplates.txt data? The whole file contents appear to have been overwritten. I've made a lot of post-fix changes to ItemTemplate.txt to fix incorrect draw order and variant counts as those errors have been discovered. There's a substantial amount of work in those fixes and simply replacing the file with a fresh export will cause a regression across all cases. I would prefer that only the required changes are made please, not replacing the entire file.

For MagicItemTemplates.txt, I've made a few comments on the file itself. My issues here are the "position" value being renamed to "index" - as this value is the record start position from binaryReader.BaseStream.Position, rather than a index. I would also prefer to preserve the raw effect data if possible as it contains more unique information that is lost when collapsing back to just "SpecialArtifactEffect". For example, how to tell the difference now between The Mehrunes Razor's "SpecialArtifactEffect" (originally 282) and The Mace of Molag Bal's "SpecialArtifactEffect" (originally 538) when it comes time to instantiate that effect in game?

I'm otherwise happy with the changes and looking forward to imported books working correctly. Just there's a lot of extraneous changes here that don't contribute to that end, which I would prefer were handled separately. :)

Allofich:
OK, I removed ItemTemplates.txt and MagicItemTemplates.txt from this PR. I'll respond to the comments you made.

Assets/Resources/MagicItemTemplates.txt
Allofich: I created the MagicItemTemplates.txt file by using the "Generate Magic Item JSON" button in the editor, so that change was done automatically because it's "index" in the file. I looked at GitHub's blame page for MagicItemsFile.cs (https://github.com/Interkarma/daggerfall-unity/blame/8bd34943b3afb0248c5bc593f0808c4b4672e4a2/Assets/Scripts/API/MagicItemsFile.cs) and it looks like it's been that name for a long time now. Anyway, I don't have a preference about the name if you want to change it!

Assets/Resources/MagicItemTemplates.txt
Allofich: I didn't realize that you had done manual edits to the file, I thought it was just automatically created.

As for this particular name, I can see an argument for changing it or not. I think maybe whoever made it that way thought of "Clavicus" as being his "first name". It's the same with "The Mehrunes Razor" (rather than "The Mehrunes Dagon Razor". Although it's "The Mace of Molag Bol" and not "The Mace of Molag".) I just checked the Clavicus Vile quest and he's only referred to as "Clavicus Vile", as one uninterrupted name, there. I think it's like that in the other TES games, too, but I generally try to avoid using the later games to retrofit things to Daggerfall. I would probably just leave it as it is in classic, but if you think it should be "The Masque of Clavicus Vile", I'm fine with that, too.

Assets/Resources/MagicItemTemplates.txt
Allofich: No data is being discarded! The enchantment for the Mace of Molag Bal is 1A 02, which was being read as it's decimal form, 538. But classic treats it as 2 values, the first (1A) is the wider category of enchantment (which I called "type") and the second (02) is interpreted depending on the enchantment category. In this case 1A (26 in decimal) is the category for unique artifact effects like the mace of molag bal absorbing strength,  so I called it "SpecialArtifactEffect". The second value (which I called "param") defines which of the unique artifact effects it is. "0" is the Masque of Clavicus (Vile), "1" is The Mehrunes Razor, "2" is the Mace of Molag Bal effect, etc.)

We could get rid of the enum if you'd prefer it to stay as numbers, (e.g. "26, 0") but I don't think that what is considered by classic like: Artifact Effect 0, Artifact Effect 1, Artifact Effect 2 should be stored as "26", "282", "538".

Allofich:
By the way, what prompted me to re-create the ItemTemplates.txt and MagicItemTemplates.txt files was that I noticed that since I changed how enchantments are being read from MAGIC.DEF message is appearing in the editor that MagicItemTemplates.txt has failed to load. I then noticed that automatically creating ItemTemplates.txt made a larger file and included some of the changes in this PR to item templates, so I thought I was simply providing the "latest versions" of these files by re-generating them. It does seem that they need to somehow be updated.

Assets/Resources/MagicItemTemplates.txt
Interkarma: That's interesting haha! I have no memory of changing that. :) I thought I exported that data from my offline test code, which is full of one-off code that never made it to git.

Assets/Resources/MagicItemTemplates.txt
Interkarma: Let's just leave it as-is then. :)

Assets/Resources/MagicItemTemplates.txt
Interkarma: No worries, thanks for explaining. :)

Interkarma:
> By the way, what prompted me to re-create the ItemTemplates.txt and MagicItemTemplates.txt files was that I noticed that since I changed how enchantments are being read from MAGIC.DEF message is appearing in the editor that MagicItemTemplates.txt has failed to load

It sounds like we need to regen MagicItemTemplates.txt in any case, so let's go with that. I believe I've only made the single edit to MagicItemTemplates.txt data (adding Clavicus Vile's full name) and I'm happy to leave that as per classic once file is regenerated.

For ItemTemplates.txt, there's a bunch of incremental fixes over time from users reporting clothing variant and draw order issues. It sounds like we need to regenerate this one as well, but need to preserve the manual edits too. Let me think on this one.

Are you happy for me to otherwise merge this PR? Then we can tackle the job of regenerating ItemTemplates.txt without losing any of the other changes.

Interkarma:
Looking back through the history of the file, it appears this isn't the first time it's been regenerated either, just the first time in about 2 years. Probably best outcome is to regenerate file then manually make edits for those post fixes.

I also believe there were edits prior to moving to full project structure in 2016. I'll need to dig back pre July 2016 to see if any manual edits were lost then as well.

Interkarma:
I've now regenerated files from scratch to make sure they have latest schema and manually replayed post fixes from the following key dates:

DONE - 26 Mar 2016
https://github.com/Interkarma/daggerfall-unity/commit/0e4990ceb6da28ddee09836174b0fd25f23cbb07#diff-1d7ea411f7d4c43b046e88f2154bed6f

DONE - 30 Mar 2016
https://github.com/Interkarma/daggerfall-unity/commit/d3085d2420ec10b061faf933b6c5fff92d20671f#diff-1d7ea411f7d4c43b046e88f2154bed6f

DONE - 12 Oct 2016
https://github.com/Interkarma/daggerfall-unity/commit/028f8898a0957997d1488124dd7206e017236f46#diff-762a0c068d3d5815e9217db805bfe12a

DONE - 12 Oct 2016
https://github.com/Interkarma/daggerfall-unity/commit/33ff16f4d747dabfb27d1ab5ac7515c925e6b29f#diff-762a0c068d3d5815e9217db805bfe12a

DONE - 12 Oct 2016
https://github.com/Interkarma/daggerfall-unity/commit/f00e860871e1155a31cf2fcf926a754b9bd0df58#diff-762a0c068d3d5815e9217db805bfe12a

DONE - 10 Sep 2017
https://github.com/Interkarma/daggerfall-unity/commit/043a769f3e472aa4667d75ba44ccc6fb3184c20e#diff-762a0c068d3d5815e9217db805bfe12a

DONE - 09 Oct 2017
https://github.com/Interkarma/daggerfall-unity/commit/f070e754d0a65f747e71b1173691abc5be869dd4#diff-762a0c068d3d5815e9217db805bfe12a

There are a couple of other changes I need to review, but I think they were also schema changes that should be included in the regenerated file.

At this time, ItemTemplates.txt and MagicItemTemplates.txt should be in the correct schema and have all important post fixes applied.

PR #842, opened 2018-08-12
Fix doors incorrectly rotating on load

Allofich:
Fixes https://forums.dfworkshop.net/viewtopic.php?f=24&t=765.

Doors that were mid-open when a game was saved will not play the open sound again when that game is loaded, and they will only open as far as they are supposed to.

Currently in master the "durationScale" that is supposed to make doors only rotate for the remaining duration was only affecting the "time" of the rotation, meaning how fast the door rotated, but not how much it actually rotated.


Interkarma:
Thank you for the fix! 👍

PR #844, opened 2018-08-18
Fix Save Explorer not opening if game isn't running

Allofich:
My previous code to import MAPSAVE data from classic saves (https://github.com/Interkarma/daggerfall-unity/commit/fd988d9b12c4ae3cbae5e48ef30f85924f843d9f) is causing Save Explorer to fail with a null exception if you try to open it without the game running in the editor. This happens because `PlayerGPS.dfUnity` isn't instantiated.

We could instantiate it, but I noticed that the MAPSAVE import makes a noticeable delay in Save Explorer as all the up-to-six saves are processed. Since Save Explorer doesn't do anything with that data yet anyway, I just disabled the import for Save Explorer. Ideally the import should be optimized, but I don't know how complex of a fix that would be.

### DIFF
--- a/Assets/Scripts/API/Save/SaveGames.cs
+++ b/Assets/Scripts/API/Save/SaveGames.cs
@@ -178,8 +178,9 @@ namespace DaggerfallConnect.Save
         /// Opens the save game index specified.
         /// </summary>
         /// <param name="save">Save index</param>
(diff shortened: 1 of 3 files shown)


Allofich:
Added a fix for https://forums.dfworkshop.net/viewtopic.php?f=24&t=892&p=10950&hilit=save+explorer#p10950.

Let me know if you want me to split it out to a different PR.

PR #845, opened 2018-08-18
Class creator bug fixes

numidium:
Includes the special advantage/disadvantage UI issue, max HP not carrying over, and spell points always being zero. Also fixes oversight in classic that allows more than 1 spell point multiplier.


Interkarma:
Thank you for continuing to refine this! 👍

PR #847, opened 2018-08-19
Handle used weapon hand for save/load

Allofich:
Fixes https://forums.dfworkshop.net/viewtopic.php?f=24&t=936.

Also imports used hand from classic saves.

Existing DF Unity saves will be loaded as if the right hand was being used. Saves made after this PR is applied will load with the correct hand.


Interkarma:
Good fix, thank you! :)

PR #848, opened 2018-08-21
Swimming improvements

MeteoricDragon:
- Implements a smooth underwater fog,  if the player is partially underwater, the fog fades partially in, until the camera is under.  This eliminates the blinking sensation that happens when exiting/entering water.

 - commented out a toggle for global fog because it was there to fix an issue with solid camera clear not working in Unity 5.5 because we're not unity 5.5 anymore, and it was preventing underwater fog from working.

- The player will no longer toggle crouch while using crouch key to swim down if player is in water.  Also, if player enters water and is crouching, will automatically stand.

- In some situations, the player could be stuck in the water while levitating.  This simple change allows the player to exit the water if they are levitating.

### DIFF
--- a/Assets/Scripts/Utility/CameraClearManager.cs
+++ b/Assets/Scripts/Utility/CameraClearManager.cs
@@ -23,7 +23,7 @@ public class CameraClearManager : MonoBehaviour
     public CameraClearFlags cameraClearExterior = CameraClearFlags.Depth;
     public CameraClearFlags cameraClearInterior = CameraClearFlags.Color;
     public Color cameraClearColor = Color.black;
-    public bool toggleGlobalFog = true;
+    //public bool toggleGlobalFog = true;
 
     GlobalFog globalFog;
     bool lastInside = false;
@@ -53,12 +53,13 @@ public class CameraClearManager : MonoBehaviour
                 mainCamera.backgroundColor = cameraClearColor;
                 //Debug.Log("Camera clear set to inside");
 
+                // Comment this out, We're not Unity 5.5 anymore - MeteoricDragon
                 // Disable global fog inside
                 // This fixes an issue with solid camera clear not working in Unity 5.5
-                if (toggleGlobalFog && globalFog)
-                {
-                    globalFog.enabled = false;
-                }
+                //if (toggleGlobalFog && globalFog)
+                //{
+                //    globalFog.enabled = false;
+                //}
 
                 lastInside = isInside;
             }
@@ -69,10 +70,10 @@ public class CameraClearManager : MonoBehaviour
                 //Debug.Log("Camera clear set to outside");
 
                 // Enable global fog outside
-                if (toggleGlobalFog && globalFog)
-                {
-                    globalFog.enabled = true;
-                }
+                //if (toggleGlobalFog && globalFog)
+                //{
+                //    globalFog.enabled = true;
+                //}
 
                 lastInside = isInside;
             }   
--- a/Assets/Scripts/Game/LevitateMotor.cs
+++ b/Assets/Scripts/Game/LevitateMotor.cs
@@ -90,7 +90,8 @@ namespace DaggerfallWorkshop.Game
             {
                 // Do not allow player to swim up out of water, as he would immediately be pulled back in, making jerky movement and playing the splash sound repeatedly
                 if ((direction.y > 0) && (playerMotor.controller.transform.position.y + (50 * MeshReader.GlobalScale) - 0.93f) >=
-                (GameManager.Instance.PlayerEnterExit.blockWaterLevel * -1 * MeshReader.GlobalScale))
+                (GameManager.Instance.PlayerEnterExit.blockWaterLevel * -1 * MeshReader.GlobalScale) &&
+                !playerLevitating)
                     direction.y = 0;
 
                 Entity.PlayerEntity player = GameManager.Instance.PlayerEntity;
--- a/Assets/Scripts/Game/Player/ClimbingMotor.cs
+++ b/Assets/Scripts/Game/Player/ClimbingMotor.cs
@@ -14,6 +14,7 @@ namespace DaggerfallWorkshop.Game
         private PlayerMotor playerMotor;
         private LevitateMotor levitateMotor;
         private CharacterController controller;
+        private PlayerHeightChanger heightChanger;
         private bool failedClimbingCheck = false;
         private bool isClimbing = false;
         private float climbingStartTimer = 0;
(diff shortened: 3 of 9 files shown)


Allofich:
The player has to be able to crouch while swimming (in dungeon water, at least), because there is a part in the Castle Wayrest dungeon where you have to do that. What's wrong with being able to crouch while swimming, anyway?

Allofich:
> I can make it so that the player can crouch when grounded and swimming.

That should be enough to allow getting past that part in the Castle Wayrest dungeon.

MeteoricDragon:
I reverted the crouch changes for now.  I've got an idea that involves actually just putting the player into a swimming size.  Which means whenever they start swimming, they shrink down to crouch size, to simulate swimming with their body horizontally which makes more sense than walking upright anyway.  This PR is ready now.

MeteoricDragon:
Implemented a simple system that makes the player crouch when entering the water without a loss in speed, and uncrouch when exiting.  I noticed the bug where the collision flags don't detect sides when trying to climb out of water, btw.  It feels good and gets rid of the need to worry about crouching using the same key to descend.

Nystul-the-Magician:
your changes will break any mods that need to alter fog settings - this is the critical code that will break them:
`        public void ResetFog()
        { 
            sky.SetSkyFogColor(sky.skyColors);
            RenderSettings.fogMode = originalFog;
            RenderSettings.fogStartDistance = 0;
            RenderSettings.fogEndDistance = 2400;
        }`
you should never set fog to hardcoded values, instead you should get the complete set of fog settings/values before changing it for underwater fog mode and later restore the old values when player leaves water.

please fix this so that the pull request won't break anything

MeteoricDragon:
@Nystul-the-Magician  Ok, I'm investigating.

MeteoricDragon:
@Nystul-the-Magician  I've investigated the code and found that the startDistance and endDistance are unnecessary and will remove them.  I also changed the originalFog to use RenderSettings.FogMode, But the SetSkyFogColor() line uses colors set by the DaggerfallSky already.  Unless I'm missing something?

Interkarma:
Thanks for making those changes @MeteoricDragon, I'm looking forward to playing with this and merging. Excellent work! :)

Thank you also for looking at the issue raised by @Nystul-the-Magician. If I understand correctly, it should be OK to set the fog colour inside dungeons as the DaggerfallSky component continuously sets this in exteriors. Nystul, please just let me know if I've misunderstood and this needs further refinement. Thank you both.

MeteoricDragon:
@Interkarma Thanks, I'm looking for feedback on two things.

The new auto-crouching, and the color of the underwater fog.    I made it kind of an aqueous blue.  And that is a good color if you're going underwater at the Grand Reef, but is it a realistic color for underwater dungeon?

I'm aware of a new problem that is made more obvious by the auto crouching
 

- if you go towards a ceiling that's declined with a stairway underwater, you can poke your head way out of the ceiling.  I tried crouch-swimming without this PR enabled, and the same problem occurred.  So I knew it wasn't because of my code.

MeteoricDragon:
Oh my gosh, Git, are you serious?  I saw that @Interkarma merged the PRs into master and decided I wanted to rebase onto master.  So I did the following:

Git Checkout master
git merge upstream/master
and all the commits from the PR went to my master.
I pushed local master to my fork's master.
git checkout SwimmingImprovements
used teamExplorer and right clicked SwimmingImprovements, selected "rebase Onto" and selected "master"
opened github desktop with swimmingImprovements still checked out, and clicked the button to sync with my repository, and it pulled on first click, then second click it pushed commits to my fork.

If I'm not mistaken, Rebase onto rewrites history and merge doesn't.  Which is why I used rebase when putting my branch onto the master branch that was just updated...  I didn't want the extra commits to show up in my PR.  I thought I was finshed making these messes...

Is this PR going to just apply my commits if merged to master?  Since master already merged the other PR's?
Why then,

Interkarma:
All good mate. :) I'm still learning as I go along as well, so not sure exactly where things went wrong.

What I'll do is merge this into a new branch equal to master right now and we'll see what happens. Worst case we can clean up and maintain your changes.

MeteoricDragon:
I think I might have fixed it, but it's hard to tell if I fixed it before or after you just changed the base branch to swimmingImprovements and then merged commit dd0f734 into Interkarma:SwimmingImprovements.  Because it said you did it like 32 seconds ago when I came to this page.  When I look at the history now, it looks pretty clean.

Assets/Scripts/Utility/CameraClearManager.cs
```diff
@@ -53,12 +53,13 @@ void Update()
                 mainCamera.backgroundColor = cameraClearColor;
                 //Debug.Log("Camera clear set to inside");
 
+                // Comment this out, We're not Unity 5.5 anymore - MeteoricDragon
                 // Disable global fog inside
                 // This fixes an issue with solid camera clear not working in Unity 5.5
-                if (toggleGlobalFog && globalFog)
```
Interkarma: I remember this issue - will need to review if problem still exists in Unity 2018.x and find another solution if so. No problems keeping commented for now.

MeteoricDragon:
What I did to fix it is right clicked on commit 4b810db7: fix incompatibility bug, and selected reset -> reset and delete changes --hard, and then went to Git Bash and typed "git checkout SwimmingImprovements" then "git push --force" and it resulted in what you see above.

Interkarma:
@MeteoricDragon Good work, glad we got past it! As long as you're happy with the current branch post-merge, I'll get this over to master soon.

> And that is a good color if you're going underwater at the Grand Reef, but is it a realistic color for underwater dungeon?

I recall classic Daggerfall's dungeon water is blue-ish as well. As long as the two are fairly close, it's fine. It does make sense for it to be a more murky colour though. If you expose waterFogColor with a get/set property then modders can change this anyway.

> I think the geometry of the dungeons allow this and there's nothing that can stop it.

I'd say so, there are lots of spots like this where game geometry isn't very well crafted. I'm not concerned about this for now. There's only so much we can do. :)

@Nystul-the-Magician are you also OK with the PR now? The only part remaining is the SetFogColor() which should be fine once player leaves dungeon.

Nystul-the-Magician:
about the sky.SetSkyFogColor(sky.skyColors) - as far as I remember esky disables DaggerfallSky - so as consequence this is not happening:

> DaggerfallSky component continuously sets this in exteriors

esky will set a custom color by itself when active (we will have to check if it only making changes to it when player is outdoors)

so as I said the right way to do it is to store the old value and restore it. 


>  I've investigated the code and found that the startDistance and endDistance are unnecessary and will remove them
I would have expected that these values need to be pretty low for underwater fog...

Interkarma:
Thanks for the feedback Nystul! :)

PR #849, opened 2018-08-23
Match enemy attack animations to classic

Allofich:
This one is a nice improvement. I figured out how to read the `.ANC` files in `MONSTER.BSA`, which define the frame sequences of animations and frames on which damage attempts or arrow shots occur, and probabilities for different frame sequences. I added this information in to the enemy definitions and use it now for enemy attack animations. This allowed me to remove the "hit frame" stopgap solution I put in earlier. In removing that I needed to adjust the calculation for attack timing, and that ended up with raising the distance from which enemies will attempt attacks. It could still use refinements but it makes enemy attacks less predictable and IMO more realistic, and makes it not so easy to just mow them down and never get hit by melee attacks. (Although the game may be tough in the beginning when your attacks are slow and missing a lot, I think in current master it may be too easy once you can start reliably hitting opponents. This should help with that.)

We could read in the  `.ANC` files but I don't see an advantage to doing so. Also if we used it exactly like classic we would inherit some animation bugs from classic, so it was easier to just re-create the data in the enemy definitions. The only data in those files that is of interest to us is the attack animations.

There are also sequences for each spell-casting enemy for their casting animations. I'll add those when enemy spells are in.

As a bit of a side-note, I'd like to get enemy classes and the player to start shooting 3d arrows (as in classic, although it's not easy to see them because of the lower resolution) instead of just doing instant hits. Has the progress on spells made it easy to create projectiles in the world? Are you interested in implementing 3d arrow projectiles yourself, Interkarma?

BTW, some mistakes in the classic files I found were:

1) The rat is unusual in having more than 4 animation frames in its walk, but it only uses 4 in classic. Since everything else uses 4, they probably forgot to configure the rat to use them all. You'll see in classic it has a jerky animation because of this, unlike in DF Unity.

2) The orc shaman has the most possible types of attack frame sequences at 5 types, but 2 of them aren't used in classic. For some reason when those 2 would be used, instead the spell casting or standing still animations are used. I configured DF Unity to use those 2 unused animation sequences as was probably intended.

### DIFF
--- a/Assets/Scripts/DaggerfallUnityStructs.cs
+++ b/Assets/Scripts/DaggerfallUnityStructs.cs
@@ -204,11 +204,20 @@ namespace DaggerfallWorkshop
         public bool ParrySounds;                    // Plays parry sounds when attacks against this enemy miss
         public int MapChance;                       // Chance of having a map
         public string LootTableKey;                 // Key to use when generating loot
-        public int HitFrame;                        // Frame of attack animation at which hit on target is attempted
         public int Weight;                          // Weight of this enemy. Affects chance of being knocked back by a hit.
         public bool CastsMagic;                     // Whether this enemy casts magic. Only used for enemy classes.
         public bool SeesThroughInvisibility;        // Whether this enemy sees through the shade, chameleon and invisibility effects.
         public int SoulPts;                         // Number of enchantment points in a trapped soul of this enemy
+        public int[] PrimaryAttackAnimFrames;       // Animation sequence to play when doing primary attack
+        public int ChanceForAttack2;                // Chance to use PrimaryAttackAnimFrames2 for an attack
+        public int[] PrimaryAttackAnimFrames2;      // Alternate animation sequence to play when doing primary attack
+        public int ChanceForAttack3;                // Chance to use PrimaryAttackAnimFrames3 for an attack
+        public int[] PrimaryAttackAnimFrames3;      // Alternate animation sequence to play when doing primary attack
+        public int ChanceForAttack4;                // Chance to use PrimaryAttackAnimFrames3 for an attack
+        public int[] PrimaryAttackAnimFrames4;      // Alternate animation sequence to play when doing primary attack
+        public int ChanceForAttack5;                // Chance to use PrimaryAttackAnimFrames3 for an attack
+        public int[] PrimaryAttackAnimFrames5;      // Alternate animation sequence to play when doing primary attack
+        public int[] RangedAttackAnimFrames;        // Animation sequence to play when doing bow & arrow attack
     }
 
     /// <summary>
--- a/Assets/Scripts/Game/EnemyAttack.cs
+++ b/Assets/Scripts/Game/EnemyAttack.cs
@@ -33,8 +33,6 @@ namespace DaggerfallWorkshop.Game
         DaggerfallMobileUnit mobile;
         DaggerfallEntityBehaviour entityBehaviour;
         float meleeTimer = 0;
-        bool isMeleeAttackingPreHitFrame;
-        bool isShootingPreHitFrame;
         int damage = 0;
 
         void Start()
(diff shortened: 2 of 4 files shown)


Interkarma:
This is very awesome. Good work working out the ANC format and adapting it across.

Interkarma:
@Allofich, I'm sorry mate, this one now has some conflicts after merging your enemy weights PR.

Interkarma:
> Are you interested in implementing 3d arrow projectiles yourself, Interkarma?

This should already be good to go with the DaggerfallMissile class. I built this in a fairly generic way to handle general missiles like arrows in additional to spell missiles.

It's been a while since I built this, so might have overlooked something you need. If anything is missing, just let me know what's required and I'll be more than happy to assist. :)

Allofich:
The merge conflicts are fixed now.

Allofich:
> This should already be good to go with the DaggerfallMissile class. I built this in a fairly generic way to handle general missiles like arrows in additional to spell missiles.

> It's been a while since I built this, so might have overlooked something you need. If anything is missing, just let me know what's required and I'll be more than happy to assist. :)

I just looked briefly at it but it seems like its quite specific to spells at the moment. You have prefabs for the different spell elements, and you instantiate the one you need, so I guess we'd want a new prefab for arrows? There's a 3d arrow of a model somewhere in classics' 3d object game files (I identified it once but I don't know if I made a note of that somewhere...) that we would want to use rather than a 2d texture, and I assume we wouldn't be using the spell-specific properties in DaggerfallMissile like "explosion radius" or "touch range". Also I see a comment in EntityEffectManager.c about spell missiles needing to be decoupled from the player to allow AI opponents to use them.

I'm not really up to trying to get a hold of all that at the moment, so I'll probably work on other stuff. Also you're still working on spells so you might still want to redesign parts of the DaggerfallMissile class.

Back when I first put in enemy ranged attacks I experienced with creating the arrows as 3d objects and shooting them out, which looked pretty good (looks a lot better than just the instant hit we have now) but I didn't know how to add collision in Unity and check for hitting entities or obstacles, so I didn't submit anything.

Nystul-the-Magician:
the arrow model is model #99800

Allofich:
> the arrow model is model #99800

Thanks!

Interkarma:
> I just looked briefly at it but it seems like its quite specific to spells at the moment. 

DaggerfallMissile is generally missile-agnostic. It naturally has support for spells, but it was written with any projectile in mind. You could shoot cows if you wanted to. And I don't mean shoot AT cows - I'm talking about a literal in-game cow cannon replete with wailing moos as they sail into the distance. :)

> so I guess we'd want a new prefab for arrows?

Yep, just need a prefab with some required components (DaggerfallMissile, DaggerfallAudioSource, SphereCollider, Rigidbody, and Light). Look to the spell missiles for prefab setup. You configure everything in the DaggerfallMissile Inspector on your prefab and shouldn't even need to touch the missile code. My main sin is assuming a few things such as lights being present when they don't need to be. I should have made lights optional at least.

The concept is that DaggerfallMissile does all the heavy lifting regarding movement, colliding, and gathering target entities. You only need to instantiate the arrow prefab, give it a directional vector, consume its OnComplete event, and call DoMissile(). If/when it hits something, your event will be fired with a list of any DaggerfallEntity targets found by the missile. Your missile can perform touch, single target, or AOE strikes depending on settings.

One other idea I haven't played with yet - it should be possible to setup the Rigidbody to be non-kinematic and use gravity, so that arrows arc over time. Likewise could lob heavy projectiles over short distances with a tight arc. With a little more work in the missile class, we could do stuff like bouncing missiles or missiles with custom movement controlled by another component. There's a lot of possibilities in there.

Of course, it's perfectly OK if you want to do something else for a while, just go where your interest takes you. I'm happy to setup the arrow prefab and clean up DaggerfallMissile a little to get things started, and we can look at this again later. :)

Allofich:
OK, well that's good if it's ready to be used for shooting 3d objects (arrows) without spell-specific properties.

> Of` course, it's perfectly OK if you want to do something else for a while, just go where your interest takes you. I'm happy to setup the arrow prefab and clean up DaggerfallMissile a little to get things started, and we can look at this again later. :)

Probably will do that, yes. I think the only "missiles" in the game are at-range spells and arrows, right? Since you've done one of those already, it seems to make sense that you would have the best idea of how to also do the other and incorporate them together nicely. I might also take a stab at it if I feel like it.

Allofich:
The game seems a little too hard after this PR. Since enemies are willing to attack from a greater range now, this results in them attacking quite frequently in combat, instead of giving up and resetting the melee timer every time they are out of range. I think I know how the melee timer works in classic, and maybe if I can get that in soon, it will alleviate the issue.

I saw you update the version, but if released as is we might see people saying they can't get out of Privateer's Hold. We could just get rid of the longer melee distance in the AI decision, but for now at least, I like it, so another quick fix could be to just raise the value the melee timer resets to. But, anyway, I'll try to get classic melee timer behavior/values in before the end of the month and maybe that will be enough.

Interkarma:
Thanks for the feedback. I'm running up new builds now mainly for users to test new font rendering setup. I'll tag this as a "font test" or something and we can spin up another round of regular pre-alpha builds once the AI has been refined a little. :)

Allofich:
OK, I won't rush myself, then. :)

PR #850, opened 2018-08-23
Shorten enemy stealth check range

Allofich:
Since I allowed enemies to detect the player through walls with the stealth check (check between the player and the enemy's stealth stats), they are detecting from a little too far away. Someone who did a video on YouTube was wondering if the game was bugged. Adjusting it down to the value in this PR seems to help for now. In classic the actual stealth check uses a distance of 1024 but I think it's also limited by the distance of enemies before they are unloaded, which I don't fully understand but I think is around 768 in interiors (1024 in exteriors).

### DIFF
--- a/Assets/Scripts/Game/EnemySenses.cs
+++ b/Assets/Scripts/Game/EnemySenses.cs
@@ -196,7 +196,7 @@ namespace DaggerfallWorkshop.Game
 
         public bool StealthCheck()
         {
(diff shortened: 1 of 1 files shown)


MeteoricDragon:
I was noticing the effects of their increased detection.  There was a save I have in Privateer's hold nearby a bat.  The bat couldn't see me whenever I loaded the save.  but after merging some changes to my fork, I noticed the bat could find me all of a sudden.

PR #852, opened 2018-08-25
Recreate classic way of choosing random enemies

Allofich:
This implements the issue raised in https://forums.dfworkshop.net/viewtopic.php?f=24&t=560, which was closed as "intended". It should also address the issue raised here: https://www.reddit.com/r/daggerfallunity/comments/96pmzc/the_enemies_in_daggerfall_unity_are_way_off_and/

Classic selection is enabled by default. Previous random selection can be turned on by enabling "Alternate Random Enemy Selection" at the startup screen. (except the alternate, undead-heavy cemetery list for DF Unity is commented out and classic's list is being used now)

Basically how it works: As you know, there are `RANDOM MON` markers that make random enemies and `MONSTER` markers that make fixed enemies. Random enemies, though, are not completely random in classic. Most seem to have a value set in their "flags" part of the rdb resource data, and this is used to choose from a list of enemy types that is created for that dungeon (which is determined using dungeon ID as a seed for the random generator, with the player's level as an input that determines the result). There are some that don't have any value for the "flags" data, in which case it's a random choice from 1 to 6, without a set seed to use, so these vary every time you leave/enter the dungeon.

This doesn't affect "random encounters" that spawn intermittently in exteriors or that interrupt player rest. I might look at those later.

### DIFF
--- a/Assets/Resources/defaults.ini.txt
+++ b/Assets/Resources/defaults.ini.txt
@@ -80,3 +80,4 @@ LypyL_ModSystem=True
 MeshAndTextureReplacement=False
 CompressModdedTextures=True
 NearDeathWarning=True
+AlternateRandomEnemySelection=False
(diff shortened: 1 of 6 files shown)


Allofich:
BTW, I've left in the  behavior we were using as an option, but I'm fine with it being removed entirely if you want to keep down the number of options on the startup screen. Looks like there isn't much room left, unless it can scroll or something.

PR #854, opened 2018-08-29
Implement classic melee timer and reflexes

Allofich:
Things this does:

- Implements classic melee timer
- Fix what seems to be oversight in classic where higher speed enemies would instead attack less
- Implements reflexes as in classic. This is in classic, too, but like so many things does not seem to work correctly (I swear, it's like they purposely left the game with a random thing missing/reversed, etc. in every game mechanic so we'd have the fun of patching the holes), because the value it refers to in the enemy object data seems to always be 0 and is not ever set by anything, so no matter what reflexes setting you choose it seems enemies attack at "very high" reflexes in classic. Luckily, although having this correctly refer to the reflexes setting slows down attacks at, for example, the "average" setting of 2, I think the above fix more or less offsets this, as many enemies have high speed attributes. At any rate the resultant attack rates seemed OK in testing.

BTW reflex setting is also supposed to affect skill increases, but I don't think that's in yet. It may also have another effect on attacks that isn't in yet.

- Matched enemy melee range to classic (128 Daggerfall distance units). Doing this also automatically updated the enemy prefabs in other ways, although I'm not sure if those updates actually affect anything or not.

### DIFF
--- a/Assets/Prefabs/Scene/DaggerfallEnemy [Game Serializable].prefab
+++ b/Assets/Prefabs/Scene/DaggerfallEnemy [Game Serializable].prefab
@@ -83,9 +83,11 @@ MeshRenderer:
   m_Enabled: 1
   m_CastShadows: 1
   m_ReceiveShadows: 1
+  m_DynamicOccludee: 1
   m_MotionVectors: 1
   m_LightProbeUsage: 1
   m_ReflectionProbeUsage: 1
+  m_RenderingLayerMask: 4294967295
   m_Materials:
   - {fileID: 0}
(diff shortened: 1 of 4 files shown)


Interkarma:
Impressive as always. :) It's great to have the reflex setting functional, I think it goes a long way to helping players work inside their reaction time. Also good to know it's less buggy than classic!

>Doing this also automatically updated the enemy prefabs in other ways, although I'm not sure if those updates actually affect anything or not.

This should be fine. Unity does this to me as well anytime I touch a prefab after updating versions.

PR #855, opened 2018-08-30
Swimming climb up

MeteoricDragon:
The last commit in this PR has lots of experimental changes to the climbing system.  The others are the groundwork for a new climbing system mostly.

Things to try with this code:  

- Climbing outdoor buildings.
- Climbing dungeon walls.
- Climbing dungeon walls while aimed at doors. Curious results.
- Climbing onto dry land from water.
- Climbing starting from underwater if feeling adventurous.

### DIFF
--- a/Assets/Scripts/Game/LevitateMotor.cs
+++ b/Assets/Scripts/Game/LevitateMotor.cs
@@ -27,7 +27,7 @@ namespace DaggerfallWorkshop.Game
         PlayerSpeedChanger speedChanger;
         Camera playerCamera;
         float moveSpeed = 4.0f;
-        CollisionFlags collisionFlags = 0;
+        private ClimbingMotor climbingMotor;
 
         public bool IsLevitating
         {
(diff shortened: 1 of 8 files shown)


Allofich:
Why no speed penalty for climbing while swimming? I haven't tried your PR but my first impression would be it would feel more realistic.

I just tried climbing in the water in classic and the player is slowed down if they climb while submerged compared to when climbing out of the water.

MeteoricDragon:
@allofich I think the levitate motor might be active at the same time and if you aim down slightly, it will fight your ascent.

Interkarma:
Thank you for this! I'll stick it in its own branch now.

MeteoricDragon:
@Allofich I just tested climbing underwater and noticed no slower ascent than climbing buildings outdoors.  I didn't have success climbing dungeon walls unless I was aimed at a door or other special walls or objects. Honestly, I don't know what's special about those objects, except for the fact that they collide with the CapsuleCast function and other walls do not. I had pictures posted in the forum with descriptions about what I could or couldn't detect colliders on with CapsuleCast.

PR #859, opened 2018-09-02
Fixes Falling in water when climbing

MeteoricDragon:
Fixes Falling in water while climbing by adding a buffer that the player cannot fail climbing above water level.

Allows the player to look left and right a little bit while climbing, but doesn't reduce the shallowness of the angle the player needs to look to start climbing.

Uses a CapsuleCast to tell if the player should stop climbing and start going forward, and gets the direction of the ledge the player will be climbing onto.  Forces the player's horizontal move direction towards the ledge directly.

This leaves the skill check open.  @Interkarma  what are you thinking you'll want for the skill check system?  I'm thinking the current skill check system is insufficient?  a player with 5 climbing skill should have terrible trouble climbing but with the current system, a player with 50 climbing skill would have a 50% chance of falling every skill check.  And falls have no real consequence, the player just slips a bit.

### DIFF
--- a/Assets/Scripts/Game/Player/ClimbingMotor.cs
+++ b/Assets/Scripts/Game/Player/ClimbingMotor.cs
@@ -14,6 +14,7 @@ namespace DaggerfallWorkshop.Game
         private PlayerMotor playerMotor;
         private LevitateMotor levitateMotor;
         private CharacterController controller;
+        private PlayerEnterExit playerEnterExit;
         //private PlayerHeightChanger heightChanger;
         private bool failedClimbingCheck = false;
         private bool isClimbing = false;
@@ -22,6 +23,7 @@ namespace DaggerfallWorkshop.Game
         private uint timeOfLastClimbingCheck = 0;
         private bool showClimbingModeMessage = true;
         private Vector2 lastHorizontalPosition = Vector2.zero;
+        private Vector3 ledgeDirection = Vector3.zero;
         public bool IsClimbing
         {
             get { return isClimbing; }
@@ -31,6 +33,7 @@ namespace DaggerfallWorkshop.Game
             playerMotor = GetComponent<PlayerMotor>();
             levitateMotor = GetComponent<LevitateMotor>();
             controller = GetComponent<CharacterController>();
+            playerEnterExit = GetComponent<PlayerEnterExit>();
             //heightChanger = GetComponent<PlayerHeightChanger>();
         }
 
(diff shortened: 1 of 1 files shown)


MeteoricDragon:
As the player climbs up, if he's angled slightly to the side, the forward movement will horizontally move the player, causing the climb to fail because he's outside the initial horizontal check difference when the ClimbingCheck method runs.  The CapsuleCast purpose is not primarily to detect that there's a wall there, but to direct the forward impulse perpendicularly to the wall and prevent horizontal movement that makes the player fail the initial climbing check.  It gets the normal vector of the wall, inverts it, and makes that the direction that the player's forward impulse direction, Perpendicular to the wall, preventing the horizontal movement that causes the climb to fail.

Interkarma:
Thanks for the explanation, looking forward to testing

MeteoricDragon:
It looks like there's also a problem with climbing high walls from water.  You fall into the water when you fail skill check.  But no fall if you started on dry land.

Interkarma:
I agree this feels much nicer on top of the other other fixes. I think this is working fairly solidly now. Good job!

@Allofich is there anything else you'd like to add or review as the original author of climb system?

Allofich:
I haven't really had a chance to try this stuff yet. I also want to examine a bit more how classic works. If I have anything to add I'll open up a PR or a forum thread or something.

MeteoricDragon:
@allofich i got a question.  Theres a block of code I left largely untouched since refactoring the climbing code.  I'm wondering why it is there.  It seems that it is only called if in game minutes have passed.

`
                     // Initial check to start climbing
                     if ((gameMinutes - timeOfLastClimbingCheck) > 18)
                     {
                         Entity.PlayerEntity player = GameManager.Instance.PlayerEntity;
                         player.TallySkill(DFCareer.Skills.Climbing, 1);
                         timeOfLastClimbingCheck = gameMinutes;
                         if (UnityEngine.Random.Range(1, 101) > 95)
                         {
                             if (UnityEngine.Random.Range(1, 101) > player.Skills.GetLiveSkillValue(DFCareer.Skills.Climbing))
                             {
                                 isClimbing = false;
                                 failedClimbingCheck = true;
                             }
                         }
}`

MeteoricDragon:
``

> 
                     // Initial check to start climbing
                     if ((gameMinutes - timeOfLastClimbingCheck) > 18)
                     {
                         Entity.PlayerEntity player = GameManager.Instance.PlayerEntity;
                         player.TallySkill(DFCareer.Skills.Climbing, 1);
                         timeOfLastClimbingCheck = gameMinutes;
                         if (UnityEngine.Random.Range(1, 101) > 95)
                         {
                             if (UnityEngine.Random.Range(1, 101) > player.Skills.GetLiveSkillValue(DFCareer.Skills.Climbing))
                             {
                                 isClimbing = false;
                                 failedClimbingCheck = true;
                             }
                         }
}

Allofich:
That was an attempt to replicate classic's check on the climbing skill, but it's probably not correct, I need to revisit it. The idea was it was an initial check against the player's climbing skill to begin the climbing process, separate from the check whether the player successfully continues to climb after starting. I'll revisit it in a while and perhaps remove it if I can't find a purpose for it.

MeteoricDragon:
@allofich I have an Idea that I'd like to try out in the climbing code. I wanted to mention this before I started working on it in case you already have something that you're working on with the climbing code so that I don't create code before you get the chance to submit what you did. Are you working on anything at the moment?

Allofich:
I'm not working on climbing at the moment. I'm OK if you want to work on climbing, but I would like to get classic's skill tallying in for it, so I'd prefer if you don't put in something really different from how classic works.

MeteoricDragon:
@Allofich I plan to do something exactly like the classic.  according to UESP description of Daggerfall climbing.

Allofich:
Ok, well no problem then. I'm also trying to get classic behavior, minus bugs.

PR #863, opened 2018-09-04
Fix out-of-bounds array access in older saves

Allofich:
Seems that at some point we changed from storing region IDs of factions from 0-based to 1-based. Maybe I had something to do with that, I don't remember. This was causing an IndexOutOfRangeException in RegionPowerAndConditionsUpdate(), which subtracted 1 from the region ID because I wrote it assuming the newer, 1-based behavior.

This PR fixes this issue by returning to using 0-based, fixing 1-based saves to be this way. Alternatively we could make them all 1-based, but I prefer 0-based for a few reasons:

1) Classic also converts the 1-based region IDs in FACTION.TXT to 0-based internally, and saves them this way in SAVEVARS.DAT. If we decide to import those region IDs from SAVEVARS.DAT (we aren't now), it will be more straightforward to use the same values.

2) Using 0-based allows me to change a bunch of ugly `(region -1)` expressions to just `region`.

3) 0-based matches the region numbers on UESP.

### DIFF
--- a/Assets/Scripts/Game/Serialization/SerializablePlayer.cs
+++ b/Assets/Scripts/Game/Serialization/SerializablePlayer.cs
@@ -210,7 +210,25 @@ namespace DaggerfallWorkshop.Game.Serialization
         public void RestoreFactionData(FactionData_v2 factionData)
         {
             PlayerEntity entity = playerEntityBehaviour.Entity as PlayerEntity;
-            entity.FactionData.FactionDict = factionData.factionDict;
+            // There are a mixture of saved games that were made using the 1-based region IDs
+            // straight from FACTION.TXT, while others made earlier used 0-based IDs.
+            // Here we convert saves with 1-based to be 0-based. 0-based is assumed by code using these IDs.
+            if (factionData.factionDict[201].region == 18) // To check for 1-based, see if Daggerfall is 18 rather than 17
+            {
+                Debug.Log("Updating 1-based faction region IDs to 0-based.");
+                entity.FactionData.FactionDict = new Dictionary<int, DaggerfallConnect.Arena2.FactionFile.FactionData>();
+                foreach (int key in factionData.factionDict.Keys)
+                {
+                    var dict = factionData.factionDict[key];
+                    if (factionData.factionDict[key].region != -1)
+                        dict.region--;
+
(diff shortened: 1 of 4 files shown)


Allofich:
I think I may know what happened. I think straight DF Unity saves were always using 1-based, but saves imported from classic used the values from SAVEVARS.DAT, which are IDs which classic had converted to 0-based, which is what it uses internally. Then, I think with this commit https://github.com/Interkarma/daggerfall-unity/commit/b48b9d41ccfb1ad0a943cd7fd1620269e97220f9, the 0-based values stopped being imported from SAVEVARS.DAT, and instead whatever values were in the already present faction data at time of load were retained. Saves made from imported classic saves before that point still had their 0-based values, while pure DF Unity saves and any new saves, DF Unity or imported from classic, after then had 1-based values.

After that I wrote RegionPowerAndConditionsUpdate(), which assumed the then-current 1-based values. This caused an IndexOutOfRangeException exception when an old 0-based value save was loaded, when RegionPowerAndConditionsUpdate() did `(region - 1)` for a region of `0` and tried to access a negative array index.

Anyway, that should all be fixed now.

Assets/Scripts/Game/Serialization/SerializablePlayer.cs
```diff
@@ -210,7 +210,25 @@ public object GetFactionSaveData()
         public void RestoreFactionData(FactionData_v2 factionData)
         {
             PlayerEntity entity = playerEntityBehaviour.Entity as PlayerEntity;
-            entity.FactionData.FactionDict = factionData.factionDict;
+            // There are a mixture of saved games that were made using the 1-based region IDs
+            // straight from FACTION.TXT, while others made earlier used 0-based IDs.
+            // Here we convert saves with 1-based to be 0-based. 0-based is assumed by code using these IDs.
+            if (factionData.factionDict[201].region == 18) // To check for 1-based, see if Daggerfall is 18 rather than 17
```
Interkarma: This is clever.

Interkarma:
That's a curly one. Good work. I don't think this will affect anything else that plugs into faction data, such as the quest system, as it uses the IDs rather than faction indices. I'm just going to stick in a branch briefly so I can do some regression testing against a few other systems, but I think it will be fine.

PR #864, opened 2018-09-05
Match non-LOS sense of player to classic

Allofich:
Should fix https://forums.dfworkshop.net/viewtopic.php?f=24&t=1227, in as much as it generally matches the range to classic.

This could maybe be tidied up a little but I wanted to get it in before the release which is probably upcoming.

Background of the issue:
In classic, enemies can detect the player without having LOS, through walls, etc., depending on a roll against the player's stealth skill. This is why they sometimes open doors and come out of rooms at you, etc. I implemented this a while ago but not the correct range at which the check should happen. Enemies were approaching the player from a lot further than in classic, and enemies weren't being found in their expected spots in Privateer's Hold, etc. This may also have been making it more difficult to find a spot away from enemies to rest. In classic the stealth check itself has a big range but it's limited by the spawn/despawn range, because AI characters won't be active to do the check unless they are near the player. It turns out a byte in the spawn markers, from 0 through 6, determines the X-range and Y-range from the player at which an enemy spawns/despawns (the distances for spawning and despawning are a little different). That is recreated by this PR now, using these distances to limit the range of their stealth detection, effectively recreating classic behavior, but without actually spawning and despawning them.

### DIFF
--- a/Assets/Scripts/Game/EnemySenses.cs
+++ b/Assets/Scripts/Game/EnemySenses.cs
@@ -38,6 +38,7 @@ namespace DaggerfallWorkshop.Game
         DaggerfallActionDoor actionDoor;
         float distanceToActionDoor;
         bool hasEncounteredPlayer = false;
+        bool wouldBeSpawnedInClassic = false;
         bool detectedPlayer = false;
         uint timeOfLastStealthCheck = 0;
         bool blockedByIllusionEffect = false;
@@ -45,6 +46,13 @@ namespace DaggerfallWorkshop.Game
 
         float classicUpdateTimer = 0f;
         bool classicUpdate = false;
+
+        float classicSpawnDespawnExterior = 4096 * MeshReader.GlobalScale;
+        float classicSpawnXZDist = 0f;
+        float classicSpawnYDistUpper = 0f;
+        float classicSpawnYDistLower = 0f;
+        float classicDespawnXZDist = 0f;
+        float classicDespawnYDist = 0f;
 
         GameObject Player
         {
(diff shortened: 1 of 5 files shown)


ajrb:
You're doing great work mate - been looking though your PRs and wanted to say. One suggestion for this code is you could consider putting all the numbers into an array and get rid of the switch. May be dumb suggestion, but see what you think.

Allofich:
That's probably a good idea, I just wanted to get the functionality going. I'll also probably move this stuff out of the stealth function later, because knowing when an enemy would be spawned or despawned should also allow us to replicate classic's "enemies nearby" check, and maybe other behavior as well.

Allofich:
Applied classic spawn/despawn result to AreEnemiesNearby() check.

Interkarma:
This is an excellent change. I'm often impressed how you find clever ways to squeeze more classic-like outcomes from these systems.

Assets/Scripts/Game/EnemySenses.cs
```diff
@@ -116,6 +144,52 @@ void FixedUpdate()
             else
                 classicUpdate = false;
 
+            // Update whether enemy would be spawned or not in classic.
+            // Only check if within the maximum possible distance (Just under 1094 classic units)
+            if (classicUpdate && distanceToPlayer < 1094 * MeshReader.GlobalScale)
```
Allofich: Just realized, that if the player teleported from being close to an enemy to outside this distance, the enemy's "spawned in classic" state would be stuck at whatever it was.

PR #865, opened 2018-09-05
Improvements for mod settings editor

TheLacus:
* Store mod that is being edited for the next session.
* Make save on exit without confirmation optional.
* Fixed an issue which could cause new values to be stored incorrectly to presets if added to settings and immediately saved without a "refresh".


Interkarma:
Thank you!

PR #866, opened 2018-09-05
Allow gold stored in wagon to be deposited at a bank

ajrb:


Interkarma:
I think this is a good fix, cheers!

PR #867, opened 2018-09-06
Climb update

MeteoricDragon:
Fixes all known bugs with climbing, including falling when climbing a long wall out of the water.  Slips are no longer meaningless.  Implements a skill check system like the system described in UESP for classic DF.

Player now has a chance to climb successfully, a chance to slip, and a lower chance to regain hold and continue climbing.  All chances can be offset by having a high skill in climbing.

70% base chance of success for continuing to climb per skill check.  A player with 50 climbing skill would boost this to 85% and 75 would boost to 92.5%

20% base chance of regaining hold (at 37% of climbing skill check's wait time).  a player with 50 skill would boost this chance to 60%, 75 skill boosts it to 80%

As an added bonus, once the player is climbing, they can swivel around and look in any direction as long as they hold the forward key and keep in mind when they reach the top they'll move in the direction they are facing!  They can also slip and catch and won't move away from the wall while looking elsewhere.

### DIFF
--- a/Assets/Scripts/Game/Player/ClimbingMotor.cs
+++ b/Assets/Scripts/Game/Player/ClimbingMotor.cs
@@ -84,22 +84,6 @@ namespace DaggerfallWorkshop.Game
                         showClimbingModeMessage = false;
                         isClimbing = true;
                     }
-
-                    // Initial check to start climbing
-                    if ((gameMinutes - timeOfLastClimbingCheck) > 18)
-                    {
-                        Entity.PlayerEntity player = GameManager.Instance.PlayerEntity;
-                        player.TallySkill(DFCareer.Skills.Climbing, 1);
-                        timeOfLastClimbingCheck = gameMinutes;
-                        if (UnityEngine.Random.Range(1, 101) > 95)
-                        {
-                            if (UnityEngine.Random.Range(1, 101) > player.Skills.GetLiveSkillValue(DFCareer.Skills.Climbing))
-                            {
-                                isClimbing = false;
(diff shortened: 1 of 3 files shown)


MeteoricDragon:
@Allofich  I changed the player's Climbing speed to PlayerSpeed / 3 according to what you said in another commit thread.

Interkarma:
Thank you for all your work on refining the climb system. It's so much better than it was even a few weeks ago.

MeteoricDragon:
You're welcome.   By the way, what do you think of being able to stop climbing and hold position by releasing the forward key, and being able to jump in a direction the player is facing?  Or even being able to climb down or left or right?  It's not like classic, it's definitely a more modern control scheme.  Of course the entire time they are moving they have a chance to fall.  I could probably build a toggle setting for it in the options.

Interkarma:
I have no problems with a "modern climbing" option if you want to develop it, and happy for this to be enabled by default. It might not be like classic, but players can just levitate wherever they like anyway so it's not like it breaks anything.

This would in fact add a very nice roleplaying touch for high-skill thieving characters and acrobats. Anyone without the skills would just slip and fall. I feel this can only make the game more fun. :)

Allofich:
> By the way, what do you think of being able to stop climbing and hold position by releasing the forward key, and being able to jump in a direction the player is facing? Or even being able to climb down or left or right?

I agree that sounds cool, but yeah, it should be an option. And IMO being able to jump from climbing doesn't seem like something that every character should be capable of, only really good climbers (unless it's a very tiny jump).

PR #868, opened 2018-09-06
Remove support for obsolete modsettings format

TheLacus:
Initially mod settings file was named _modfilename.ini.txt_, then was changed on Lypyl suggestion to _modsettings.ini.txt_ to avoid breaking if mod filename was changed. It's been a while since we moved from plain textboxes to a json-based window with typed controls, so i think it's time to remove support for the oldest one.

I'm confident no mod is using it but tell me if i'm wrong,  I want to be sure not to break any existing mod  :)

I'm not in a hurry to stop supporting _modsettings.ini.txt_, but i slightly changed the log message printed during the automatic local upgrade to reflect this change.


Interkarma:
Sounds good to me. :)

PR #872, opened 2018-09-09
4 Directional Climbing and jumping to and from walls.

MeteoricDragon:
Implements a toggleable, true by default advanced climbing system.

4 directional climbing, the ability to pause climbing, Look around in any direction.
The ability to jump off a wall when climbing, onto another wall (skill check required).  Can also initialize a climb by jumping from the ground onto a wall with a skill check.

Known issue:  The checkbox setting on the gameplay section of the advanced settings page is out of margin, but still clickable.

### DIFF
--- a/Assets/Resources/defaults.ini.txt
+++ b/Assets/Resources/defaults.ini.txt
@@ -80,4 +80,5 @@ LypyL_ModSystem=True
 MeshAndTextureReplacement=False
 CompressModdedTextures=True
 NearDeathWarning=True
+AdvancedClimbing=True
 AlternateRandomEnemySelection=False
--- a/Assets/Scripts/Game/Player/AcrobatMotor.cs
+++ b/Assets/Scripts/Game/Player/AcrobatMotor.cs
@@ -14,6 +14,7 @@ namespace DaggerfallWorkshop.Game
 
         PlayerMotor playerMotor;
         FrictionMotor frictionMotor;
+        ClimbingMotor climbingMotor;
         Transform myTransform;
 
         private float fallStartLevel;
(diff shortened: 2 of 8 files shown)


Interkarma:
This sounds like its going to be very fun for acrobats. Looking forward to testing it out.

Also happy there's an option for it so hardware classic players can disable. Cheers!

MeteoricDragon:
Anyone who has played the NES classic Ninja Gaiden will appreciate feeling like a ninja climbing the walls.

MeteoricDragon:
I would also like to enable the player to round corners if the wall bends around for inside and outside corners.  I know that not all walls intersect at right angles in DF so the algorithm would need to detect the hit normal of the adjacent wall is at.  If there was a way that I could detect where the wall was when the player moves right or left...

Basically, the way I envision it is when the player moves left or right, a series of x capsulecasts search for nearby walls almost touching the player and if it finds out that the first wall touching the player has the same hit normal that the wall they are climbing does, then it does nothing.  If a different wall is found, it lerp rotates and lerp moves the player to that position somehow.

Interkarma:
That sounds fun, and provided all the advanced stuff is optional and doesn't break basic climbing, I don't have a problem with it. 👍

Interkarma:
BTW, could you have a look at the bug raised in this post as well?

https://forums.dfworkshop.net/viewtopic.php?f=23&t=1228&start=20#p15089

I was just able to launch myself high into the air climbing a flat roofed building in Daggerfall, so it's not just weirdly angled landing points. This is a hard one to reproduce.

roof-launch

MeteoricDragon:
I think I could clamp the movedirection.y to go no higher than the (maximum player speed / 3) * 2.  Which would be the maximum climbing speed times the maximum climbing speed boost bonus.  What is the max player speed?

PR #873, opened 2018-09-13
Support for mod localization

TheLacus:
This PR aims to provide a standard localization interface for mods.

### Concept
A mod called **modname** is automatically linked to a file named **mod_modname.txt** in _StreamingAssets/Text_ and/or a file name **textdatabase.txt** inside the mod bundle. The table in _StreamingAssets_ takes precedence.

### Settings
The file **settings.json** bundled with a mod provides informations used by the mod manager to draw controls in the mod settings window and store values to a local file. This includes setting names and descriptions/tooltips. These text string are seked automatically in the language table using the following patterns:

```
Mod.Description
Settings.SectionName.Name
Settings.SectionName.Description
Settings.SectionName.KeyName.Name
Settings.SectionName.KeyName.Description
Presets.PresetTitle.Title
Presets.PresetTitle.Description
```
If a key is missing, the default value from **settings.json** is used. This means there is no need to provide an english table. No support is required from individual mods, this is all managed by the mod manager.

### Mod Strings
The Mod instance includes a method called `Localize()` which can be used to seek additional localized strings from the text table. The key can be a string or a succession of strings wich are concatenated.
```
string Localize(string key) // key, text
string Localize(params string[] keyParts) // a.b.c, text 
```

`Fountain.Drink, Press key to drink at the fountain.`
`string text = mod.Localize("Fountain", "Drink").`

If a key is not found in the table from the Text folder, is seeked in the fallback table inside the mod.  This means that no additional files other than the dfmod are required for a standard english version. Mod developers only need to export strings to a text file and build it as an asset.

### DIFF
--- a/Assets/Scripts/Utility/Table.cs
+++ b/Assets/Scripts/Utility/Table.cs
@@ -13,6 +13,7 @@ using UnityEngine;
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.IO;
 
 namespace DaggerfallWorkshop.Utility
 {
@@ -87,6 +88,15 @@ namespace DaggerfallWorkshop.Utility
         public Table(string[] lines)
         {
             LoadTable(lines);
+        }
+
+        /// <summary>
+        /// Load constructor.
+        /// </summary>
+        /// <param name="text">Text of data file.</param>
+        public Table(string text)
+        {
+            LoadTable(text.Split(new[] { "\r\n", "\r", "\n" }, StringSplitOptions.None));
         }
 
         #endregion
@@ -372,6 +382,20 @@ namespace DaggerfallWorkshop.Utility
             }
 
             return result;
+        }
+
+        #endregion
+
+        #region Public Helpers
+
+        /// <summary>
+        /// Ensures that a string can be used as a value for a table entry.
+        /// </summary>
+        /// <param name="text">A candidate for an entry value.</param>
+        /// <returns>A safe string.</returns>
+        public static string SanitizeEntryValue(string text)
+        {
+            return text != null ? text.Replace(",", string.Empty) : string.Empty;
         }
 
         #endregion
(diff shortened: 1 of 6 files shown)


Assets/Scripts/Utility/Table.cs
```diff
@@ -376,6 +386,20 @@ public int GetInt(string columnName, string key)
...
         #endregion
 
+        #region Public Helpers
+
+        /// <summary>
+        /// Ensures that a string can be used as a value for a table entry.
+        /// </summary>
+        /// <param name="text">A candidate for an entry value.</param>
+        /// <returns>A safe string.</returns>
+        public static string SanitizeEntryValue(string text)
```
Interkarma: Something I've been meaning to do for a while is add extra intelligence to Table for supporting string literals so that commas don't break the schema. Would it help you if I bring this forward?

Interkarma:
This is an amazing start, well done!

Assets/Scripts/Utility/Table.cs
TheLacus: It would be great! I'm curious to know what you have in mind. I imagine they could be enscaped with a slash (no changes needed to current tables) or the whole string could be put inside quotes (a bit more readable).

Interkarma: Cheers, I'll work on that sooner then. :) My thinking was just to enclose literals in quotes and expect literals to be declared in the schema. For example:

    schema: *id, group ,$notes
    0, widgets, "These are quality widgets, yes really!"

The reason for declaring string literals in schema is that table might otherwise want to use double-quotes for some other reason. I don't want to assume that or we end up with same issue comma has now.

Will that get the job done?

TheLacus: I updated my comment at the same time you posted yours but i got my answer anyway ahah
I agree leaving the choiche to the schema is the safest option 👍

PR #874, opened 2018-09-13
Fix enemy animations

Allofich:
I noticed that there were still several issues with enemy animations, but I think (and hope) this should resolve all the problems.

1) Enemy female thief has a small animation frame in the idle animations when looking diagonally. There used to be a fix for this but I removed it in https://github.com/Interkarma/daggerfall-unity/pull/483 because I noticed the female thief being stretched too large when attacking diagonally. The old fix was for all diagonal animations but it's only the idle animations where the problem is. Also, better than stretching the sprite as the old fix did, there is a correct-size diagonal idle, misplaced among the hurt animations for some reason. Now that is used and it looks correct in-game.

2) In that PR I also removed a fix for the rat, because I noticed it being flipped when it shouldn't. On closer inspection it needed a more customized fix, because it's just the 3 mid-way frames between "towards player" and "away from player" that needed flipping.

 3) I also confirmed wrong-way animations for the werewolf, slaughterfish and scorpion. It also would have been wrongly flipped for other enemy types using "move" animations for their idle that we weren't specifically doing a correction flip on, meaning the wereboar and dragonling.

4) Fixed the ghost and wraith. Both needed the sideways animation for their move/idle and their attack to be flipped.

5) There was a logic error here:

            ```
            // Scorpion Idle and Move animations need to be inverted
            if (summary.Enemy.ID == (int)MobileTypes.GiantScorpion &&
                summary.EnemyState == MobileStates.Idle || summary.EnemyState == MobileStates.Move)
                flip = !flip;
            ```
     
      which was always being true for any moving enemy, which led to the moving animations being defined as needing to be flipped when they didn't really need to be. This is also why !HasIdle enemies using move animations for their idle were seemingly needing to be flipped (I had set giant bat, imp, dreughs and lamias to be flipped this way, because I had noticed them all being reversed, but it was actually all !HasIdle enemy types)

### DIFF
--- a/Assets/Scripts/Internal/DaggerfallMobileUnit.cs
+++ b/Assets/Scripts/Internal/DaggerfallMobileUnit.cs
@@ -390,17 +390,8 @@ namespace DaggerfallWorkshop
             // Check if orientation flip needed
             bool flip = summary.StateAnims[orientation].FlipLeftRight;
 
-            // Some characters' idle animations need to be inverted. Add as needed.
-            if ((summary.Enemy.ID == (int)MobileTypes.GiantBat
-                || summary.Enemy.ID == (int)MobileTypes.Imp
-                || summary.Enemy.ID == (int)MobileTypes.Dreugh
-                || summary.Enemy.ID == (int)MobileTypes.Lamia)
-                && summary.EnemyState == MobileStates.Idle)
-                flip = !flip;
-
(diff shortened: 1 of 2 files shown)


Interkarma:
Thanks for all your hard work cleaning this up. I know how fiddly those animations are. 👍

PR #875, opened 2018-09-13
Fix reading classic magic item names

Allofich:
Items in classic saves should only have their names read up until the null terminator. This is because when classic creates a magic item, it first creates the base, non-magic item, including its name, and then it overwrites the name with the magic item name. Since the magic item names use the %it macro in place of the original names, they can be shorter than the original names, leaving the tail of the original name after the null terminator.

In my case I have a save game with "Cursing Champion straps". The name in SAVETREE.DAT is "Cursing %it aps", because "Cursing %it" was written over the "Champion str" part of "Champion straps". It showed up in Daggerfall Unity as "Cursing Champion straps aps". With this PR it shows correctly.

### DIFF
--- a/Assets/Scripts/API/Save/ItemRecord.cs
+++ b/Assets/Scripts/API/Save/ItemRecord.cs
@@ -125,7 +125,12 @@ namespace DaggerfallConnect.Save
 
             // Read native item data
             parsedData = new ItemRecordData();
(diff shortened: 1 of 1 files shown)


Interkarma:
I never picked that. Thanks for fixing!

PR #877, opened 2018-09-14
Update game settings window

TheLacus:
Gameplay settings window was overflowing so i moved some settings to their own page. Took the occasion to add audio settings. There are some other settings from ini to add, but first i wanted to make sure this quick fix is in the next build.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallAdvancedSettingsWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallAdvancedSettingsWindow.cs
@@ -91,6 +91,10 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
         HorizontalSlider weaponSensitivity;
         HorizontalSlider movementAcceleration;
         TextBox weaponAttackThreshold;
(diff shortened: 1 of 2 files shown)


TheLacus:
ok, added some options and moved around things a bit. Now there is more room for additional settings and the group structure is more similar to the ini than before, to avoid confusion.

I think some tooltips could benefit of a revision to be more descriptive but still short, and it looks like there is a spacing issue for the tab buttons, i'm going to check it.

Maybe you already know it, but the ini itself has some settings out of place: SoundVolume and MusicVolume are under Controls.

TheLacus:
Just a long text overlapping neighbors, not it looks cleaner.

MeteoricDragon:
Thanks, I was going to try and figure out how to organize it but you apparently know it better

Interkarma:
Thank you for working on that. I'll merge and take a look! 👍

Interkarma:
Just one minor thing - the "modSystem" key in GameSettings.txt table is duplicated on lines 19 and 75. Easy enough to fix, but I wasn't sure which one you wanted to keep, so just letting you know. :)

PR #878, opened 2018-09-14
Implement classic music selection

Allofich:
This recreates classic music selection when playing with a General MIDI midi device, since we were so close to it already. I did notice one video on YouTube where someone remarked that the music in Privateer's Hold was different from classic, but it matches with this PR, as should all dungeons.

The music played when in exteriors or taverns is determined by the game day. So a particular date in the game always plays deterministic songs. Temples play a song determined by the faction of that temple. Dungeons play songs depending on a value only being referred to so far as "Unknown2", and the region number they are in. If you load a classic save and compare importing it to DF Unity, you should have the same music playing in all cases, except when it's snowing and possibly when at night or outside a dungeon.

The results aren't exactly the same because what seem to be a couple mistakes in classic are fixed (as they already are in master). One is that general MIDI in classic doesn't play 21.HMI at night or outside dungeons, that entry in the song list is just a duplicate of 10.HMI, even though when using FM MIDI it does use 21FM.HMI for that entry in the list.

The other is that classic does not use SNOWING.HMI, the 4th and last song in the snow music list. When choosing a snow song it only picks from the first 3. Considering that, as with other songs, the FM music list fills out that same slot (by using a duplicate FMOVER_C.HMI entry), and that SNOWING.HMI sounds fine, I think this might have been a mistake in classic. Because of using 4 songs instead of 3, though, the song selection for snowing at daytime doesn't match classic.

### DIFF
--- a/Assets/Prefabs/World/SongPlayer.prefab
+++ b/Assets/Prefabs/World/SongPlayer.prefab
@@ -60,54 +60,69 @@ AudioSource:
   rolloffCustomCurve:
     serializedVersion: 2
     m_Curve:
-    - serializedVersion: 2
+    - serializedVersion: 3
       time: 0
       value: 1
       inSlope: 0
       outSlope: 0
       tangentMode: 0
-    - serializedVersion: 2
+      weightedMode: 0
+      inWeight: 0.33333334
+      outWeight: 0.33333334
+    - serializedVersion: 3
       time: 1
       value: 0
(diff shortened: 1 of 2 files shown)


Interkarma:
This is outstanding, thank you! Perfect music selection in dungeons and in world was one of those things that bothered me a little bit, especially in PH. I'm very happy to see this patch roll into the game.

PR #879, opened 2018-09-14
Initial version of painting info construction

ajrb:
Painting titles match classic.
Artist names don't match classic. 
Image not displayed yet.

### DIFF
--- a/Assets/Scripts/API/PaintFile.cs
+++ b/Assets/Scripts/API/PaintFile.cs
@@ -3,19 +3,14 @@
 // Web Site:        http://www.dfworkshop.net
 // License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
 // Source Code:     https://github.com/Interkarma/daggerfall-unity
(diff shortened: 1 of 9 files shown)


ajrb:
@Interkarma could you check the PaintFile class and its usage please? I copied one of the existing DF file access classes and then stripped most of the code out. I had no idea what i was doing or if what's there correctly fits the pattern you created for the DF file access classes.

Interkarma:
That looks good to me Hazelnut. I'll merge this now and take a look at adding image support to message box over weekend. Great work!

ajrb:
happy for you to suggest how you want it done and then I can implement the image support instead of you so you can keep focussed on fun stuff (magic) - either way is fine with me.

Interkarma:
I just merged a very minor update to add an ImagePanel property to DaggerfallMessageBox. This should at least help you get started as you can control the entire Panel itself.

ajrb:
Cool, I will take it from there then. Thanks. :-)

PR #883, opened 2018-09-16
Implement shoplifting

ajrb:
Guards currently spawn outside the shop due to min spawn distance I think. I like this in one respect that you can't simply flee to the exit and escape like you can in classic, but it does feel like you are getting away with stealing until you leave and see the dozens of guards waiting for you...

Allofich, can you check the formula I added please? I took it from chronicles.  PlayerEntity.SpawnCityGuards() needs more work, as you probably already know.

### DIFF
--- a/Assets/Scripts/Game/Formulas/FormulaHelper.cs
+++ b/Assets/Scripts/Game/Formulas/FormulaHelper.cs
@@ -229,6 +229,18 @@ namespace DaggerfallWorkshop.Game.Formulas
             return Mathf.Clamp(chance, 5, 95);
         }
 
+        // Calculate chance of being caught shoplifting items
+        public static int CalculateShopliftingChance(PlayerEntity player, DaggerfallEntity na, int shopQuality, int weightAndNumItems)
+        {
+            Formula_2de_2i del;
+            if (formula_2de_2i.TryGetValue("CalculateShopliftingChance", out del))
+                return del(player, null, shopQuality, weightAndNumItems);
+
+            int chance = 100 - player.Skills.GetLiveSkillValue(DFCareer.Skills.Pickpocket);
+            chance += shopQuality + weightAndNumItems;
+            return Mathf.Clamp(chance, 5, 95);
+        }
+
         // Calculate how many uses a skill needs before its value will rise.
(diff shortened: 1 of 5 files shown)


Allofich:
> PlayerEntity.SpawnCityGuards() needs more work, as you probably already know.

Yeah, I do know that it's pretty rough. I think classic uses LOS checks with pedestrians outside, which is something I wanted to get around to with that eventually, and also "spawning" guards out of guard civilians.

I'll take a look at the shoplifting formula when I can.

Interkarma:
Thank you for this! 👍

Allofich:
The actual formula used by classic is a little different, but I consider the formula in DF Chronicles superior.

What classic does is compare a 0 to 100 roll against
`pickpocketSkill` - `(cost >> 5)` - `(total weight of items >> 2`, with no clamping between 5 and 95.

This means that, unless I'm missing something, it's impossible to steal anything worth around 3,000 gold or so or more. Sure enough, I tried a couple times with a character with 100 pickpocketing and failed both times.

I think there probably should be a clamp. Not only because the DF Chronicles says it, but also because the other thief skills (lockpicking and pickpocketing) use it, and it seems to be an overall design choice that you never have 0% or 100% success rate. I recommend keeping the formula we have now.

One other thing, is that classic tallies the player's stealth skill on a failed shoplifting attempt. I think we should tally, but the pickpocket skill instead. That's what's actually used in the formula, and with the pickpocket spamming of classic removed, it's not so easy to raise that skill anymore. Stealth raises all the time whether you want to or not, just by being around enemies. I'm not sure whether it should tally for both success and failure or just failure, but doing just for failure kind of makes sense, like you only learn something new if you did something wrong.

ajrb:
UESP says that pickpocket skill is not affected by shoplifting, but I am happy to add it. Generally we also tally successful skill uses though I thought.

Agree about the formula, the class of shop and number+weight seem to be more pertinent than price.

Allofich:
For regular pickpocketing, the tally happens regardless of success or failure. Probably that's the best option (I was just saying I could kind of see a rationale for doing it on failures, but that's probably not intuitive to the player anyway).

Allofich:
> UESP says that pickpocket skill is not affected by shoplifting

Right, it's not affected, stealth is. But it's the skill that is used, not stealth, so that seems kind of off.

PR #885, opened 2018-09-17
Climbing wrap

MeteoricDragon:
Fix climbing teleportation bug by preventing the situation that causes it.  Force the player to finish climbing if almost done climbing, e.g. no mesh collider is in front of him.
Works nicely.

Implements a wrapping around corners system that allows the player to wrap around wall corners.

Debug Rays are still drawn because there may be errors left.  Please only comment them out and don't remove.

Implements luck into the climbing factor.

### DIFF
--- a/Assets/Scripts/DaggerfallUnityEnums.cs
+++ b/Assets/Scripts/DaggerfallUnityEnums.cs
@@ -593,4 +593,20 @@ namespace DaggerfallWorkshop
         QuestSpawn = 11,        // Quest spawn marker (Foe/Person resources)
         QuestItem = 18,         // Quest item marker (Item resource)
     }
+
+    /// <summary>
+    /// Phases of the moons.
+    /// </summary>
+    public enum LunarPhases
+    {
+        None = -1,
+        New = 0,
+        OneWax = 1,
+        HalfWax = 2,
+        ThreeWax = 3,
+        Full = 4,
+        ThreeWane = 5,
+        HalfWane = 6,
+        OneWane = 7,
+    }
 }
--- a/Assets/Scripts/Game/MagicAndEffects/EntityEffectBundle.cs
+++ b/Assets/Scripts/Game/MagicAndEffects/EntityEffectBundle.cs
@@ -24,6 +24,7 @@ namespace DaggerfallWorkshop.Game.MagicAndEffects
         EffectBundleSettings settings;
         DaggerfallEntityBehaviour casterEntityBehaviour = null;
         DaggerfallUnityItem fromEquippedItem = null;
+        bool fromPoison = false;
 
         #endregion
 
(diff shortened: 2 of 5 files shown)


MeteoricDragon:
looks like only 2 duplicate commits this time! good.

MeteoricDragon:
@Interkarma I have an additional Idea.  I'll describe it in pseudocode and let me know if you think it's good.

In ClimbingMotor, In ClimbingCheck

if (playerMovingBackward && !Running, && !isRiding && !isSwimming && isFalling && !isJumping && SomeOtherConditionIDon'tKnowYetButChecksForWallInFrontOfPlayerAfterHeFell)  
Make Player start climbing the wall.

Basically it enables player to climb down a wall after they've already climbed up it :)

Interkarma:
>Basically it enables player to climb down a wall after they've already climbed up it :)

I think that's a decent idea. Considering you've already built the logic to wrap around corners, it makes sense to allow climbing down. Seems like a good fit for the advanced climbing model.

One problem I've always had with this in games (i.e. on ladders) is moving backwards towards an edge makes me feel uncomfortable. It's like a trust exercise where someone has to catch me, but in this case it's cold, uncaring code. :)

MeteoricDragon:
Well, I was thinking change the acrobatic motor so that when the player is backing up and starts falling, make him fall straight down until the climbing system finds the wall, then trigger the start climbing.

The only drawback is that the player won't be able to back off ledges walking and if they back down big steps they might catch them?  I'll have to see what happens.

Interkarma:
As long as it's partitioned into advanced climbing feature set, I'm happy for you to experiment and see where it goes. :)

There's something related I should raise. I'm considering raising the "Anti Bump Factor" on FrictionMotor to 2.5. This will resolve issue where player launches off steep ramps they are supposed to walk down and fall instead. One side effect of this change is that player will fall faster when walking of a vertical edge.

If you try setting "Anti Bump Factor" to 2.5 in FrictionMotor Inspector, you should find the player will automatically fall faster and more vertically hen they walk backwards off a vertical edge. This might keep them close enough to the wall for climbing to kick in based on checks. If that's all you require, then it's another reason for me to make the change besides keeping player glued to those ridiculously steep ramps.

And don't forget the air control toggle. Players might have air control enabled which will make them continue to move away from surface after they walk off it. This would extend the distance requirement to trigger climbing check after walking backwards off a roof or similar.

Interkarma:
I'm just sending this to own branch first for testing. Will merge into master soon. Thanks for all your hard work on this. 👍

MeteoricDragon:
> One problem I've always had with this in games (i.e. on ladders) is moving backwards towards an edge makes me feel uncomfortable. It's like a trust exercise where someone has to catch me, but in this case it's cold, uncaring code. :)

maybe we can make it secure uncaring code then :)

PR #888, opened 2018-09-17
Add race/gender attack and hit sounds

Allofich:
On by default, but can be turned off. Happens a percentage of the time when melee attacking or being hit, both for the player and for enemy classes.

Male sounds are pretty straight-forward, with unique sounds for each race.
Female sounds were a mixture. Some sounds are duplicates, and some races only seemed to have attack sounds or only pain sounds. I spent a fair bit of time listening to the sounds and working out a solution that ends up working pretty well. It basically works out to humans, elves, khajiit and argonians as 4 groups, with various bits of crossover.

Just for the purposes of sounds, each enemy class is randomly assigned a race (Breton through High Elf), to add some variety. This isn't saved or loaded at the moment, and is just intended to be for that character's voice, not to actually serve as the race.

### DIFF
--- a/Assets/Resources/defaults.ini.txt
+++ b/Assets/Resources/defaults.ini.txt
@@ -85,3 +85,4 @@ AlternateRandomEnemySelection=False
 DungeonAmbientLightScale=1.0
 NightAmbientLightScale=1.0
 PlayerTorchLightScale=1.0
+CombatVoices=True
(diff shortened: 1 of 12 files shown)


Interkarma:
I can see you've put a lot of work into this. Looking forward to playing with it.

PR #889, opened 2018-09-17
 Implement selling magic items at fence in TG 

ajrb:
Also disabled jumping with cart which I always intended to do.

@Allofich It currently re-uses the same calculation for cost. However UESP states "Appraisers will buy magical items from any member who is at least a Filcher in the Guild. They offer higher base prices than other merchants. The final price for your items is affected by your reputation with the faction." This implies that the function will need guild rank to be passed instead of building quality. If you give me the equation for the formula I'm happy to implement it... you do seem to have a few plates spinning right now. :-)


Interkarma:
Looks good, thank you! 👍

PR #890, opened 2018-09-18
Fixes

Allofich:
Fixes poison enum problems. Sorry about that.

Also fixes/adjustments to the combat voices. I had left off dark elf female in one part. Also I switched out male high elf on NPCs for male wood elf. The high elf sound sounds a little "breathing into the mic"-ish and might be okay for 1st person but doesn't sound very good coming from third-person.

Adjusted the "heavy damage" female pain sound to play based on character's max health compared with damage instead of looking only at the damage amount.

I might later add sounds for civilians, some non-class enemies, etc. I'll have to check how it feels in-game.

### DIFF
--- a/Assets/Scripts/Game/EnemySounds.cs
+++ b/Assets/Scripts/Game/EnemySounds.cs
@@ -69,7 +69,7 @@ namespace DaggerfallWorkshop.Game
 
             RaceForSounds = (Entity.Races)Random.Range(1, 6);
 
-            // Start attrack timer
+            // Start attract timer
             StartWaiting();
         }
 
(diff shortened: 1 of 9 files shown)


Assets/Scripts/Game/EnemySounds.cs
```diff
@@ -69,7 +69,7 @@ void Start()
 
             RaceForSounds = (Entity.Races)Random.Range(1, 6);
 
-            // Start attrack timer
+            // Start attract timer
```
ajrb: Is this some kind of new romance option? You have 10 seconds to attract your enemy... lol

(hope you find my poking fun to be funny and not obnoxious)

Allofich: Hey, that wasn't even me who put that comment in :) I just figured I might as well fix the spelling error.

ajrb: Well it is now a correctly spelled English word. I suspect it should be attack rather than attract though - that I what I was pointing fun at.

ajrb: Oh, okay. well my 'joke' was really lame then. lol

Allofich: Haha, no worries :)

Interkarma:
Thank you for the fixes! I was playing with combat vocalisations last night and found myself really enjoying the feature.

I've always found the khajiit roar/purr/growl blended with a human voice a bit disturbing, but that's nothing to do with your work. :)

Assets/Scripts/Game/EnemySounds.cs
Interkarma: This looks like one of my little idiosyncrasies / typos.

Interkarma:
@Allofich are you happy for me to start livening up poisons with effect system, or do you want to do more work first?

Allofich:
I wasn't planning on doing anything else with the poisons right now. 

The only other thing that I had at all in mind, is that there are two "DaggerfallAction" types, not in DF Unity yet, one of them that tries to give the player a random disease and one that tries to give them a random poison (any of the 12, both weapon poisons and drugs) , but I don't know if or where these are actually used in the game. To get these in it would be nice to have some simple function that can be called to just try to inflict one of those on the player, so maybe keep that in mind if it is relevant to your design.

Allofich:
Oh, I guess we'll also want to be able to import poisons and diseases (disease was working but I think it's disabled right now) from classic saves, not that it's imperative, but it's nice to have. So, maybe another thing to keep in mind.

Interkarma:
>I wasn't planning on doing anything else with the poisons right now.

Thank you mate, I'll merge my branch over and finish up in master then.

>Oh, I guess we'll also want to be able to import poisons and diseases (disease was working but I think it's disabled right now) from classic saves, not that it's imperative, but it's nice to have. So, maybe another thing to keep in mind.

This should be easy to implement. For each disease:

    Diseases diseaseType = <required disease>;
    EntityEffectBundle bundle = GameManager.Instance.PlayerEffectManager.CreateDisease(diseaseType);
    GameManager.Instance.PlayerEffectManager.AssignBundle(bundle);

And for each poison:

    Poisons poisonType = <required poison>;
    EntityEffectBundle bundle = GameManager.Instance.PlayerEffectManager.CreatePoison(poisonType);
    GameManager.Instance.PlayerEffectManager.AssignBundle(bundle);

If we import a classic save with a disease in progress, will the attributes imported reflect the pre-disease values or the current value as modified by disease? We might need some special handling to ensure we import the correct base values for attributes as unmodified by diseases.

>The only other thing that I had at all in mind, is that there are two "DaggerfallAction" types, not in DF Unity yet, one of them that tries to give the player a random disease and one that tries to give them a random poison (any of the 12, both weapon poisons and drugs) , but I don't know if or where these are actually used in the game. To get these in it would be nice to have some simple function that can be called to just try to inflict one of those on the player

Happy for this to go in whenever, but maybe don't give it priority until we can find where this happens in game.

>That reminds me, the CastSpell DaggerfallAction is one that we might want pretty soon. There are some main quest locations that have objects that cast Levitate on the player (when leaving Scourg Barrow, in the Mantellan Crux, maybe more), helping out players without that ability.

Agreed. This is something I'd like to get in shortly after the quest actions related to spells (which I want to tackle in a few weeks at most).

Right now just trying to plough through all those spell effects. :)

PR #892, opened 2018-09-18
Climbing speed refactor

MeteoricDragon:
Quick refactor of climbing speed to PlayerSpeedChanger

### DIFF
--- a/Assets/Scripts/Game/Player/ClimbingMotor.cs
+++ b/Assets/Scripts/Game/Player/ClimbingMotor.cs
@@ -17,6 +17,7 @@ namespace DaggerfallWorkshop.Game
         private CharacterController controller;
         private PlayerEnterExit playerEnterExit;
         private AcrobatMotor acrobatMotor;
+        private PlayerSpeedChanger speedChanger;
         private bool overrideSkillCheck = false;
         private bool isClimbing = false;
         private bool isSlipping = false;
@@ -87,6 +88,7 @@ namespace DaggerfallWorkshop.Game
             controller = GetComponent<CharacterController>();
             playerEnterExit = GetComponent<PlayerEnterExit>();
             acrobatMotor = GetComponent<AcrobatMotor>();
+            speedChanger = GetComponent<PlayerSpeedChanger>();
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


MeteoricDragon:
@Interkarma I thought of something,  If I give players the ability to climb down from a platform, they will be able to climb down on moving platforms.  And I'm pretty sure if they attempt that... Not sure what would happen.

Interkarma:
They'd possibly just come unstuck, the moving platform check is done based on their grounded state. Could always check for static flag and disallow climbing mobile mesh objects.

But yeah, lots of edge cases to worry about this one. Nobody is going to be unhappy if that feature doesn't make it in BTW. You've done great work on this already, much more than we had in classic. :)

MeteoricDragon:
I'm going to investigate making the playermotor more attachment-friendly so that levitation doesn't cause the player to pass through platforms.  Are there any platforms in Daggerfall that move horizontally?  The only one that I know of is one that moves vertically in Privateer's Hold.

Interkarma:
Yes there's a few. Once particular dungeon block has an entrance opening to a large area with a horizontally moving platform up high. N0000025.RDB is one example. This is often found right near the entrance. Just need to find a dungeon with this block.

In the meantime, you can view that block in Daggerfall Modelling and trigger the action by mouse over (action objects have a blue outline) and pressing R to activate.

MeteoricDragon:
Best case scenario all forms of movement will be compatible with moving platforms.

Interkarma:
I'll definitely snag a save if I can find one. I might even have one in the 400-odd saves I've created at different stages of testing.

Saves are in the persistent data path in a Saves folder. This is the path shown at the bottom of startup UI before you hit Play. On windows this is:

    C:\Users\Username\AppData\LocalLow\Daggerfall Workshop\Daggerfall Unity\

And then inside Saves. You'll see the save index on your save/loadUI at top-right of screenshot. The Bug Report Guidelines topic has all of this too.

https://forums.dfworkshop.net/viewtopic.php?f=24&t=313

Interkarma:
Below a save where player is standing on a horizontally moving platform. To trigger the platform, switch to editor view and select the object below player, then hit the "Activate" button in Inspector for Daggerfall Action component. I've also added a screenshot with everything highlighted.

Use console with "levitate on" and "tgm" to set yourself up. Or "showspellmaker" to create an actual spell. Good luck.

horz-platform

[SAVE425.zip](https://github.com/Interkarma/daggerfall-unity/files/2403341/SAVE425.zip)

MeteoricDragon:
Thanks, I'll continue this on this topic: https://forums.dfworkshop.net/viewtopic.php?f=23&t=1228&p=15317#p15317

PR #897, opened 2018-09-20
SDF support for exterior automap

Nystul-the-Magician:
implemented font rendering with textlabel class for exterior automap - supports SDF font rendering now
there is one bug with tooltips - their text is not rendered if tooltip is too big for automap window (background is rendered though) - I am looking into this but it is a bit tricky - related to restrictedRenderArea but could not find the reason for it

hopefully  this pull request works - I tried to rebase before issuing it - this is my first try to rebase and had to create a new local branch since git would not allow me to push it into existing local branch - I don't understand it quite right - but it looks like it should be ok with only one commit now

### DIFF
--- a/Assets/Scripts/Game/ExteriorAutomap.cs
+++ b/Assets/Scripts/Game/ExteriorAutomap.cs
@@ -1,4 +1,4 @@
-﻿// Project:         Daggerfall Tools For Unity
+// Project:         Daggerfall Tools For Unity
 // Copyright:       Copyright (C) 2009-2018 Daggerfall Workshop
 // Web Site:        http://www.dfworkshop.net
 // License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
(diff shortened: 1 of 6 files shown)


Interkarma:
Thank you Nystul. :)

I'll just pop this in a branch for now. It's hard to see the actual changes in this one, for some reason the files appear to be completely rewritten.

Will review as I can.

Nystul-the-Magician:
I also saw this - looks like it is the result of rebasing everything before issuing the pull request.
would you like me to do a normal merge instead of rebase - I think the history would be present then but there are some commits with comments "progress made" where I was too lazy to write more detailed explainations. I have a branch from before rebasing so this would be no problem.

It is a good idea to test this and play around with it in a branch on your side!

Interkarma:
That would be great if you could mate. I don't mind all the minor history items, it's mainly so I can see changes properly. Cheers! :)

PR #898, opened 2018-09-20
Daedra summoning initial implementation 

ajrb:
Daedra summoning is working but doesn't play the vid files and uses normal message boxes instead for the time being. The new class DaggerfallDaedraSummonedWindow is currently unused. It's a stub which will play the video and display text and accept Y/N entry - this will replace the message boxes.

Also:
* Fixed all of the Daedra Lord/Lady quests.
* Added pickpocket skill tally for shoplifting.
* Tidied up macrosExpandedAlready mechanism.
* Added faction flags accessor methods and associated enum.
* Fixed static message box helper methods to use mcp's.
* Fixed quest completion macro expansion bug.


Interkarma:
Wow epic PR! Thank you mate, I'll work through this when I can. :)

ajrb:
No rush, I'm going to be away most of the coming week.

Assets/Scripts/Game/Entities/PlayerEntity.cs
```diff
@@ -1404,7 +1404,7 @@ public void RegionPowerAndConditionsUpdate(bool updateConditions)
                         }
 
                         // Chance for new ruler
-                        if ((factionData.FactionDict[key].flags & 0x10) == 0)
+                        if (factionData.GetFlag(key, FactionFile.Flags.RulerImmune))
                         {
```
Allofich: This flag check seems reversed. The code in this if() should run if the 0x10 flag is not set, not if it is.

ajrb: Crap, sorry mate. You're quite right, glad you spotted that! Will fix asap.

PR #900, opened 2018-09-21
sdf font rendering support for exterior automap (no rebase, just merge - still too much for diff how it seems)

Nystul-the-Magician:
sdf font rendering support for exterior automap
no rebase, just merge - still too much for diff how it seems
anyway I fixed the tool tip issues mostly (except that I have this question (I use defaultToolTip for the moment): https://forums.dfworkshop.net/viewtopic.php?f=23&t=1281

**I would suggest to test and play around in a branch for now ;) at least until we are happy with the tooltip issues mentioned below **

worth noting are these 2 changes to ToolTip class:
line 157-160 was necessary to fix this issue: https://imgur.com/31aMknB
fixed it looks like expected: https://imgur.com/23eNChC

line 140-145 fixes this issue: https://imgur.com/tbvOsd4

I tested and it seems to work ok with tooltips in other windows
there is this issue: https://imgur.com/LiUTYs7 -
a leftover after tool tip hides again but tooltip was displayed outside its parent panel - I might need your help here - but I experience the issue also without the changes to ToolTip class
another problem is that when the automap is open the version string in the top right screen corner (or better said the part outside the automap panel) is sometimes drawn, sometimes not - see screenshots above - the "flickering" is coming from my changes, but before it was always cut (which is also suboptimal)


Interkarma:
I'd really love to see a proper diff, if possible. I can't see what's actually changed, it looks like the whole file has been replaced. :(

Nystul-the-Magician:
don't know why this happened - maybe too many changes at one time in a commit?
will try to manually create a commit series where not the whole file gets replaced...

PR #902, opened 2018-09-22
Fix LevitateMotor Collisions via hack

MeteoricDragon:
Refactor IsStandngStill to a property.
make hitting head bounceback an acrobatic method.
Rename a couple local variables in PlayerGroundMotor

Hacks the inability of CharacterController to detect collisions while stationary by moving him a little tiny bit up, down, left, right, forward, back directions an imperceptible amount by calling the movement function in groundmotor.  Now a moving platform can come from any direction and push against the player and it will move the player appropriately.


Interkarma:
Cheers! I've pushed this into a branch for testing. Will merge it to master soon.

Interkarma:
The vibration hack works a treat. I haven't been able to produce any unintended side effects and player is pushed by platforms as expected. I'm really happy with how this is working now. Great job! :)

PR #903, opened 2018-09-22
sdf font rendering support for exterior automap (hopefully with history now

Nystul-the-Magician:
sdf font rendering support for exterior automap (hopefully with working history now)

I fixed the tool tip issues mostly (except that I have this question (I use defaultToolTip for the moment): https://forums.dfworkshop.net/viewtopic.php?f=23&t=1281

**I would suggest to test and play around in a branch for now ;) at least until we are happy with the tooltip issues mentioned below **

worth noting are these 2 changes to ToolTip class:
line 157-160 was necessary to fix this issue: https://imgur.com/31aMknB
fixed it looks like expected: https://imgur.com/23eNChC

line 140-145 fixes this issue: https://imgur.com/tbvOsd4

I tested and it seems to work ok with tooltips in other windows
there is this issue: https://imgur.com/LiUTYs7 -
a leftover after tool tip hides again but tooltip was displayed outside its parent panel - I might need your help here - but I experience the issue also without the changes to ToolTip class
another problem is that when the automap is open the version string in the top right screen corner (or better said the part outside the automap panel) is sometimes drawn, sometimes not - see screenshots above - the "flickering" is coming from my changes, but before it was always cut (which is also suboptimal)


Interkarma:
That looks so much clearer, thank you mate. :)

I'll work through this as I can and get merged into another branch so we can solve remaining issues.

Assets/Scripts/Game/UserInterface/DaggerfallFont.cs
```diff
@@ -134,19 +139,6 @@ public DaggerfallFont(string arena2Path, FontName font = FontName.FONT0003)
 
         #region Glyph Rendering
 
-        public int GetSDFGlyphWidth(int ascii)
```
Interkarma: Please don't remove this. It's part of the DaggerfallFont feature set. Just because nothing needs it right now doesn't mean that it shouldn't be available. :)

Nystul-the-Magician: ah sry, thought I added this one, since I wrote several similar helper functions when I first tried to use old Texture2D approach

Interkarma: All good! I didn't have it in the right spot anyway, so it looked out of place. :)

PR #906, opened 2018-09-23
Implement AI target selection and infighting

Allofich:
This is one I was looking forward to doing, and although it could still do with improvement, mainly in enemy pathfinding, it works pretty well. Please give it some testing, though, as it touches a lot of things and could break something. I think I've fixed all the problems I encountered. There's a lot of raycasting going on but I've tried to avoid too much FPS drop.

I worked on this to prepare for the Sanguine Rose artifact, which allows the player to summon a Daedroth, but I ended up implementing an "enemy infighting" mode that I think is fun to play with.

What's in the PR:
1) Enemies now evaluate targets and choose one, instead of always targeting the player. (Unless "enemy infighting" is turned on, they reject all targets but the player. Later I intend to figure player allies into this, to allow the Sanguine Rose to work). Target selection is based on classic, then tweaked to get better results.
2) Optional "enemy infighting" mode (default on), where enemies of differing "teams" (don't really like that name, would like to call them "factions" but that has already has a different meaning) will attack each other if they see each other. They are grouped so similar enemies (if there are any) are on the same team.
3) Enemies can hit each other now, not just the player.
4) Pacified/guard types will ignore the player but, if "enemy infighting" is on, still fight any enemy teams.
5) Weapon-wielding AI will use hand-to-hand if enemy is immune to their weapon.
6) New console command "nt" activates "NoTarget" mode, where enemies will ignore player.
7) Fixed a mistake in the "wouldBeSpawnedInClassic" code that sometimes caused enemies to wrongly get a "false".

### DIFF
--- a/Assets/Game/Addons/UnityConsole/Console/Scripts/DefaultCommands.cs
+++ b/Assets/Game/Addons/UnityConsole/Console/Scripts/DefaultCommands.cs
@@ -31,6 +31,7 @@ namespace Wenzil.Console
             ConsoleCommandsDatabase.RegisterCommand(HelpCommand.name, HelpCommand.description, HelpCommand.usage, HelpCommand.Execute);
             ConsoleCommandsDatabase.RegisterCommand(LoadCommand.name, LoadCommand.description, LoadCommand.usage, LoadCommand.Execute);
             ConsoleCommandsDatabase.RegisterCommand(GodCommand.name, GodCommand.description, GodCommand.usage, GodCommand.Execute);
+            ConsoleCommandsDatabase.RegisterCommand(NoTargetCommand.name, NoTargetCommand.description, NoTargetCommand.usage, NoTargetCommand.Execute);
             ConsoleCommandsDatabase.RegisterCommand(Suicide.name, Suicide.description, Suicide.usage, Suicide.Execute);
             ConsoleCommandsDatabase.RegisterCommand(ShowDebugStrings.name, ShowDebugStrings.description, ShowDebugStrings.usage, ShowDebugStrings.Execute);
             ConsoleCommandsDatabase.RegisterCommand(SetWeather.name, SetWeather.description, SetWeather.usage, SetWeather.Execute);
@@ -228,6 +229,26 @@ namespace Wenzil.Console
                     PlayerEntity playerEntity = GameManager.Instance.PlayerEntity;
                     playerEntity.GodMode = !playerEntity.GodMode;
                     return string.Format("Godmode enabled: {0}", playerEntity.GodMode);
+                }
+                else
+                    return error;
+            }
+        }
+
+        private static class NoTargetCommand
+        {
+            public static readonly string name = "nt";
+            public static readonly string error = "Failed to set NoTarget mode";
+            public static readonly string usage = "nt";
+            public static readonly string description = "Toggle NoTarget mode";
+
+            public static string Execute(params string[] args)
+            {
+                PlayerEntity playerEntity = GameManager.Instance.PlayerEntity;
+                if (playerEntity != null)
+                {
+                    playerEntity.NoTargetMode = !playerEntity.NoTargetMode;
+                    return string.Format("NoTarget enabled: {0}", playerEntity.NoTargetMode);
                 }
                 else
                     return error;
--- a/Assets/Resources/defaults.ini.txt
+++ b/Assets/Resources/defaults.ini.txt
@@ -86,3 +86,4 @@ DungeonAmbientLightScale=1.0
 NightAmbientLightScale=1.0
 PlayerTorchLightScale=1.0
 CombatVoices=True
+EnemyInfighting=True
--- a/Assets/Scripts/DaggerfallUnityEnums.cs
+++ b/Assets/Scripts/DaggerfallUnityEnums.cs
@@ -255,6 +255,34 @@ namespace DaggerfallWorkshop
     }
 
     /// <summary>
+    /// Mobile teams.
+    /// </summary>
+    public enum MobileTeams
+    {
+        PlayerAlly,
+        Vermin,
+        Spriggans,
+        Bears,
+        Tigers,
+        Spiders,
+        Orcs,
+        Centaurs,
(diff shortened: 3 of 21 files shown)


Interkarma:
I love this! It looks good on a read. Will pop it in a branch and play for a while to see if I can find any other issues.

Allofich:
The main issue I can foresee for this right now is that DFUnity quests that create enemies of varying types in the same space will have the issue of them fighting each other. For example, in your quest with the necromancer at a cemetery, they could end up fighting with the undead. I don't think any Bethesda quests will have much of an issue, because they usually (always?) send only a single enemy type at the player at a time.

For now, workaround is to disable infighting for these quests (or allow it if you're just messing around and want to watch them fight). Adding the ability for quests to specify the "team" of a spawned enemy might solve this.

Interkarma:
That makes sense. I'll play with that idea when I can. I take it that enemies in dungeons from different teams won't spontaneously attack each other? Although that does seem kind of awesome. I'm itching to try a battle royale with as many enemies as possible to load test this system - I just think it's cool as hell.

While we're talking enemy AI, are you interested in setting up enemy spellcasters soon? This is basically ready to go, possibly minus a couple of effects. I'm happy to provide any help if needed, just let me know. :)

Allofich:
> I take it that enemies in dungeons from different teams won't spontaneously attack each other?

They do spontaneously attack each other. However, the way it works, they only start picking targets when the player is near and only look nearby, so the fighting shouldn't be excessive, you'll just occasionally find them fighting and can also lure opposing types near to each other and get them to fight each other.

> I'm itching to try a battle royale with as many enemies as possible to load test this system - I just think it's cool as hell.

While working on it I was testing by having the castle guards all be put onto different teams and making them all fight each other. It was pretty cool. I thought about putting in a cheat console command for fun that would randomly assign all AI to different teams so they'd all start fighting each other. I kind of want people to first have the fun of naturally running across a few battles or seeing opponents attack each other instead of gorging on a cheat command right away and losing the novelty, though. :)

> While we're talking enemy AI, are you interested in setting up enemy spellcasters soon? This is basically ready to go, possibly minus a couple of effects. I'm happy to provide any help if needed, just let me know.

Definitely, would love to get this in. I briefly looked at the spell missile code but there was a TODO comment about needing to disassociate it with the player so that AI could use it, so I thought maybe it still wasn't ready yet.

In `EntityEffectManager.cs`

```
            // TODO: Split missile generation from player spell casting so monsters can also cast spells
            // Using player as sole testing platform for now
```

Interkarma:
>They do spontaneously attack each other. However, the way it works, they only start picking targets when the player is near and only look nearby, and while testing I felt that the amount of infighting wasn't excessive, you'll just occasionally find them fighting and you can lure opposing types near to each other and get them to fight each other.

That's cool. I think it adds some more interesting behaviours to the mobs. We'll see what the feedback is like.

>I kind of want people to first have the fun of naturally running across a few battles or seeing opponents attack each other instead of gorging on a cheat command right away and losing the novelty, though. :)

I'll let people discover this one themselves if you like. It will be a surprise to anyone not following comments on git. Although, I'm dying setup something huge outdoors with hundreds of random enemies going at it. I don't think I'll be able to help myself. :)

>Definitely, would love to get this in. I briefly looked at the spell missile code but there was a TODO comment about needing to disassociate it with the player so that AI could use it, so I thought maybe it still wasn't ready yet.

It only needs a small amount of work. I'll get this into shape and kick things off in a new branch with a couple of basic example to grease the wheels for you.

Interkarma:
This is awesome! I kited skeleton near the throne lift in PH around the corner into orc room and watched the two of them go at it. This may not be in classic, but it adds a whole new dynamic to the game.

I did receive an exception when attacking skeleton after it killed the orc. It just looks like a string.format issue missing one or more args. Haven't chased it down yet though.

> FormatException: Index (zero based) must be greater than or equal to zero and less than the size of the argument list.
System.String.FormatHelper (System.Text.StringBuilder result, IFormatProvider provider, System.String format, System.Object[] args) (at /Users/builduser/buildslave/mono/build/mcs/class/corlib/System/String.cs:1925)
System.String.Format (IFormatProvider provider, System.String format, System.Object[] args) (at /Users/builduser/buildslave/mono/build/mcs/class/corlib/System/String.cs:1873)
System.String.Format (System.String format, System.Object[] args) (at /Users/builduser/buildslave/mono/build/mcs/class/corlib/System/String.cs:1868)
UnityEngine.DebugLogHandler.LogFormat (LogType logType, UnityEngine.Object context, System.String format, System.Object[] args) (at C:/buildslave/unity/build/Runtime/Export/Logger/DebugLogHandler.cs:9)
UnityEngine.Logger.LogFormat (LogType logType, System.String format, System.Object[] args) (at C:/buildslave/unity/build/Runtime/Export/Logger/Logger.cs:113)

Allofich:
> I'll let people discover this one themselves if you like. It will be a surprise to anyone not following comments on git. Although, I'm dying setup something huge outdoors with hundreds of random enemies going at it. I don't think I'll be able to help myself. :)

I'm fine with you letting people know about it or doing a giant battle or something, I was just talking about what I considered when thinking about putting in a cheat code like that.

Allofich:
Another potential problem I thought about, though, and one that could affect Bethesda quests, is quest target mobs being attacked by other mobs. I wonder if the "on-hit" quest event will trigger when this happens. That, and if the target mob dies, could lead to some out-of-place messages happening.

Don't know how much of an issue this is, but one sort of hackish solution would be to put any quest-placed mobs onto a special team that is ignored by and ignores other teams. Perhaps, that would be the default, such as for Bethesda quests. In custom quests, you could also maybe specify a team.

There might be other, better solutions, too.

Interkarma:
That's a good point, thank you. The InjuredFoe quest action only checks that target foe was injured, not how. I agree with you that foes involved with quests should be excluded from infighting for now.

One easy way to do this is just check if enemy has the QuestResourceBehaviour component added. This is only assigned to enemies and objects involved in quests. If this component is present then enemy should not become involved with infighting system as aggressor or target.

PR #910, opened 2018-09-24
Merge sdf_automap_new to master

Interkarma:
I'm happy with this to be in master now. I'll tackle the version text issue soon.

@Nystul-the-Magician Thanks for all the work on this, it looks great! 👍


Nystul-the-Magician:
cool! the positive side effect of the switch to textlabel class instead of Texture2D is that I was able to remove a lot of code - it should be easier for others to understand what is going on now

PR #912, opened 2018-09-25
Misc. fixes

Allofich:
1) Quest AI will no longer target or be targeted by other AI. This solves the issue for now but I think we should ideally allow for spawning quest AI that can also fight each other.
2) Expanded AI to target if it can see player, as it could before, instead of limiting it to the classic spawn range. This is because the spawn is pretty small vertically, and sometimes there could be enemies in plain view that wouldn't react. I don't really like tying the AI "activating" it's target-seeking according to having the player in sight, but it works well enough. I also tried just using a wider radius around the player but then enemies were fighting a little too much and killing each other out of view of the player so I settled on player LOS for now, which is what the old behavior before the "infighting" PR was anyway.
3) Fixed issue with AI playing combat sounds, miss sounds at wrong timing.
4) Fixed referring to wrong gender value for doing AI voices, which caused mismatches between appearance and voice.
5) Fixed not playing AI hurt sounds when AI hit each other.
6) Classic only uses gender flag for fixed spawns. Doing same now for DF Unity.

### DIFF
--- a/Assets/Scripts/Game/EnemyAttack.cs
+++ b/Assets/Scripts/Game/EnemyAttack.cs
@@ -61,9 +61,6 @@ namespace DaggerfallWorkshop.Game
 
         void Update()
         {
-            if (entityBehaviour.Target == null)
-                return;
-
             // If a melee attack has reached the damage frame we can run a melee attempt
             if (mobile.DoMeleeDamage)
             {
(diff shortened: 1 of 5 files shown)


Interkarma:
Thank you for continuing work on this. Looks good on a read. I'll merge to master and play around with it some more. 👍

PR #913, opened 2018-09-25
Climb down from top2

MeteoricDragon:
Simply back up from any ledge at any angle and you'll automatically attach yourself to the wall below via a C-shaped movement


MeteoricDragon:
In the process of creating this, I discovered a bug that if you walk off a ledge, you'll fall faster than if you jumped off the ledge.

MeteoricDragon:
There is an issue.  If you tap backwards and you've released the backwards key by the time the player starts falling, the IsRappelling boolean won't become true, and the player will back off the ledge and fall.  Try it and see what I mean.  I don't know how to counter that problem.

Interkarma:
> In the process of creating this, I discovered a bug that if you walk off a ledge, you'll fall faster than if you jumped off the ledge

This one is not a bug. :) It's a side-effect increasing antibump setting that keeps player glued to those ridiculously steep downramps in dungeons so they don't fall off while running down them. Since we can't fix the mad geometry, we have to resolve in other ways through simulation. Unfortunately this increases drop speed for now. I mentioned this one to you a little while back when you first raised this feature idea.

>There is an issue. If you tap backwards and you've released the backwards key by the time the player starts falling, the IsRappelling boolean won't become true, and the player will back off the ledge and fall. Try it and see what I mean. I don't know how to counter that problem.

Not sure it's a problem. It just means player can still tap backwards off ledge if they want to drop and hold down key if they want to climb. Seems like a feature to me. :)

Interkarma:
This works well! It does seem to introduce another teleport bug, though not as bad as the first one. Steps to reproduce:

1. Climb to top of building.
2. Rappel backwards from top of building and keeping holding down backwards key all the way to ground.
3. Hit jump to detach from wall, and player is teleported across town.

I was able to make that happen every time, so hopefully an easy one to repro and fix. Great work! 👍

MeteoricDragon:
I think that in order to fix this I'm going to have to use a raycast to measure distance from the ground while moving down, and if it's something small like ... 0.3f, I'll prematurely drop the player down to avoid that problem edge, because whatever it is, it's acting as unforgiving as the other teleportation bug when nearly finishing climbing.  I'll just have to figure how I want to reuse some existing code to accomplish the task, without making the system more confusing.

Interkarma:
Sounds good. :)

One other minor thing, I noticed player is able to start rappel when Advanced Climbing is disabled (then they drop to ground). I feel that purists probably wouldn't want the rappel to even start with classic climbing.

But apart from those, your climbing system is working really well. I'm having a total blast exploring the world in this new way, and I'm eyeing off every vertical surface in dungeons thinking where I can climb to. It's seriously great fun. 👍

PR #914, opened 2018-09-27
Climb down from top2

MeteoricDragon:
Fixes climbing teleportation upon hitting the ground.  Also makes you dismount wall when close enough to the ground.

Fixed bugs where rappel and grasp wall in air would still work even if disabled in settings


MeteoricDragon:
Resolved a minor conflict that wasn't really a conflict at all, but github said it was.

Interkarma:
Thank you for fixes. I'll test and let you know any feedback.

MeteoricDragon:
It appears that rappel mode triggers when not even needed,   If you back up off tables, or thrones, or down stairs sometimes even.  I'll fix that.

PR #915, opened 2018-09-27
Biography text

numidium:
This enables the "History" button on the character sheet and allows you to see your character's backstory.


Interkarma:
Exciting! Looking forward to testing this out. Thank you for all your work on this. 👍

Interkarma:
@numidium I think this might have introduced a minor bug. When starting a new character and selecting Custom class, the biography questionnaire will throw an exception on either option (fast start or answer questions). I've created a bug report on forums with exception details. Fortunately easy to reproduce.

https://forums.dfworkshop.net/viewtopic.php?f=24&t=1304

PR #916, opened 2018-09-28
Fix quest giver greeting context for guilds + new mod hooks

ajrb:
Nystul should check this and give his blessing before is merged, but I'm pretty sure this is correct.


Nystul-the-Magician:
thanks for fixing
changes looking good to me ;)

ajrb:
Added some functions for mods:
* GetNumItems() added to ItemCollection
* OnTrade event added to Trade window
(needed for Archaeologists Guild)

Interkarma:
Everyone seems happy, will get this merged now. Great work mate. 👍

PR #917, opened 2018-09-30
Match name generation closer to classic

Allofich:
Changes:

1) Painting author generation matches classic.
2) Change to name generation
- Name sets re-identified to what they appear to be in classic
= Khajiit names supported
= Imperial names moved in order to "Argonian" index, as in classic
= "Noble" is a monster type
NameGen.txt edited to match above and match ordering of `NAMEGEN.DAT`
- Name generation rules matched to classic
= Changes to Redguard and Nord names (end in "sen"), and monster names (including nobility titles)
- Should always be possible to get the same names as classic now as long as the same seed is provided to DFRandom.rand()

Notes:
- I'm not sure how classic picks monster names for quests. For now I set it to choose from Monster1 or Monster2, and not the nobility titles, which are in Monster3, since nobility titles obviously may not fit for all cases. I don't know if classic quests somehow specify the namebank, I'll try to look into that. I just tried checking the mummy wrapping quest in classic by repeatedly talking to the questgiver. It came up twice and both times it was a name with a nobility title.

- Also, the monster name code in classic includes taking gender into effect, which is replicated here. I think, though, classic just always passes in "male" for monster names, so the female name endings are never used. Would be a nice enhancement to support this in DF Unity.

- Names of NPCs in interiors still don't match yet.

### DIFF
--- a/Assets/Resources/NameGen.txt
+++ b/Assets/Resources/NameGen.txt
@@ -332,193 +332,7 @@
             }
         ]
     },
-    "Noble": {
-        "setCount": 4,
-        "sets": [
-            {
-                "setIndex": 0,
-                "parts": [
-                    "Lord",
-                    "Count",
-                    "Baron",
-                    "Viscount",
-                    "Marquis",
-                    "Duke"
-                ]
-            },
-            {
-                "setIndex": 1,
-                "parts": [
-                    "Vlad",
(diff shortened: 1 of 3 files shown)


Interkarma:
Nice one! Thank you for your work on this.

PR #922, opened 2018-10-01
Fix new custom class bug

numidium:
It was a simple fix. I changed some stuff around when adding the player backstory and neglected the poor custom class.


Interkarma:
Thank you for the quick fix!

PR #923, opened 2018-10-02
Motor fixes

MeteoricDragon:
Fixes 2 bugs

Rappelling no longer happens if you step back off short heighted walls or objects.
Anti-bump gravity now is used when going down slopes, but not when walking off cliffs

Does this by Creating a small component class for playerAdvanced called PlayerStepDetector whose value is accessed by AcrobaticMotor and ClimbingMotor.  The component class detects what the player is about to step onto by measuring the distance of RaycastHit by SphereCast downward.  The origin of this ray is offset by the player Movement vector so it detects the distance to the floor that the player is stepping onto instead of the floor directly beneath.  The component class is for efficiency and code reuse, Like VitalsChangeDetector.  Kills two birds with one stone.  Poor birdies.  Good stone.

PlayerAdvanced.prefab was auto generated with some code changes that I'm not certain are neccessary, @Interkarma please tell me if   This change is neccessary: 
```
systemTimerUpdatesPerSecond: 0.055
systemTimerUpdatesDivisor: 0.0549254
```

### DIFF
--- a/Assets/Prefabs/Player/PlayerAdvanced.prefab
+++ b/Assets/Prefabs/Player/PlayerAdvanced.prefab
@@ -41,6 +41,7 @@ GameObject:
   - component: {fileID: 114220683863726122}
   - component: {fileID: 114033680799267548}
   - component: {fileID: 114397629012878356}
+  - component: {fileID: 114025923971531530}
   m_Layer: 0
   m_Name: PlayerAdvanced
   m_TagString: Player
@@ -611,7 +612,7 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   limitDiagonalSpeed: 1
-  systemTimerUpdatesPerSecond: 0.055
+  systemTimerUpdatesDivisor: 0.0549254
   smoothFollower: {fileID: 4832112164805246}
   smoothFollowerLerpSpeed: 25
 --- !u!114 &11400008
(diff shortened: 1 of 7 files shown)


Interkarma:
This looks good. I like how you combined the two problems and found a component-based solution. I'll test this out when I can.

>PlayerAdvanced.prefab was auto generated with some code changes that I'm not certain are neccessary, @Interkarma please tell me if This change is neccessary:

That all looks OK. Most of its just the new component, and Unity has a habit of shifting numbers slightly on save.

>Kills two birds with one stone. Poor birdies. Good stone.

🤣

Allofich:
The value change is from a recent PR of mine. That value is to replicate the BIOS timer which is used by classic for various timings, although I haven't really used it for much. I updated the value to be slightly more accurate and renamed it because the name I had given it before was inaccurate.

Interkarma:
Found a small bug with recent changes. Rappelling backwards off a short drop will lock up player's ability to jump or climb again. One of the states might be stuck.

Here's a save where I seem able to reproduce every time. Just walk backwards off this low roof and then try to jump or climb.

[SAVE436.zip](https://github.com/Interkarma/daggerfall-unity/files/2444718/SAVE436.zip)

Interkarma:
The problem with bumping off steep ramps in dungeons is back again with these changes. The player falls to their death when running at full speed. Here's a save game to reproduce. Just run forwards.

[SAVE405.zip](https://github.com/Interkarma/daggerfall-unity/files/2444747/SAVE405.zip)

Might need to increase the distance of slope check in step detector.

My initial solution of increasing antibump factor was fairly straightforward but it got the job done with no added complexity. The only downside was player would fall faster when walking off sheer ledges. I'm happy for a more comprehensive solution, but just be sure to check it against worst-case scenario of a 100 SPD character running down one of these ramps. :)

MeteoricDragon:
I fixed the bug of climbing underwater while overweight, and the bumping down the slide by cranking up anti-bump to 20.75f.  I'm working on the bug for rappelling, I'll fix later, It's midnight here, need to sleep.

Interkarma:
>I fixed the bug of climbing underwater while overweight, and the bumping down the slide by cranking up anti-bump to 20.75f. I'm working on the bug for rappelling, I'll fix later, It's midnight here, need to sleep.

Thank you mate, and all good. Have a great sleep!

Interkarma:
FWIW, an antibump of around 12f seems to work fairly well also for these ramps. I'm not sure if a larger or smaller value is any benefit considering your stepmanager is applying smarter gravity to player. Anyway, we'll work it out later. :)

PR #925, opened 2018-10-02
fix: only stack stackable items

petchema:
Make IsStackable() even slightly more conservative about what can be
stacked, by not allowing quest items stacking;
That done, stacking no longer need quest items special handling (after a
FindExistingStack() fix, though).

I did allow stacking for "quest" money and standard money, because I
think some quests are paid that way in classic Daggerfall; But that may
be totally irrelevant in DFU.

It would be great to allow stacking potions, because non caster that
depend on potions can have quite a few in their inventory.


Interkarma:
Thanks for the extra work on this.

PR #926, opened 2018-10-02
Fix not deleting bio save file when deleting a save

ajrb:
@numidium Hope you don't mind me fixing this, I just happened to spot it while I was testing that new characters were initialised with mages guild as enemy of achaeologists.


numidium:
Not at all, @ajrb. I likely would not have noticed that until much later.

Interkarma:
I forget this one all the time too. Thanks for the catch Hazelnut.

PR #928, opened 2018-10-03
Highlight ListBox items with MouseOver events.

MeteoricDragon:
A graphical tweak that makes mouseover actions on list boxes highlight the text to help people see what they are clicking better, and worry less about if they're actually clicking on an item above the one they meant to point click.  It affects saved games list, spell list, talk topics, etc.  I think It turned out quite well.

I tested: Saved games list, spellbook list, Buy spells list, talk topics list.

### DIFF
--- a/Assets/Scripts/Game/DaggerfallUI.cs
+++ b/Assets/Scripts/Game/DaggerfallUI.cs
@@ -43,9 +43,10 @@ namespace DaggerfallWorkshop.Game
         public static Color DaggerfallDefaultShadowColor = new Color32(93, 77, 12, 255);
         public static Color DaggerfallAlternateShadowColor1 = new Color32(44, 60, 60, 255);
         public static Color DaggerfallDefaultSelectedTextColor = new Color32(162, 36, 12, 255);
+        public static Color DaggerfallBrighterSelectedTextColor = new Color32(254, 56, 18, 255);
         public static Color DaggerfallUnityStatDrainedTextColor = new Color32(190, 85, 24, 255);
         public static Color DaggerfallUnityStatIncreasedTextColor = new Color32(178, 207, 255, 255);
-        public static Color DaggerfallDefaultTextCursorColor = new Color32(154, 134, 0, 200);
+        public static Color DaggerfallDefaultTextCursorColor = new Color32(255, 110, 110, 200);
         public static Color DaggerfallUnityDefaultToolTipBackgroundColor = new Color32(64, 64, 64, 210);
         public static Color DaggerfallUnityDefaultToolTipTextColor = new Color32(230, 230, 200, 255);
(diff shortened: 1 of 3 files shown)


Interkarma:
This looks simple enough on a read. I'll test it out. Cheers.

MeteoricDragon:
In regards to number two, that's correct. Because you are keeping the mouse on the listBox the highlightedIndex doesn't clear.  I'd like to figure out how to set the highlighted index to -1 when the mouse isn't pointing at an entry.

Interkarma:
That's OK, the second issue isn't a big deal, although would be nice to refine when you can.

Assets/Scripts/Game/DaggerfallUI.cs
```diff
@@ -43,9 +43,10 @@ public class DaggerfallUI : MonoBehaviour
         public static Color DaggerfallDefaultShadowColor = new Color32(93, 77, 12, 255);
         public static Color DaggerfallAlternateShadowColor1 = new Color32(44, 60, 60, 255);
         public static Color DaggerfallDefaultSelectedTextColor = new Color32(162, 36, 12, 255);
+        public static Color DaggerfallBrighterSelectedTextColor = new Color32(254, 56, 18, 255);
         public static Color DaggerfallUnityStatDrainedTextColor = new Color32(190, 85, 24, 255);
         public static Color DaggerfallUnityStatIncreasedTextColor = new Color32(178, 207, 255, 255);
-        public static Color DaggerfallDefaultTextCursorColor = new Color32(154, 134, 0, 200);
```
Interkarma: I just realised you overwrote the text cursor colour here. This is the blinking I-beam in text entry, and it shouldn't be this color. I'll reset this one now.

pink-ibeam

TheLacus:
> 
> 
> This looks simple enough on a read. I'll test it out. Cheers.

@Interkarma I noticed that you often merge a PR to test and then provide feedback. If this is not a personal preference, you might want to locally fetch the PR branch instead. This allows to test before accepting a PR and without affecting master with potentially unwanted changes.

```
git fetch upstream pull/928/head:HighlightListBox
git checkout HighlightListBox

// then delete without publishing
git branch -D HighlightListBox
```

Then new changes can be added to the same PR, which is accepted when you think is ready to be pushed to master. Just a simple suggestion that i think might make things simpler to manage :)

Interkarma:
@TheLacus Thank you for the info. I often do the equivalent in GitHub Desktop or send larger PRs to their own branch.

Overall, it's a personal preference to merge fast and fix later. Especially when it's a small change I can follow easily like this one. I don't mind finding small problems later, and feel that I find more bugs by having code in play sooner.

I also find you guys are really great at coming back to fix problems by yourselves or with minimal prompting. There's only been a couple of times where someone has broken something and vanished, resulting in me needing to fix it later. If it wasn't for the mature contributors we have, I'd probably manage things a bit differently. :)

PR #929, opened 2018-10-04
ListBox quick tweaks

MeteoricDragon:
Now the highlight disappears when you aren't pointing at an entry.

Now the highlighted text on parchment background is slightly more readable through increased red value and the shadow of the text.


MeteoricDragon:
I'd like to add one more commit, It's to enable player to climb underwater if they are overencumbered when swimming.  they can climb overencumbered when above water so naturally...  But I don't know where the code is that makes the player sink.

Interkarma:
I believe this is just applied as a downward force line after evaluating line 81 of LevitateMotor.

    GameManager.Instance.PlayerEnterExit.IsPlayerSwimming && (GameManager.Instance.PlayerEntity.CarriedWeight * 4) > 250)

Grouping this with an && !IsClimbing check of some kind should do the trick 👍

Interkarma:
I'll just merge the listbox stuff to test now.

PR #930, opened 2018-10-04
Fixes bio text not reloading when loading a new game

numidium:
Just a small bug fix.


Interkarma:
Thank you for the fix! 👍

PR #931, opened 2018-10-04
Bug fixes

MeteoricDragon:
Fix can't climb while sinking with overweight-ness.
Fix Anti-bump factor being too low.
Fix Rappel bug.

Discovered new bug.  When climbing onto a house with a gabled roof, the player will continue moving forward until they reach the top of the gable.  Harmless but quirky.  

I'm really unsure how to fix it...  The controller detects side collision with it, or a capsule cast detects it.  And the code is set to keep the player moving until they've stepped onto the ledge if they get past a certain height when climbing.  It does that to prevent teleporting into the sky.

### DIFF
--- a/Assets/Scripts/Game/LevitateMotor.cs
+++ b/Assets/Scripts/Game/LevitateMotor.cs
@@ -28,6 +28,7 @@ namespace DaggerfallWorkshop.Game
         PlayerMotor playerMotor;
         PlayerSpeedChanger speedChanger;
         PlayerGroundMotor groundMotor;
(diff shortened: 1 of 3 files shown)


Interkarma:
Thank you for the fixes! 👍 

> Discovered new bug. When climbing onto a house with a gabled roof, the player will continue moving forward until they reach the top of the gable. Harmless but quirky.

I've noticed this one too and it didn't bother me. For some reason, I actually find it a satisfying finale to the climb. We'll see what feedback is, but I don't think it's something you should worry too much about.

PR #938, opened 2018-10-07
Impose restriction on enemies approaching player

MeteoricDragon:
If a ground enemy sees the player at a greater pitch than 45 degrees up or down, they will wait before pursuing,  If it is a flying enemy, they will move straight up or down to get within pitch range before approaching the player.  Since all enemies are 2-dimensional, this prevents them from disappearing underneath or above the player.

I'd like anyone who's available to test it and let me know what they think.  I tested ground enemies and they work,  I can't find a good situation with swimming or flying enemies to test it.  I can't cast levitate on guards to test it either.

### DIFF
--- a/Assets/Scripts/Game/EnemyMotor.cs
+++ b/Assets/Scripts/Game/EnemyMotor.cs
@@ -369,6 +369,20 @@ namespace DaggerfallWorkshop.Game
 
             var motion = transform.forward * moveSpeed;
 
+            if (!senses.TargetIsWithinPitchAngle(45.0f))
+            {
+                if (flies || isLevitating || swims)
+                {
+                    if (!senses.TargetIsAbove())
+                        motion = -transform.up * moveSpeed;
+                    else
+                        motion = transform.up * moveSpeed;
+                }
+                else
+                    return; // maybe change mobile state to stationary too?
+            }
+
+
             // Prevent rat stacks (enemies don't stand on shorter enemies)
             AvoidEnemies(ref motion);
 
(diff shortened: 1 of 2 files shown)


Allofich:
Since you're using `toTarget` and `lastKnownTargetPosition`, this will work for AI vs AI combat as well. This is a good thing, I'm just letting you know since you only mentioned the player.

Allofich:
I gave it a test, but it feels kind of robotic to me how they will move forward and stop exactly as I do when I'm lowering.

I think maybe a better solution is to just make enemies use 2D distance to the player to decide when to stop moving towards them.

Allofich:
Another idea: Add a little buffer area, so for example, they start moving when the target is x distance away, but will move until they are x - 0.5 distance away. I think something similar is done for approaching targets normally. That way they don't move exactly at the same time that the distance to the target changes.

MeteoricDragon:
@allofich How about those changes?

Allofich:
Gave it a quick test, and the behavior of enemies waiting for the player to descend seems better now.

I did see a slaughterfish sort of jittering up and down against another slaughterfish that was in the way. I don't know if that was caused by this PR but I don't recall seeing it before.

MeteoricDragon:
> I did see a slaughterfish sort of jittering up and down against another slaughterfish that was in the way. I don't know if that was caused by this PR but I don't recall seeing it before.

I don't know either. Was the jittering slaughterfish underneath the other or was it mating with the one on top?

Allofich:
I think it was too busy trying to kill me to be mating, unless it was some kind of super slaughterfish. Anyway it's something that can be tweaked later, whether it is due to this PR or not. I don't think it's a blocker to merging.

Interkarma:
I'll switch this one to 0.6 branch, and will test soon.

PR #940, opened 2018-10-07
Faster flame animations

petchema:
The right speed is a bit subjective, but from memory, and watching flames on video,
it feels that torches and other flames are not fast enough.

(Unless gravity is lower in Tamriel, but from the look of physics in the
game it doesn't seem so).

I know we entered feature freeze for 0.5 (beside spells), but this is a one liner, and flames look painfully slow without this patch...


petchema:
Mmmh actually it looks right for torches and braziers, but original speed may be right for candres...

PR #942, opened 2018-10-09
Fix St-st-st-stuttering bumps on gently rolling hills

MeteoricDragon:
The minimum range for applying anti-bump gravity was too high and the max range is increased to 1.10f to compensate for adjusted minRange


petchema:
Works for me, thank you :+1:

Interkarma:
Thanks for fix! :)

PR #944, opened 2018-10-10
Tally magic skills when player casts spells

numidium:
This is my first time touching the effects system. If I'm doing something wrong then feel free to let me know.

### DIFF
--- a/Assets/Scripts/Game/MagicAndEffects/EntityEffect.cs
+++ b/Assets/Scripts/Game/MagicAndEffects/EntityEffect.cs
@@ -326,6 +326,7 @@ namespace DaggerfallWorkshop.Game.MagicAndEffects
             this.caster = caster;
             SetDuration();
             SetChanceSuccess();
+            TallyMagicSkill(caster);
         }
 
         /// <summary>
@@ -342,6 +343,7 @@ namespace DaggerfallWorkshop.Game.MagicAndEffects
             variantCount = effectData.variantCount;
             currentVariant = effectData.currentVariant;
             effectEnded = effectData.effectEnded;
+            TallyMagicSkill(caster);
         }
 
         /// <summary>
@@ -650,6 +652,14 @@ namespace DaggerfallWorkshop.Game.MagicAndEffects
         void SetChanceSuccess()
         {
             chanceSuccess = RollChance();
+        }
+
+        void TallyMagicSkill(DaggerfallEntityBehaviour caster)
+        {
+            if (caster == GameManager.Instance.PlayerEntityBehaviour && properties.MagicSkill != DFCareer.MagicSkills.None)
+            {
+                GameManager.Instance.PlayerEntity.TallySkill((DFCareer.Skills)properties.MagicSkill, 1);
+            }
         }
 
         #endregion


Interkarma:
Thank you for making a start on this. :)

I'd recommend not performing skill tally in effect base. This would cause player to get a free skill increase every time they load a game, as effect is resumed on load. As a general rule, effects shouldn't make direct changes to entity state.

There's a few other little gotchas as well.

Effects should only get a skill tally when they target an entity, including self. Casts that miss a target (e.g. shooting fireballs into air or at walls) shouldn't be tallied. EntityEffectManager.AssignBundle() is a better spot. Try the following around line 342 before absorb check.

    // Tally magic skills
    if (sourceBundle.CasterEntityBehaviour == GameManager.Instance.PlayerEntityBehaviour)
    {
        GameManager.Instance.PlayerEntity.TallySkill((DFCareer.Skills)effect.Properties.MagicSkill, 1);
    }

We also need to ensure skill tally for effect isn't counted multiple times when AoE hits several targets at once. I think best solution here is to add a bool in EntityEffectBundle at line 69.

    /// <summary>
    /// Gets or sets if skills have been tallied for this bundle.
    /// </summary>
    public bool SkillsTallied { get; set; }

Then back in EntityEffectManager.AssignBundle() line 342, change code to the following:

    // Tally magic skills
    if (sourceBundle.CasterEntityBehaviour == GameManager.Instance.PlayerEntityBehaviour && !sourceBundle.SkillsTallied)
    {
        GameManager.Instance.PlayerEntity.TallySkill((DFCareer.Skills)effect.Properties.MagicSkill, 1);
        sourceBundle.SkillsTallied = true;
    }

As the same bundle payload instance is handed to each target entity for instantiation, the flag will ensure skills are only tallied once for that instance.

I don't think we need to worry about other effects being counted (e.g. diseases, poisons) as they shouldn't have a caster set. But probably can't hurt to exclude non-spell bundles in the above check just in case. So something like the below at line 342 ought to have us covered.

    // Tally magic skills
    if (sourceBundle.CasterEntityBehaviour == GameManager.Instance.PlayerEntityBehaviour && !sourceBundle.SkillsTallied && sourceBundle.Settings.BundleType == BundleTypes.Spell)
    {
        GameManager.Instance.PlayerEntity.TallySkill((DFCareer.Skills)effect.Properties.MagicSkill, 1);
        sourceBundle.SkillsTallied = true;
    }

I hope that helps! Let me know if anything I can do to assist. I'd also be interested in feedback from @Allofich around magic skill tallies specific to classic.

numidium:
Thanks @Interkarma for your insights. This should get me on the right track. I considered doing the tally in AssignBundle but I thought I remembered Daggerfall increasing your skills even if the spells didn't hit a target. On further analysis I see that's not true.

I'll close this and start a new one, using your post as a reference.

Allofich:
> Effects should only get a skill tally when they target an entity, including self. Casts that miss a target (e.g. shooting fireballs into air or at walls) shouldn't be tallied. 

Do you just mean that as a preference? Classic doesn't have any such requirement. In fact, it tallies the skill as soon as a spell is "readied", before you've even fired, and untallies it (tallies it -1) if you abort before releasing the readied spell. Or, for an instant cast, as soon as it is cast.

The ID of the magic skills raised on a cast in classic is gotten like the following, using a `SpellRecord.SpellRecordData`:

```
TallyMagicSkill(SpellRecord.SpellRecordData spell)
{
  byte[] MagicSkills =  [ 0x19, 0x17, 0x16, 0x1B, 0x1A, 0x18 ]; // The six magic skills
  byte[] MagicSkillsIndices = [ 0, 2, 3, 1, 2, 2, 3, 2, 0, 1, 1, 2, 3, 5, 4, 5, 3, 3, 1, 3, 1, 4, 4, 5, 5, 0, 1, 0, 0, 5, 0, 4, 0, 4, 4, 0, 3, 3, 0, 4, 4, 4, 5, 3, 3, 0, 0, 4, 4, 4, 4];

  for (int i = 0; i < 3; ++i) // 3 possible effects per spell
  {
    if (spell.effects[i].type != -1) // an effect exists in this slot
    {
      TallySkill(MagicSkills[MagicSkillsIndices[spell.effects[i].type]]);
    }
  }
}
```

Note that that list of skills used is 51 bytes long and at the end includes skills for effects that aren't in the game. The full list of effects, in the order that matches to the above, is :

Paralysis
Continuous Damage
Create Item
Cure
Damage
Disintegrate
Dispel
Drain
Elemental Resistance
Fortify Attribute
Heal
Transfer
Soul Trap
Invisibility
Levitate
Light
Lock
Open
Regenerate
Silence
Spell Absorption
Spell Reflection
Spell Resistance
Chameleon
Shadow
Slowfall
Climbing
Jumping
Free Action
Lycanthropy/Polymorph
Water Breathing
Water Walking
Dimunition
Calm
Charm
Shield
Telekinesis
Astral Travel
Etherealness
Detect
Identify
Wizard Sight
Darkness
Recall
Comprehend Languages
Intensify Fire
Diminish Fire
Wall of Stone
Wall of Fire
Wall of Frost
Wall of Poison

Allofich:
> I thought I remembered Daggerfall increasing your skills even if the spells didn't hit a target. On further analysis I see that's not true.

Did you test that and that was the case?

Interkarma:
> Do you just mean that as a preference? Classic doesn't have any such requirement.

In my early spell testing, I was unable to raise a magic skill just by casting at nothing. I had to physically target something (either self or another entity). Mind you, I did this back in around 2015/2016 so it's been a while. Happy to be wrong here, it would simplify things.

@numidium I'd take Allofich's word over mine here. In this case, the best place to tally is in EntityEffectManager.PlayerSpellCasting_OnReleaseFrame() event. This is the point where spell is physically cast. If player cancels spell, this event is never called (so no need to tally/untally).

>The ID of the magic skills raised on a cast in classic is gotten like the following

That's what the effect.Properties.MagicSkill is for in DFUnity. Keep in mind there's potential for an infinite number of custom effects, so the classic handling won't be enough. But it's very similar on our end anyway. Just need to loop through effects in bundle at release time and tally,

I think I'll stub this out quickly before bed and we can build on it from there. Thank you both.

Allofich:
> That's what the effect.Properties.MagicSkill is for in DFUnity.

Sure, I was just putting the info out there. I realize you would probably want to have the skill used as a property of the effect and not as a loose byte array like it is in classic.

Interkarma:
Thank you for the info, it's very helpful. And much preferred to my faulty memory and/or testing process. This is how we get more done together. :)

Allofich:
Alright, good to know. Thanks for responding.

PR #945, opened 2018-10-10
Fix anti-bump being applied while climbing and slipping

MeteoricDragon:
Anti- bump was being applied while climbing dungeon walls, and made the player zip to the ground when slipping and climbing.


Interkarma:
Thanks for fix!

PR #946, opened 2018-10-10
Clean more files

Allofich:
More file cleanup, this time it's

Scripts/External
Scripts/Game/Entities
Scripts/Game/Formulas

This includes a minor functional fix. I noticed while cleaning up the condition code in `PlayerEntity.cs` that the  alliance and rivalry changes, and the new ruler changes to regions were not taking effect (I wasn't properly updating the faction data dictionary). Only changes to power were taking effect. Of course we still need the rumors so this stuff is actually visible to the player.

Again, will pause here until you're OK with it. These files are more volatile than the API ones but if any merge conflicts do come up they shouldn't be hard to fix, as these are all very simple changes.

Question: I'm simplifying conditionals like

```
if (something == true)
```
and
```
return (something ? true : false)
```
to just be
```
if (something)
````
and
```
return (something)
```

and I think I prefer the shorter versions, but do you prefer having "true" and "false" written out?

Fixes:
Fix incorrect function summary and param tags
Spacing clean up and whitespace removal
Remove deprecated DaggerfallDisease
Fix shadowed variables
Add newline at end of files
Simplify conditional expressions
Revise inaccurate comments and TODOs
Merge nested if statements
Remove redundant int to int casts
Make value const
A little safer null checking in damage calculation
Remove out-of-date comments and TODOs
Fix region alliances, rivalries, ruler data to actually update

### CODE REVIEW

**Comment by ajrb** [comment]
I know you were not asking me, but when has that ever stopped me from offering my opinion, eh? Anyway, I agree with the shorter versions 100% it's more meaningful and elegant. I used to get angry when outsourced development code would come in with 
` 
bool thing = someStuff();
if (thing == true)
  return true;
else
  return false;

`

File: Assets/Scripts/Game/Entities/PlayerEntity.cs [hunk 1]
```diff
@@ -842,27 +847,39 @@ public int GetGoldAmount()
...
         public int DeductGoldAmount(int amount)
         {
-            if (amount <= goldPieces) {
+            if (amount <= goldPieces)
+            {
                 goldPieces -= amount;
-            } else {
+            }
+            else
+            {
```
ajrb: It's going to expand a lot of compressed blocks like this, since it's my coding style to do this for small blocks of code as I find it more readable to not use extra lines in these situations.

If the team decision is to have them expanded, that is perfectly fine, I'm just raising it so a decision can be taken. For all I know my style has been annoying everyone. Anyway we should go into it eyes wide open.

Interkarma:
>I'm simplifying conditionals like...

Both are one-liners and functionally equivalent, it's just a matter of style. Personally, I find the slightly longer one-line format a bit easier to skim through quickly as the explicit outcomes give the condition "shape" faster in my mind when rapid-reading code. I'm not quite as bad as @ajrb's example though (you meanie, Hazelnut 🤣). Anyway, the changes are fine with me.

I'm happy with general cleanup, but I'm not confident about changing simple style layout that would compile down identically anyway, and may be specific to authors preference (see Hazelnut's comment). I'd rather the original author be able to jump back into their code and get straight back into flow because they recognise their code immediately. Psychologically, I also think it helps devs retain a feeling a ownership of their code, which is important for a community project where no-one is being paid to cut code a certain way.

Maybe just leave those blocks in place as written. I sometimes do this as well, although I seem to do it less these days. Stuff like this shouldn't really matter, and doesn't seem to change the readability much either way.

ajrb:
I wasn't being a meanie intentionally, that was a true story that happened a lot by so called 'professionals' and legitimately drove me mad, and I wasn't drawing a parallel honest. :)

Regarding the block layout, I hope my comment was clear that I don't mind either way. Some prefer consistency and others (like me) prefer whatever is most compact while still being easily parsed. I'm happy if the consensus is to make it all consistent, I was just saying it should be a conscious group decision to do this.

Allofich:
> Regarding the block layout, I hope my comment was clear that I don't mind either way. Some prefer consistency and others (like me) prefer whatever is most compact while still being easily parsed.

My intention right now is to go in and remove that change. I'm trying to keep these to just changes that everyone (well, Interkarma at least) is OK with. This was a just a change that the autoformat in Visual Studio did. The majority of the code doesn't use that kind of formatting so I thought it was OK as a change to bring in line with that, but it's no big deal either way.

Allofich:
I took out the block changes.

Interkarma:
>I wasn't being a meanie intentionally, that was a true story that happened a lot by so called 'professionals' and legitimately drove me mad, and I wasn't drawing a parallel honest. :)

I know, I was just being cheeky. :)

Thanks for all your work on this one Allofich!

ajrb:
> > I wasn't being a meanie intentionally, that was a true story that happened a lot by so called 'professionals' and legitimately drove me mad, and I wasn't drawing a parallel honest. :)
 
> I know, I was just being cheeky. :)

Yeah, it definitely gave me a chuckle. :-)

PR #947, opened 2018-10-11
Fix free room for member of region knightly order

Allofich:
The knightly order code was still using +1 when checking the region ID, although the FactionData now subtracts 1 from the IDs from `FACTION.TXT` when importing (like classic does) and the +1 check is no longer necessary.


Interkarma:
Good catch, thank you! 👍

PR #948, opened 2018-10-11
Give player starting spells

numidium:
Implements the classic starting spell lists.

### DIFF
--- a/Assets/Scripts/Game/Utility/StartGameBehaviour.cs
+++ b/Assets/Scripts/Game/Utility/StartGameBehaviour.cs
@@ -702,14 +702,6 @@ namespace DaggerfallWorkshop.Game.Utility
                 spellsToAdd.Add(startingSpells[spellSetIndex].SpellsList[i]);
             }
 
-            // Get standard list of spells
-            List<SpellRecord.SpellRecordData> standardSpells = DaggerfallSpellReader.ReadSpellsFile(Path.Combine(DaggerfallUnity.Instance.Arena2Path, "SPELLS.STD"));
-            if (standardSpells == null || standardSpells.Count == 0)
-            {
-                Debug.LogError("Failed to load SPELLS.STD while assigning starting spells.");
-                return;
-            }
-
             // Add spells to player from standard list
             foreach (StartingSpell spell in spellsToAdd)
             {
--- a/Assets/Scripts/Game/Utility/StartGameBehaviour.cs
+++ b/Assets/Scripts/Game/Utility/StartGameBehaviour.cs
@@ -665,67 +665,71 @@ namespace DaggerfallWorkshop.Game.Utility
 
         void SetStartingSpells(PlayerEntity playerEntity)
         {
-            // Spell indices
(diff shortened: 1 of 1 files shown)


Interkarma:
Cool! It will be great to have this in finally.

Just something to note, I exported starting spells to Resources/StartingSpells.txt in JSON format. It's keyed by career index and spellid. This was extracted from FALL.EXE with @Allofich's help on where to find data.

I notice the spell loadout by character in your implementation is a bit different to the one exposed by .exe. I believe the data in StartingSpells.exe is the correct loadout classic uses.

Would it be possible to revise implementation to use the StartingSpells.txt JSON? I think this might even simplify implementation a bit.

numidium:
Sure, that will work perfectly. The less hard-coding the better. I was going by the list of starting spells on UESP but it looks like they've steered me wrong unfortunately.

Allofich:
Custom only gets spells if a primary or major skill is a magic skill.

numidium:
@Allofich @Interkarma I'm running into an odd issue with the SPELLS.STD file. Most of the indices there match up with StartingSpells.txt if you subtract the index by one but there are a few that don't. Balyna's Balm, for example, has a SpellID of 97 in StartingSpells.txt but is at index 86 in my standardSpells array. Should I hard code corrections for the mismatched IDs or is there another way around this quirk?

Interkarma:
You might be mixing up the index with the spellid. The spellid is a value read from the spell record itself. It should correspond to the STD# row from table in this page.

http://en.uesp.net/wiki/Daggerfall:SPELLS.STD_indices

Now that I think about it, I'm not sure I've added a helper to convert spellid back to a bundle. Let me take a look in a few hours when I get off work and I'll bridge these for you. :)

numidium:
> Now that I think about it, I'm not sure I've added a helper to convert spellid back to a bundle.

I was able to make that conversion in my latest commit but a helper would make my method a lot cleaner. Much appreciated.

Interkarma:
Sounds like a plan. 👍

Interkarma:
Actually, I have already done this. Check out EntityEffectBroker.GetClassicSpellRecord(int id, out SpellRecord.SpellRecordData spellOut).

Just feed it an ID and out param, and it will eject the SpellRecordData.

Sorry, I've been a bit distracted today with work. :)

Assets/Scripts/Game/Utility/StartGameBehaviour.cs
```diff
@@ -659,6 +663,74 @@ void ResetWeaponManager()
...
+                spellsToAdd.Add(startingSpells[spellSetIndex].SpellsList[i]);
+            }
+
+            // Get standard list of spells
+            List<SpellRecord.SpellRecordData> standardSpells = DaggerfallSpellReader.ReadSpellsFile(Path.Combine(DaggerfallUnity.Instance.Arena2Path, "SPELLS.STD"));
+            if (standardSpells == null || standardSpells.Count == 0)
+            {
+                Debug.LogError("Failed to load SPELLS.STD while assigning starting spells.");
+                return;
+            }
```
Interkarma: I think you can ditch the above block. No need to call ReadSpellsFile() now.

Interkarma:
Looks good, I'll take it for a spin. Thanks for all your work on this one. 👍

PR #950, opened 2018-10-13
Video import: Upgrade to VideoPlayer 

TheLacus:


Interkarma:
Wow, thank you for doing that! 👍

PR #952, opened 2018-10-14
File cleanup

Allofich:
Moving on in sequential order, this is cleanup on:
`Scripts/Game/Guilds`
`Scripts/Game/Items`

Use implicit array initializers
Merge nested if statements
Add end of file newlines
Fix variable shadowing
Clean up spacing, remove whitespace
Omit redundant comparison
Make sure foreach iteration is on expected type
Remove unused parameter in ReceiveHouse()
Add missing `<summary>` tag
Change !(A == B) to (A != B)
Revise comments
Fix incorrect function summary params
Remove redundant variable assignment
Remove redundant textAsset to textAsset casts

### DIFF
--- a/Assets/Scripts/Game/Guilds/Guild.cs
+++ b/Assets/Scripts/Game/Guilds/Guild.cs
@@ -34,9 +34,9 @@ namespace DaggerfallWorkshop.Game.Guilds
 
         #region Static Data
 
-        public static int[] rankReqReputation = new int[] {  0, 10, 20, 30, 40, 50, 60, 70, 80, 90 };
-        public static int[] rankReqSkillHigh = new int[]  { 22, 23, 31, 39, 47, 55, 63, 71, 79, 87 };
-        public static int[] rankReqSkillLow = new int[]   {  4,  5,  9, 13, 17, 21, 25, 29, 33, 37 };
+        public static int[] rankReqReputation = {  0, 10, 20, 30, 40, 50, 60, 70, 80, 90 };
+        public static int[] rankReqSkillHigh =  { 22, 23, 31, 39, 47, 55, 63, 71, 79, 87 };
+        public static int[] rankReqSkillLow =   {  4,  5,  9, 13, 17, 21, 25, 29, 33, 37 };
 
         #endregion
 
@@ -105,7 +105,7 @@ namespace DaggerfallWorkshop.Game.Guilds
                 if (rep < rankReqReputation[r] || high < 1 || low + high < 2)
                     break;
             }
-            Debug.LogFormat("rep: {0} high#: {1} low#: {2} new rank: {3}", rep, high, low, r-1);
+            Debug.LogFormat("rep: {0} high#: {1} low#: {2} new rank: {3}", rep, high, low, r - 1);
             return --r;
         }
 
@@ -345,7 +345,7 @@ namespace DaggerfallWorkshop.Game.Guilds
         /// </summary>
         protected class GuildMacroDataSource : MacroDataSource
         {
-            private Guild parent;
+            private readonly Guild parent;
             public GuildMacroDataSource(Guild guild)
             {
                 parent = guild;
--- a/Assets/Scripts/Game/Guilds/GuildManager.cs
+++ b/Assets/Scripts/Game/Guilds/GuildManager.cs
@@ -91,7 +91,7 @@ namespace DaggerfallWorkshop.Game.Guilds
                 default:
                     Type guildType;
                     if (customGuilds.TryGetValue(guildGroup, out guildType))
-                        return (int) guildType.GetProperty("FactionId").GetValue(null, null);
+                        return (int)guildType.GetProperty("FactionId").GetValue(null, null);
                     else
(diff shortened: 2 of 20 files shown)


Assets/Scripts/Game/Guilds/Guild.cs
```diff
@@ -81,7 +81,7 @@ public virtual TextFile.Token[] UpdateRank(PlayerEntity playerEntity)
                         tokens = TokensPromotion(newRank);
                     else if (newRank < 0)       // Expulsion
                         tokens = TokensExpulsion();
-                    else if (newRank < rank)    // Demotion
+                    else                        // Demotion
```
ajrb: I would prefer this is left as an explicit condition please.

Assets/Scripts/Game/Guilds/GuildManager.cs
```diff
@@ -258,7 +259,7 @@ public void RestoreMembershipData(Dictionary<int, GuildMembership_v1> data)
             ClearMembershipData();
             if (data != null)
             {
-                foreach (FactionFile.GuildGroups guildGroup in data.Keys)
+                foreach (FactionFile.GuildGroups guildGroup in data.Keys.OfType<FactionFile.GuildGroups>())
```
ajrb: Is this really neccessary?

Assets/Scripts/Game/Guilds/KnightlyOrder.cs
```diff
@@ -142,13 +141,13 @@ public override TextFile.Token[] TokensPromotion(int newRank)
             return DaggerfallUnity.Instance.TextProvider.GetRSCTokens(GetPromotionMsgId(newRank));
         }
 
-        private int GetPromotionMsgId(int rank)
+        private int GetPromotionMsgId(int newRank)
```
ajrb: Why does this need changing? This method doesn't care if rank is a new one or not, hence why I named the parameter just rank.

Assets/Scripts/Game/Guilds/GuildManager.cs
TheLacus: `OfType` is useful to filter by type old collections like `ArrayList`, which are now deprecated with the introduction of generics.

I believe `OfType` is not doing anything here because the cast is always succesful. If you want to check that an int value is defined with an enum, you need to use something like `.Where(x => Enum.IsDefined(typeof(FactionFile.GuildGroups), x))`, but this comes with a big performance cost.

Assets/Scripts/Game/Guilds/Guild.cs
Allofich: OK, I'll put it back to how it was.

Assets/Scripts/Game/Guilds/GuildManager.cs
Allofich: I'll undo this change. Sorry, I'd meant to bring this one up when submitting the PR but forgot to.

Here is the web site for the warning that led to me changing that:
https://rules.sonarsource.com/csharp/RSPEC-3217

They call it "critical" and it sounded like a worthwhile change/guideline to keep to, but unless we have a general consensus for it I'll ignore this one.

Assets/Scripts/Game/Guilds/KnightlyOrder.cs
ajrb: Oh you mean it is the same name as an internal member? Yes, that's true so I understand now.

Assets/Scripts/Game/Guilds/GuildManager.cs
TheLacus: Foreach loop makes a cast due to [historical reasons](https://ericlippert.com/2013/07/22/why-does-a-foreach-loop-silently-insert-an-explicit-conversion/). It's a bad idea to make use of it because it can hide a possibly bad design, so this is a good point to rise.

The cast done here is from int to enum, which is always valid. The issue is that it's easy to forget to handle unknown values in an appropriate way. @ajrb might want to check this is the case and maybe see if values can be directly received and stored as enum instead of getting int values.

Allofich: Thanks for the insight @TheLacus. I'll leave this one to @ajrb if he would like to pursue it further and decide whether to make any changes.

ajrb: I'll take a look at some point, thanks for the info.

Assets/Scripts/Game/Guilds/KnightlyOrder.cs
ajrb: Well I'm fairly comfortable hiding members in local scope, since you can always do this.rank to access the member. I think the rename is fine though.

Allofich:
Couple things I need to address. Please don't merge yet.

Allofich:
Addressed those issues. Also rebased to master and changed target to master. Let me know if you want me to target a different branch.

Interkarma:
This looks good, thank you. I'm OK with changes, will merge now.

PR #953, opened 2018-10-14
First pass at enemy casting

Allofich:
Here is the start of enemy casting.

Implemented:
- Enemies assigned spells and spell points like in classic
- Enemies shoot a ranged spell like in classic

Not implemented:
- Spell casting animations
- Touch spells
- Some additional AI logic present in classic (don't cast a spell while it is still active on the target, intelligent enemies may stop casting when target has reflected a spell)

I'm sending a PR now because there is an issue maybe you can help with. For some reason enemies only cast a single spell, then stop. The logic in EnemyMotor.cs is correctly continuing to select spells for casting but something is stopping any more than 1 from shooting.

I had to make a few adjustments related to `EntityEffectManager.cs`:

- The `IsEnemyEntity` was always failing, because everything but the player is just type `None`.
- The  `EntityEffectManager` peered with enemies needed the missile prefabs assigned for instantiation.

### DIFF
--- a/Assets/Prefabs/Scene/DaggerfallEnemy [Game Serializable].prefab
+++ b/Assets/Prefabs/Scene/DaggerfallEnemy [Game Serializable].prefab
@@ -3,9 +3,9 @@
 --- !u!1 &131734
 GameObject:
   m_ObjectHideFlags: 0
-  m_PrefabParentObject: {fileID: 0}
+  m_CorrespondingSourceObject: {fileID: 0}
   m_PrefabInternal: {fileID: 100100000}
-  serializedVersion: 5
+  serializedVersion: 6
   m_Component:
   - component: {fileID: 488118}
   - component: {fileID: 3367684}
@@ -21,9 +21,9 @@ GameObject:
 --- !u!1 &140354
 GameObject:
   m_ObjectHideFlags: 0
-  m_PrefabParentObject: {fileID: 0}
+  m_CorrespondingSourceObject: {fileID: 0}
   m_PrefabInternal: {fileID: 100100000}
-  serializedVersion: 5
+  serializedVersion: 6
   m_Component:
   - component: {fileID: 481572}
   - component: {fileID: 11410784}
@@ -51,7 +51,7 @@ GameObject:
 --- !u!4 &481572
 Transform:
   m_ObjectHideFlags: 1
-  m_PrefabParentObject: {fileID: 0}
+  m_CorrespondingSourceObject: {fileID: 0}
   m_PrefabInternal: {fileID: 100100000}
   m_GameObject: {fileID: 140354}
   m_LocalRotation: {x: 0, y: 0, z: 0, w: 1}
@@ -65,7 +65,7 @@ Transform:
 --- !u!4 &488118
 Transform:
   m_ObjectHideFlags: 1
-  m_PrefabParentObject: {fileID: 0}
+  m_CorrespondingSourceObject: {fileID: 0}
   m_PrefabInternal: {fileID: 100100000}
   m_GameObject: {fileID: 131734}
   m_LocalRotation: {x: 0, y: -0.32127059, z: 0, w: 0.94698745}
@@ -78,7 +78,7 @@ Transform:
 --- !u!23 &2302536
 MeshRenderer:
   m_ObjectHideFlags: 1
-  m_PrefabParentObject: {fileID: 0}
+  m_CorrespondingSourceObject: {fileID: 0}
   m_PrefabInternal: {fileID: 100100000}
   m_GameObject: {fileID: 131734}
   m_Enabled: 1
@@ -113,14 +113,14 @@ MeshRenderer:
 --- !u!33 &3367684
 MeshFilter:
   m_ObjectHideFlags: 1
-  m_PrefabParentObject: {fileID: 0}
+  m_CorrespondingSourceObject: {fileID: 0}
   m_PrefabInternal: {fileID: 100100000}
   m_GameObject: {fileID: 131734}
   m_Mesh: {fileID: 0}
 --- !u!82 &8249844
 AudioSource:
   m_ObjectHideFlags: 1
-  m_PrefabParentObject: {fileID: 0}
+  m_CorrespondingSourceObject: {fileID: 0}
   m_PrefabInternal: {fileID: 100100000}
   m_GameObject: {fileID: 140354}
   m_Enabled: 1
@@ -215,7 +215,7 @@ AudioSource:
 --- !u!114 &11410784
 MonoBehaviour:
   m_ObjectHideFlags: 1
-  m_PrefabParentObject: {fileID: 0}
+  m_CorrespondingSourceObject: {fileID: 0}
   m_PrefabInternal: {fileID: 100100000}
   m_GameObject: {fileID: 140354}
   m_Enabled: 1
@@ -226,10 +226,11 @@ MonoBehaviour:
   EnemyType: 15
   EnemyReaction: 0
   EnemyGender: 0
+  ClassicSpawnDistanceType: 0
 --- !u!114 &11424130
 MonoBehaviour:
   m_ObjectHideFlags: 1
-  m_PrefabParentObject: {fileID: 0}
+  m_CorrespondingSourceObject: {fileID: 0}
   m_PrefabInternal: {fileID: 100100000}
   m_GameObject: {fileID: 140354}
   m_Enabled: 1
@@ -242,7 +243,7 @@ MonoBehaviour:
 --- !u!114 &11435454
 MonoBehaviour:
   m_ObjectHideFlags: 1
-  m_PrefabParentObject: {fileID: 0}
+  m_CorrespondingSourceObject: {fileID: 0}
   m_PrefabInternal: {fileID: 100100000}
   m_GameObject: {fileID: 140354}
   m_Enabled: 1
@@ -254,7 +255,7 @@ MonoBehaviour:
 --- !u!114 &11439840
 MonoBehaviour:
   m_ObjectHideFlags: 1
-  m_PrefabParentObject: {fileID: 0}
+  m_CorrespondingSourceObject: {fileID: 0}
   m_PrefabInternal: {fileID: 100100000}
   m_GameObject: {fileID: 140354}
   m_Enabled: 1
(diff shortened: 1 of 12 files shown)


Allofich:
The only casting once issue was because of the `castInProgress` being set to true and never cleared. I commented it out for now as `readySpell` being set to null after cast accomplishes the same thing for now.

Allofich:
Added first pass at enemy touch spells.

As an aside, you no doubt have experienced in classic how some enemies, like Vampire Ancients, will practically insta-skill you with spells, not even letting the camera fall to the floor before hitting you with another spell. This is because even though the original game uses the MeleeTimer cooldown for enemy touch spells, unlike with regular attacks touch spells do not reset the counter, so they hit you with all of their touch spells as fast as the AI updates.

For DF Unity I'm resetting the melee timer as was probably supposed to have happened.

Allofich:
Enemy spell bundle is failing to be created for the spell `Spell Shield`. Is this expected? You can test by going into Orsinium because Orc Shamans have it.

Allofich:
Added paralyze from spiders and scorpions. Also fixed a bug where you couldn't use magic items if silenced or if you didn't have the spell points for what the item would cost if it were a spell (although the spell points wouldn't be removed if you did use the item, you still had to have them)

Allofich:
Changed on-touch, on-caster logic to be for on-touch, around-caster. The logic doesn't make sense for on-caster spells (having to be near target is arguable OK, but checking whether the effect is on the target doesn't make sense for on-caster spells).

On-caster logic will have to be implemented somewhere else.

Allofich:
AI only casts now if effects are not already on the target.

In classic, the check is for whether the same spell ID is already on the target. I couldn't see a way to do that, so I compared by checking whether all the effects of the spell to be cast are already present.

This is potentially smarter than classic. For example, if an enemy had 2 paralysis spells, it would try to cast both of them in classic, but with this PR it would see that paralysis is already on from one of the spells and not bother with the other. However, enemy spell lists in classic may be set up in such a way that there is no overlap of effects between spells, in which case this won't matter unless mods change the spell layouts.

Also, the way of this PR is potentially not smart if spells can stack. Can they?

MeteoricDragon:
One thing I remember about spell-casting enemies in daggerfall is they will cast touch target spells rapidly at the exact instant they are within range of their target.  This was off-putting considering and kind of robotic.  Proposed solution: Make enemies ai cast their spells spaced out at random times.

MeteoricDragon:
> Also, the way of this PR is potentially not smart if spells can stack. Can they?

I think that spells can stack if it is the same spell being cast.  I recall casting some spells to increase the duration of an effect in classic.  There might be a cap to how many times a spell can stack though.

Allofich:
> One thing I remember about spell-casting enemies in daggerfall is they will cast touch target spells rapidly at the exact instant they are within range of their target. This was off-putting and kind of robotic. Proposed solution: Make enemies ai cast their spells spaced out at random times.

Yes I think that's actually a bug in classic. I mentioned it above.

> As an aside, you no doubt have experienced in classic how some enemies, like Vampire Ancients, will practically insta-skill you with spells, not even letting the camera fall to the floor before hitting you with another spell. This is because even though the original game uses the MeleeTimer cooldown for enemy touch spells, unlike with regular attacks touch spells do not reset the counter, so they hit you with all of their touch spells as fast as the AI updates.

MeteoricDragon:
Well, that's a pitfall of a lot of games in that era.  They didn't sync with time to make sure they weren't doing it too fast :)  they were too worried about the computer being fast enough!

Interkarma:
Lots happening overnight, well done! I'll work through your comments and respond when I can. Might have to spread over next two days though, have a couple of busy ones again.

Interkarma:
I just restored castInProgress as its part of player animation handling. But it doesn't need to be evaluated for monsters, so moved this into the player block of CastReadySpell(). This should fix the problems for monsters and still suppress player casting through their own animations.

Interkarma:
>Enemy spell bundle is failing to be created for the spell Spell Shield. Is this expected?

Effect isn't implemented yet. I'll get to it shortly.

>Also, the way of this PR is potentially not smart if spells can stack. Can they?

Incumbent spells (usually anything with a duration) will stack duration on secondary casts. You can use EntityEffectManager.FindIncumbentEffect<T>() to see if there's an IEntityEffect matching T on that entity. In the back-end, this is similar to what you're doing anyway. I'm happy for monsters to have their "intelligence" here if that's preferable to you.

Interkarma:
>Added paralyze from spiders and scorpions. Also fixed a bug where you couldn't use magic items if silenced or if you didn't have the spell points for what the item would cost if it were a spell (although the spell points wouldn't be removed if you did use the item, you still had to have them)

Thank you for that, an oversight on my part.

Interkarma:
>As an aside, you no doubt have experienced in classic how some enemies, like Vampire Ancients, will practically insta-skill you with spells, not even letting the camera fall to the floor before hitting you with another spell.

I know it all too well! Had always planned to improve this, and happy to hear you feel the same. 👍

MeteoricDragon:
> BTW, MeteoricDragon, while testing I've been running into issues with the Player I think getting stuck in rappel mode or something, where I start gliding all over the room until I get to a wall and get out of climbing mode. 
If you were to back up off a ledge, and immediately turn away from the ledge before the controller can raycast forward to the ledge, it might happen.  I need a better way that can get the hitnormal of the wall the player is rappeling onto.

Allofich:
> You can use EntityEffectManager.FindIncumbentEffect() to see if there's an IEntityEffect matching T on that entity. In the back-end, this is similar to what you're doing anyway. I'm happy for monsters to have their "intelligence" here if that's preferable to you.

Looks like I could replace the for-loops I used with EntityEffectManager.FindIncumbentEffect(), but I would still be doing it in the same spot because I wanted to find whether all effects of the chosen spell were already on the target, not just a single effect.

> Incumbent spells (usually anything with a duration) will stack duration on secondary casts.
What about different spells that use the same effect, like two different fire-damage spells? They would both apply their damage amounts at the same time, right?
Classic compares the "index" of the spell being cast, and only casts if a spell record with that index is not on the player. So two different fire-damage spells could both be cast in succession in classic, whereas in the implementation I used in this PR, only one would be cast at a time, since the effect type is checked.

If `ClassicSpellRecordDataToEffectBundleSettings()` were to assign `spellRecordData.index` to a value in the bundle, we could replicate the classic behavior.

Interkarma:
>Looks like I could replace the for-loops I used with EntityEffectManager.FindIncumbentEffect(), but I would still be doing it in the same spot because I wanted to find whether all effects of the chosen spell were already on the target, not just a single effect.

I was pressed for time and did not explain well. What I was trying to say is that I'm happy for monsters to have their own "intelligence" here (as in here, in this situation). Do whatever you see fit for their AI, and however you want to implement it. My only requirement is that it still fits within my broader design. :)

>What about different spells that use the same effect, like two different fire-damage spells?

It depends on the effect and if they inherit from IncumbentEffect or not. Incumbent effects have three interesting virtual methods:
* BecomeIncumbent() is called when this is the first instance of an incumbent effect on the entity (e.g. player casts levitate on themselves).
* AddState() is called when there is already an incumbent effect on the entity (e.g. player casts levitate again to top-up duration).
* IsLikeKind() is called so the effect can return true/false whether it's a match for the other instance of the same effect. It's possible for the effect to make a decision based on its own logic and become a new instance instead of adding state.

These three methods put together are incredibly cool and useful. They let the effect itself determine its own behaviour. It could just be to increase duration of the incumbent (e.g. topping up a levitate) or it could do something _completely novel_. I'm not sure anyone else has realised just how flexible my entity-effect system is yet, as I'm the only one cranking out effects so far. This will change once its opened up to modders and we have someone doing for effects what JayH is doing for quests. This is just the start right now. :)

But to try and answer your question with an example - a "Continuous Damage Health" fire effect would stack duration, as its an incumbent effect and uses AddState() to that end. Whereas a "Damage Health" fire effect would not as it's just a basic instant effect that comes and goes in a single blast.

The thing to understand (and it took me ages to really get this) is that "spells" don't really matter. Not one solitary jot. The player can create an unlimited number of spells with an infinite variety of stock and custom effects via the mod system. A "spell" is just a container. It's the _effects_ that really matter. That's why I call them bundles instead of spells. And why it's the "magic effect" system instead of the "magic spell" system. There's a deep core thoughtful core in all of this once you see it. :)

>Classic compares the "index" of the spell being cast, and only casts if a spell record with that index is not on the player. So two different fire-damage spells could both be cast in succession in classic, whereas in the implementation I used in this PR, only one would be cast at a time, since the effect type is checked.

If you look down the list of spell enemies can cast, it's basically one effect per bundle, right? With my design, effects are autonomous and effects make decisions for themselves. There's almost no need for the AI to even worry about this. The effect will handle its payload completely independently of whatever entity it originated from. This is actually kind of liberating - now you can chuck anything at anything and everything still just works.

But if you really want to emulate classic here in enemy AI spell selection, I'd be happy to add an "int ClassicIndexTag {get; set;}" property to bundles. You could then tag the bundle payload before release from enemy and later make decisions depending on what tags are present. But the effects would just behave as per their own logic anyway. 

Of course, this wouldn't have any meaning once people can mod enemy bundle payloads, and my feeling is it's really not even needed. I'd consider what the classic way of an enemy selecting next spell really means in the broader context. Is it really worth tagging bundles back to classic spell from SPELLS.STD or does this just limit what people can do in future?

Anyway, it's getting late here and I need to get some sleep. Hopefully I've been able to explain a bit better this time. I'm enjoying all the work you've done so far from the very brief test I had this morning. /salute

Allofich:
> But if you really want to emulate classic here in enemy AI spell selection, I'd be happy to add an "int ClassicIndexTag {get; set;}" property to bundles...Of course, this wouldn't have any meaning once people can mod enemy bundle payloads

If it won't be a general solution, let's not bother with it. It doesn't really make any logical sense anyway that two "Continuous Damage Health" fire effects would be OK to cast over each other if they were different spells but not OK if they were coming from the same spell. And we're already differing from classic anyway since I've fixed a few bugs (no casting all touch spells instantanteously bug, not using "touch spell" logic for on-self casts), so no big deal.

Anyway I'm fine for this to be merged at any time you are. I was starting to look at adding AI for when the target has spell reflection on but looks like spell reflection isn't implemented yet. (Even if it is and I overlooked it, I'm still OK for merge).

Interkarma:
This is working well for me. I'll merge and make a start on remaining effects when I can. Great work!

PR #954, opened 2018-10-14
Refine enemy attack range

Allofich:
A while ago I set enemies to attack from somewhat beyond their actual range. I think I overdid it, though, and while watching a Youtube video I got the impression the player was learning to wait until the enemy swung, then run in and hit in the cooldown afterwards. That's not necessarily a bad thing, but I don't want enemies to be too easy to predict. This PR reduces the extra distance and also actually takes change of target position into consideration, so there shouldn't be enemies attacking from well out of range when the player is stationary, for example, something that I could see happen with the previous implementation (the one currently in master).


Allofich:
Working on additional AI improvements but it's too hard to avoid merge conflicts with the spells PR so I'll wait until that is merged.

PR #958, opened 2018-10-16
Fix ResetWarDataForRegion using regionId - 1

Allofich:
The comparison against -1 keeps the function from running on "Random Ruler", which has region of -1. (Regions of -1 in FACTION.TXT are the one exception that doesn't get corrected down 1 to become 0-based. This is already accounted for in the import.)

The problem was the

`regionData[regionID - 1]`

statements, written back when the faction IDs were 1-based. I think I only looked for `regionID + 1` when fixing these and didn't think to check for `regionID - 1`.


Interkarma:
Thanks for fix!

PR #959, opened 2018-10-16
AI enemy fixes

Allofich:
Must have accidentally removed this earlier.

### DIFF
--- a/Assets/Scripts/Game/EnemyAttack.cs
+++ b/Assets/Scripts/Game/EnemyAttack.cs
@@ -251,6 +251,8 @@ namespace DaggerfallWorkshop.Game
                 else
                     SendDamageToPlayer();
             }
+            else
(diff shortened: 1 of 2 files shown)


Allofich:
Fixed a few more:

- Enemies shot ranged spells too fast, because it was using Daggerfall Unity's update rate and not classic's update rate
- Enemies with ranged spells now hold ground when shooting at range, as in classic.
- On-touch spells cast from same distance as melee now. `stopDistance`, which they were using, was too close to the target so they wouldn't cast unless you let them get right next to you.
- Enemies now evaluate whether they have enough spell points to cast a spell before doing spell-related behavior.
- Enemies are more willing to change target when they already have a target but are attacked by someone else. Priority of seen targets vs. nearby targets was adjusted a little. Basically they were a bit too focused on chasing down their existing targets and I felt they ignored other targets who were closer to them too much.

PR #961, opened 2018-10-16
Fix shop titles with %rt macro

ajrb:
Removed the duplicate title generation method.


Interkarma:
Nice one, thank you for the fix!

BTW, that spelling error in region list has probably been around since 2001. I've been carting that same list forward since I built Daggerfall Cartographer (my first DF city explorer) in around 2001. Some of the code written back then is still used in DFU to this day, although its practically unrecognisable now.

ajrb:
It's a rich history that has led (nearly) to a fantastic triumph of your dedication! BTW I read the region name the same way you spelled it, so it was quite nice to know I'm not the only one who mis-reads words like this.

PR #962, opened 2018-10-16
Catch exception in GetGuild method and just log it as it can happen due to bad faction data.

ajrb:
Unfortunately faction data has some erroneous entries that are marked as guilds (Knightly Orders) and don't match a valid Deity. This was happening when trying to talk to anyone if the PlayerEnterExit.factionID got set to one of these factions.

### DIFF
--- a/Assets/Scripts/Game/Guilds/GuildManager.cs
+++ b/Assets/Scripts/Game/Guilds/GuildManager.cs
@@ -208,11 +208,17 @@ namespace DaggerfallWorkshop.Game.Guilds
         /// </summary>
         public Guild GetGuild(int factionId)
         {
-            FactionFile.GuildGroups guildGroup = GetGuildGroup(factionId);
-            if (guildGroup == FactionFile.GuildGroups.None)
+            try {
+                FactionFile.GuildGroups guildGroup = GetGuildGroup(factionId);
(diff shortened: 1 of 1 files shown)


Interkarma:
Thank you. One less exception is always a good thing. 👍 

The FACTION.TXT file is so riddled with problems (and different problems in different versions of the game) that its one of the more common causes of game not starting up for new users. I've been considering for a while now making this a DFU specific file and just ignoring the one in user Arena2 folders. Then we could slowly resolve issues in faction data and everyone would be on the same page with faction data at start.

This wouldn't be compatible with old translation mods that edited name in faction data. But we can probably add a better method of doing that anyway without touching this big messy file other than for fixes.

ajrb:
I would have supported that a year ago, but now I have second thoughts as I think we've nearly worked though all these now and it's not nothing that DFU is driven off the original datafiles IMO!

Interkarma:
I want to support as much of the old game data as possible too. The only bits I've split out have been tables and strings baked into the .exe, quests, and pre-converting HMI to MID. There's been a good reason for each one, either technical or design for access to modders. You're right the faction data is a bit iffy in this case, which is why I haven't done anything with it yet.

I'll keep monitoring things as we get closer to alpha. I'd rather not make this change right now either.

PR #963, opened 2018-10-17
Cleanup

Allofich:
Next two in the list.

`Scripts/Game/MagicAndEffects/Alteration`
`Scripts/Game/MagicAndEffects/Destruction`

Very minor changes. Same type of fixes as in the earlier PRs. The only new one is that I changed one function, that had an `int` return value that is never used to a `void` function. The function, `IncreaseMagnitude`, is called 2 times within the class, `DrainEffect`.

No whitespace changes, they were already perfect in that regard.

`ElementalResistance.cs` and `VampiricFortifyEffect.cs` both had unused variables, now commented out. Maybe take a look at those if you like, Interkarma.

Comment out unused variables
Add end of file newline
Use array initializers
Remove redundant conditionals
Remove int to int casts
Make function with unused return value void


Interkarma:
Cheers, will review when I can.

Interkarma:
I'm happy with these changes. Thanks again for all your work on this!

PR #964, opened 2018-10-17
Import books from modding locations

TheLacus:
BOK*.TXT files are seeked from StreamingAssets/Books and inside mod bundles.
This allows to provide overrides via the asset-injection framework without replacing core files on disk.

### DIFF
--- a/Assets/Scripts/API/BookFile.cs
+++ b/Assets/Scripts/API/BookFile.cs
@@ -102,6 +102,17 @@ namespace DaggerfallConnect.Arena2
         }
 
         /// <summary>
+        /// Open a book file from binary data.
+        /// </summary>
+        /// <param name="data">Byte array to parse as a book file.</param>
+        /// <param name="name">Name to describe book.</param>
+        public void OpenBook(byte[] data, string name)
+        {
+            bookFile.Load(data, name);
+            ReadHeader();
+        }
+
+        /// <summary>
         /// Reads the TextResource tokens for this page record.
         /// </summary>
         /// <param name="page">Page index.</param>
--- a/Assets/Scripts/API/FileProxy.cs
+++ b/Assets/Scripts/API/FileProxy.cs
@@ -103,9 +103,7 @@ namespace DaggerfallConnect.Utility
         /// <param name="name">Name, filename, or path  to describe memory buffer.</param>
         public FileProxy(byte[] data, string name)
         {
-            fileBuffer = data;
-            managedFilePath = name;
-            fileUsage = FileUsage.UseMemory;
+            Load(data, name);
         }
 
         #endregion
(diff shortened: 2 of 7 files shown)


ajrb:
I'm not a big fan of breaking the fourth wall like that personally MD, but lets see what others think.

JayH2971:
I'd be in favor of an in-universe "Guide to the Iliac Bay" with tips like disease and poison, travel time, resting, enemy weapon resistances, guilds, and so on. It could be given at character creation. If a writer were needed, we could pool up ideas on the forums and I could write it.

Allofich:
I'm not a fan of this being part of stock DF Unity. I don't like breaking 4th wall either and original Daggerfall already documents itself. Going beyond those basics also could enter `is this a tip or an exploit` territory.

JayH2971:
I think Daggerfall's documentation is one point most people never found. The main quest was programmed to give you hints only during the optional tutorial, which due to a bug expired before you ever made it to a town. If you decided to talk to peasants, they'd give you scant hints as rumors. Everything combined, it sounds like an astronomically low number of people would ever get a well-rounded introduction to Daggerfall. I certainly didn't.

I'm not a dev and I can't make this decision, but I do believe the average player would find a use for it. It wouldn't have to be comprehensive or overly detailed, but just informative of major sticking points the average DF player should know in order to survive.

Allofich:
I've no problem with a mod!

TheLacus:
We could as well expose `DaggerfallBookshelf` and allow mods to provide new books. Something like a static property as `IBooksProvider` or just a callback that gives ids/names used to populate bookshelf, which are later loaded by `TextProvider`. 

Even better if contributions from multiple mods can be all used. One method to push a list of names (usable by all mods), and another with the choice behaviour.

Allofich:
> I think Daggerfall's documentation is one point most people never found. The main quest was programmed to give you hints only during the optional tutorial, which due to a bug expired before you ever made it to the destination town. If you decided to talk to peasants, they'd give you scant hints as rumors. Everything combined, it sounds like an astronomically low number of people would ever get a well-rounded introduction to Daggerfall. I certainly didn't.

I hope we have that quest fixed for DF Unity. I think it is, right? Even if it is, yeah, the game could maybe have used a more thorough introduction. I never really had a problem with it, but I can see that point.

I think there will be users who want to play with only "canon" material, regardless of the quality of any alternatives, and being the only playable remake of the game so far, I think any alternative content is best having some way to opt-out of it. But I think your idea of a book could be quite cool, and you said it would be in-universe, too.

ajrb:
There are a few places where we will want to allow books from mods to be integrated into the game. Bookshelves, shops and random drops.

I like the concept of BookProvider you propose, with 1 for classic books and more for bundles added by mods. However it should not be tied to shelves specifically, or buildings or guilds. I'll happily change the implementation of the Bookshelves and Shop shelves to select from extended lists of books once that facility is provided by mod system. I will leave the drops to someone who knows that code.

JayH2971:
> I hope we have that quest fixed for DF Unity. I think it is, right?

Yeah, that part is much better off in DFU. Interkarma took care of it early in main quest development.

TheLacus:
Good point about random drops and other books locations. I think the first step is to make a list of all possible choices that are selected evenly.

A future improvement could be to allow mods to provide a rarity weight like later games. I think this should be left to mods implementation, this is why i suggested to expose locations and/or other details.

PR #967, opened 2018-10-18
Improve enemy combat behavior

Allofich:
Main goal of this PR is to address melee combat in Daggerfall always being the enemy relentlessly advancing while the player backpedals and swings, and make combat a little more interesting.

- AI now sometimes backs off while in combat. This is influenced by level difference with the target, and whether the AI takes significant damage from a hit. Some enemy types (undead, except for the "smarter" ones like vampires and liches) never back off, and city guards don't unless the player has started resisting arrest (while they are still in "arrest the player" mode). They always choose to advance if the target is paralyzed or showing their back.

- Added a timer to "pad" changes in AI movement behavior, so it doesn't react instantly to things like the player entering or leaving range.

- Slightly reduced touch-spell range, as AI seemed a little too eager to use them before really closing with the player

- Fixed (I think) a problem with enemies incorrectly determining distance to their targets due to comparing the distance to their old target to their new target after changing, when they should have been comparing the change in position of a single target. Symptom: AI suddenly swinging at the air while changing targets

- Improved AI response to target out of 45 degree angle. They will now more reliably stop and will also back up if the degree gets even tighter (out of 55 degree angle)

- Tweaked AI evaluation of targets. They don't place quite as much value on seeing the target now (i.e., more emphasis on distance) and will only not switch targets when attacked if their current target is very close to them.

This stuff will probably still need tweaking, but I'm happy enough with it now to submit as a PR. I might also put an option to disable this stuff and stick to a simpler "classic" mode AI, but I want to get this into a solid state and put it out for people for testing and feedback if that's alright with you Interkarma.


Allofich:
Added gravity for stationary enemies if they have a combat target. Not sure if this is the best way, but it doesn't hurt the fps and solves the issue of enemies floating in the air and shooting arrows, etc. which I have sometimes seen.

Allofich:
Added a setting in enhancements. If enabled, the new behavior of this PR and any other additions to combat AI for DF Unity will be enabled, otherwise it will be as close as I can get it to classic AI.

Also adjusted weapon range to actually be the same as classic now (I was a little off before), and removed the shorter on-touch spell range, as the slightly shorter weapon range now is close enough for touch spells to hit.

Interkarma:
Sounds awesome, thank you! I've run out of time to review this one properly. I'll target this in a couple of days, as soon as I can make time.

PR #970, opened 2018-10-20
Fix incorrect starting skill values

Allofich:
Fix for https://forums.dfworkshop.net/viewtopic.php?f=24&t=1366.

The issue: `startingSkillSum`, used to compute when the character gains a level, was being set in DF Unity in `AssignCharacter()`, which was being computed before skills from biography bonuses were added. In classic, it isn't computer until after character creation finishes. This meant not only that the skill bonuses from the character background (if they were a primary, major or minor skill) were being considered as skills raised during active gameplay, getting you to levels 2 and 3 sooner, but it also meant the player would have a lower `startingSkillSum` than they should have, so they would always level up a bit sooner than they should have been.


Interkarma:
Thank you for the fixes!

PR #971, opened 2018-10-20
Stop fast travel countdown when canceling

Allofich:
Fixes https://forums.dfworkshop.net/viewtopic.php?f=24&t=742.

It's not realistic that you can cancel and "turn back time" after days have already supposedly passed, but it's good enough for now and is convenient, anyway.

As an aside, in classic, the background world events are actually being run for each day as the counter goes down, so it isn't just decorative like it is for us. And so of course, you can't cancel. We could do that, too, and it might speed up loading of a new location, although as I recall the background event processing doesn't really take up significant time. If we do that, we'd need to remove the ability to cancel, or store all changes to be undone when canceling (would probably be a pain to do that). Also it would be kind of cool if canceling just stopped the countdown where it was, leaving the player somewhere mid-way on the trip.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTravelPopUp.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTravelPopUp.cs
@@ -369,6 +369,12 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             }
         }
 
+        public override void CancelWindow()
+        {
+            doFastTravel = false;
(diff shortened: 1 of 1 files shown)


Interkarma:
Thanks for the fix. Happy to keep this simple for now. I agree that it would be awesome to dump player somewhere half way if they cancel travel. Could even add some hooks for mods to interrupt travel in this manner, e.g. for a random encounters during fast travel mod. Anyway, simple is best for now.

PR #972, opened 2018-10-20
Fix wrong text for Pacify effect

Allofich:
https://forums.dfworkshop.net/viewtopic.php?f=24&t=1363


Interkarma:
Thanks so much for fixing. Something must have distracted me mid-stream, I'm normally decent at not making basic copy-paste errors like this one.

PR #973, opened 2018-10-20
Don't stack enchanted ingredients

petchema:
Totally forgot about those!
https://forums.dfworkshop.net/viewtopic.php?f=28&t=1274&start=10#p16257


Interkarma:
Me too! Thanks! I also saw your comment about equipped gems on forums and to be careful in item maker. Thank you. :)

PR #976, opened 2018-10-23
Climbing motor changes

MeteoricDragon:
This is Still a WIP.  I'm submitting this PR so you or anyone else can have a taste at what I'm doing so far.

Things you can do:

- Climb into the ceiling, and wait a second for hanging mode to activate.
- While hanging, press against a wall squarely to activate climbing mode and disengage hanging.
- Look and move while hanging in any direction on level ceilings.
- hang on the underside of small eaves.

Things you can't do... yet.

- climb over the eaves
- fail at hanging - (for now, I want to succeed at hanging until I fix some more bugs)
- Reset gravitational speed when recovering from slipping while climbing, or after grabbing the walls.
- Strafe into the wall while hanging without plummeting to your death.


Interkarma:
Great! Looking forward to playing with it. :)

Will likely leave this one open for a couple of releases until we roll into 0.6 proper, and you're happy it's past the WIP stage.

MeteoricDragon:
It might work like this.  Player hangs for 2 seconds before he fails a skill check and goes into LosingGrip mode.  During LosingGrip, he cannot move for about  .5 seconds, and at the end of the .5 seconds he needs to pass a check to see if he actually completely loses grip before falling.  It gives the sudden feeling of "uh oh, I might fall now!"

Interkarma:
On the topic of climbing over eaves, someone on forums has reported its impossible to climb out of the coffin tunnel you drop down in Scourg Barrow. This is apparently possible to climb out of in classic. The angled walls appear to be the culprit here.

https://forums.dfworkshop.net/viewtopic.php?f=24&t=1437

Allofich:
I thought in classic it was only possible to climb out if you picked a wall there that was flat. Will check later.

Interkarma:
Thanks Allofich. I haven't confirmed bug report either, I just wanted to drag MD into the conversation as he's the climbing mastermind. It would be good to get some confirmation when we can. :)

I've also implemented "cast spell" activators now, so users can get a free levitate out of that hole if they need one. I don't think it's a show-stopper either way.

Interkarma:
> I might be able to fix this along with the changes for hanging. I'm going to need to use the same system that gets used to find the adjacent wall when side climbing.

This might be even easier to fix that expected. Check my testing notes in below post.

https://forums.dfworkshop.net/viewtopic.php?f=24&t=1437&p=16806#p16806

The problem only started after 26 Aug 18, so it's fairly recent. I just think some of the changes in advanced climbing or mantling detection might be getting stuck by the walls narrowing in around player as they climb. The north surface itself is perfectly flat and player can't climb those angled walls in classic either.

Maybe just need to tune radius of those casts or add a bit of special handling to the climb system? Hopefully not so hard to fix now. :)

Allofich:
Checked that spot in classic. You can only climb out if you use the flat wall. The two walls on the side start flat but start to taper inward as you get near the top. When you reach the part where it starts to taper, the climb upward will stop and you won't any further up, or you will fall through the wall out into the "void".

IMO this works fine as it is. Figuring out how to get out of this pit, including choosing the right wall, is part of the original game. I'd say leave it out for the non-advanced climbing system at least.

Interkarma:
>IMO figuring out how to get out of this pit, including choosing the right wall, is part of the original game. I'd say leave it out for the non-advanced climbing system at least.

Agree with this. And as per my testing notes above, the inability to climb the flat wall is a fairly recent regression shortly after Aug 26 when it was last working.

MeteoricDragon:
If there were a way to repeatedly get the normal of the wall the player is climbing, then the climbing movement vector could be adjusted using Vector3.ProjectOnPlane(MoveDirection, PlaneNormal), but the normal is decided when the player starts climbing.  That was the best time to decide it because it's the only time the player is guarunteed to be facing the wall and thus the only time that controller.transform.forward would find the wall.

Interkarma:
I don't think it's to do with wall normal. Please check my notes in the forum link above.

It's a perfectly flat wall, the player starts climbing fine and only gets stuck half way up as the passage narrows around them on the sides. It's just going to be some of the other climbing logic finding the space too narrow at that point. You don't have to worry about climbing angled walls or anything like that, the player can definitely only climb the flat wall. :)

MeteoricDragon:
@Interkarma I trust you that it's a flat wall, but I'm still mind-boggled that it is flat, but looks not flat.

Interkarma:
Interesting. Here's an in-editor screenshot of the north wall taken from the base looking up with brightness increased. Looks flat from that angle.

north-wall

However, move the camera down a little and we can see there _is_ a very subtle difference in angle, as the lower part of the wall cuts away before the upper part. So you're right, there's definitely a change in normal there. Your eye is better than mine (and probably most people) here just from eyeballing it. :)

north-wall-2

Still, the normal set at start of climb is still very close to what's needed from top to bottom. Is it possible to add a couple of degrees of tolerance that would not be blocked by the slight change in angle of north wall, but would still be blocked by steeper inward angle of east and west climbs?

MeteoricDragon:
I can experiment with some things.  But i need to fix some other bugs before I can do that.

Interkarma:
>I can experiment with some things. But i need to fix some other bugs before I can do that.

For sure, no priority on this one. I added the "cast spell" activators yesterday so users could click the nearby banner and levitate out specifically to help work around the issue. This one isn't a blocking issue for normal play at this time.

The gliding bug when stuck in rappel mode is probably the main bug to fix, and even then it's fairly rare. Your whole climbing system is fairly epic and I'm having so much fun with it. :)

MeteoricDragon:
An update: The player can now hang after climbing and pushing his head into the ceiling for a little while.  And they can also grapple their way into a wall and kind of catch it as they start falling,  I'd like to make it so that they simply start climbing from hanging position, though.  I'm working on that.

MeteoricDragon:
I found out that when going from hanging to climbing the code tries to force the player into climbing but immediately after forcing him the players climb is canceled because the sides Collision flag hasn't been set. It hasn't been set even though the player has been pushing into the wall from a hanging position.

MeteoricDragon:
What's left is transitions for special cases between climbing and hanging.

- Climb down wall and there's ceiling in front of you underneath the wall, Transition climb to hang.
- hang backwards up to wall (this will be useful for scaling eaves.)

Theres also special cases where the player could be climbing or hanging, and enter a door.  PlayerEnterExit should clear their climbing status somewhere in that class.

MeteoricDragon:
@Interkarma @Allofich It says merged More Cleanup #1028 Can I still commit changes to this branch for merging later?

Interkarma:
>@Interkarma @Allofich It says merged More Cleanup #1028 Can I still commit changes to this branch for merging later?

Yep, keep on plugging. Allofich didn't clean up any of the files which were part of your PR.

Interkarma:
>Player can now rappel up walls and over eaves.  Press backward while hanging.

Awesome! It's exciting watching all this come together. :)

MeteoricDragon:
Thanks! I found a bug. If you climb and start hanging on ceiling and go to another wall, when you try to start climbing the second wall you retain the movement direction from the first wall when you release from the ceiling. Also need to make a movement that swoops down under and up to transition from climbing to hanging underneath

MeteoricDragon:
@Interkarma How can I temporarily disable the player's vertical and/or horizontal mouse look movement? individually or together?

MeteoricDragon:
I'm working on the rappel movements.  The player is able to mantle some small eaves, but when trying to do other eaves he glitches and warps back a few feet.

Interkarma:
Hey mate! I'm about to close off 0.5 and 0.6, and move into 0.7. Are you happy for me to merge this into master or would you like to keep working on PR some more first?

I don't mind if things are a bit broken at the start of a new point release cycle, as long as game is mostly playable. Users can always fallback to 0.5 stable if there are too many issues.

MeteoricDragon:
Yes. go ahead.

Interkarma:
Cheers. Will do one final test and get the code merged later today.

MeteoricDragon:
rappelling if you climb under a wall to a ceiling beneath doesn't attach you to the ceiling, by the way.  I've got a newer version of code that does it.

MeteoricDragon:
Ok, there's the updated code.  Now you can technically traverse between all surfaces. but there are some bugs that must be fixed.  

- When swooping under a wall to a ceiling, it doesn't say hanging mode immediately until you start moving.
- Sometimes when swooping under to a ceiling you get multiple hanging mode messages.
- When swooping over to a gabled roof, sometimes the player gets forced into a permanent crouch and cannot uncrouch, even across saves until restarting the game.

Interkarma:
Thank you. :) I've had a good play with this now and it's clear you've put a huge amount of love and work into this. It all works quite well, I had very few issues in my sessions. Users can always disable advanced climbing if they strike the problems you're still working on.

I'll get this merged into master now so it goes into next round of builds for everyone to play with. I'm looking forward to having this in my games. 👍

Interkarma:
I just noticed an odd problem with rappel. If I walk backwards off that mound in the middle of Daggerfall's shopping district then try to climb a nearby building, rappel mode will trigger briefly. When this happened, I was also forced into a crouch and fell through world while trying to stand up again. Quite the surprise! 🤣 

I'm still having issues climbing that surface in Scourg Barrow as well, just so you know. :)

MeteoricDragon:
@Interkarma I see, i'll look into that. Did you have any trouble mantling small eaves?  Push into the eave, then press and hold back.

Interkarma:
No troubles with that at all. I worked out how to do it after a few attempts. I also spent some time climbing around the ceiling of a dungeon and feeling very agile. You're doing good work here. :)

PR #977, opened 2018-10-24
Geographic context backgrounds & HUD arrow count

ajrb:
Both are toggled by a new setting.
Arrow count position is centered because I could not figure out how to ensure it was just left of the compass where it looks brilliant. I found that it disappeared under the compass at different resolutions. It would be nice to move it if someone can help.


Interkarma:
Looking forward to checking this out. :)

Your arrow count position will need to be positioned relative to the compass and take into account the LocalScale compass is using. I'm about to hit bed, but I'll put together an example when I can find time tomorrow that should help.

PR #978, opened 2018-10-24
Cleanup

Allofich:
Next one's in the list, going through the script files sequentially.
There isn't much to clean up in the magic effect files, so I went ahead and did all of the rest of them.

`Scripts/Game/MagicAndEffects/Effects/Diseases`
`Scripts/Game/MagicAndEffects/Effects/Illusion`
`Scripts/Game/MagicAndEffects/Effects/Mysticism`
`Scripts/Game/MagicAndEffects/Effects/Poisons`
`Scripts/Game/MagicAndEffects/Effects/Powers`
`Scripts/Game/MagicAndEffects/Effects/Restoration`
`Scripts/Game/MagicAndEffects/Effects/Thaumaturgy`

The changes are all straightforward, the most significant ones (still just simple cleanup) are in `Teleport.cs`. Just in case I tested the "Recall" spell after the changes and it worked fine, whether teleporting within the same area or to a different one. 

Add end of file newlines
Revise comment
Merge nested if statements
Remove redundant array initializer
Remove redundant return
Remove redundant override method
Remove redundant fetching of SerializablePlayer and PlayerEnterExit in
Teleport.cs
Remove redundant boolean check
Use readonly keyword where appropriate
Simplify logic expression in PoisonEffect.cs
Comment out unused variables


Assets/Scripts/Game/MagicAndEffects/Effects/Mysticism/Teleport.cs
```diff
@@ -231,15 +223,6 @@ private void PlayerEnterExit_OnRespawnerComplete()
             if (caster == null || caster != GameManager.Instance.PlayerEntityBehaviour)
                 return;
 
-            // Get peered SerializablePlayer and PlayerEnterExit
```
Interkarma: This does seem redundant here - but I think I added this for an edge case. I just can't recall exactly the situation right now. I think the event can fire out of order to effect starting, which means these references are not otherwise collected.

Rather than remove this completely, I'd prefer a CacheReferences() call in place of this block. It would also be helpful if CacheReferences() returned true when both references are successfully collected. Then in place of this block we could use:

    if (!CacheReferences())
        return;

That would make me feel a bit better when handling this event.

Assets/Scripts/Game/MagicAndEffects/Effects/Mysticism/Teleport.cs
```diff
@@ -164,11 +158,10 @@ void TeleportPlayer()
                 // Need to load some other part of the world again - player could be anywhere
                 PlayerEnterExit.OnRespawnerComplete += PlayerEnterExit_OnRespawnerComplete;
                 playerEnterExit.RestorePositionHelper(anchorPosition, false);
-            }
 
-            // When moving anywhere other than same interior trigger a fade so transition appears smoother
-            if (!sameInterior)
```
Interkarma: That's good. And since we're not checking this bool twice, can probably do away with:

    bool sameInterior = IsSameInterior();
    if (sameInterior)
    ...

and just change if statement to:

    if (IsSameInterior())
    ...

It cleans up another line and removed a value checked only once now.

Interkarma:
Cheers mate. Will merge this soon.

PR #979, opened 2018-10-24
Stacking and equipping

petchema:
Don't stack equipped items with other items; and when asked to equip a stack, only equip one of the items.

Limitations:
- when unequipping an item, no attempt is done to stack it with other similar items;
- characters may have equipped stacked items already; Is it worth fixing while loading a gamesave?

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
@@ -1217,7 +1217,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
                 // Shouldn't happen
                 if (oneItem == null)
                     return;
-                playerEntity.Items.AddItem(oneItem, noStack: true);
+                playerEntity.Items.AddItem(oneItem, preferredOrder, true);
                 item = oneItem;
             }
             // Try to equip the item, and update armour values accordingly
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
@@ -1210,6 +1210,16 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
                 }
                 return;
             }
+            // If more than one item selected, equip only one
+            if (item.IsAStack())
+            {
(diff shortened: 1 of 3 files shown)


Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
```diff
@@ -1210,6 +1210,16 @@ protected void EquipItem(DaggerfallUnityItem item)
                 }
                 return;
             }
+            // If more than one item selected, equip only one
+            if (item.IsAStack())
+            {
+                DaggerfallUnityItem oneItem = playerEntity.Items.SplitStack(item, 1);
```
ajrb: Maybe could have SplitStack() add the item to the internal collection before returning the new item. Then this becomes:

if (item.IsAStack())
    item = playerEntity.Items.SplitStack(item, 1);

Just a suggestion. I think the default case should be splitting inside the ItemCollection - it can always be removed if needed for some circumstance.

ajrb:
Looks good to me, I like the approach. Made a suggested change - see what you think.

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
petchema: That simplifies that case, but will make other cases more awkward later; I'm thinking stack splitting to only drop/pick part of a stack, for example.
Leaving that responsibility to the caller is also more flexible, it can decide what position to use to put the items back in the list, if it so wishes.

petchema:
Figured out how to re-stack unequipped items

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
ajrb: Fair enough :)

petchema: Actually I went ahead and tried to implement user-controlled stack splitting (using control keys), and since the inventory uses TransferItem() between lists, your suggestion matches better existing code.
I'm only transferring one item at a time, I'm not sure how to ask the user how many items to pick/drop
https://github.com/petchema/daggerfall-unity/compare/stacking-and-equipping...petchema:stack-splitting

PR #980, opened 2018-10-25
Add 3d arrow projectiles

Allofich:
This replaces the temporary instant arrow hits with actual 3d projectiles.

This is probably messier than it should be, as I put it together with trial and error. For example, I couldn't use the existing collider used by `DaggerfallMissile`, because it won't work with the concave shape and I had to make the rigidBody kinematic. Instead I use a sphere ray cast for the arrows. Interkarma, please feel free to gut this implementation if you don't like it. Functionally, it works, though.

All existing arrow code was adapted to work with it. I had to account for various things, like getting the player weapon at the time of shooting, not the time of damage calculation, since they might unequip their bow before an arrow hits the target.

Known issues:
- The arrow hit sound plays without any doppler because the same code as it used for spell projectiles is used, and they aren't using doppler. (Seems to me like they should, even if classic doesn't)

- When looking up or down while shooting an arrow, the arrow spawns offset from the center of the screen where it should spawn from.

- There is a merge conflict with the Serializable enemy prefab in the AI combat PR, which will need to be fixed. I can update the PR as needed or you could do it yourself if you want to.

- The arrows can sometimes displace enemies or the player a little bit, I think because of collision with the 3d.

### DIFF
--- a/Assets/Scripts/Game/DaggerfallMissile.cs
+++ b/Assets/Scripts/Game/DaggerfallMissile.cs
@@ -79,8 +79,11 @@ namespace DaggerfallWorkshop.Game
         float initialRange;
         float initialIntensity;
         EntityEffectBundle payload;
+        bool isArrow = false;
+        GameObject goModel = null;
 
         List<DaggerfallEntityBehaviour> targetEntities = new List<DaggerfallEntityBehaviour>();
+        RaycastHit arrowHit;
 
         #endregion
 
@@ -126,6 +129,12 @@ namespace DaggerfallWorkshop.Game
         {
             get { return caster; }
             set { caster = value; }
+        }
+
+        public bool IsArrow
+        {
+            get { return isArrow; }
+            set { isArrow = value; }
         }
 
         /// <summary>
@@ -183,6 +192,33 @@ namespace DaggerfallWorkshop.Game
                 }
             }
 
+            // Setup arrow
+            if (isArrow)
+            {
+                // Create and orient 3d arrow
+                goModel = GameObjectHelper.CreateDaggerfallMeshGameObject(99800, transform);
+
+                Vector3 adjust;
+                // Offset up so it comes from where it looks like it should from on the enemy sprites
+                if (caster != GameManager.Instance.PlayerEntityBehaviour)
+                    adjust = new Vector3(0, 0.6f, 0);
+                else
+                {
+                    // Offset forward to avoid collision with player
+                    adjust = caster.transform.forward * 0.6f;
+                    // Adjust slightly downward to match bow animation
+                    adjust.y -= 0.1f;
+                    // Adjust to the right or left to match bow animation
+                    if (!GameManager.Instance.WeaponManager.ScreenWeapon.FlipHorizontal)
+                        adjust += caster.transform.right * 0.15f;
+                    else
+                        adjust -= caster.transform.right * 0.15f;
+                }
+
+                goModel.transform.localPosition = adjust;
+                goModel.transform.rotation = Quaternion.LookRotation(GetAimDirection());
+            }
+
             // Ignore missile collision with caster (this is a different check to AOE targets)
             if (caster)
                 Physics.IgnoreCollision(caster.GetComponent<Collider>(), this.GetComponent<Collider>());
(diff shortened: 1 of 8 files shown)


Allofich:
Tweaked the position of the arrow a bit (it closely matches the sprite when shooting with a level view now) and slightly slowed the arrows as they were a little too hard to see.

Allofich:
Made arrow impact use doppler.

Interkarma:
> The arrow hit sound plays without any doppler because the same code as it used for spell projectiles is used, and they aren't using doppler. (Seems to me like they should, even if classic doesn't)

I started with doppler for spell missiles and removed it because it just made spells sound weak and ineffective at range. It may not be "realistic", but removing doppler makes them feel punchier and more energetic on impact. They're much more satisfying without.

Interkarma:
Good job on this, it looks fine on a read. I'll play with it when I can and offer any feedback. Thanks for putting all this together! 👍

Allofich:
> If you use a BoxCollider, Unity should create a minimum fitting box around the projectile. 

I tried using a BoxCollider instead of a SphereCollider, assuming that's what you mean, but the error still happens.

```
Non-convex MeshCollider with non-kinematic Rigidbody is no longer supported in Unity 5.
If you want to use a non-convex mesh either make the Rigidbody kinematic or remove the Rigidbody component.
```

Seems like the problem is with the Rigidbody component, not the collider type.

Interkarma:
>I tried using a BoxCollider instead of a SphereCollider, assuming that's what you mean, but the error still happens.

Can you explain what the error you're encountering is exactly? I'm not really clear on the problem you encountered. I'll try to help if I can. :)

Allofich:
I quoted it, the message starting with "Non-convex MeshCollider"

Same as with the SphereCollider, if I make the Rigidbody have "Is Kinematic" checked, the error won't appear, but `OnCollisionEnter()` in `DaggerfallMissile.cs` doesn't fire. That's why I looked for another solution and used `Physics.SphereCast`.

Interkarma:
BoxCollider and SphereCollider aren't based on MeshCollider, the convex flag doesn't apply to them.

I'll take a look when I can and help. It just sounds like some of the prefab properties need to be tweaked. Have anotner look at how the spell missiles are set up. They just use a basic SphereCollider and collisions work ok without pushing things around or needing to use sphere casts.

We'll sort it out. :)

Interkarma:
Apart from easy 3D model helpers, the missile class was written from the start to support arrows - any kind of missile in fact. Its missile agnostic. All the collision handling you need is already in there for you. Either a SphereCollider or BoxCollider will be fine.

I'll add an example for you when I can over weekend and we can rework PR from there. :)

Assets/Scripts/Game/DaggerfallMissile.cs
```diff
@@ -183,6 +192,27 @@ private void Start()
...
+            if (isArrow)
+            {
+                // Create and orient 3d arrow
+                goModel = GameObjectHelper.CreateDaggerfallMeshGameObject(99800, transform);
+
+                Vector3 adjust;
+                // Offset up so it comes from where it looks like it should from on the enemy sprites
+                if (caster != GameManager.Instance.PlayerEntityBehaviour)
+                    adjust = new Vector3(0, 0.6f, 0);
+                else // Offset forward to avoid collision with player and to the right to match bow animation
```
ajrb: Remember there's an option to switch weapons to left hand.. looks wierd with a bow but is possible.

Allofich: Thanks, fixing it now.

Allofich: This part is fine, as that option just draws the weapon to the left, but bows are still equipped in the right hand. But the arrow won't spawn in the right place, so I have to fix that.

Allofich:
I'm done with it for now.

> I'm going to revert some of the changes to the missile class

Go ahead!

Allofich:
> I'll send this to an "arrows" branch on my side

I see that branch already exists, so I'll re-target it to that.

ajrb:
> > Enjoy and relax mate - DFU stuff can wait.
> 
> I ended up having a great weekend. Good to feel less under pressure again for once. :) Didn't quite get those builds out, but taking care of that this morning.

Very glad to hear it. You have to pace yourself man, this is a marathon not a sprint... and we would be screwed if you burned yaself out!

PR #981, opened 2018-10-25
Implemented a Notebook tab and Finished Quests tab of Journal window

ajrb:
* Assume will go in 0.6, but didn't target that branch, is up to you to decide.

* Created a DaggerfallUI.txt strings table. Intending this to have messages for UI windows that only have a few strings. Heavy string UI windows like SpellMaker or Automap will still have their own. Let me know if you'd rather have one per window even if there's only a handful of messages. Seems that journal window now has quite a few, more than originally anticipated.

* Even though PlayerNotebook is a member of PlayerEntity it serialises to it's own file and controls this separate from the player serialisation.

* I added a message that pops up if player tries to travel while indoors like the one for transport changes. This isn't in classic but is both more consistent and helps a player clicking on a quest in the journal understand why it doesn't bring up the map. (bug report avoidance!)

* Enabled shadows for text in conversation panel on talk window. This is so I can change the shadow colour to indicate which text chunks will be copied. Hope you don't mind Nystul, it was the most elegant solution I could find.

### DIFF
--- a/Assets/Game/Addons/UnityConsole/Console/Scripts/DefaultCommands.cs
+++ b/Assets/Game/Addons/UnityConsole/Console/Scripts/DefaultCommands.cs
@@ -1428,7 +1428,7 @@ namespace Wenzil.Console
         {
             public static readonly string name = "add";
             public static readonly string description = "Adds n inventory items to the character, based on the given keyword. n = 1 by default";
-            public static readonly string usage = "add (book|weapon|armor|cloth|ingr|relig|gold|magic|drug) [n]";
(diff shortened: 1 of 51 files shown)


ajrb:
This is targetting master, as when I target 0.6 branch it's way behind so this PR had a lot of changes that are just from master branch since I am up to date. Feel free to merge into 0.6 tho!

Interkarma:
I'll definitely merge this to master. I think it's a really exciting upgrade and people should be able to play with it sooner rather than later.

Thanks for all your hard work on this, it's very impressive!

PR #982, opened 2018-10-26
ItemTemplate fixes

ajrb:
These replace edits done by IC ages ago that got lost mid august when IK regenerated them. (56f4d2d4529a984984de16ebc9888b170b83d0f8)

Most edits were re-done, but two were missed:

* Unit weights all had .0 added to the end, which doesn't matter since this is handled by the macro handler so not changed this.

* Fixed book display where they show using holy tome icon. Populating player tex archive + record and then changed GetRandomBook to select one of the 3 variants that are allowed instead of just using the same icon for every book. One of these is the spellbook icon... since classic only uses one icon for books (assuming a bug given variant number) and I think this will be confusing I have reduced variants to 2.

* Also changed the icon for LoC to an unused icon. (was going to be for deeds I think)

### DIFF
--- a/Assets/Resources/ItemTemplates.txt
+++ b/Assets/Resources/ItemTemplates.txt
@@ -5515,7 +5515,7 @@
         "isOneHanded": false,
         "isIngredient": false,
         "worldTextureArchive": 209,
-        "worldTextureRecord": 8,
+        "worldTextureRecord": 7,
         "playerTextureArchive": 0,
         "playerTextureRecord": 0
     },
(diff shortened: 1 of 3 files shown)


ajrb:
So on further testing it's only books from saves before the templates were updated. So I suppose this is up to opinion as to whether we want to do this, turns out it's not a bug really - just a version issue with the saves from before mid aug.

This would mean that we get two variants of book icons, but if that's not desirable - this PR can be binned.

PR #983, opened 2018-10-27
Stack splitting

petchema:
(patch on top of stacking-and-equipping)
Allow user to pick or drop part of a stack by maintaining control key pressed while clicking on the stack.

Message box can probably be improved.


petchema:
Not ready, {Wagon}CanCarry() checks happen too early, they're missed occasions for splitting stacks in shops, some callers do not expect TransferItem() to be cancelled interactively,...

PR #987, opened 2018-10-27
Make all sound effects respect volume setting

Allofich:
Fixes https://forums.dfworkshop.net/viewtopic.php?f=24&t=1009.

I can't reproduce the issue there about music playing even when at 0 volume. For me it turns off completely. But the issues of sound effects ignoring the volume setting are fixed.

Was a bit tricky to figure out because sounds are played in several different ways:
1) Goes through a PlayOneShot function in DFAudioSource. In this case, DFAudioSource handles applying the volume setting and the call to PlayOneShot can ignore the volume setting. (No fixes needed here)
2) Goes through PlayOneShot in AudioSource. In this case, the volume setting needs to be applied in the call. (This needed to be done for enemies' PlayAttackSound() and for sounds in character creation)
3) Goes through Play() on an AudioSource. In this case, the Audiosource's "volume" needs to be set to match the volume setting. This had to be done in a few different places, and should fix weather sounds/crickets, DaggerfallAction sounds, quest sounds, and riding sounds.

I think the only sound that doesn't respect the sound volume setting now is .VID files. Not sure how their sound is handled.

### DIFF
--- a/Assets/Scripts/Game/AmbientEffectsPlayer.cs
+++ b/Assets/Scripts/Game/AmbientEffectsPlayer.cs
@@ -95,6 +95,9 @@ namespace DaggerfallWorkshop.Game
                 StartWaiting();
             }
 
+            // Update sound volume
+            dfAudioSource.AudioSource.volume = DaggerfallUnity.Settings.SoundVolume;
+
             // Start rain loop if not running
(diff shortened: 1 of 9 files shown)


Allofich:
.VID files now also use the volume setting.

PR #990, opened 2018-10-28
Fix parsing of character background questions

Allofich:
Fix for https://forums.dfworkshop.net/viewtopic.php?f=24&t=1365.


Interkarma:
Awesome, thank you!

PR #991, opened 2018-10-28
Temporary fix for saves with missing caster loadID

Allofich:
A temporary workaround for https://forums.dfworkshop.net/viewtopic.php?f=24&t=1374.

In my experience this happens _every_ time you save with a spell having been cast by an enemy, whether on the player or on another enemy. I'm kind of surprised this hasn't been brought up more on the forum.

I don't know what the underlying cause is, but this treats the symptom and at least allows for loading an affected saved game and continuing play.

By the way, a somewhat pedantic note, but the failure in `GetCasterReference` was throwing an exception. I changed it to `LogError` to indicate the same level of "error", but where this is used for resuming effects, a failure was just a `LogWarningFormat` so there I used a same level `LogWarning`, but unless I'm missing something it seems like these should either both be "errors" or both be "warnings".

Interkarma if you have a fix in the works, you can ignore this PR if you want.

### DIFF
--- a/Assets/Scripts/Game/MagicAndEffects/EntityEffectManager.cs
+++ b/Assets/Scripts/Game/MagicAndEffects/EntityEffectManager.cs
@@ -1521,6 +1521,12 @@ namespace DaggerfallWorkshop.Game.MagicAndEffects
                         continue;
                     }
 
+                    if (instancedBundle.caster == null)
+                    {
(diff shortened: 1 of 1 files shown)


Interkarma:
This is fine for now, thank you. :)

I'm working on properly deserializing caster references for enemies. Will get this completed soon.

PR #994, opened 2018-10-29
Fix for ship interior persistence and headings for finished quests

ajrb:
See: https://forums.dfworkshop.net/viewtopic.php?f=24&t=1426


ajrb:
Screw GitHub. All I wanted was to get the master changes into my repo FFS.

PR #995, opened 2018-10-30
Use virtual size for Panel border textures

TheLacus:
Allow a virtual size to be provided for panel border textures. This brings support for override of SPOP.RCI.


Interkarma:
Thank you! :)

PR #996, opened 2018-10-30
Fix for ship interior persistence and headings for finished quests

ajrb:
See: https://forums.dfworkshop.net/viewtopic.php?f=24&t=1426


Interkarma:
Thanks for all your work on this. :)

PR #999, opened 2018-10-31
Item Icon variants - books and LoC

ajrb:
* Fixed book display where they show using holy tome icon in old saves. Populating player tex archive + record and then changed GetRandomBook to select one of the 3 variants that are allowed instead of just using the same icon for every book. One of these is the spellbook icon... since classic only uses one icon for books (assuming a bug given variant number) and I think this will be confusing I have reduced variants to 2.

* Also changed the icon for LoC to an unused icon. (was going to be for deeds I think)

### DIFF
--- a/Assets/Resources/ItemTemplates.txt
+++ b/Assets/Resources/ItemTemplates.txt
@@ -5515,7 +5515,7 @@
         "isOneHanded": false,
         "isIngredient": false,
         "worldTextureArchive": 209,
-        "worldTextureRecord": 8,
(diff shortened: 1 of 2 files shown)


ajrb:
Please see https://forums.dfworkshop.net/viewtopic.php?f=20&t=1438 and let me know what you all think about this.. yes or no?

Allofich:
I'm OK with it, too. I like the LoC icon, and it being able to differentiate it from letters is useful.

PR #1001, opened 2018-11-01
More cleanup

Allofich:
More simple cleanup. This finishes out the `MagicAndEffects` folder.

Biggest change is that I commented out a chunk of code in `MagicCandleBehaviour.cs` that doesn't appear to do anything, as `myBillboard` doesn't seem to be used by anything. I didn't notice any difference in the candle in-game with or without the code, but please confirm, Interkarma, and let me know if you want me to restore, delete, or leave commented that code.

Whitespace cleanup
Make abstract class constructor not public
Add end-of-file newline
Remove unused variable assignments
Fix local variables hiding class variables
Comment out code in MagicCandleBehaviour that does nothing
Match summary parameter names to function parameters


Interkarma:
>Biggest change is that I commented out a chunk of code in MagicCandleBehaviour.cs that doesn't appear to do anything, as myBillboard doesn't seem to be used by anything.

While the default properties are almost identical to those being set, I like the clarity of setting up the billboard properties at creation time in this case. At minimum, please leave the following in place. Cheers. :)

    myBillboard = go.GetComponent<DaggerfallBillboard>();
    myBillboard.FramesPerSecond = candleFramesPerSecond;
    myBillboard.GetComponent<MeshRenderer>().receiveShadows = false;

Allofich:
Still, this code appears to do nothing. i.e. setting myBillboard.FramesPerSecond to a huge number there doesn't change the FPS that it animates in-game. So unless I'm missing something it seems like it should be removed or re-implemented somewhere else where it has an effect.

Interkarma:
>Still, this code appears to do nothing. i.e. setting myBillboard.FramesPerSecond to a huge number there doesn't change the FPS that the candle animates in-game.

Had a look at this, cause is in DaggerfallBillboard class line 135:

    else if (summary.Archive == Utility.TextureReader.LightsTextureArchive) speed = lightFps;

The FramesPerSecond property is being overwritten because the candle is part of the light archive, and that animation speed is hardcoded into the billboard class.

That's fine then, feel free to remove that code from MagicCandleBehaviour as per your original PR. Sorry for the hassle mate. :)

Allofich:
No problem. The code is removed (not commented out) now.

PR #1002, opened 2018-11-01
Make post-pause input delay for 1/12 a second

Allofich:
Fixes https://forums.dfworkshop.net/viewtopic.php?f=24&t=1377&start=10 by changing from a frame-based solution to a time-based one. Based on my testing there that 5 frames worked with a 60 fps area (1/12 of a second) I made the wait be that long. The issue is fixed for me.

Feel free to reject if you had a different solution in mind, Interkarma.


Interkarma:
This should be fine. I've been kicking around different ways of handling this and a time delay was one solution. I'm perfectly happy with this for now. Thank  you. :)

PR #1004, opened 2018-11-01
Fix cancellation of generation windows

Allofich:
Fixes
https://forums.dfworkshop.net/viewtopic.php?f=24&t=1420
and 
https://forums.dfworkshop.net/viewtopic.php?f=24&t=1418

Also fixes a new issue I found, which is that if you pressed ESC at the screen where you choose whether questions will be automatically answered or not, it would treat it as if you had selected automatic and proceed to the naming screen.


Interkarma:
Thanks for the fix!

PR #1008, opened 2018-11-03
Add missing paralysis resist logic

Allofich:
Fixes https://forums.dfworkshop.net/viewtopic.php?f=24&t=1443. Only way I saw to do it was to add special handling for paralysis in the effect manager. If the logic was within `Paralyze.cs`, I think the effect would have started and we'd have to undo the magic rounds that had been added, etc. If you have a better solution in mind, feel free to do it instead.

Also fixes saving throws. I was wondering why a custom class character with immunity to paralysis was not resisting paralysis from some Giant Scorpions I was testing with. Turned out the paralysis spell they use (Spider Touch) has both the paralysis and magic flags. I messed up with the saving throw logic before and made it only check for one and only one flag type, so it didn't work with 2 flag types. Now, like classic, it works with multiple flags.

Also made the player immune to paralysis in god mode, because it gets in the way of testing.

### DIFF
--- a/Assets/Scripts/Game/Formulas/FormulaHelper.cs
+++ b/Assets/Scripts/Game/Formulas/FormulaHelper.cs
@@ -941,41 +941,53 @@ namespace DaggerfallWorkshop.Game.Formulas
             int biographyMod = 0;
 
             PlayerEntity playerEntity = GameManager.Instance.PlayerEntity;
-            if (effectFlags == DFCareer.EffectFlags.Paralysis)
-                toleranceFlags = target.Career.Paralysis;
-            if (effectFlags == DFCareer.EffectFlags.Magic)
+            if ((effectFlags & DFCareer.EffectFlags.Paralysis) != 0)
             {
-                toleranceFlags = target.Career.Magic;
-                if (target == playerEntity)
-                    biographyMod = playerEntity.BiographyResistMagicMod;
+                toleranceFlags |= target.Career.Paralysis;
+                // Innate immunity if high elf. Start with 100 saving throw, but can be modified by
+                // tolerance flags. Note this differs from classic, where high elves have 100% immunity
+                // regardless of tolerance flags.
+                if (target == playerEntity && playerEntity.Race == Races.HighElf)
+                    savingThrow = 100;
             }
-            else if (effectFlags == DFCareer.EffectFlags.Poison)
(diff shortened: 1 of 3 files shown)


numidium:
If the player is a High Elf with critical weakness to paralysis then they will still be immune, correct? It works that way in classic and people have mentioned in the forums that it's a bit of an exploit. I'm wondering if, going forward, I should make it so you can't add the paralysis CW in the class creator to prevent this. Thoughts?

Interkarma:
Another option is for paralysis CW to cancel out High Elf immunity in checks. There's always variation in populations, and these elves are the 0.1% that didn't inherit immunity from their parents, poor devils. But on the plus side, they were able to develop other gifts in its place while running from spiders and scorpions.

@Allofich Thanks for fixing this. I was looking at tackling this a bit further down the roadmap, but have no problems with the bundle being outright blocked in AssignBundle().

BTW, I already have handling in place for Free Action that operates in isolation and works when casting Free Action either before or after being paralysed. Rather than just block or remove the effect, it allows player to move normally during paralysis - so long as Free Action lasts. If Free Action expires before Paralysis effect then player will resume being paralysed. I like this implementation as it makes effect distinct from Cure Paralysis, and adds a new angle to magical battles where your Free Action might not last as long as the Paralysis you've been inflicted by. I do understand classic just blocks the effect when Free Action is present however, so happy for his to remain in place here. :)

Allofich:
Updated.

- I'm no longer doing anything with FreeAction. Based on your description, Interkarma, I'm assuming it's already handled.

- High Elf immunity is now handled differently from classic. Instead of starting with a base saving throw value of 50 like in every other case, high elves start with 100 (immunity) if a paralysis flag is being used. Advantages/disadvantages for paralysis can then effect this as follows:

- Immunity: +50 to saving throw (granting 100% for a non high-elf using just this advantage)
- Critical weakness: -50 to saving throw (granting 0% for a non-high elf using just this disadvantage. A high elf will still have 50% resistance)
- Low tolerance: -25 to saving throw
- Resistant: +25 to saving throw

This seems like a decent solution to me, as it allows mixing and matching the flags and gets rid of exploits using immunity for all cases, not just the high elf. It gets a little nonsensical for a person to have both "immunity" and "critical weakness" to something, so we could also forbid picking contradictory selections in the custom class maker, and maybe that looks a little better. This works, though, and there might be some difficulty with the other solution that doesn't immediately come to mind.

PR #1009, opened 2018-11-03
Fix overly fast video playback

Allofich:
Fixes https://forums.dfworkshop.net/viewtopic.php?f=24&t=507.

The problem was that `DaggerfallVideo` seemed to assume that there would always be an audio frame to use to update `nextEventTime`, but there are parts in several of Daggerfall's videos where there are stretches of only video frames. Another issue was that the audioBuffer length when the last audio frame played could be very small, leading to short `frameDelay`. Enforcing a minimum `frameDelay` and accounting for parts where there are no audio frames fixes the speed ups at the end of `ANIM0000.VID` (the book intro) and `ANIM0005.VID` (meeting Lysandus's ghost). It may also slightly improve the timing of the Bethesda logo intro (you can see the sparkle disappear at the end now) and the end frames of some endings.

I tested all the video files in the game with this PR and they all play fine.

### DIFF
--- a/Assets/Scripts/API/VidFile.cs
+++ b/Assets/Scripts/API/VidFile.cs
@@ -31,6 +31,7 @@ namespace DaggerfallConnect.Arena2
         const int paletteDataLength = 768;
         const int paletteColorCount = 256;
         const int paletteMultiplier = 4;
+        const float minFrameDelay = (740f / sampleRate);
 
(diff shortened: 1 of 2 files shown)


Interkarma:
Thanks for fixing! :) This is one of those items that I've just been pushing down the list for ages now. Great to know the problem has been resolved.

PR #1010, opened 2018-11-03
Improvements to PresetPicker

TheLacus:
Changed the preset picker window to follow a more consistent style and be more comfortable to use.

presetpi cker


Interkarma:
One change at  time, this just keep getting better. Thank you. :)

PR #1011, opened 2018-11-04
AI: Don't walk off ledges, LOS check from eye level

Allofich:
AI stops at ledges (unless they are very shallow) and does LOS check from around eye level.

### DIFF
--- a/Assets/Prefabs/Scene/DaggerfallLootContainer [Game Serializable].prefab
+++ b/Assets/Prefabs/Scene/DaggerfallLootContainer [Game Serializable].prefab
@@ -3,18 +3,18 @@
 --- !u!1 &101432
 GameObject:
   m_ObjectHideFlags: 0
-  m_PrefabParentObject: {fileID: 0}
+  m_CorrespondingSourceObject: {fileID: 0}
   m_PrefabInternal: {fileID: 100100000}
-  serializedVersion: 4
+  serializedVersion: 6
   m_Component:
-  - 4: {fileID: 421668}
-  - 114: {fileID: 11486862}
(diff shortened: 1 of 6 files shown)


Allofich:
Added various fixes:

- Fixed a few null exceptions in AIAttack if the target became null
- Made loot items ignore raycasts so that they don't block AI eye-sight or arrows
- Improved ray-casting from eye-level to go to target eye-level, to account for short targets
- Optimized target selection to not bother checking LOS with targets that will be rejected anyway
- Fixed up slight logic mistake in arrow shooting, which I think may alleviate issue I was having of archers slowly inching off of ledges

Allofich:
A few additional changes to make arrows, and enemy arrows specifically, shoot from where they should and not collide with geometry.

Interkarma:
This is generally working well and I've merged. Something I've noticed is that enemies will get stuck at the bottom of stairs and won't climb after player. For example, drag the skele in PH to bottom of stairs and he won't go back up now.

And dragging enemies lengthwise along walls get them stuck also. I think the LOS check clips through the geometry and stops them from sliding along the wall after player like they would before.

That aside, this is a nice change. Enemies feel smarter not blindly walking into abyss now. :)

Interkarma:
>Unfortunately my computer died the other day, so I'm out of commission for a while. I'll return to this when I can, or you can try the fixes I mentioned above if you want to make a release or something.

It's all good mate. No pressure on this stuff. :) Good luck getting your PC sorted out!

Allofich:
Thanks! Hopefully I'll have Daggerfall Unity and everything set up again before too long.

PR #1015, opened 2018-11-08
Fix spurious "you cannot carry anymore stuff."

petchema:
When, while you're selling items, you put items back into your open
wagon, you can sometimes get a "you can't carry that much weight" kind
of message and yet the items are correctly removed from the selling
selection.

The culpit is the seemingly harmless

    int canCarry = CanCarryAmount(item);
    if (usingWagon)
    {
        canCarry = Math.Max(canCarry, WagonCanHoldAmount(item));
    }

because CanCarryAmount() an WagonCanHoldAmount() not only return how
many instances of a given item still fit in character's inventory or
wagon, but they also display the above message if none fits; So they
cannot be freely composed, like above.

While I think I understand the intention behind that max() ("you can
pick the item if it fits in either your pack or your wagon"), I think it
currently has another bug, because the item is not put in your pack if
it doesn't fix in the wagon; So under the right conditions you can stack
pretty much everything in your wagon.

This patch remove that fallback logic entirely, which is the simplest
fix.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTradeWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTradeWindow.cs
@@ -596,11 +596,11 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             // Handle click based on action
             if (selectedActionMode == ActionModes.Select)
             {
-                int canCarry = CanCarryAmount(item);
+                int canCarry;
                 if (usingWagon)
-                {
-                    canCarry = Math.Max(canCarry, WagonCanHoldAmount(item));
-                }
+                    canCarry = WagonCanHoldAmount(item);
+                else 
+                    canCarry = CanCarryAmount(item);
                 if (windowMode == WindowModes.Buy)
                 {
                     TransferItem(item, remoteItems, basketItems, canCarry, equip: true);


petchema:
Was that logic there to match classic Daggerfall or something?

Interkarma:
Makes sense to me. @ajrb Are you happy with this change mate?

ajrb:
I'll need some time to try and recall how it was intended to work before it was changed to handle stacks.

ajrb:
So the original code before stacking changes was:

                if (CanCarry(item) || (usingWagon && WagonCanHold(item)))
                {
                    if (windowMode == WindowModes.Buy)
                    {
                        TransferItem(item, remoteItems, basketItems);
                        EquipItem(item);
                    } else {
                        TransferItem(item, remoteItems, localItems);
                    }
                }

Pretty simple, you can only select an item to buy if you can physically carry it. Seems to be allowing purchase of items into the wagon, but the purchase code then stuff everything into the player inventory. Definitely a mis-match here. Looks like I need to tidy this up which I will do soon. Actually can't recall if I was trying to match classic here or just implementing sensible behaviour. Give me a bit longer please.

ajrb:
Thanks for being so patient with me. I found time tonight to go through the history and check classic behaviour etc. I've made this change and a few other fixes and will create a PR. Glad you led me to the issues, cheers.

Interkarma:
Cheers Hazelnut! As your recent PR also includes this fix, I'll close this one. @petchema Thanks for bringing this one up and for fix. :)

PR #1016, opened 2018-11-09
TextBox overwrite

petchema:
Sometimes you want to open an text input box with a default value, but do not want it to get in the way of user input; For example, when renting a room in a tavern, you want to suggest a default of 1 day, but if the user types 3 you want the number of days to become 3, not 13.
Most of the time the current system works well enough, because the most common default value ("0") doesn't hurt as a prefix of integer input (except it's not very pretty). But in some cases like above, it really doesn't work so well.

This patches modifies TextBox so that if the user starts typing a new value, the previous value is first cleared; But if the user first starts with any kind of editing action (arrow keys, backspace,...) the value is modified instead, like before.
This is a cheap emulation of full fledged input text boxen, that have such behavior because the whole previous value is selected by default; But implementing text selection looks overkill for our purpose.

I also noticed DefaultText, that seems to have been envisioned for that purpose (given GetResultText logic), but
- DefaultText is used for a different purpose (put a "Enter your character name here" kind of filling text in empty input text boxes);
- GetResultText doesn't seem to be used;
- there's no option for textbox input validation callbacks to receive ResultText instead of Text as "input" parameter.
So maybe GetResultText could also be dropped along with this patch?


petchema:
Looking at all the uses of TextBox, I'm no longer sure it's a better behavior in all cases. Maybe it should only be enabled on a case-by-case basis

Interkarma:
I did something similar when first implementing edit boxes and ended up just reverting to classic behaviour. I'm happy to try different things though, if you'd like to keep working on it?

petchema:
Haha, I should have known you experimented with this *also* ;)
Well, obviously the current state of affairs bothers me somewhat.
Beside the case I already mentioned, I'm also thinking about an improvement with stack splitting. Stack splitting is not only triggered if you ctrl-click, but also automatically if your inventory is almost full and you try to pick some stack (say, gold); It indicates the max that you can carry as usual, but I think that in this case the default value should also be the max. But I ran into current TextBox behavior.
If I continue with this patch, I'll try to keep it simple; Not only because it fixes a small issue, but also because I'm still not confident with UI stuff.
Consistency is usually an important goal in interfaces, but I think it can make sense that some TextBox have overwrite and others don't, so my current idea is to keep most TextBox unchanged and only enable overwrite for some specific ones.
Was your previous attempt more advanced than that?

ajrb:
It is annoying but it matches classic. I don't like the idea that text input will have different behaviour in different places, consistency is king to my mind.

petchema:
It seems that the most annoying cases are integer input.
Thinking more out-of-the-box, instead of tweaking TextBox maybe that will require a dedicated control to address this need (slider ?)...

petchema:
Closing the PR as this is probably a dead end...

ajrb:
Having an option for a better numeric entry control would be nice, I think is used in half a dozen places at least though. I think DF purists will want to classic type entry boxes to be available still.

PR #1017, opened 2018-11-10
Use fog from Post-Processing Stack

TheLacus:
Fog component from Post Processing Stack is used, as a replacement of fog effect from old Standard Assets.
Other post processing effects are still disabled but can be enabled by mods, as it was before.


Assets/Scenes/DaggerfallUnityGame.unity
```diff
@@ -753,9 +753,27 @@ Prefab:
       propertyPath: m_Enabled
       value: 1
       objectReference: {fileID: 0}
-    m_RemovedComponents: []
+    m_RemovedComponents:
+    - {fileID: 114583161180529142, guid: 53a45b46ca97a974f8e677f68c2a85e6, type: 2}
```
TheLacus: I don't know what the purpose of this is, i can't find any information.

Assets/Scripts/Game/UnderwaterFog.cs
```diff
@@ -18,8 +14,6 @@ public class UnderwaterFog
         public UnderwaterFog()
         {
             mainCamera = GameManager.Instance.MainCamera;
-            globalFog = mainCamera.GetComponent<GlobalFog>();
-            globalFog.enabled = true;
```
TheLacus: I don't think fog is ever disabled. In that case i believe `RenderSettings.fog = true` should be the way to do it.

Assets/Scenes/DaggerfallUnityGame.unity
Interkarma: I'm not sure about this one, but it's probably OK.

Assets/Scripts/Game/UnderwaterFog.cs
Interkarma: That's fine. We'll need to evaluate any fog changes a bit differently now anyway.

PR #1019, opened 2018-11-11
"You cannot sleep now." if player is levitating

petchema:
Daggerfall classic displays that message if player is swimming, but also if he's levitating.


Interkarma:
Sensible! Thanks for your recent fixes. :)

PR #1020, opened 2018-11-11
stack potions

petchema:
Requested in forums
https://forums.dfworkshop.net/viewtopic.php?f=12&t=807

Potions require an extra check on typeDependantData, I hope I got it right

### DIFF
--- a/Assets/Scripts/Game/Items/DaggerfallUnityItem.cs
+++ b/Assets/Scripts/Game/Items/DaggerfallUnityItem.cs
@@ -534,7 +534,7 @@ namespace DaggerfallWorkshop.Game.Items
 
         /// <summary>
         /// Determines if item is stackable.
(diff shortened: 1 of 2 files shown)


Interkarma:
I'd prefer to hold off stacking using the classic recipe ID. I want potion recipes to be extensible via custom effects and mods, which means the classic recipe IDs will only take us so far. I've made a start on a new way of tracking recipes in PotionRecipe.cs that uses comparable lists and ID hashes rather than a single ID.

It's probably fine to use the classic recipe ID for now while potions aren't implemented and all we have is a "stub" item for player to carry. But this will need to be upgraded later and I'd rather not deal with pre-stacked potions in save games before I've had a chance to fully work through the systems. :)

petchema:
Fine with me, I'll close this PR so it doesn't stay in the way.

PR #1021, opened 2018-11-11
Implemented ImportedComponentAttribute

TheLacus:
One of the big limitations of the Mod Sytem is that MonoBehaviours provided my a mod can't be serialized to a prefab, because their type is not known by Unity when they are loaded from the assetbundle. The solution is to add components at runtime with `AddComponent()` but this is quite bothersome, and also impossible when a gameobject is instatiated by Asset-Injection without mod intervention.

As a workaround Lypyl introduced an interface to serialize custom scripts and provide a simple way to deserialize them, but it still requires the mod intervention so it doesn't fully satisfy all needs. For this reason i decided to give it a try with my own take on this issue.

## Component

Imported Monobehaviours needs to be assigned the `ImportedComponentAttribute`. Thats's all.

```
[ImportedComponent]
public class ExampleComponent : MonoBehaviour
{
    public string Title;
    public int Number;
}
```

## Serialization

The mod builder checks prefabs for components with the attribute and remove them on the bundled clone (the original is left unaltered), not before saving the required data to a json file named **prefabName.imported.json**. This is similar to what Lypyl was doing.
The final result looks like this:

examplecomponent

```
{
    "Components": [
        {
            "Title": "Elephant",
            "Number": 24,
            "$type": "ExampleComponent"
        }
    ],
    "Children": {
        "Sphere": {
            "Components": [
                {
                    "Title": "Dog",
                    "Number": 234,
                    "$type": "ExampleComponent"
                }
            ],
            "Children": null
        }
    }
}
```

There are two things to notice:
+ Components without the attribute are not saved here because they are already managed by Unity with the **.prefab**.
+ FullSerializer adds `$type` properties with the full type name.

## Deserialization

When a prefab is loaded with `Mod.GetAsset()`, the mod manager automatically look for the **imported.json** file. This is the tricky (and somewhat hacky) part:
Instead of giving the gameobject instance to FullSerializer (which wouldn't work), we traverse the `fsData` dictionary to read the `$type` properties and retrieve the actual type at runtime from the compiled mod assembly. At this point we can make a new instance with `AddComponent()` and let FullSerializer do its work on it.

## Conclusion

Manually using `AddComponent()` is always the fastest option but if this is not possible or is otherwise not desiderable, the `ImportedComponentAttribute` can automatize this process.

### DIFF
--- a/Assets/Game/Addons/ModSupport/Editor/CreateModEditorWindow.cs
+++ b/Assets/Game/Addons/ModSupport/Editor/CreateModEditorWindow.cs
@@ -1,4 +1,4 @@
-﻿// Project:         Daggerfall Tools For Unity
+// Project:         Daggerfall Tools For Unity
 // Copyright:       Copyright (C) 2009-2018 Daggerfall Workshop
 // Web Site:        http://www.dfworkshop.net
 // License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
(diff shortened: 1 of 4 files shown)


Interkarma:
Wow this has serious potential to automate some things! As always, your work on the mod system is incredible. :)

PR #1022, opened 2018-11-11
Handle asynchronous loading of audioClips

petchema:
Currently we can skip the first use of audioclips loaded from discrete files, because they're loaded asynchronously and may not be ready at the time of AudioSource.PlayOneShot().
See bug report https://forums.dfworkshop.net/viewtopic.php?f=24&t=1414
This patch wraps calls to PlayOneShot() with a coroutine so sound is played when ready, with a delay up to 150ms.

There's some code duplication here, could DaggerfallWorkshop.Game.Questing.Actions.PlaySound use DaggerfallWorkshop.DaggerfallAudioSource?

### DIFF
--- a/Assets/Scripts/Game/Questing/Actions/PlaySound.cs
+++ b/Assets/Scripts/Game/Questing/Actions/PlaySound.cs
@@ -27,14 +27,12 @@ namespace DaggerfallWorkshop.Game.Questing.Actions
         /// or play sound (soundname) x y
         /// the second number is not currently used for anything - purpose is unkown.
 
-        const float audioClipMaxDelay = 0.150f; //give up if sound takes longer to load
-
         public string   soundName;              //used to lookup sound index in sound table
         public int      soundIndex;
         public uint     interval;               //how often to play; measured in game minutes
         public int      unknown;                //according to Tipton's documentation, doesn't do anything
         public ulong    lastTimePlayed = 0;     //last time sound was played
-        public AudioClip clip;
+        public AudioClip clip;                  //preloading
 
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


Interkarma:
>There's some code duplication here, could DaggerfallWorkshop.Game.Questing.Actions.PlaySound use DaggerfallWorkshop.DaggerfallAudioSource?

The QuestMachine also carries a DaggerfallAudioSource, I don't mind the call being changed to use that component's PlayOneShot() instead - it should be equivalent.

petchema:
I didn't find any DaggerfallAudioSource in QuestMachine, only an AudioSource in PlaySound.
So instead I replaced that AudioSource with a DaggerfallAudioSource.
Now since DaggerfallAudioSource.PlayOneShot takes a soundIndex, and takes care of caching, the benefit of the clip field of PlaySource is just preloading, should I remove it?

Interkarma:
>I didn't find any DaggerfallAudioSource in QuestMachine

I was referring to game object in scene. If you check the QuestMachine Inspector in Unity Editor, you'll see it's carrying a DaggerfallAudioSource component. So what you've done here is perfect. :)

    DaggerfallAudioSource source = QuestMachine.Instance.GetComponent<DaggerfallAudioSource>();

petchema:
Beginner's Luck ;)
I modify code in MonoDevelop and only use the Unity Editor as a compiler...

Any advice for the clip field I mentioned previously?

Interkarma:
>Any advice for the clip field I mentioned previously?

Sorry I didn't answer. Yep, I think that should be fine now. :)

PR #1023, opened 2018-11-12
Stream ogg songs from disk

petchema:
MidnightPrince assembled a kickass soundscape for Daggerfall Unity
https://forums.dfworkshop.net/viewtopic.php?f=14&t=1347#p16078

All songs are replaced with OGG music, and there's a noticeable frames drop when songs (re)start.
This patch uses Unity audioClip streaming for digital songs.

### DIFF
--- a/Assets/Scripts/Utility/AssetInjection/SoundReplacement.cs
+++ b/Assets/Scripts/Utility/AssetInjection/SoundReplacement.cs
@@ -57,7 +57,7 @@ namespace DaggerfallWorkshop.Utility.AssetInjection
         /// <returns>True if song is found.</returns>
         public static bool TryImportSong(SongFiles song, out AudioClip audioClip)
         {
-            return TryImportAudioClip(song.ToString(), ".ogg", out audioClip);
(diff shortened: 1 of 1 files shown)


Midknightprince:
I likey !

Assets/Scripts/Utility/AssetInjection/SoundReplacement.cs
```diff
@@ -78,7 +78,7 @@ public static bool TryImportMidiSong(string filename, out byte[] songBytes)
         /// <summary>
         /// Import sound data from modding locations as an audio clip.
         /// </summary>
-        private static bool TryImportAudioClip(string name, string extension, out AudioClip audioClip)
+        private static bool TryImportAudioClip(string name, string extension, out AudioClip audioClip, bool streaming = false)
```
TheLacus: I'm ok with this. Just a small note: the convention is to place the out parameter in the last position. I would prefer if you move `streaming` before it and just pass true and false respectively.

petchema: Done. Somehow I was expecting to have more places to update...

PR #1024, opened 2018-11-12
Fix for level up bug in enhanced vitals indicators

ajrb:
@MeteoricDragon You should probably check this change and test that you're happy with the fix. Bug report https://forums.dfworkshop.net/viewtopic.php?f=24&t=1479


Interkarma:
Thank you mate.

PR #1034, opened 2018-11-18
Prevent talk menu from becoming unresponsive

numidium:
If the "okay" button is clicked while "things" is selected or if the "people" list is empty then the talk menu no longer responds to double clicks/okay button clicks.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTalkWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTalkWindow.cs
@@ -1258,8 +1258,11 @@ namespace DaggerfallWorkshop.Game.UserInterface
                 return;
             inListboxTopicContentUpdate = true;
 
-            if (index < 0 || index >= listboxTopic.Count) 
+            if (index < 0 || index >= listboxTopic.Count)
+            {
+                inListboxTopicContentUpdate = false;
                 return;
+            }
 
             listboxTopic.SelectedIndex = index;
             TalkManager.ListItem listItem = listCurrentTopics[index];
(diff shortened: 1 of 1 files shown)


numidium:
> could you explain why you use flag inListboxTopicContentUpdate to fix this bug
When debugging I noticed that inListboxTopicContentUpdate was getting stuck on "true" whenever I would click "okay" for either "things" or an empty list of "people." Therefore, SelectTopicFromTopicList would return from the if clause every time I would try to ask about a topic.
I tried replicating the "previous list" issue you described after making this change but I'm able to navigate the talk menu normally.

Nystul-the-Magician:
I try to reproduce your mentioned bug without the changes - and I can't - how do I repoduce it? clicking "okey" while "things" is selected or if "people" list is empty both does not make anything unrespondable :/

numidium:
So after you click "okay" on "things" or "people" and then attempt to ask about the location of a store/tavern/etc. the NPCs still give you directions? If so that's very odd. I'm using the Windows version if that makes any difference.

Anyone else experiencing this or is it just me? @Interkarma @ajrb @Allofich

Nystul-the-Magician:
see my upper comment - you were right and I was wrong - your fix is perfectly valid - sry for my lack of understanding the issue :D

numidium:
No worries. Thanks for verifying. :)

Nystul-the-Magician:
don't forget to put your name into the contributers list (if you want).

Nystul-the-Magician:
I will also add TheExceptionist to contributers since he submitted a fix for this but I did not understand what the issue was back then and he never answered my questions about it so I rolled back the changes after not seeing any differences and my fear of braking the previous list navigation stuff again :/

PR #1035, opened 2018-11-18
AI target movement prediction

Allofich:
With enhanced AI on, enemies now predict target movement.

This makes them a bit more active in combat, and makes guards a little harder to avoid.
This also allows enemies to lead their targets with arrows and spells.

Also fixed reversed check when AI was checking if other AI had back turned.

Prediction includes iterative prediction when the target is unseen. For example, if the player ducks moves behind a building, the AI assumes the player, while out of sight, is continuing to move in the last direction it saw the player move. Right now this feature doesn't work so great because of lack of obstacle avoidance, which is the next thing I'm working on.

Also, adding AI leading targets made the issue of the player riding on arrows more prominent, because arrows are more likely to hit the player at an angle now, rather than directly head-on. I added a hack to address this: arrows from AI will raycast in the direction of the target rather than directly forward (no change for the player's arrows). Seems to work well in testing. If we can use collision for the arrows this hack can be removed.

### DIFF
--- a/Assets/Scripts/Game/DaggerfallMissile.cs
+++ b/Assets/Scripts/Game/DaggerfallMissile.cs
@@ -81,6 +81,7 @@ namespace DaggerfallWorkshop.Game
         EntityEffectBundle payload;
         bool isArrow = false;
         GameObject goModel = null;
+        EnemySenses enemySenses;
 
         List<DaggerfallEntityBehaviour> targetEntities = new List<DaggerfallEntityBehaviour>();
         RaycastHit arrowHit;
@@ -190,6 +191,12 @@ namespace DaggerfallWorkshop.Game
                 {
                     UseSpellBillboardAnims(elementType);
                 }
+            }
+
+            // Setup senses
+            if (caster != GameManager.Instance.PlayerEntityBehaviour)
+            {
+                enemySenses = caster.GetComponent<EnemySenses>();
             }
 
             // Setup arrow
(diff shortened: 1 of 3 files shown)


Allofich:
Added AI obstacle avoidance. It could still use improvement but I've gotten it working decently enough that it's looks okay and is fun to play with.

Interkarma:
This is mind boggling. Very cool indeed!

Allofich:
Thanks, but it definitely needs more work. I'm finding they struggle to get through doors, for example. I'll play with it more later.

Interkarma:
That's fine, I know you'll come back and refine it later. :)

A couple of small items of feedback I've noticed playing with it for a couple of hours now.
* Enemies try to route around billboard flats (e.g. static NPCs in guilds) that don't block player movement. This actually looks good when it works smoothly and lends more intelligence to the mobile. But when NPC is close to another obstacle (such as static geometry or another NPC) they can struggle to path around.
* I'm not sure, but sometimes it feels like the new "back up" combat action will happen during navigation around obstacles. The mobile appears indecisive as they switch around between backing up and navigating.
* They have problems with narrow space and tight corners, but you're aware of that anyway.

But this really is an achievement. I thought about something like this way back at the start of project and quickly put it in the "too hard" basket. Considering how rudimentary Daggerfall's mobiles are the bar for AI was already quite low. Now you've taken this challenge head-on and, refinement aside, absolutely kicked it square through the goal posts. It's an impressive feat.

PR #1040, opened 2018-11-19
Bugfixes

ajrb:
* Resetting lightsource and spellbook.
* Correcting Mages rank 8 level message.
* Fix teleporting to restore building data.
* Add torch to items list of add console command.
* Fix horseriding in dungeons bug.
* Fix map reading in Alik'r Desert bug. (possibly fixes other problems with this region)
* Allow light sources to be used or equipped.

### DIFF
--- a/Assets/Scripts/Internal/PlayerGPS.cs
+++ b/Assets/Scripts/Internal/PlayerGPS.cs
@@ -518,7 +518,7 @@ namespace DaggerfallWorkshop
             currentClimateIndex = dfUnity.ContentReader.MapFileReader.GetClimateIndex(x, y);
             currentPoliticIndex = dfUnity.ContentReader.MapFileReader.GetPoliticIndex(x, y);
             climateSettings = MapsFile.GetWorldClimateSettings(currentClimateIndex);
-            if (currentPoliticIndex > 128)
+            if (currentPoliticIndex >= 128)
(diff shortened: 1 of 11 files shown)


ajrb:
The change to PlayerGPS line 512 needs reviewing by @Interkarma before this is merged. I am unsure of the consequences of this change.

Assets/Scripts/Internal/PlayerGPS.cs
```diff
@@ -518,7 +518,7 @@ private void UpdateWorldInfo(int x, int y)
             currentClimateIndex = dfUnity.ContentReader.MapFileReader.GetClimateIndex(x, y);
             currentPoliticIndex = dfUnity.ContentReader.MapFileReader.GetPoliticIndex(x, y);
             climateSettings = MapsFile.GetWorldClimateSettings(currentClimateIndex);
-            if (currentPoliticIndex > 128)
+            if (currentPoliticIndex >= 128)
```
Interkarma: This should be fine. In fact, I think you've fixed a much older bug into the bargain. The politic index is the value read from [POLITIC.PAK](https://en.uesp.net/wiki/Daggerfall:POLITIC.PAK) map. UESP states:

>For values from 128 through and including 233, subtract 128 from the value and to derive the Region Number for that pixel.

So >= 128 should be the correct solution here, with region numbers being zero-based in this case.

Thanks for the fix. :)

PR #1041, opened 2018-11-22
Can always put arrows into cart

petchema:
Since arrows weight is not accounted for, they can always be put back into cart
Bug report  https://forums.dfworkshop.net/viewtopic.php?f=24&t=1532


ajrb:
Looks good.

Interkarma:
Thank you!

PR #1043, opened 2018-11-23
Clean up EnemySenses and fix regressions in classic AI

Allofich:
Clean up for EnemySenses and fixes 2 regressions in classic AI from the recent changes.

Regressions:
1) Classic AI wasn't finding target's position anymore
2) Classic AI would stop at obstacles but not try to get around them.

1) is fixed, and for 2) obstacle avoidance code is used for classic as well now. Classic actually has some simple obstacle avoidance code of its own that is kind of similar, but it's not very effective. Just using the same obstacle avoidance code.

Regarding the cleanup, in addition to tidying things up, illusion spells (invisibility, chameleon, shade) now work for AI as well (meaning, AI will not see other AI with these effects).

Also, 2 questions for Interkarma.

1) Do you have a preference if private methods have the `private` keyword or omit it? I see it both ways in the files I've been looking at. I get the impression that it's been omitted more often, so I matched that for this cleanup, but if you would prefer it to be used, I can do it that way instead.

2) How do you feel about using auto-properties? If we use auto-properties, this PR will change from reducing line count by about 30 lines, to reducing it by about 100 lines.
See:
https://github.com/Interkarma/daggerfall-unity/compare/master...Allofich:cleanup2


Interkarma:
Thank you for the PR. :)

>Do you have a preference if private methods have the private keyword or omit it?

No firm preference here. I almost frequently used private keyword in older code, but have mostly stopped using it in last couple of years. Now it mainly comes in from VS-generated code or just to match the style of code around a change. No complaints if you wish to remove superfluous private keywords in cleanup.

>How do you feel about using auto-properties?

I don't mind auto properties, I'm trying to using them more frequently in recent code. Something about them feels a bit off for some reason, though. Probably just because I'm so adapted to the long form convention and change is hard for my stupid human brain. But I'm happy if you want to use auto-properties on clear candidates. :)

Allofich:
I'll look at doing an auto-properties PR on the files I've done cleanup on so far (don't want it to be just this one file). If I feel good about it, I'll submit it.

PR #1045, opened 2018-11-23
Ambient sound cleanup and made sounds 3D

Allofich:
Water sound now played 3D at a point on the water surface
Ambient sounds played 3D around the player

The water sound was already fake 3D but I had put a TODO for making it actual 3D.
The ambient sounds seem to have been 2D on purpose but I think 3D sounds better. It comes from a random direction around the player now. Let me know if you like it or you want them to stay 2D.

Cleanup:
Merge nested if statements
Remove redundant yield break
Remove unneeded variable for classic update loop
Use const keyword
Remove unused variable assignments
Add contributors
Add end-of-file newline


Interkarma:
Oh wow, the ambient sounds can be _genuinely disturbing_ now. I didn't realise how accustomed I'd become to tuning out ambients because they played right on top of me. Now that creaking door could _really_ be something opening a door just around the corner. And hearing the wind, jingle coins, etc. are more disturbing. Gives me a new reason to play with headphones on. Great work!

PR #1046, opened 2018-11-23
Swimming fixes

Allofich:
Mainly, this fixes that the player was able to cheat past the swimming encumbrance limit by looking up and moving forward while pressing jump or float up.

This is fixed by this PR by not allowing up/down movement from forward/reverse, which also matches classic.

GodMode now ignores the encumbrance limit since it can be a nuisance when testing.

Interkarma, if you'd prefer to be able to move up and down using forward/reverse, I can try to think up a solution that keeps that ability. In real life you kind of have to specifically aim for it to swim up or down (well I do anyway) so I don't mind float up/float down being a specific action here. Another idea is to use this float up/float down behavior for swimming, but allow forward/reverse for levitate. Right now this PR is following classic and doing float up/float down for both swimming and levitation.

Also swimming/levitate now uses the swimming/levitate speed for up/down, not fixed a fixed speed like classic. Classic's speed is still noted in a comment, but I can't see why, for example, a better swimming player character can swim faster horizontally but not vertically.

Another small difference from classic: In classic even when encumbered in water, you can stop the downward movement and stay at your current height by holding the float up key, although you can't go up. We can't know for sure, but I suspect this was probably not intended, considering the "You are carrying too much to stay afloat" message (although that message is from the demo and didn't make it to the release version) and that doing this trick allows you to "stay afloat" at your current height, so this PR doesn't allow it.

### DIFF
--- a/Assets/Scripts/Game/LevitateMotor.cs
+++ b/Assets/Scripts/Game/LevitateMotor.cs
@@ -77,11 +77,14 @@ namespace DaggerfallWorkshop.Game
 
             // Up/down
             Vector3 upDownVector = new Vector3 (0, 0, 0);
-            if (InputManager.Instance.HasAction(InputManager.Actions.Jump) || InputManager.Instance.HasAction(InputManager.Actions.FloatUp))
-                upDownVector = upDownVector + Vector3.up;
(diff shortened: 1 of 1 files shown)


Allofich:
No it's in for levitation as well right now, same as classic.
I'll see if I can make it apply just to swimming.

Allofich:
OK it's just for swimming now.

Interkarma:
Thank you for updating so levitation still uses directional flight. :)

PR #1048, opened 2018-11-23
AI fixes

Allofich:
Flyers and swimmers now use obstacle avoidance
AI now checks for a clear path, not just LOS, before shooting spells


Allofich:
Updated AI obstacle avoidance. They should be better at navigating tight spaces and following the player.

Interkarma:
So much great work in here! Are you happy for me to merge in next day or so, or would you like it left open a bit longer (as I can see you're still tweaking)?

Allofich:
Merging now is fine! 
Would be good if you can confirm that FPS is acceptable for you. Lots of ray casting. It is OK for me, but I will definitely try to optimize and simplify this later.

Allofich:
Meaning, I don’t see a noticeable drop, but I think it probably did drop a little.

Allofich:
Added fix for (3) in https://forums.dfworkshop.net/viewtopic.php?f=24&t=1519. (1) is intended and (2) is less of an issue now, and the subject of general improvements to the AI, so I think that bug report can be closed once this is merged.

PR #1049, opened 2018-11-24
Apply proficiency flags to custom classes

numidium:
Addresses issue described here:
https://forums.dfworkshop.net/viewtopic.php?f=24&t=1535


Interkarma:
Thank you for fixing! :)

PR #1050, opened 2018-11-25
Make triggered spells closer to classic

Allofich:
Partway fix for https://forums.dfworkshop.net/viewtopic.php?f=24&t=1545.

This fixes 2) and fixes 1) as far as only putting "casterOnly" spells on the player.

To complete the fix, we need to shoot non-casterOnly spells at the player from the triggering object, at the strength of the player's level. Interkarma I think you would know better how to do that. Would you like to finish out this fix by implementing that?

For 2), in classic a counter gets set to 1000 when triggering a spell trigger, then it is lowered by an amount on each trigger. After getting <= 0 the spell fires again. I think the amount may vary by FPS. I matched the amount so that the banner in Scourg Barrow triggers every 22 clicks, which is what I was getting in testing classic.

Attaching a video of Direnni Tower trap behavior from classic.
[Direnni Tower.zip](https://github.com/Interkarma/daggerfall-unity/files/2613259/Direnni.Tower.zip)

### DIFF
--- a/Assets/Scripts/Internal/DaggerfallAction.cs
+++ b/Assets/Scripts/Internal/DaggerfallAction.cs
@@ -55,6 +55,7 @@ namespace DaggerfallWorkshop
 
         AudioSource audioSource;
         ActionState currentState;
+        float cooldown;
 
         //lookup for action type12, temp. storing them here
         static Dictionary<int, string[]> actionTypeTwelveLookup = new Dictionary<int, string[]>()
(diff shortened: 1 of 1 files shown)


Interkarma:
This is awesome, thank you!

>Interkarma I think you would know better how to do that. Would you like to finish out this fix by implementing that?

Sure thing, and thank you so much for video to help me find a repro spot. I really appreciate it. :)

PR #1051, opened 2018-11-26
 fixed bug with fog settings being changed by UnderwaterFog script breaking mods

Nystul-the-Magician:
underwater fog script no longer changes fog settings when outside water or even worse when player is in an exterior location
underwater fog script now stores old fog values before entering water and restores them when leaving water
works with mods now


Interkarma:
Thank you Nystul! :)

PR #1052, opened 2018-11-26
Updates TweenToStart & TweenToEnd methods in DaggerfallAction

Lypyl:
Previously these methods were using the class member Duration instead of the function parameter duration for tweening.


Interkarma:
Good catch, cheers!

PR #1053, opened 2018-11-27
Arrive from expected direction after fast travelling

petchema:
Player can be relocated "to a random marker" in 6 cases:
- when entering/exiting buildings and dungeons;
- after fast travel;
- when going from/to ship;
- when teleported;
- when expelled by court;
- using "location" console command.

This patch implements start coordinates saving, and selection of facing
based on those coordinates as a new RelocationMethod; And then use this
method for fast travel case only.

There's some comment near entering/leaving buildings
(PlayerEnterExit.Respawner) that suggests that the closest start marker
should be chosen, but it's unclear if this new RelocationMethod could
help.


Interkarma:
Nice one! This works really well and still leaves the random behaviour intact for other systems that need it as a fallback.

I just spend around 10 minutes travelling around for the fun of it, and cracked a big grin when I could travel to Privateer's Hold from up north and landed on the northern start marker of PH looking south. It just felt _right_. Great work!

PR #1054, opened 2018-11-28
DaggerfallMissile cleanup, fix arrow spawn

Allofich:
Generally fixes the remaining issue for https://forums.dfworkshop.net/viewtopic.php?f=24&t=1235 of the arrows shot at extreme angles not looking like they come from the right spot. If you look completely up or down (probably very rare) it's still slightly off, but I think maybe that's an issue/limitation of the main camera or something as that is what I am using for positioning the arrow spawn now (rather than the player location).

Arrows do hit at extreme angles so that part of the bug report is not an issue, and I think it can be closed.

Also did a light cleanup of DaggerfallMissile.cs.

Add contributor
Add end of file newline
Fix variable shadowing
Remove whitespace
Remove unneeded boolean

Fix arrows to look correct when shot by player


Interkarma:
Thank you mate!

PR #1055, opened 2018-11-28
Cleanup PlayerActivate.cs

Allofich:
Simple cleanup, no bug fixes.

Add contributor
Match values in .cs file to those set in Unity editor
Merge if statements
Fix local variables hiding parameters
Simplify boolean returns
Clean up whitespace
Remove unneeded namespace references


Assets/Scripts/Game/PlayerActivate.cs
```diff
@@ -45,18 +43,19 @@ public class PlayerActivate : MonoBehaviour
...
-        // Maximum distance from which different object types can be activated, in classic distance units
-        public float DefaultActivationDistance = 128;
-        public float DoorActivationDistance = 128;
-        public float TreasureActivationDistance = 128;
-        public float PickpocketDistance = 128;
-        public float CorpseActivationDistance = 150;
-        //public float TouchSpellActivationDistance = 160;
-        public float StaticNPCActivationDistance = 256;
-        public float MobileNPCActivationDistance = 256;
+        // Maximum distance from which different object types can be activated, converted from classic units (divided by 40)
```
Interkarma: I wonder if we should use value * MeshReader.GlobalScale here? It's unlikely, but there's always a chance someone might change the global scale later. It would also lead code readers back to the GlobalScale const value in MeshReader with full comments on why we divide by 40.

Allofich: Updated to use ```value * MeshReader.GlobalScale```.

I also set the values as consts, no longer public variables editable from the editor.

The way the public variables editable from the editor were before (in master branch):
RayDistance: Unity distance value. Overridden in the file so had no effect.
Activation distance values: Classic distance value (then multiplied by MeshReader.GlobalScale)

My original idea with the cleanup here (in addition to no longer overriding RayDistance so it would actually have an effect) was to unify these all as unity distances, so the expected value entered from the Unity editor would be used.

To show that MeshReader.GlobalScale is being used, we could unify the values all as classic ones, with anyone changing them having the responsibility to understand they are not the usual Unity distance values, but classic distance values. Seems undesirable.

So, considering the above, and that I figure probably nobody is modifying these properties from the editor anyway, I decided to just make them private variables. Let me know if the current solution is OK with you.

PR #1056, opened 2018-11-29
introduced param raisePromoteTerrainEvent in function PromoteTerrainData

Nystul-the-Magician:
default always true
with raisePromoteTerrainEvent set to false function can be called without raising the PromoteTerrainEvent - needed for mod to prevent a bug/issue related to event raising and consumption in different mods
discussed here: https://forums.dfworkshop.net/viewtopic.php?f=20&t=1554&p=17765#p17765

### DIFF
--- a/Assets/Scripts/Terrain/DaggerfallTerrain.cs
+++ b/Assets/Scripts/Terrain/DaggerfallTerrain.cs
@@ -1,4 +1,4 @@
-﻿// Project:         Daggerfall Tools For Unity
+// Project:         Daggerfall Tools For Unity
 // Copyright:       Copyright (C) 2009-2018 Daggerfall Workshop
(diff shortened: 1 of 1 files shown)


Nystul-the-Magician:
this might be caused by a bug of the real grass mod as pointed out by lypyl here: https://forums.dfworkshop.net/viewtopic.php?f=20&p=17773#p17773

I reported the bug in the real grass mod thread (https://forums.dfworkshop.net/viewtopic.php?f=27&p=17777#p17777)

should we wait for a fix for the real grass mod before introducing this change? It might make the need for this change obsolete...

Nystul-the-Magician:
Lypyl pointed me to a workaround (try-catch block around the function call in my mod) that does not require changes to the core code - so I will discard this pull request

PR #1058, opened 2018-11-29
Don't allow aquatic enemies to spawn out of water

Allofich:
Fixes https://forums.dfworkshop.net/viewtopic.php?f=24&t=1515.

The problem was that this particular slaughterfish is a fixed enemy, and no check was done in DF Unity for fixed enemies spawning out of water, only for random enemies.
Classic doesn't allow any aquatic enemies to spawn out of water, regardless if they are fixed or random.


Interkarma:
Good old Daggerfall. Whenever the system looks sensible, there's always at least one outlier in the dataset to prove the exception. Thank you for fixing this one. :)

PR #1059, opened 2018-11-30
Prevent guards from spawning in dungeons

Allofich:
Fixes https://forums.dfworkshop.net/viewtopic.php?f=24&t=1549.

The actual behavior to match classic needs a little more work, but this fixes the reported bug at least.


Allofich:
Implemented NPC guards. Guards will appear in town at the same rate as in classic, and will turn into enemyMobiles when crimes happen around them in the same manner.

Allofich:
Update: Since unlike classic, mobileNPCs continue to spawn even when enemyMobile guards have spawned, new guard mobileNPCs will be updated to turn into enemyMobiles if there are any enemyMobile guards around.

Interkarma:
Thank you so much for working on this. Really looking forward to trying this one out.

PR #1060, opened 2018-11-30
Fix bugs with lightsources

ajrb:


Interkarma:
Thank you for the fixes! This light source feature is good fun btw. :)

PR #1061, opened 2018-12-01
Limits console output string from growing too large for TextMesh

Lypyl:
Bug report:

https://forums.dfworkshop.net/viewtopic.php?f=24&t=1544


Interkarma:
Thank you mate! :)

PR #1062, opened 2018-12-01
fix debuff blinking bug

petchema:
Fix attempt (947acf6) was necessary but not sufficient;
AlignIcons() must also make use of spells' poolIndex

Bug report
https://forums.dfworkshop.net/viewtopic.php?f=24&t=1517&p=17434


Interkarma:
Cheers for taking a look at this. :)

PR #1063, opened 2018-12-01
Close spell merchant interface upon spell double-click only

petchema:
Bug report https://forums.dfworkshop.net/viewtopic.php?f=24&t=1375

It seems classic behavior has not been consistent between releases or something; But it matches the version I used, what I've seen in several streams, seems useful and it looks more consistent with other shops.

The patch is short but I'm not confident in my UI understanding, so please review.


petchema:
I don't think I can change pull request title, because it's now inaccurate; This patch now emulates classic: automatically close spell merchant window on spell double-click.

Allofich:
> I don't think I can change pull request title?

In my PRs, I can see an "Edit" button I can click over to the right of the PR title that lets me change the title.

PR #1065, opened 2018-12-02
Vital indicators fixes

petchema:
This patch fixes:
- reflecting relative health decrease when max health increase after a level up;
- gauges sometimes staying stuck with some highlighted change until stat really changes.

It detect, handle and propagate resets (discontinuities).
Standard code flow can then handle the continuous, normal case.

I checked that HUDVitals receives reset events correctly, other
VitalChangeDetector listeners may benefit from handling reset events
too.


Interkarma:
Thank you! :)

PR #1067, opened 2018-12-03
Adds travel map settings to save state & sets travel map settings from load data

Lypyl:
https://forums.dfworkshop.net/viewtopic.php?f=24&p=17866#p17866


Allofich:
I think the bug reporter wanted classic behavior, which is to save and load those settings per saved game.

Lypyl:
> 
> 
> I think the bug reporter wanted classic behavior, which is to save and load those settings per saved game.

Yeah, I see that now.  It never occurred to me the travel map settings were getting saved, especially considering the filter buttons on the map aren't.

Lypyl:
This should now save the travel map settings like in Daggerfall.  One difference is that the state location filter buttons will be saved, when they aren't in Daggerfall.  This seems like a bug / oversight and should instead be consistent with the fast travel settings.  If the original behavior is preferred I can change it.

Interkarma:
This is great, thank you! :)

PR #1070, opened 2018-12-07
Adjust running skill tally rate

Allofich:
Discussed in https://forums.dfworkshop.net/viewtopic.php?f=4&t=1296.

In general, I think we should try to avoid changing game content for things like rebalancing purposes, but the running skill tallies so quickly that it seems like it may actually be a bug or oversight (maybe even a DOSBox timing emulation issue or something). The tally in classic happens at the same rate as the main game update loop. Midknightprince there says all of his characters are at 100 running. My character from a regular playthrough of classic also has 100 running (no other skills maxed out), even though I didn't try to get it.

A suggestion on the forum of 1/4 the current rate seemed fine when I gave it a quick test.

### DIFF
--- a/Assets/Prefabs/Player/PlayerAdvanced.prefab
+++ b/Assets/Prefabs/Player/PlayerAdvanced.prefab
@@ -596,15 +596,6 @@ MonoBehaviour:
   m_Script: {fileID: 11500000, guid: 581803dfddb0b7741868a443507efd24, type: 3}
   m_Name: 
   m_EditorClassIdentifier: 
-  RayDistance: 2.25
-  ActivateDistance: 2.25
-  DefaultActivationDistance: 128
-  DoorActivationDistance: 128
-  TreasureActivationDistance: 128
-  PickpocketDistance: 128
-  CorpseActivationDistance: 150
-  StaticNPCActivationDistance: 256
-  MobileNPCActivationDistance: 256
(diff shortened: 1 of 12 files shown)


ajrb:
So you are saying the running tally in classic was dependent on the speed of the PC? I think that makes it prime candidate for adjustment for DFU in that case.

Allofich:
Tried 2 saves from https://en.uesp.net/dagger/dagsave.shtml.

"Morthag Iron Heart" save, dated May 1998, is lvl 13 and has running (minor skill) of 87%, highest skill, while primary skills are in the 60s and 50s.

"Lord Soth" (werewolf) save, dated Dec. 1997, is lvl 4 and has running (minor skill) of 25% (after adjusting for werewolf +30 bonus). Primary skills are in 40s and 30s.

Not really conclusive, but it does look like in both these cases running was advancing faster than other skills. My character with 100% running (from DOSBox) is only lvl. 11, so it's possible there is a timing issue where it's advancing even faster in DOSBox than it did on 90s-era computers. Another possibility, though, and maybe most likely, is that it's just because I was using running more because I switched to the VIEWPLYR.DAT control scheme with WSAD movement and Space Bar for running (recommended by Bethesda in the readme on the CD), rather than the default of arrow keys and P for running, so running was easier to use.

PR #1071, opened 2018-12-08
fix afternoon billboards shadows

petchema:
Fixes afternoon billboards shadows, could it be that simple?
Bug report https://forums.dfworkshop.net/viewtopic.php?f=24&t=653


Interkarma:
Nice one! Such a simple fix too. :)

PR #1072, opened 2018-12-08
Loose objects cleanup fix

petchema:
Not sure it's the final word on that problem, but there's a bug in
CollectLooseObjects() because it uses RemoveAt(i) in a for loop, so if
two consecutive items need to be removed the second will be missed
because RemoveAt(i) will have shifted it down

Bug report https://forums.dfworkshop.net/viewtopic.php?f=28&p=18052#p18051


ajrb:
Nope, the predecrement makes no difference - have edited the code suggestion... that's what happens when you write code without thinking or trying it. Your version works fine but it might be more obvious by subtracting 1 from count and then using the increment argument. I leave that to your discression. Sorry for misleading.

Interkarma:
Cheers, another silly mistake on my part. Appreciate the fix. :)

PR #1073, opened 2018-12-08
 Do not use atlas for MobilePersonBillboard imported textures 

TheLacus:


TheLacus:
I tested it with King of Worms's textures and met no issues.

If performance becomes a problem we could use TextureArrays, but we can't make one for the whole archive because records have different sizes. We would need five arrays, one for each record.

Alternatively we can request the same size for all textures and use uvs appropriately, at least for npcs. Enemies have more important variations so this is probably not viable.

petchema:
No issue to report with Villagers Overhaul either
So cool, thanks!

Interkarma:
That was quick! Thank you for that. :)

HybOj:
Gods be praised Lacus!! Thank You!
Last thing we need is to make the trees from a DFMOD package to appear ingame.
Sofar, they appear only when they are injected as a loose files.
When packed in DFMOD package, only the low res vanila sprites are shown ingame.
(KoW)

PR #1074, opened 2018-12-08
Calculate book values like classic

ajrb:
Also added book replacement check to the item MCP for author. I considered moving this check inside BookFile class so it didn't need to be done in every place the class is used, but there may be a good reason this wasn't done so I chose to leave it alone.


Interkarma:
Cheers mate!

PR #1077, opened 2018-12-09
Don't show lock messages for AI

Allofich:
Messages would show when AI tried to open a locked door.


Allofich:
Not necessary. The issue only appeared for https://github.com/Interkarma/daggerfall-unity/pull/1048 but I can avoid the issue there.

PR #1078, opened 2018-12-09
Fix importing classic save startingLevelUpSum

Allofich:
Classic save import has been incorrectly importing the startingLevelUpSum since
https://github.com/Interkarma/daggerfall-unity/commit/222c142625b37cb42d96d2c485461fa39da469ca,
messing up the leveling of imported characters. This should fix it.


Interkarma:
Good catch, cheers!

PR #1080, opened 2018-12-10
factorize capacity computations

petchema:
Factorize capacity computations (player carry capacity, wagon hold
capacity, gold dropping into wagon), using floats to avoid rounding down
bugs.
Code that makes horse, wagon and arrows weightless was also factorized.

Gold dropping message was left as-is to preserve Daggerfall classic
behavior; However, when dropping gold into wagon, a post capacity check
was added: if the wagon cannot accomodate the amount, only the gold that
fits is dropped and a message is displayed.

Bug report https://forums.dfworkshop.net/viewtopic.php?f=24&t=1587

I found another bug while factorizing: gold weight was counted twice in
wagon, because it was handled with a special case when the common case
was sufficient.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
@@ -135,6 +135,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
         readonly string damRep = TextManager.Instance.GetText(textDatabase, "damRep");
         readonly string arSrc = TextManager.Instance.GetText(textDatabase, "arSrc");
         readonly string arRep = TextManager.Instance.GetText(textDatabase, "arRep");
+        readonly string wagonFullGold = TextManager.Instance.GetText(textDatabase, "wagonFullGold");
 
         const string baseTextureName = "INVE00I0.IMG";
         const string goldTextureName = "INVE01I0.IMG";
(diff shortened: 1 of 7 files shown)


Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
```diff
@@ -1156,9 +1157,19 @@ private void DropGoldPopup_OnGotUserInput(DaggerfallInputMessageBox sender, stri
...
+            {
+                // Check wagon weight limit
+                int wagonCanHold = ComputeCanHoldAmount(playerGold, DaggerfallBankManager.goldUnitWeightInKg, ItemHelper.wagonKgLimit - remoteItems.GetWeight());
+                if (goldToDrop > wagonCanHold)
+                {
+                    goldToDrop = wagonCanHold;
+                    DaggerfallMessageBox messageBox = new DaggerfallMessageBox(uiManager, this);
+                    messageBox.SetText(String.Format(wagonFullGold, wagonCanHold));
+                    messageBox.ClickAnywhereToClose = true;
+                    messageBox.Show();
```
ajrb: Maybe you could use this to shorten this boilerplate code for all the MsgBoxes you added:

DaggerfallUI.MessageBox(text);

ajrb:
Excellent work Pango, looks good.

ajrb:
Perfect, very happy how you sorted this out.

Interkarma:
Very clever! Thank you. :)

PR #1084, opened 2018-12-12
Fix trying to find undiscovered locations from notebook

ajrb:
Fixes issue raised in https://forums.dfworkshop.net/viewtopic.php?f=24&t=1578


petchema:
It seems it is now too strict, you should be able to fast travel to towns you haven't visited yet for instance.
Maybe DaggerfallQuestJournalWindow.CanFindPlace and DaggerfallTravelMapWindow.checkLocationDiscovered should match?

PR #1086, opened 2018-12-13
Merging 0.6 to master

Interkarma:
I'm happy with where 0.5 stable is at right now. Time to start breaking stuff again. :)


TheLacus:
Looks like some lines of `UnderwaterFog` i removed in 0.6 are now [present again](https://github.com/Interkarma/daggerfall-unity/commit/b4603c3b808c2a54552e40c79371b7e8711788d2#diff-d7f100f96bd5dddfa64aecf561cf7328).

Interkarma:
I just merged a quick change to make game compile again. Could I get your help with a quick review of fog code in master now that everything back together? :)

TheLacus:
I think it's all ok in code now and i didn't find any issue in game  ;)

Interkarma:
>I think it's all ok in code now and i didn't find any issue in game ;)

Cheers! I did notice a couple of quirks this morning while playing around. First is a case where fog doesn't seem to activate:
1. Restart game > load outdoor save
2. Use "set_weather 3" and note fog not applied to scene

But this seems to work better when using the following:
1. Run game > load dungeon save near water
2. Jump into water and note underwater fog activates
3. Load outdoor save (without restarting game)
4. Use "set_weather 3" and note fog is applied to scene

Something seems to activate when jumping into water first that isn't on a fresh start. Above is the same with 0.5 stable as well for me, so I'm probably just noticing an older bug for the first time. :)

I also noticed new fog does not apply to static billboard batches (trees, lamps, etc.) in new builds. Seems to work normally in editor though. Maybe a dummy material is needed to force something into builds? Screenshot below is taken from a local build.

fog-billboards-builds

TheLacus:
I can replicate both issues, but i think i also see it in 0.5:

fog

TheLacus:
This is also 0.5, same save but in editor.

fog1

Interkarma:
Thanks for the detailed response Nystul! I kind of prefer the thicker fog from exponential as a weather event, it feels closer to the foggy weather I remember in classic. But naturally don't want to interfere with the smaller amount of atmospheric fogging in normal scenes. It just feels like foggy weather should be a bit soupier.

>What it does not do is that it restores settings when a different savegame is loaded... we should take this into account (I think we should restore settings when a savegame is loaded as well, what do you think?

I think consistency is probably best to start with. We can worry about tuning up appearance of foggy days later. 👍

Nystul-the-Magician:
i noticed that without global fog, even when fog is fixed the skybox does not get affected by fog :/
is this a limitation of standard fog?

TheLacus:
Look in the Resources folder, there is a file named `Post-Processing.asset`, with the option `Exclude Skybox`. I left it as default because it seemed to me that global fog in 0.5 also didn't affect skybox, but apparently i tried to recreate the bugged behaviour rather than the intended one. 

You can uncheck this option to enable it ;)

Nystul-the-Magician:
is this an asset in use by dfunity core? I was not aware of this

Nystul-the-Magician:
i unchecked the excludeSkybox setting in my latest pull request related to fixing fog

PR #1087, opened 2018-12-14
Fix premature closing of message boxes in live builds

ajrb:
Would like this backported to 0.5 otherwise players using the 0.5 stable build will not see any messages for guild promotions at all.


Interkarma:
No problems mate, and thank you for the fix. I've spun up a 0.5-stable branch from its last commit and cherry picked your changes into that branch after merging them to master.

I'll setup new build targets for 0.5-stable branch soon and update builds with these backported changes included.

PR #1089, opened 2018-12-15
Page by page scrolling using scrollbar's trough

petchema:
It could be a matter of taste, but it makes troughs provide an additional and
useful feature compared to scroll buttons.


Interkarma:
I don't mind this change, cheers. :)

PR #1090, opened 2018-12-15
Delay inventory scroll down

petchema:
Bottom items are shifted up when an item is removed from a list; But if the end of list is displayed, the whole list can also at the same time be scrolled down, which can become confusing, specially with enhanced lists.
This patch disables scrolling while removing items from enhanced lists; Scrolling is updated when the mouse pointer is moved away from the list.
They're no doubt ergonomic tradeoffs with that kind of change, so this is just a proposal


ajrb:
Sorry @petchema  & @Interkarma, I really don't like this change. It creates more confusion IMO and is worse than the original behaviour. I was just removing a bunch of recipes from inventory to cart and thought I'd done them all then a load more popped into existance (i.e. the list scrolled up) when I moved my mouse off the scroller.

It may well be that this is because I am used to the old behaviour though. Thoughts?

Interkarma:
I wanted to pay with it for a while and see what feedback was like. I'm finding it a little jarring also, but like you thought it might be because I was used to the previous way. It's a simple one to reverse in any case.

Will make a call on this one after a little while longer.

petchema:
I just noticed it's the behavior of classic too, maybe I should have started by checking that

PR #1091, opened 2018-12-15
Only flash crosshair twice after selecting a location

numidium:
In classic the yes/no popup which shows on the travel map does not wait for the full four flashes of the crosshair to appear. This change makes the travel UI faster and closer to classic.


Interkarma:
Cheers! :)

PR #1093, opened 2018-12-15
Updates RDBLayout to recognize models with the "NEW" tag as action doors

Lypyl:
Models 55007, 55024, 55018 have this tag and always seem to be action doors. 

There are 2 doors (55007) in Orsinium that also have rotation actions which causes them to operate strangely - original Daggerfall has the same problem.

Relevant bug report:

https://forums.dfworkshop.net/viewtopic.php?f=24&t=1597


Interkarma:
Thanks Lypyl! Based on my code comments, I had a suspicion that NEW could be universally treated as an action door. Thank you for confirming and wrapping that up for me.

PR #1095, opened 2018-12-15
TextureReplacement: Improvements for FPSWeapon

TheLacus:
* Mirror texture rect when required.
* Use texture key instead of string for dictionary.


Interkarma:
Wonderful, thank you! :)

PR #1096, opened 2018-12-16
fixed fog settings (WeatherManager class related)

Nystul-the-Magician:
- WeatherManager class uses different fog settings for different weather types now (fog mode, density, etc. stored per weather type now)
- "set_weather 3" now works again

### DIFF
--- a/Assets/Resources/Post-Processing.asset
+++ b/Assets/Resources/Post-Processing.asset
@@ -27,7 +27,7 @@ MonoBehaviour:
   fog:
     m_Enabled: 1
     m_Settings:
-      excludeSkybox: 1
+      excludeSkybox: 0
   antialiasing:
     m_Enabled: 0
     m_Settings:
--- a/Assets/Scenes/DaggerfallUnityGame.unity
+++ b/Assets/Scenes/DaggerfallUnityGame.unity
@@ -1711,11 +1711,6 @@ MonoBehaviour:
   StormSunlightScale: 0.35
   SnowSunlightScale: 0.5
   WinterSunlightScale: 0.65
-  SunnyFogDensity: 0.0005
-  OvercastFogDensity: 0.0005
-  RainyFogDensity: 0.0001
-  SnowyFogDensity: 0.001
-  HeavyFogDensity: 0.05
 --- !u!1 &1437448953
 GameObject:
   m_ObjectHideFlags: 0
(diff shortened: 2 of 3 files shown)


Nystul-the-Magician:
there are some smaller issues related to post processing fog together with vanilla sky (I am looking into a solution right now):
the problem is:
if fog profile setting excludeSkybox is true - skybox looks correct for sunny weather, but incorrect for foggy weather (skybox still visible)
if excludeSkybox is false - skybox looks ok for foggy weather (not visible) but is faded a bit (which can be noticed easily) since fog affects it - setting fog density (in lighting tab) to a smaller value helps with the skybox but than trees in the distance are not faded anymore correctly

I am working on a solution with several post processing profiles for different weathers

Nystul-the-Magician:
implemented different post processing profiles to allow correct fog settings depending on weather type
this should solve the issue mentioned above

TheLacus:
Nystul, can't you just toggle that specific value on the profile when needed? I don't understand why you make a copy of the whole profile. I think it should be enough to reassign the fog setting struct :)

Nystul-the-Magician:
I tried to do this, but you can not set a single settings (I have not found a way) - please let me know if you know of a way

TheLacus:
`FogModel.Settings` is a struct. You need to make a local copy on the stack, edit it and then reassign it back to field in class instance.

```
var fogSettings = profile.fog.settings;
fogSettings.excludeSkybox = true;
profile.fog.settings = fogSettings;
```

Nystul-the-Magician:
thanks will update

Nystul-the-Magician:
changed the code according to your suggestion, thanks

I am still not happy with exponential fog on dfunity standard view distance (without mods) - linear just looks better, thinking about rewriting WeatherManager a bit also storing fog mode for every weather...

TheLacus:
Thank you. This allows to edit options for other effects at runtime, without the need to be aware of a profile switch :)

Nystul-the-Magician:
my lack of c# knowledge striked once again here ;)

Nystul-the-Magician:
ok I implemented the collection of fog settings per weather type now - this allows linear fog for sunny weather and exponential fog for heavy fog weather and any possible combination for any weather, mods can overwrite fog settings for every weather ;)

Interkarma:
Thanks for all your work on this Nystul! I'm hoping all these kinds of changes will settle down now through until 1.0 so we can just focus on refinement. You're probably getting sick of fixing fog at this point. :)

Interkarma:
Just something to note is that classic sky background no longer appears in game.

image

The editor also no longer shows grids, etc. in scene view.

image

Are you able to take a look at these two items? Thank you mate. :)

Nystul-the-Magician:
I did not experience these issues - very strange - will take a look

PR #1097, opened 2018-12-17
Identify classic vampire data, vampire quests

Allofich:
This provides functionality/information for vampire mechanics from classic. Interkarma, please feel free to refactor anything you like, or if this can't merge cleanly with local WIP code you have, just use it as a reference or something.

What's in here:

- Identify and import 3 vampire-related data pieces from classic saves.
1) Last time that need to kill was satiated
2) Vampire clan
3) Whether or not initial vampire quest has been initiated

This data is not serialized in PlayerEntity yet. 1) and 2) are commented out in PlayerEntity, but 3) is used for the quest logic.

- Uncommented Vampire race and temporarily using Breton graphics, to allow for opening classic vampire saves to aid in testing against classic.

(Note that if you load a classic vampire, they will gain several levels because their skill bonuses are imported as if they are the real skill levels. Classic temporarily undoes vampire/werecreature skill bonuses when calculating skill advances/level ups, then reapplies them. This is something we'll need to handle when importing classic vampires)

- Run vampire quests the same as classic + bug fixes.
Non-cure quests can appear every 38 days, as I think they were supposed to, instead of every 266 days which I think is how often they can happen in classic. This has been noted in a comment in PlayerEntity for a while now. Cure quests can appear every 84 days, same as classic. Quests can occur at the same chances as in classic. The initial quest only happens once, unlike classic. (If it's possible to re-become a vampire after being cured, probably the initial quest should happen again, haven't done that yet)

The initial quest and cure quest are always the same, but there are also other quests. Unfortunately, I've never touched the quest code in DF Unity before, so I'm not sure how to handle selecting a quest through it. I put together a simple function that calls one of the quests, but it's hard-coded. I'm sure we'll want to move that. Werecreatures only have the cure quest, and I put that in as well, commented out.

BTW, I think the current quest selection incorrectly treats the 4th and 5th characters of a quest name as the required faction reputation to get that quest. Actually the 4th char is either a level requirement or a rank requirement. The char's ASCII value - 48 is compared to the level/rank. (As I did in the function I added for the vampire quests).

Rank requirement:
Fighters Guild
Dark Brotherhood
Temple
Thieves Guild

Level requirement:
Knightly Order
Mages Guild
Royalty
Coven
Merchant
Vampire

The only purpose I know of the 5th char is that if it is 'X' or 'Y' it will not be selected when ChildGard is on. Ones with 'X' all either have Daedra Seducers or Harpies in them (maybe harpies had or were planned for a more explicit graphic before?). No classic quests use 'Y'.

Examples:
N0B40Y07 = Mages Guild, available to level 4 and higher.
O0B2XY04 = Thieves Guild quest that can have a daedra seducer. Censored in ChildGard mode.

### DIFF
--- a/Assets/Scripts/Game/Entities/PlayerEntity.cs
+++ b/Assets/Scripts/Game/Entities/PlayerEntity.cs
@@ -40,6 +40,10 @@ namespace DaggerfallWorkshop.Game.Entity
         bool preventNormalizingReputations = false;
         bool isResting = false;
 
+        bool hasStartedInitialVampireQuest = false;
+        //byte vampireClan = 0;
+        //uint lastTimeVampireNeedToKillSatiated = 0;
+
         const int testPlayerLevel = 1;
         const string testPlayerName = "Nameless";
 
@@ -364,12 +368,11 @@ namespace DaggerfallWorkshop.Game.Entity
                 if (((i + lastGameMinutes) % 54720) == 0) // 38 days
                 {
                     RegionPowerAndConditionsUpdate(true);
-                    // GetVampireOrWerecreatureQuest(); Non-cure quest
+                    GetVampireOrWereCreatureQuest(false);
                 }
 
-                // TODO: Get vampire/werecreature quest
-                //if (((i + lastGameMinutes) % 120960) == 0) // 84 days
-                //    GetVampireOrWerecreatureQuest(); cure quest
+                if (((i + lastGameMinutes) % 120960) == 0) // 84 days
+                    GetVampireOrWereCreatureQuest(true);
             }
 
             // TODO: Right now enemy spawns are only prevented when time has been raised for
@@ -435,6 +438,62 @@ namespace DaggerfallWorkshop.Game.Entity
             {
                 haveShownSurrenderToGuardsDialogue = false;
             }
+        }
+
+        // Recreation of vampire/werecreature quest starter from classic
+        // Notes:
+        // 1) In classic the initial quest can happen multiple times but this is probably a mistake.
+        // 2) Additional quests will come regardless of whether the initial quest was completed or failed, as the bool is set
+        //    when the initial quest begins, not on successful completion.
+        void GetVampireOrWereCreatureQuest(bool isCureQuest)
+        {
+            if (Race == Races.Vampire)
+            {
+                if (isCureQuest)
+                {
+                    if (DFRandom.random_range_inclusive(10, 100) < 30)
+                        Questing.QuestMachine.Instance.InstantiateQuest("$CUREVAM");
+                }
+                else if (hasStartedInitialVampireQuest)
+                {
+                    // Do a quest starting with "P0B" for player's level
+                    // TODO: Merge with rest of quest-loading code
+                    string[] vampireQuests = { "P0B00L01", "P0B00L03", "P0B00L04", "P0B00L06", "P0B01L02", "P0B10L07", "P0B10L08", "P0B10L10", "P0B20L09" };
+                    List<string> vQList = new List<string>(vampireQuests);
+                    if (DFRandom.random_range_inclusive(1, 100) < 50)
+                    {
+                        bool found = false;
+                        string choice = string.Empty;
+                        while (!found && vQList.Count > 0)
+                        {
+                            int index = UnityEngine.Random.Range(0, vQList.Count);
+                            choice = vQList[index];
+                            // 4th letter of name is the minimum level the player must be to get the quest
+                            if (choice[3] - 48 <= level)
+                            {
+                                found = true;
+                            }
+                            else
+                                vQList.RemoveAt(index);
+                        }
+
+                        if (found)
+                            Questing.QuestMachine.Instance.InstantiateQuest(choice);
+                    }
+                }
+                else if (DFRandom.random_range_inclusive(1, 100) < 50)
+                {
+                    Questing.QuestMachine.Instance.InstantiateQuest("P0A01L00");
+                    hasStartedInitialVampireQuest = true;
+                }
+            }
+            /*else
+            {
+                if (playerIsWereCreature && isCureQuest && DFRandom.random_range_inclusive(1, 100) < 30)
+                {
+                    Questing.QuestMachine.Instance.InstantiateQuest("$CUREWER");
+                }
+            }*/
         }
 
         public bool IntermittentEnemySpawn(uint Minutes)
@@ -721,6 +780,9 @@ namespace DaggerfallWorkshop.Game.Entity
             this.timeForDarkBrotherhoodLetter = character.timeForDarkBrotherhoodLetter;
             this.darkBrotherhoodRequirementTally = character.darkBrotherhoodRequirementTally;
             this.thievesGuildRequirementTally = character.thievesGuildRequirementTally;
+            this.hasStartedInitialVampireQuest = character.hasStartedInitialVampireQuest != 0;
+            //this.vampireClan = character.vampireClan;
+            //this.lastTimeVampireNeedToKillSatiated = character.lastTimeVampireNeedToKillSatiated;
 
             BackStory = character.backStory;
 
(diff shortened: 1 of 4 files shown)


ajrb:
I used the info from the UESP which you can find here: https://en.uesp.net/w/index.php?title=Daggerfall:Quest_hacking_guide&oldid=1256929

This has since been changed by a user called KShrimp and the name doesn't link to any info about who this is. They seem to have been making a lot of edits this year which is odd to me that they have not mentioned this on DFU forums - seems unlikely someone with such indepth DF knowledge would not be aware of the remake project.

Did you get your info from UESP or did you analyse classic code? I must admit I really hope that it was correct before the recent changes and that I don't need to re-work the quest dispensing code. I'd like to check the classic code to figure out whether the new info on UESP is correct or whether the old info was right. It will be a pain having to re-work it all and update all the tables I generated and then get Jay to update all his quest packs etc so definitely want to be sure first.

As for the vampire quests and guild stuff, I plan to work on that once Interkarma is ready. It will get rolled into the quest management code, so very happy for you to hardcode stuff temporarily, I can move it when the time comes.

JayH2971:
I just made a new character in classic, hacked in 95 reputation with the Mages Guild and all skills at 4. I'm an Apprentice and I'm receiving quests such as the Baltam Greyman and the Guard the Guild, which are both for 10+ reputation. I can provide a save later today.

ajrb:
Thanks Jay, I'd like to think that this person hasn't updated UESP with a load of complete crap info, but I suspect that may be the case - especially since no one mentioned this on DFU forums which I would have expected given how the word of mouth about it has been recently.

UESP, while is often slighly wrong or incomplete, is invaluable information so this is bad news if true. I had not noticed these edits until this evening when Allofich said he thought the intepretation of the quest filenames I implemented was incorrect.

Interkarma:
Thanks Allofich! It's good to have that data from classic saves particularly.

An approach I'm taking while building out vampirism is to see just how much of the heavy lifting I can accomplish in the effect system vs. hardcoding everything. Although some of the vampire stuff will still end up hard-coded (it's part of core after all), I'd like to offer some hooks for custom character changes through the effect system. This means I have a slightly different approach to assigning portrait and race to what I envisioned earlier on.

I'll leave this PR open while I'm working through the concepts. I probably won't hit all the goals I'm thinking of in first iteration anyway. It's most important just to get the vamp/were gameplay playable.

JayH2971:
Whichever you prefer is fine. Seeing all this deep science about Daggerfall is quite fun :D

ajrb:
Fair enough Allofich, and thanks for confirming. So looks like JShrimp has some pretty good knowledge - it would be nice to have them on the forums. They have edited quite a lot of the DF pages over past 6 months, wonder if there's anything else we've implemented incorrectly by following UESP that they have now corrected?

I mean I even implemented the wildcarding for the X that indicates childguard since that's what the old page said. Was pretty tricky, and I feel like leaving that functionality available to new DFU quests even if we change the classic ones to match. So, I'll look at layering both ways via quest tables - see if that's viable.

Allofich:
Sure. BTW I updated the top comment. I had mistakenly thought read the Dark Brotherhood quest filenames as having "Y" in the 5th character byte, but they don't. No classic quests do. Also I found that those with "X" are not only quests with Daedra Seducers, but also ones with harpies. Maybe harpies had or were planned for a different graphic that had nudity or something.

Also, although classic appears to not do anything with it, I think I know what "1" (as opposed to "0")in the 5th character means: All of the ones I have looked at (haven't looked at all of them) are appropriate as one-time quests. For example, the quest-giver refers to their lover, or the quest is referred to as a "one in a lifetime opportunity", the quest with Baltham Greyman, etc. Look at them and you will see they all seem to make sense as one-shots.

ajrb:
So the UESP page says this:

> The minimum reputation (in ASCII representation) with the quest giver required to receive the quest, or minimum player level for certain guilds. Values can range from '0' (0x30) to '9' (0x39) although it is recommended that you not exceed '7' (0x37) for the majority of quests. A value of '5' (0x35) would indicate the quest can only be received by characters with at least 50 reputation.

But you're not sure if it's player level or guild rank when is a guild quest yet?

The most compelling part of what you're discovered is that 5th character which indicates childguard or one time quests. That makes complete sense and makes me feel I should change it - still need to look at it all, for now staying focussed on potions!

Allofich:
> But you're not sure if it's player level or guild rank when is a guild quest yet?

It depends on the quest giver whether it is player level or guild rank. The basic pattern is that it is guild rank for guilds, and level when the quest giver is of a faction that the player can't join and rise ranks in. Which makes sense.

The 2 oddballs are the Mages Guild and knightly orders, which use player level for some reason (leading to the minor issue that you can get the highest level quests right away if you join these guilds later in the game. Not necessarily a problem I suppose.)

I'm guessing the mages guild was probably one of the first guilds implemented. It existed in Arena unlike most of the other "guilds" (Arena only has mages guild and temples). Building type 11 is the only dedicated guild-type building and is only used for the mages guild, and they are the only guild that gets its own music in the final game. Arena didn't have guild ranks, as the guilds can't be joined. I don't know if Arena quests are level-based, but if they are, they may have started out doing it the same way in Daggerfall, and the Mages Guild got level-based quests. Then, intentionally or not, they may have kept it that way. Maybe something similar for knightly orders. Or my speculation could be wrong, who knows.

BTW, I said above that the reputation method would space out the quests more, but I was only thinking of the level-based comparison when I said that. Since ranks can only rise about once a month, when comparing against rank the quests would be spaced out plenty in classic. Also giving quests by rank would be consistent with later TES games.

ajrb:
Yeah I think I agree with you, all guilds should use rank. 

The only exception for me is using level for quests like noble ones.. I quite liked that it's based on rep right now. So nobles give quests to those they respect, but it does raise the question of how you get a rep with nobles without them offering quests. Basing on level is probably better.

It is irritating that I implemented a load of reputation code with wildcards etc. It's pretty cool but based on incorrect info. Jay, do you think there's any worth in keeping a rep based mechanism for quest packs etc? I'm thinking probably not, but interested in your opinion as quest master!

ajrb:
Thinking some more about it, I'm quite happy to make the code changes but I'm really not wanting to re-test it all as thoroughly as I did originally. I spent hours and hours testing everything. Hopefully @petchema will pick up any screwups I make as he tends to do. :-)

JayH2971:
I've used reputation requirements to ensure balance in my quests. I tend not to tell everything that's going to happen at the start of a quest, which means I could pop in a vampire or a gargoyle without anyone knowing (my knight quest where you find out the assassin is a Daedra Seducer is a prime example I'm thinking of). I believe this can be adequately satisfied by a level-based progression instead.

The other main attraction for rank-based questing to me is that I've written basic errand quests and simple killing actions as rep 0 quests (as you would for a guild initiate) and enabled more complex ones for rank 50+ (when the player already has the gist of how the guild actually functions). This prevents me from blindsiding a new player with a very complex quest that might poison their perception of the guild. However, a level-based approach would likely not cause many further problems, since most people run out to join all their guilds at level 1-3 anyway. I don't know of anyone who waits until level 15 and then decides to join the Fighters Guild.

tldr: Just fine with a leveling progression for quests.

Interkarma:
@Allofich I'll merge this one now, and I'm planning a couple of changes under the hood. I'm setting up racial overrides via the effect system rather than having vampires completely hard-coded. This will allow for mods to create custom race transformations down the track.

When importing vampires, I'll remove their bonuses as this is handled by the effect system, and reset their race back to birth race. We don't need the vampire race data any longer in RaceTemplate.cs. I'll try to get everything I'm working on in as quickly as I can so you can see where I'm going with it.

Assets/Scripts/Game/Entities/PlayerEntity.cs
```diff
@@ -364,12 +368,11 @@ public override void Update(DaggerfallEntityBehaviour sender)
...
                 {
                     RegionPowerAndConditionsUpdate(true);
-                    // GetVampireOrWerecreatureQuest(); Non-cure quest
+                    GetVampireOrWereCreatureQuest(false);
                 }
 
-                // TODO: Get vampire/werecreature quest
-                //if (((i + lastGameMinutes) % 120960) == 0) // 84 days
-                //    GetVampireOrWerecreatureQuest(); cure quest
+                if (((i + lastGameMinutes) % 120960) == 0) // 84 days
```
Interkarma: I'll just comment this out again temporarily while working on the rest of vampirism. @ajrb might also want to deploy elsewhere via his quest delivery setup. This is an important bit of reference code though, thank you!

PR #1099, opened 2018-12-18
hopefully fixes the issues with vanilla skybox for other weather types than sunny

Nystul-the-Magician:
introduced bool excludeSkybox to FogSettings struct
dfunity vanilla skybox should now work for all weather types

### DIFF
--- a/Assets/Resources/Post-Processing.asset
+++ b/Assets/Resources/Post-Processing.asset
@@ -27,7 +27,7 @@ MonoBehaviour:
   fog:
     m_Enabled: 1
     m_Settings:
-      excludeSkybox: 0
+      excludeSkybox: 1
   antialiasing:
     m_Enabled: 0
     m_Settings:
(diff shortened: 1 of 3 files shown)


Nystul-the-Magician:
one thing I wanted to ask is if we should have fog settings public to inspector - I did this for now (because the old fog density value was available for change in unity inspector as well) - so you can change the weather struct for every weather type in the inspector when WeatherManager gameobject is selected but it is confusing a bit (there are default values in the code - that are overriden by the inspector values - I spend at least some minutes a few days ago to pin down where those changed values came from (when still working with fog density only)
Should we:
- get rid of the default values in code (I have a bad feeling of doing so - when the scene's fog settings values are deleted in the file this would cause the structs to be uninitialized)
- or should we disable the fog settings in the inspector (which makes editor testing much worse)
- or should we leave it as it is (and remember that there is this behaviour of overriding)

Interkarma:
Thank you so much for the quick turnaround on this! I just tested PR and everything seems to be working again. I briefly experienced the editor issue, then it stopped again. I'm thinking this is a Unity quirk.

>or should we leave it as it is (and remember that there is this behaviour of overriding)
This is probably best all around. I know it's frustrating when code values are overridden by editor settings, this used to catch me out a lot also. Once it clicked though, I remembered that any MonoBehaviour with public fields will be automatically serialized and it doesn't seem to catch me out any more.

Nystul-the-Magician:
ok really relieved that it works now ;)

PR #1100, opened 2018-12-18
new experimental ini file setting "TerrainHeightmapPixelError" introduced

Nystul-the-Magician:
new experimental ini file setting "TerrainHeightmapPixelError" introduced
this setting can be used to prevent inaccuracies with location placement on terrain - lower values are better, 1 is best
default is 5 (as before) but we can try to lower it to 1


Nystul-the-Magician:
wanted to make own pull request for the small changes to WeatherManager.cs  - commited without thinking that pr is still open... do you mind to merging it anyway? sry

PR #1104, opened 2018-12-20
Add null check for dfLocation

Allofich:
Fixes https://forums.dfworkshop.net/viewtopic.php?f=24&t=1617.

Looks like the problem was that if you went far enough away from town and the current dfLocation became null, there would be a null reference.


Interkarma:
Thank you!

PR #1105, opened 2018-12-20
Stop pending holiday text if location changes

Allofich:
Fixes https://forums.dfworkshop.net/viewtopic.php?f=24&t=1630.


Allofich:
If you're about to make builds, this is the last fix I'm doing now, you could add it in if you like. It's a minor issue, though.

Interkarma:
Sure thing, I'll merge it and rerun those builds before bed. Thanks for all this stuff. :)

PR #1107, opened 2018-12-20
Enable potion recipe use in potion maker

ajrb:
Also updated all effects to have a static public effect key. I think I got them all.


ajrb:
This finishes the potion implementation, unless the variant icons need to be enabled by a setting. I'm happy to do this, but some have expressed that we have too many settings. Not that I care overmuch about those opinions, but this really seems borderline. @Allofich what's your take from a classic purist POV?

Interkarma:
Cheers mate! I'm happy with the icon differences not to have a variant setting for now.

PR #1109, opened 2018-12-20
Fix casting spells with no spellbook

ajrb:
Fixes: https://forums.dfworkshop.net/viewtopic.php?f=28&p=18606


ajrb:
This will allow a new spellbook to be purchased and then player will have all their old spells. We would need to make the spell list attach to a specific spellbook item to match classic. Not sure it's worth it, I think most players would be grateful they don't have to re-populate it. Up to you @Interkarma.

Interkarma:
I definitely don't want player's to lose spells that way. :) In my original implementation the spellbook was also undroppable, I've always seen it as a flavour item.

PR #1111, opened 2018-12-21
Adjust light source range and light height

ajrb:
Still refining this feature. Feedback welcome. Largest issue is dungeons with large rooms and moving platforms. There's no way to see what's going on if ambient light is very low. (<0.2) But I suppose that's teh users error.


Midknightprince:
I can't see a difference in this, and 0.7.2 live build, what am I looking for ?
Do I need to scale ambient light down (7 dungeons/8 nighttime), or player torch (7)?
And is it just Outdoors or all lights sources ?

ajrb:
You probably wont see a difference with ambient light so high. I've been testing with dungeon=0.1, night=0.3 and torch=1.0

This just extends the range of the lights a little, so just wait for it to come into the main builds.

PR #1112, opened 2018-12-21
Quest dispensing update

ajrb:
- Updated quest dispensing to follow updated rules on UESP (confirmed by Allofich the binary wizard) with a unavoidable change to quest list schema. This will break existing quest packs, so I will work with Jay to ensure his stuff isn't broken for too long.
- Kept the reputation based dispensing for mods, 10+ rep but without any wildcarding
- Adult quests are restricted if player nudity is turned off (nearest thing to childguard)
- Updated the code for vampire quests
- Tracks one time quest acceptance

### DIFF
--- a/Assets/Scripts/Game/Entities/PlayerEntity.cs
+++ b/Assets/Scripts/Game/Entities/PlayerEntity.cs
@@ -26,6 +26,7 @@ using DaggerfallWorkshop.Game.Serialization;
 using DaggerfallWorkshop.Game.Guilds;
 using DaggerfallWorkshop.Game.MagicAndEffects;
 using DaggerfallWorkshop.Game.MagicAndEffects.MagicEffects;
+using DaggerfallWorkshop.Game.Questing;
 
 namespace DaggerfallWorkshop.Game.Entity
 {
@@ -383,11 +384,11 @@ namespace DaggerfallWorkshop.Game.Entity
                 if (((i + lastGameMinutes) % 54720) == 0) // 38 days
                 {
                     RegionPowerAndConditionsUpdate(true);
-                    //GetVampireOrWereCreatureQuest(false);
+                    StartVampireOrWereCreatureQuest(false);
                 }
 
-                //if (((i + lastGameMinutes) % 120960) == 0) // 84 days
-                //    GetVampireOrWereCreatureQuest(true);
+                if (((i + lastGameMinutes) % 120960) == 0) // 84 days
+                    StartVampireOrWereCreatureQuest(true);
             }
 
             // TODO: Right now enemy spawns are only prevented when time has been raised for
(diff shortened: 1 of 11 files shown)


Assets/Scripts/Game/Entities/PlayerEntity.cs
```diff
@@ -460,56 +461,43 @@ public override void Update(DaggerfallEntityBehaviour sender)
...
-            //    }
-            //}
-            ///*else
-            //{
-            //    if (playerIsWereCreature && isCureQuest && DFRandom.random_range_inclusive(1, 100) < 30)
-            //    {
-            //        Questing.QuestMachine.Instance.InstantiateQuest("$CUREWER");
-            //    }
-            //}*/
+            if (Race == Races.Vampire)
```
Interkarma: Please don't uncomment this just yet. :) There is no such thing as Races.Vampire. This is done with racial overriding instead.

ajrb: Okay I commented out both calls to this method - is this okay? 
AFAICS it would simply never have done anything if race never gets set to Vampire.

Assets/Scripts/Game/Entities/PlayerEntity.cs
```diff
@@ -460,56 +461,43 @@ public override void Update(DaggerfallEntityBehaviour sender)
...
+                    if (DFRandom.random_range_inclusive(10, 100) < 30)
+                        QuestMachine.Instance.InstantiateQuest("$CUREVAM");
+                }
+                else if (hasStartedInitialVampireQuest)
+                {
+                    // Get an appropriate quest for player's level?
+                    if (DFRandom.random_range_inclusive(1, 100) < 50)
+                    {
+                        // Get the regional vampire clan faction id for affecting reputation on success/failure, and current rep
+                        int factionId = 23; // TODO: get appropriate value - just hardcoding The Vraseth for now!
```
ajrb: I should point out this line which needs the vampire clan faction id for doling out rep gains etc upon quest completion

Assets/Scripts/Game/Entities/PlayerEntity.cs
Interkarma: Yep, the code won't do anything because it's main condition is checking for the wrong thing. That's why I just wanted it just as reference only so it can be rewritten later. I also want it to be moved out of PlayerEntity to somewhere else, such as the effect managing vampirism itself. I'd prefer the code to stay commented out, as it's basically scheduled for deletion from this point soon.

Assets/Scripts/Game/Entities/PlayerEntity.cs
Interkarma: The clan can be found from VampirismEffect incumbent after player turns, or from VampirismInfection prior to turning. This can be pulled now if you want to set it up. Vampire faction IDs are 150-158 and have a matching enum in DaggerfallUnityEnums.VampireClans.

JayH2971:
> Tracks one time quest acceptance

Does this refer to quests that will only ever be offered one time to the PC?

I'm probably starting to pick away at details before I get the executive summary so maybe I should wait :smile:

Assets/Scripts/Game/Entities/PlayerEntity.cs
ajrb: Okay will comment out all the code now. I agree it should be moved out of player entity, as should some of the other code at some point - it's getting a bit unwieldy as a class.

Assets/Scripts/Game/Entities/PlayerEntity.cs
ajrb: I will leave this for you to add when you're ready to move the code.

ajrb:
> 
> 
> > Tracks one time quest acceptance
> 
> Does this refer to quests that will only ever be offered one time to the PC?
> 
@JayH2971 Yes this is for quests that are only able to be accepted once. That's the way I'm interpreting it for now, unless @Allofich says otherwise and these quests can only ever be offered once. I figured to keep them availiable if player refuses, and only mark them as given if the quest gets accepted. Happy to change this (it would actually be simpler code) if that's the prevailing opinion.

I had written a lot here but decided it would be better posted on forums, see: https://forums.dfworkshop.net/viewtopic.php?f=25&t=1643

Allofich:
> That's the way I'm interpreting it for now, unless @Allofich says otherwise and these quests can only ever be offered once.

Just to be sure we're all on the same page, this isn't based on any code in the binary for interpreting the `1`. It's completely ignored by classic as far as I can see, it just looks for and uses the `X`. The "it only happens once" thing is my theory after going through and reading a bunch of the quests with `1` and looking for something common between them, and noticing that they all seemed to be quests that didn't make sense as repeatable quests.

As for offering/accepting, I agree that offering them more than once seems like the right solution. That's how all the later games work. Considering Daggerfall's main quest line, though, it might not be how classic would have done it. But yeah, any modern game would be nice to players and allow it I think.

BTW, @ajrb, on the subject of one-time quests, is there any solution on-hand for making the vampire initiation quest have to be completed for the rest of the vampire quests to be available? I imagine that's how it was supposed to work. You do that quest, you get introduced to your clan, and they give you a ring for your membership and tell you they'll contact you with more work. But in classic, there are no checks for the quest being successfully completed and you'll get additional quests from your clan even if you fail or ignore it. Would be nice to get that working correctly.

ajrb:
Not from my code, it's simply responsible for dispensing and starting quests. I think this could be done by listening for the QuestMachine.OnQuestEnded and updating a variable when the join quest is complete. 

And yes only offering once would be simpler codewise, but I definitely don't want to do both styles depending on the rejection message. I think lets leave it as it is now.

Allofich:
> I definitely don't want to do both styles depending on the rejection message. 

I meant make them all offered only once, or all offered repeatedly, based on what the pattern of the rejection messages seemed to be.

I just gave a quick look at several of the rejection messages, and I haven't seen any that would make it odd for the quest to be offered again.

ajrb:
> Re: making vamp init quest have to be completed. Not from my code, it's simply responsible for dispensing and starting quests. I think this could be done by listening for the QuestMachine.OnQuestEnded and updating a variable when the join quest is complete.

See GuildManager class for how that handles the initiation quests for Thieves & DB. If vampires were an actual guild with ranks I would simply have copied this. Oh daggerfall, you beautiful mess.

Interkarma:
>BTW, @ajrb, is there any solution on-hand for making the vampire initiation quest have to be completed for the rest of the vampire quests to be available?

I can help with this if its not a good fit for @ajrb's quest dispensing code. :)

> If vampires were an actual guild with ranks I would simply have copied this. Oh daggerfall, you beautiful mess.

That's an interesting thought. It seems like vampires _should_ be a guild doesn't it?

Assets/Scripts/Game/Entities/PlayerEntity.cs
Interkarma: Yeah, I'd like cleanup review of PlayerEntity at some point. But happy to let the rest sit for now. Cheers mate. :)

ajrb:
Ah well, that's just me being confusing by not thinking it through fully before posting. And there was me hoping no one would notice. ;-)

Interkarma:
I really appreciate all the work on this stuff. I'm still parsing through this update but it looks good so far.

Interkarma:
This all looks good. I'll merge now so it goes into next updates and starts being playtested.

PR #1113, opened 2018-12-22
Guard improvements

Allofich:
Fix for https://forums.dfworkshop.net/viewtopic.php?f=24&t=1637

### DIFF
--- a/Assets/Scripts/Game/EnemySounds.cs
+++ b/Assets/Scripts/Game/EnemySounds.cs
@@ -1,4 +1,4 @@
-﻿// Project:         Daggerfall Tools For Unity
+// Project:         Daggerfall Tools For Unity
 // Copyright:       Copyright (C) 2009-2018 Daggerfall Workshop
 // Web Site:        http://www.dfworkshop.net
 // License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
@@ -46,7 +46,7 @@ namespace DaggerfallWorkshop.Game
         float waitCounter;
         float volumeScale = 1f;
 
-        void Start()
+        void Awake()
         {
             // Save references
             dfAudioSource = GetComponent<DaggerfallAudioSource>();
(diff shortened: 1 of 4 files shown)


Allofich:
Various improvements to guards:

- Guards spawning from MobileNPCs (including guard MobileNPCs) now are looking in the same direction as the MobileNPC was looking when they first spawn

- Hitting MobileNPC guards is now an assault, not a murder (MobileNPC guards "dying" in one hit and counting as a murder is an issue in classic as well). If the EnemyMobile guard is killed, then it become murder.

- Hitting MobileNPC guards will now instantly create an EnemyMobile guard in the same spot (intended to already be doing this, but because I skip over disabled NPCs when choosing where to spawn guards from, it wasn't happening). The EnemyMobile guard will take a hit/parry as usual. (Had to make Start() in EnemySounds into Awake() so that the dfAudioSource will be ready in time)

- Slight redo of guard spawning code. Only MobileNPC guards in classic spawn range are selected for spawning MobileEnemy guards from. I had originally made all MobileNPC guards spawn into MobileEnemy guards instantly, but after I added the code to periodically update nearby MobileNPC guards into MobileEnemy guards when the player is being chased by guards, it was no longer necessary to make them all into MobileEnemy guards immediately.

Allofich:
BTW, I added some code that I think will work OK for spells hitting MobileNPCs, but spells are currently not damaging MobileNPCs, they just go through them without colliding. I this was working before? Arrows work.

Allofich:
> I'll take a look at this before merging just in case there's a simpler fix.

This PR doesn't contain a fix. I just added some code to account for attacking guards next to the existing part for attacking MobileNPCs. Since spells aren't hitting MobileNPCs, though, it's untested.

Interkarma:
Thanks for clarifying. I haven't reviewed code, just responding to comments. Will loop back when I can to look closer. :)

Allofich:
I'll test this in a while against the new master code since spells on NPCMobiles should be working.

Allofich:
Yikes, does not work right. Don't shoot a spell at a MobileNPC with this PR with the sound on right now, or you'll get an explosion of overlapping spell impact sounds as it repeatedly hits an exception in the PR code. I'll fix it.

Allofich:
Fixed. BTW I still had a couple instances where a spell seemed to go through a target. Once with a MobileNPC and once with a MobileEnemy guard, although it worked most of the time. Might still be some tuning needed there.

PR #1117, opened 2018-12-23
Fixes missing streaming world locations

Lypyl:
Relevant bug reports:

https://forums.dfworkshop.net/viewtopic.php?f=24&t=1628

https://forums.dfworkshop.net/viewtopic.php?f=24&t=1645

Issue occurs when terrain objects become pooled & their locations are cleaned up. Because updateLocation was being set to false the locations were not being recreated.


Interkarma:
That's a fairly obvious mistake on my part now that I see it. Thank for debugging and providing such a simple and effective fix.

PR #1118, opened 2018-12-23
Only give vampire quests once initiation quest is finished

ajrb:
NOTE: Untested.


Interkarma:
Thanks for the PR! I'm now tracking this in the VampirismEffect itself, which is also where the vampire quests are delivered as part of racial override system. All nice and centralised to effect now, and no need to track this independently in quest machine. Will close this one.

PR #1120, opened 2018-12-23
Adjust large colliders to fit through doors

Allofich:
Fixes https://forums.dfworkshop.net/viewtopic.php?f=24&t=329.

I just got this through trial and error, but it doesn't seem to cause any problems.

### DIFF
--- a/Assets/Scripts/Game/SetupDemoEnemy.cs
+++ b/Assets/Scripts/Game/SetupDemoEnemy.cs
@@ -70,11 +70,16 @@ namespace DaggerfallWorkshop.Game
                     if (dfMobile.Summary.Enemy.Behaviour == MobileBehaviour.Flying)
                         controller.height /= 2f;
 
-                    // Uncomment below lines to limit maximum controller height
+                    // Limit maximum controller height
                     // Some particularly tall sprites (e.g. giants) require this hack to get through doors
-                    // However they will appear sunken into ground as a result
-                    //if (controller.height > 1.9f)
-                    //    controller.height = 1.9f;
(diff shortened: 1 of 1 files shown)


Allofich:
Had to limit the height to 1.75 in order for the thief to be able to get through the doorway in the first picture without trouble. (They can also get through with 1.8 if they go at an angle but still get stuck going straight)

Assets/Scripts/Game/Entities/PlayerEntity.cs
```diff
@@ -450,7 +450,7 @@ public override void Update(DaggerfallEntityBehaviour sender)
             HandleStartingCrimeGuildQuests();
 
             // Reset surrender to guards dialogue if no guards are nearby
-            if (haveShownSurrenderToGuardsDialogue && GameManager.Instance.HowManyEnemiesOfType(MobileTypes.Knight_CityWatch, true) == 0)
+            if (haveShownSurrenderToGuardsDialogue && GameManager.Instance.HowManyEnemiesOfType(MobileTypes.Giant, true) == 0)
```
ajrb: Did you mean to leave these changes in the PR? Looks like testing stuff to me... but i could be wrong

Allofich: Haha, yeah. I was testing some giants to make sure they could fit under the mages guild sign because I had noticed that was a case where the player can get through but AI characters couldn't. Let me fix that.

Interkarma:
Thanks Allofich! You can probably tell from the comments that was something I meant to do another pass on later and didn't get around to. I appreciate the help. :)

PR #1122, opened 2018-12-26
also set "random name" button visibility during Setup()

petchema:
... otherwise it seems to be visible for all races.
(I wonder if that's some kind of Unity regression or if that bug has always been around)


petchema:
After more tests, this patch is not correct: if I _start_ with a race that has a name generator, I don't get the Random button. But if I go back and keep selecting different races, the Random button then shows up as it should.
It's as if either the OnPush() method was not called the first time around, or is called too late to have an effect on the Random button...

Interkarma:
OnPush() should be the best place, there's likely something else happening with timing as you suggest. I'll take a look when I can. :)

PR #1124, opened 2018-12-26
Minor tutorial content fixes

petchema:
The goal is to adjust the tutorial content to be less confusing and more
helpful for new (DFU) players:
- remove references to icons-based interface, cursor movements mode and
save slots;
- mention WASD, quicksaving/quickloading and controls customization.

I tried to keep the changes minimal to preserve the classic experience.
Pace will probably be too slow for 80% of even new players, but I didn't change it
either.

### DIFF
--- a/Assets/StreamingAssets/Quests/_TUTOR__.txt
+++ b/Assets/StreamingAssets/Quests/_TUTOR__.txt
@@ -61,12 +61,12 @@ Message:  1012
 <ce>            with moving your character around. Don't worry
 <ce>          about monsters, there are none in this first room.
 <ce>                                    
-<ce>           When this text clears, move the mouse around,
-<ce>           it will change where you are looking.
+<ce>           When this text clears, move the mouse around
+<ce>            to change where you are looking.
 <ce>
 <ce>              Use the W, A, S and D keys to go around.
-<ce>          Hint: Those are the default keys, to change them later
-<ce>          use the ESC key then click on CONTROLS button.
+<ce>          Hint: Those are the default keys. You can change them later
+<ce>          using the ESC key and then the CONTROLS button.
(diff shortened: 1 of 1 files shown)


JayH2971:
Assuming Interkarma approves this, would you accept some minor grammatical corrections on my part? I believe this is very good content, just that the English could be a little better in a few cases.

petchema:
> Assuming Interkarma approves this, would you accept some minor grammatical corrections on my part? I believe this is very good content, just that the English could be a little better in a few cases.

Of course! I have no doubt you can improve what I did here

JayH2971:
Okay! Sorry, it's either me or my OS that has problems with using github directly so I'm going to have to present it as a separate text. Hope that isn't a bother for you. [Pastebin link](http://pastebin.com/bEmCa5ev)

petchema:
I was surprised to find no explanation on how to rest in the tutorial. Maybe the feature was obvious enough with the icons-based interface.
I added the sentence on controls customization also as a hint for all the actions not covered by the tutorial, but I'm not sure that's enough...

Interkarma:
I'm happy for the tutorial quest to be rewritten entirely if you see fit. Chucking those "range 0 1" for now is also a good idea. Thanks for your work on this. :)

petchema:
Thanks! I'm testing the removal of "range 0 1" before pushing the change here

petchema:
timers seem to work fine without their ranges

Interkarma:
Thanks for your work on this, and for your efforts with timings in general. With the information from Elenwel in below topic, the Clock resource will get a small review soon.

https://forums.dfworkshop.net/viewtopic.php?f=23&t=1655&start=10#p19174

PR #1125, opened 2018-12-27
Make NPC reaction rolls semi-random

numidium:
Addresses issue discussed in https://forums.dfworkshop.net/viewtopic.php?f=24&t=1500&start=10
Adds a random number (0-15 inclusive) to the reaction rating so that not all NPCs will refuse to give info if the base reaction number is below zero.


Interkarma:
Thanks for your work on this!

PR #1127, opened 2019-01-01
Fix NPCs to use parent factions for dialogue

Allofich:
Fixes https://forums.dfworkshop.net/viewtopic.php?f=24&t=1683.

The problem was the NPC in question is of the Order of the Lamp faction, which has a gguild of 10, or knightly order. The code would then try to look up a knightly order with the building faction of 40 (Mages Guild) and throw an exception.

When opening dialogue, for NPCs of a faction that is not of type 2, 7, or 9, classic will iterate up through their parents until it finds one that is of these factions (if there is one) and use that faction instead. With this behavior replicated, the NPC in the bug report will now use the parent faction of Mages Guild and give a Mages Guild greeting like in classic.

### DIFF
--- a/Assets/Scripts/Game/TalkManager.cs
+++ b/Assets/Scripts/Game/TalkManager.cs
@@ -766,7 +766,13 @@ namespace DaggerfallWorkshop.Game
             DaggerfallUI.Instance.TalkWindow.UpdateNameNPC();
 
             FactionFile.FactionData factionData;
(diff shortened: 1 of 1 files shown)


Allofich:
The bug report includes a save but when I load it I get a message about the terrain sampler and relocated outside the city. From the screenshot included it looks like some replacement textures were used. Here is a save of the same character from unmodded DF Unity in front of the same NPC.
[TalkBug.zip](https://github.com/Interkarma/daggerfall-unity/files/2719071/TalkBug.zip)

Interkarma:
Nicely done, thank you! :)

PR #1129, opened 2019-01-01
Match armor dye colors to classic

Allofich:
Fixes the second part of https://forums.dfworkshop.net/viewtopic.php?f=24&t=1660&sid=5ee3de52dcc231329e59ee68733ff5e8

DF Unity was using "Steel" dye as the default, unchanged, color, but classic uses "Silver" dye. This is used for both leather and classic. Removed the LeatherHelm fix as it's no longer necessary.

Also, DF Unity displayed Silver and Elven dyes as the same color (both as Silver), but Elven is different. Elven/Silver items from existing DF Unity saves will still appear as Silver. Also, the Orcish dye differed from classic. Any intentional changes?

Color comparisons attached. Left = master branch. Right = this PR (same as classic)

Leather
leather

Elven
elven

Orcish
orcish

### DIFF
--- a/Assets/Scripts/DaggerfallUnityEnums.cs
+++ b/Assets/Scripts/DaggerfallUnityEnums.cs
@@ -349,7 +349,6 @@ namespace DaggerfallWorkshop
         None,
         Iron,
         Steel,
-        Chain,
         Silver,
         Elven,
         Dwarven,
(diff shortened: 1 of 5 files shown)


Interkarma:
Great! Thank you for your work on this. It's nice to see these kinds of incremental improvements.

PR #1131, opened 2019-01-03
interior/dungeon fog is now independent of weather, savegame restore mechanism fix

Nystul-the-Magician:
interior/dungeon fog is now independent of outdoor weather
fog from save games should be now restored correctly


Interkarma:
Awesome! Thanks for continuing to work on this Nystul. :)

PR #1132, opened 2019-01-03
fix for issue with classic savegame import and exterior automap

Nystul-the-Magician:
described in https://forums.dfworkshop.net/viewtopic.php?f=24&p=19244#p19244)


Interkarma:
Fantastic work, cheers! :)

PR #1134, opened 2019-01-05
Bash exterior doors

numidium:
Currently uses the best open chance to determine a successful bash.  I used the same formula regardless of whether the door is locked since unlocked doors aren't automatically kicked open on the first try in classic.


Interkarma:
Nice one! Thank you for working on that. :)

PR #1135, opened 2019-01-05
fog color in interiors and dungeons is now black

Nystul-the-Magician:


Nystul-the-Magician:
btw, we could maybe go with higher density fog values for dungeons - but I did not dare to change it for the moment since I don't want to break other stuff that I did not have time to test (like darker dungeons feature)
but a DungeonFogSettings with density e.g. 0.05 looks better than the current 0.005 in my opinion - feel free to try and fine tune - the same for InteriorFogSettings (used for building interiors)

Interkarma:
Thanks for the fix! By the way, it looks like foggy weather can appear in dungeons as well.

https://forums.dfworkshop.net/viewtopic.php?f=4&t=37&p=19399#p19377

Bug or not, that does look kind of incredible. I see your point about higher fog density in dungeons. :)

Maybe "foggy dungeons" is something we can expose as an option with some randomness to density (seeded by mapid perhaps so it's the same each time player enters dungeons).

Nystul-the-Magician:
previous pr fixed foggy weather in dungeons ;)

Interkarma:
>previous pr fixed foggy weather in dungeons ;)

Ah, thank you! Sorry I missed that, I haven't jumped into review mode for this round of PRs yet. :)

We can worry about intentionally foggy dungeons later in any case. 👍

PR #1138, opened 2019-01-06
Account for change in center of collider after height adjustment

Allofich:
https://github.com/Interkarma/daggerfall-unity/pull/1120 caused a problem because the adjusted center of the collider interfered with a few calculations that used transform.position and colliders. Specifically, I noticed a Thief and an Orc were no longer able to see each other to fight and would simply stand there looking at each other.

The change in center of the collider is now accounted for.

Also, I found that the player's collider is height 1.78. Enemies now use this value as well. I confirmed the Thief in Privateer's Hold can fit through the door with this value.


Interkarma:
Cheers! This one now has a minor conflict with a previous merge. I'll let you take a look first so I don't break anything on you. :)

Allofich:
Resolved the conflict.

PR #1139, opened 2019-01-06
Spell icons modding fix (RFC)

petchema:
This is a hack so that spell icons can be modded.
But more than anything this is a call for a cleaner fix.
Forums: https://forums.dfworkshop.net/viewtopic.php?f=22&t=1698#p19359

### DIFF
--- a/Assets/Scripts/Game/UserInterface/SpellIconCollection.cs
+++ b/Assets/Scripts/Game/UserInterface/SpellIconCollection.cs
@@ -174,12 +174,16 @@ namespace DaggerfallWorkshop.Game.UserInterface
             spellIcons.Clear();
 
             // Get source atlas
-            Texture2D spellIconAtlas = DaggerfallUI.GetTextureFromImg(spellIconsFile);
-            if (spellIconAtlas == null)
+            Texture2D loadSpellIconAtlas = DaggerfallUI.GetTextureFromImg(spellIconsFile);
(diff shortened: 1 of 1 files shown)


petchema:
Thank you!

TheLacus:
> Graphics.CopyTexture with a region will not copy readable texture data for compressed formats (source texture format 12)

This warning refers to texture data on cpu side. If the texture is marked as read-only with Apply() the message is not shown in build, only in editor because a copy of texture is always kept, but is harmless .

> Graphics.CopyTexture can only copy between same texture format groups (OpenGL internal formats: src=33779, blockSize=16 ; dst=32856, blockSize=4)

Need to specify correct format in constructor :)
`Texture2D iconTexture = new Texture2D(dim, dim, spellIconAtlas.format, false);`

A limitation of copy with DXT is that size and coordinates must be a multiple of block size (4 pixels) so it's probably better to disable compression for UI images. I'll work on a PR.

TheLacus:
Note that we can't control the format of textures imported from dfmods, only from loose files. If a mod author use a compressed format i think we should assume is an intended choice and an appropriate size is used. 

I'm going to push a PR to use the correct format in constructor and write to log if size is incorrect. This will allow to use VMblast textures when resized to 1760x352. Then we can decide if we should override game setting for compression for specific images like this one when loaded from loose files :)

petchema:
Superseded by #1141

PR #1141, opened 2019-01-07
Use correct format for ICON00I0.IMG when compression is enabled

TheLacus:
See discussion for  #1139.

I also disabled mipmaps, is this ok?


petchema:
I only had time for a quick test, but it seems to work fine.
About mipmaps, I assume that means that for best results icons shouldn't be much larger than their display size? I assume this is an acceptable constraint...
In short, LGTM

TheLacus:
@petchema @Interkarma Mipmaps are disabled for most other UI images. If it was intentional to leave them enabled for this one, i can change it before PR is accepted.

PR #1143, opened 2019-01-08
Fix order of male robes vs. cloaks

Allofich:
Fixes https://forums.dfworkshop.net/viewtopic.php?f=24&t=1458.

The issue only existed for male robes with cloaks. For females the robes have a draw order value of 80 vs. 90 for the cloaks, but for males they are both 90. It seems like a minor oversight in the original data, so the best fix seemed to just be to correct the values in `ItemTemplates.txt.`

As for why it works in classic, I believe it's because in classic the items are effectively drawn by two orderings, the `drawOrder` value but also the `EquipSlot`, with smaller `EquipSlot` value items being drawn in front of items with larger `EquipSlot` values if the `drawOrder` of the two items is the same.

Modifying `PaperDoll.cs` from

```
            // Sort equipped items by draw order
            List<DaggerfallUnityItem> orderedItems = equippedItems.OrderBy(o => o.drawOrder).ToList();
```

to

```
            // Sort equipped items by equip slot and draw order
            List<DaggerfallUnityItem> orderedItems = equippedItems.OrderByDescending(o => o.EquipSlot)
                                                    .ThenBy(o => o.drawOrder).ToList();
```

fixes the robe vs. cloak issue. However, it also causes double cloaks to be drawn backwards (with the second-equipped cloak in front of the first, rather than the first in front as in classic and current master), so additional tinkering would probably be needed there. I think the cloaks work as they do in classic because the second-equipped cloak would be the larger-value slot `Cloak2` (16) compared to the first cloak, `Cloak1` (14), and as I said above items with a larger `EquipSlot` seem to be drawn behind items with smaller `EquipSlot` values if their `drawOrder` is the same.

It's possible that classic may just have coincidentally worked for the male robe + cloak combination, which hid the problem in the item template data. It's probably better to just fix the item templates.

Also fixed a description for an item flag in classic item records. That value is not just for edged weapons, but for all one-handed weapons.


Allofich:
By the way, I checked and the drawOrder for these items was 90 even before `ItemTemplates.txt` was regenerated recently. If male robe + cloak was working correctly before, I'm not sure why that would be.

Interkarma:
Thanks for the fix! :)

PR #1144, opened 2019-01-09
 optimization: GenerateSamples function now computes heights in several threads

Nystul-the-Magician:


Interkarma:
With 8 cores on my system, it now takes around 1-2ms instead of 9ms. Quite a solid improvement and no issues I can see. Great work Nystul!

PR #1145, opened 2019-01-10
reuse CheckLocationDiscovered to allow notebook fast travel

petchema:
#1084 made notebook fast travel checks too strict, for instance not allowing to fast travel to a town you didn't visit before.
Forums: https://forums.dfworkshop.net/viewtopic.php?f=28&t=1578&p=19594#p19575
This patch reuses DaggerfallTravelMapWindow.CheckLocationDiscovered(), I hope that's the right thing to do.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallQuestJournalWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallQuestJournalWindow.cs
@@ -339,8 +339,10 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             DFLocation location;
             if (DaggerfallUnity.Instance.ContentReader.GetLocation(regionName, name, out location))
             {
-                int mapPixelID = MapsFile.GetMapPixelIDFromLongitudeLatitude(location.MapTableData.Longitude, location.MapTableData.Latitude);
-                return GameManager.Instance.PlayerGPS.HasDiscoveredLocation(mapPixelID);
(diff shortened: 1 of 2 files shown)


ajrb:
Not sure I understand this change, does it mean there are two mechanisms in DFU for location discovery then? One that tracks which locations PC is aware of and another that tracks which locations PC has actually been to? If so, that's news to me.

petchema:
Exactly, and I assume that summary.Discovered flag is "stuff you're aware of", like all towns, farmsteads, temples, and some of the cemeteries at the beginning of the game. But then with code comments like
```
// TODO: This by itself doesn't account for DFRegion.LocationTypes.GraveyardForgotten locations that start the game discovered in classic
summary.Discovered = mapTable.Discovered;
```
...may not be the whole story.
Anyway, aligning notebook fast travel condition with world map fast travel condition should be safe, if the code is bug free...

ajrb:
Hmm, I honestly had not realised there were two levels of discovery. Thing is I get quests to places I have not travelled to and I can use the notebook to travel there just fine. Maybe that's beacuse I only tested dungeons and towns are different?

Interkarma:
The discovery system is only a subset of total places in game world, so it's possible for a location to be available (e.g. town is visible on the travel map) but not yet recorded in the discovery database as the player hasn't been there yet. It's safe to assume that all towns, taverns, homes, etc. can be fast travelled to without needing to be part of the discovery record. I think Pango's solution here is a good one, and I feel silly for not picking this up myself when I reviewed earlier change. :)

ajrb:
Agreed, thanks for the fix Pango! I also learned something about the discovery system when it comes to towns. Was quite confused for a while tho... :-)

PR #1146, opened 2019-01-11
Changes to SpellIconCollection for more and better icons

Alyndiar:
Modified GetSpellIcon method to allow for more than 69 icons
Modified LoadSpellTargetAndElementIcons method to calculate target and element icon sizes based on atlas size instead of hardcoding. May need more changes to allow for compressed textures as this atlas is currently forced to load as an ARGB32 texture. I need a little help to ensure that the current functions used to extract the icons for target and element are correct since the are different from those used in the spell icons extraction portion.

### DIFF
--- a/Assets/Scripts/Game/UserInterface/SpellIconCollection.cs
+++ b/Assets/Scripts/Game/UserInterface/SpellIconCollection.cs
@@ -40,7 +40,7 @@ namespace DaggerfallWorkshop.Game.UserInterface
         const string spellIconsFile = "ICON00I0.IMG";
         const string spellTargetAndElementIconsFile = "MASK04I0.IMG";
         const int spellIconsRowCount = 20;
-        const int spellIconsCount = 116;
+        const int spellIconsCount = 69;
         const int spellTargetIconsCount = 5;
         const int spellElementIconsCount = 5;
 
--- a/Assets/Scripts/Game/UserInterface/SpellIconCollection.cs
+++ b/Assets/Scripts/Game/UserInterface/SpellIconCollection.cs
@@ -29,6 +29,9 @@ namespace DaggerfallWorkshop.Game.UserInterface
     /// This ensures icons are extracted in the same order as classic at new size without any further metadata.
     /// Current design does not allow for adding any more or less than 69 icons to collection.
     /// This could be an enhancement added later but would need to be carefully considered as it crosses multiple UIs.
+    /// Note by Alyndiar : Double-checked everywhere and number of spell icons is solely decided by the SpellIconCollection
+    /// class. All information is self contained and no part of the engine makes any assumption about the number of icons.
+    /// Accordingly, this enhancement can be added without worries.
     /// </summary>
     public class SpellIconCollection : IEnumerable
     {
@@ -37,7 +40,7 @@ namespace DaggerfallWorkshop.Game.UserInterface
         const string spellIconsFile = "ICON00I0.IMG";
         const string spellTargetAndElementIconsFile = "MASK04I0.IMG";
         const int spellIconsRowCount = 20;
-        const int spellIconsCount = 69;
+        const int spellIconsCount = 116;
         const int spellTargetIconsCount = 5;
         const int spellElementIconsCount = 5;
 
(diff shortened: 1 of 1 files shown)


TheLacus:
You may want to use `git rebase -i HEAD~4` to clean history. Set all commits but the first with the `fixup` flag ;)

petchema:
I noticed that modded target and element icons are used in spellbook and spell merchant, but not the spell maker
Quick and dirty upscale for testing: http://ks369561.kimsufi.com/~petchema/gfx/Daggerfall/MASK04I0.IMG.png

petchema:
One problem with targets and elements is that the background button is part of the texture. Worse yet, unselected targets and elements in the spell maker screen are part of the background form.
So modding targets and elements icons may require more thinking than just be able to load larger textures...

Alyndiar:
> You may want to use `git rebase -i HEAD~4` to clean history. Set all commits but the first with the `fixup` flag ;)

Sorry, I don't know how to do that yet. I'm using Github Desktop and I see nothing in there to do it. I also don't know how to use command-line git. This is the first collab project I work on.

Alyndiar:
> One problem with targets and elements is that the background button is part of the texture. Worse yet, unselected targets and elements in the spell maker screen are part of the background form.
> So to be compatible with later UI modding, modding targets and elements icons may require more thinking than just be able to load larger textures...

Yes, I noticed. That is why I was asking for help on those. It works in the spellcasting/spellbook view though. It's only in the spellmaker that changes are needed. As far as I understood though, the base image is the unselected icons and only a gold filter texture is superposed on selection. Anyway, this change keeps existing functionality intact and allows for extension of the concept later on. No more changes will be needed to SpellIconCollection.cs

TheLacus:
> 
> 
> > You may want to use `git rebase -i HEAD~4` to clean history. Set all commits but the first with the `fixup` flag ;)
> 
> Sorry, I don't know how to do that yet. I'm using Github Desktop and I see nothing in there to do it. I also don't know how to use command-line git. This is the first collab project I work on.

_Repository_ > _Open in Powershell_
It's not an issue if you can't, i only suggested it to clean up the duplicate commits :)

petchema:
> It works in the spellcasting/spellbook view though.
Well, not quite. If some future UI mod uses a different stone background, or no stone background, or want to change buttons style, it'll have to come with its own targets and elements gfx.
I don't know if that's a huge deal, maybe it'd better anyway?

petchema:
> It works in the spellcasting/spellbook view though.

Well, not quite, that's what I was saying. If some future UI mod uses a different stone background, or no stone background, or changes buttons style, it'll have to come with its own set of targets and elements gfx.
I don't know if it's a huge deal, maybe it'd better anyway?

Alyndiar:
> _Repository_ > _Open in Powershell_
> It's not an issue if you can't, i only suggested it to clean up the duplicate commits :)

Well, I tried... didn't work it seems... There's a whole other science just in getting used to git... I'll try to figure it out at some point, I just don't have the head for it now...

Interkarma:
Pango is right, it's not a full solution for the element and target icons. Spellbook and spellmaker use slightly different approaches as well due to how those textures are used, and sizes are still hardcoded. I should have considered modding better when writing these interfaces to create a more unified approach.

That said, there's definitely an argument to be made that having element and target icons modable for spellbook is better than not at all. Let me think on this one for a while. :)

Alyndiar:
I'll also take a look at the spellmaker compared to spellbook once I'm finished with these nice icons.

Can I also ask how to allow the number of icons to be set from mod instead of hardcoded (const int spellIconsCount = 69;) as it is right now?

Alyndiar:
Wow! OK, I'll be eager to see all of that! I'd still propose for there to be a way (option maybe) to completely make the classic icons disappear in favor of new icons. Only for those who want that, of course.

Interkarma:
@Alyndiar What I'm setting up is support for "icon packs" in DFU. Each icon pack is one part atlas and one part metadata. The metadata lets modder define number of icons and provide hints like suggested effects for that icon. Classic icons are basically the "system icon pack" but can be replaced loose texture mods like now.

This is still very much WIP right now, I just knocked out some of the basics while I felt up to it. I'll start adding UI support and will provide more information on how this is used once its a bit further along. :)

Interkarma:
My apologies for steamrolling the PR. You've really inspired a nice feature though. :)

I'll close PR for now. We'll still need to look at texture replacement support for target and element icons, but we can handle that at a later date once the spell icons have settled in a bit.

PR #1148, opened 2019-01-13
Optimise applying climate settings for locations and asset injection checks

ajrb:
1)
If not yielding across frames, only apply climate settings once dfLocation is completely initialised instead of for every block. This needs to be verified as logically correct before merging. I'm pretty sure that it can be moved but I don't have a deep enough understanding of this code to be absolutely sure

Also improved layout times diagnostics.

This optimisation almost halves the time it takes to construct a city location when exiting buildings after loading an interior save. I stumbled across it when analysing the exterior world code to see how it might have parallelisation applied. 

2)
Only check for assets to be injected once by keeping a record of tried assets. This halves the time taken to layout terrain because of the sheer number (hundreds of thousands) of them. Most players will have asset injection switched on, and I think it's okay to check each asset once only.

Also I added a command to clear the cache of asset names, although you don't appear to be able to change assets and have them loaded by Unity without restarting the game so not sure it's particularly useful.

### DIFF
--- a/Assets/Scripts/Terrain/StreamingWorld.cs
+++ b/Assets/Scripts/Terrain/StreamingWorld.cs
@@ -585,7 +585,6 @@ namespace DaggerfallWorkshop
 
 #if SHOW_LAYOUT_TIMES
             System.Diagnostics.Stopwatch stopwatch = System.Diagnostics.Stopwatch.StartNew();
-            long startTime = stopwatch.ElapsedMilliseconds;
 #endif
 
             CollectLooseObjects(true);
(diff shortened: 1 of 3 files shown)


Assets/Scripts/Terrain/StreamingWorld.cs
```diff
@@ -773,7 +774,8 @@ private IEnumerator UpdateLocation(int index, bool allowYield)
                             go.hideFlags = defaultHideFlags;
                             go.transform.parent = locationObject.transform;
                             go.transform.localPosition = blockOrigin;
-                            dfLocation.ApplyClimateSettings();
+                            if (allowYield)
+                                dfLocation.ApplyClimateSettings();
```
ajrb: I assumed that it's important to apply climate settings before yielding for the next frame.. but I don't really know if it makes a difference.

Interkarma: The terrain and locations are loaded in around player when they're exploring outside world, and I think my reasoning was that player might see texture swaps happen (like pop-in) if climate settings were delayed all the way to end. It's been a while though, some of this stuff dates back to 2014 when it was just DFTFU. It's worth testing without though under a variety of conditions to see if any texture swap pop-in is visible. This might even turn out to be unnecessary. :)

Interkarma:
Great work Hazelnut :)

PR #1150, opened 2019-01-15
Optimise LayoutNatureBillboards loop

ajrb:
Only a tiny improvement but still worthwhile since it iterates a large number of times.
Also tidied the checks code in mesh replacement.

### DIFF
--- a/Assets/Scripts/Terrain/TerrainHelper.cs
+++ b/Assets/Scripts/Terrain/TerrainHelper.cs
@@ -1,4 +1,4 @@
-﻿// Project:         Daggerfall Tools For Unity
+// Project:         Daggerfall Tools For Unity
 // Copyright:       Copyright (C) 2009-2018 Daggerfall Workshop
 // Web Site:        http://www.dfworkshop.net
 // License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Terrain/TerrainHelper.cs
```diff
@@ -404,52 +431,27 @@ public static void LayoutNatureBillboards(DaggerfallTerrain dfTerrain, Daggerfal
...
-                    // Chance scaled by base climate type
-                    float climateScale = 1.0f;
-                    DFLocation.ClimateSettings climate = MapsFile.GetWorldClimateSettings(dfTerrain.MapData.worldClimate);
-                    switch (climate.ClimateType)
-                    {
-                        case DFLocation.ClimateBaseType.Desert:         // Just lower desert for now
-                            climateScale = 0.25f;
-                            break;
-                    }
+                    if (rect.Contains(tilePos))
```
ajrb: This should check that rect.x > 0 && rect.y > 0 will fix that tonight

ajrb: I need to add the 'null' rect test back in here otherwise it's calling contains for non-location tiles.

PR #1153, opened 2019-01-17
Make town population textures from areas with 'MountainWoods' climate value match classic

df-ferital:
In classic, areas under climate value equal or above 230 (MountainWoods) use Breton race and names for town people.
This fixes DFU to match classic.
Actually, another difference with classic I couldn't fix yet is that town people names should not be generated according to climate values. E.g. in classic, Dragontail people use Nord textures (climate value 226, Mountain) with Redguard names, whereas Wrothgarian people (same climate value) use Nord textures with Breton names.

### DIFF
--- a/Assets/Scripts/API/MapsFile.cs
+++ b/Assets/Scripts/API/MapsFile.cs
@@ -456,8 +456,8 @@ namespace DaggerfallConnect.Arena2
                     settings.GroundArchive = 102;
                     settings.NatureArchive = (int)DFLocation.ClimateTextureSet.Nature_TemperateWoodland;
                     settings.SkyBase = 16;
-                    settings.People = FactionFile.FactionRaces.Nord;
-                    settings.Names = FactionFile.FactionRaces.Nord;
+                    settings.People = FactionFile.FactionRaces.Breton;
+                    settings.Names = FactionFile.FactionRaces.Breton;
                     break;
                 case (int)Climates.Woodlands:
                     settings.ClimateType = DFLocation.ClimateBaseType.Temperate;


Interkarma:
Hey Ferital! :)

Can you please provide town names for where these mismatches occur? Keep in mind that climate map is distinct from regional map, I'll need to verify the climate pixel in classic vs DFU. Using a sample town is a good consistent way to do this.

df-ferital:
Hey Interkarma :)

Western towns of the Ilessan Hills are located in climate area 230. For example, you can check Aldhead Rock to see the mismatch. Actually, the MountainWoods area in this part of High Rock is much wider as it also encompasses eastern Glenpoint and northern Daggerfall as well.

Also, a small part of Wayrest (in the north-east) uses this same climate value.

Actually, I did some FALL.EXE reverse engineering to understand the function which loads town people flats, so I figured out that it works like this:

```
Climate value    Race number
   <=225             2 (Redguard)
     226                0 (Nord)
     227                2 (Redguard)
     228                1 (Breton)
     229                1 (Breton)
   >=230             1 (Breton)
```
I discovered that while trying to fix Redguards town people in Hammerfell (https://forums.dfworkshop.net/viewtopic.php?f=8&t=1583)

Interkarma:
Excellent! You've definitely done your leg-work then. :)

What are your thoughts on the town name selection? How do you feel this might be determined if not by climate?

df-ferital:
Don't know yet, but I'm almost sure it's in a different function. I will fire up IDA Pro and try to locate it :)

PR #1154, opened 2019-01-18
Missed changed in new Spell Icons interface

Alyndiar:
The LoadPackIcons method is using the wrong variable for number of icons to load. Error shows as blank icons for all icons in the icon pack that exceed the number of icons in the classic icon pack. 
Changed the loop to use pack.IconCount instead of SpellIconCount


Alyndiar:
Error in my git manipulations... Trying again...

Assets/Scripts/Game/UserInterface/SpellIconCollection.cs
```diff
@@ -320,7 +334,7 @@ void LoadPackIcons(SpellIconPack pack, Texture2D atlas, string path)
 
             // Read icons to their own texture (remembering Unity textures are flipped vertically)
             int srcX = 0, srcY = atlas.height - dim;
-            for (int i = 0; i < SpellIconCount; i++)
+            for (int i = 0; i < pack.iconCount; i++)
```
Interkarma: Thanks for spotting this, good catch! I'm about to spin up new builds, so I'll quickly make this change and credit you.

PR #1155, opened 2019-01-18
Rain and thunder

petchema:
Allows ambient sound effects and ambient sound loops to play at the same time.

Main issue I have with that code is that I needed to duplicate some of DaggerfallAudioSource features (to work correctly with modded sounds), so it'd better be factorized at some point...

(Lost a lot of time because the compiler was silently failing (Monodevelop added "using System" that made "Random." ambiguous) so I kept running the last version of the game that compiled, not understanding why my changes had no effect :( )

### DIFF
--- a/Assets/Scripts/Game/AmbientEffectsPlayer.cs
+++ b/Assets/Scripts/Game/AmbientEffectsPlayer.cs
@@ -11,6 +11,7 @@
 
 using UnityEngine;
 using System.Collections;
+using System;
 
 namespace DaggerfallWorkshop.Game
 {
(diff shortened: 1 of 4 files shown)


petchema:
Thunder is much weaker... is there some AudioSource parameter that I forgot to set?

Interkarma:
I'll take a look when I can. :) I'm about to spin up some new builds. I'll leave this one pending until we've have time to work through it a bit more.

Interkarma:
I love the change and it works well. Are you happy for me to merge?

Congrats on fixing this one too, I know it's been bothering you for a while!

petchema:
I'm glad you like it!
Maybe thunder should be spatialized farther away so it doesn't break if the player moves too much? I haven't tested that, and I'm not sure how to adjust volume correctly for distant sources...

There's also always room for improvement (fix known bugs like crickets with snow weather, maybe use a pool of AudioSources for one shot ambient sounds so they can overlap,...) but those can be discussed in later patches

petchema:
Okay, I guess it's ready for merging

PR #1159, opened 2019-01-19
Match dialogue reactions close to classic

Allofich:
Although https://forums.dfworkshop.net/viewtopic.php?f=24&t=1500 seems that it may have already been fixed by numidum3rd, this also should fix it. It also implements formula and tallying of skill for etiquette and streetwise in dialogue, addresses some TODOs about when responses are used, and compacts the code significantly by cutting out unnecessary things and implementing directly as classic does.

Daggerfall Unity was using a single reaction value for dialogue, but classic calculates two different reactions. First there is a reaction when clicking an NPC that determines if the NPC will talk to you, and the initial greeting they give you, which is affected by the biography reaction roll. This one was basically already working right, and uses values like -20 through 30, with 0 being default.

For deciding the tone that NPCs answer the player and whether they refuse to give info or not, though, a separate value is calculated that is either 0 (dislike player, refuse info), 1 (neutral) or 2 (like player). This is affected by personality, speech tone (etiquette or streetwise), and type of question asked.

It is kind of strange how it's divided like this, with different factors influencing it, but at least it should fix the issue. We could always improve or have some alternate method of determining reaction to the player, but if we do so I think we should try to get classic's behavior in there and work on top of that.

### DIFF
--- a/Assets/Scripts/Game/TalkManager.cs
+++ b/Assets/Scripts/Game/TalkManager.cs
@@ -98,10 +98,13 @@ namespace DaggerfallWorkshop.Game
         const int minVeryLikeReaction = 30;
 
         readonly short[] questionTypeReactionMods = { 10, -5,  0, 0, 0, -10, -5, -5 };
-        readonly ushort[] answersToDirections =     { 7250, 7265, 7280, 7251, 7266, 7281, 7252, 7267, 7282, 7253, 7268, 7283, 7254, 7269, 7284,
-                                                      7255, 7270, 7285, 7256, 7271, 7286, 7257, 7272, 7287, 7258, 7273, 7288, 7259, 7274, 7289};
-        readonly ushort[] answersToNonDirections =  { 7250, 7265, 7280, 7251, 7266, 7281, 7252, 7267, 7282, 7253, 7268, 7283, 7254, 7269, 7284,
-                                                      7260, 7275, 7290, 7261, 7276, 7291, 7262, 7277, 7292, 7263, 7278, 7293, 7264, 7279, 7294};
+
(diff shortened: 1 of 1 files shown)


Allofich:
I realized it was kind of rude of me not to send a mention to @numidium, since this PR undoes most of his recent PR. I'm guessing/hoping, numidium, that you're not too attached to that as it was just a fix for the forum issue linked above, right?

numidium:
No worries, @Allofich. My PR was just a temporary solution until someone figured out how to match it up with Classic. And now you’ve done just that. 👍

Allofich:
@numidium, thanks. @Nystul-the-Magician, yes this is based on reverse-engineering the original game.

An issue with the original game, and this PR, is that since two separate reactions are calculated for 1) starting dialogue/initial greeting and 2) NPC's attitude in answering questions, they can mismatch each other.

Maybe it can kind of make sense if you interpret it as 1 being their "initial impression" of you, then 2 being their impression of you once you actually start having a conversation, and so the result can differ, but in that case I would think that, realistically, when you talk to them a second time their greeting should match their newer impression of you from 2 in the first talk session. I'm fine with just having classic's behavior, though. 

This disconnect between the two reactions, though, does fix the reported bug in the forum (in a different way from how numidium did) since the reaction roll modifier from character generation only affects NPC's willingness to start dialogue/initial greeting and not question answering.

Allofich:
Applied the fix (already in master) for the Underworld social group to the textIDs I got from classic.

Interkarma:
I'm still impressed how you regularly manage to do this. It's just incredible and we're all very grateful to have you.

PR #1160, opened 2019-01-20
Initial implementation of custom doors for buildings

TheLacus:
A pretty basic implementation, but should allow people interested in this field to experiment a bit with towns modding.

### DIFF
--- a/Assets/Scripts/Game/PlayerActivate.cs
+++ b/Assets/Scripts/Game/PlayerActivate.cs
@@ -22,6 +22,7 @@ using DaggerfallWorkshop.Game.Banking;
 using DaggerfallWorkshop.Game.Guilds;
 using DaggerfallWorkshop.Game.MagicAndEffects.MagicEffects;
 using System.Collections.Generic;
+using DaggerfallWorkshop.Utility.AssetInjection;
 
 namespace DaggerfallWorkshop.Game
 {
@@ -240,7 +241,7 @@ namespace DaggerfallWorkshop.Game
                     if (doors && playerEnterExit)
                     {
                         StaticDoor door;
-                        if (doors.HasHit(hit.point, out door))
+                        if (doors.HasHit(hit.point, out door) || CustomDoor.HasHit(hit, out door))
                         {
                             // Check if close enough to activate
                             if (hit.distance > DoorActivationDistance)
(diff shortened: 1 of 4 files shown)


Interkarma:
I love that you've made a start on this. Will have a play with it and merge as soon as I can.

TheLacus:
I'm working on interior to exterior transition but i think i would be more elegant if we could edit StaticDoor array for exteriors instead of adding a separate door system.

Is it possible to convert trigger transform data to meaningful StaticDoor.centre and StaticDoor.size? Maybe what i'm saying doesn't make sense, but i need to do the reverse of this:
`Vector3 centre = doors[i].ownerRotation * doors[i].buildingMatrix.MultiplyPoint3x4(doors[i].centre) + doors[i].ownerPosition;`

If not i'll keep working as i'm doing now, but i'll need to handle saves inside building with custom door positions.

Interkarma:
Sounds reasonable to unify this rather than add another layer. One of the tricks will be that new models/blocks might have more/less doors than the original. And there are combined/uncombined pathways to consider. Perhaps its time just to use trigger colliders in place of my static door array?

I've always felt the static door array was overly complicated and possibly isn't saving me as much time and memory as hoped. It could be the epitome of premature optimisation. :)

What are your thoughts on shifting to a more Unity-like way of doing this?

TheLacus:
My first intention was to integrate additions within current framework but I'm ok with this, it could provide a more future-proof solution. I'll see how to best implement transition to exteriors :)

Interkarma:
I've been playing with replacing static door array with box collider triggers. This has an impact of around 10-12ms per city block, so large cities like Daggerfall take another 160ms or so to layout. It adds several hundred additional GameObjects to scene, which also requires a little bit more RAM. It turns out the static doors setup is more efficient overall.

There are some extra considerations too. Something I'd forgotten is that doors need to carry metadata about the building and interior to support transitions. This would need to be moved to a peer component on the collider object. Definitely possible, but a few other systems plug into this so it would be a somewhat broad change with a lot of risk of introducing regressions.

I think for now it's best just to work within the existing framework as you had intended. I still feel that my original design could be improved, but it's probably not as bad as I thought after looking into it again after a few years. It's just not very flexible - bit your changes are a big step towards improving this.

Are you happy for me to merge this PR and we can keep improving the existing framework for now?

TheLacus:
I'm happy to have the PR merged. The main issue that was raised on forums is that door must be positioned exactly as in classic to trigger checks on collision. This limitation is now solved so custom door  component should be already usable in most situations.

The next step is to use custom positions for interior to exterior transition to allow increased customization. Do you think is viable to edit the StaticDoor array with data created from custom triggers, with some implementation changes, or is better to store them separately ? 

If we want to provide a smooth process on mod changes or uninstall we might also need to follow the example of terrain sampler and store additional metadata to saves made in a building, such as the guid of the mod that owns the door and a version number. Otherwise it could be enough to do a collision check on position after the exterior scene is loaded.

PR #1162, opened 2019-01-23
TextureReplacement: disable emission maps on non-emissive lights

TheLacus:
The light archive contains emissive and non emissive textures. If we want to allow replacing individual textures, we need to load the atlas and override only the imported textures. This includes the emission atlas.


Interkarma:
Thanks for fixes!

PR #1164, opened 2019-01-26
Non-quest classic rumor implementation

Allofich:
Features:
1. Imports non-quest rumors from classic, including their time limit, flags, etc. Quest rumors are not imported since this is pointless unless we also import quest status from classic saves.
Note that due to a mistake in classic (which I've mentioned before), non-quest rumors are very rarely updated in classic and you will often not have any in saved games where you have progressed in the game. You can be guaranteed to have non-quest rumors for import if you make a new character, for about the first game month until the rumors expire.

2. Non-quest rumors now have time limits, the same as those in classic. They last for about a month.

3. Flag data from classic data recreated. This has things set up for signboard messages, whenever clicking on signboards becomes supported.

4. Non-quest rumors are now created for the various background events that occur, same as classic. Several of these are signboard messages and can't be seen yet.

5. Some simplification and correction to macro resolution for rumors. Rumors store the relevant factions and region IDs at the time of their creation in classic, and this is recreated now.

6. Some rumors that are always created on every background events update in classic (about witches, daedra, vampires, thieves guild and dark brotherhood, artifacts, etc.) and were probably supposed to always be available, are created now.

When you first load a saved game, you may have to first do some fast traveling and get the background events update to fire before the new non-quest rumors are generated. Until then NPCs will just say "I try not to spread rumors", etc. This should only be an issue until new saves are made with the updated rumor data.

### DIFF
--- a/Assets/Scripts/API/MapsFile.cs
+++ b/Assets/Scripts/API/MapsFile.cs
@@ -44,7 +44,7 @@ namespace DaggerfallConnect.Arena2
         /// <summary>
         /// All region names.
         /// </summary>
-        private readonly string[] regionNames = {
+        private static readonly string[] regionNames = {
             "Alik'r Desert", "Dragontail Mountains", "Glenpoint Foothills", "Daggerfall Bluffs",
             "Yeorth Burrowland", "Dwynnen", "Ravennian Forest", "Devilrock",
             "Malekna Forest", "Isle of Balfiera", "Bantha", "Dak'fron",
@@ -209,7 +209,7 @@ namespace DaggerfallConnect.Arena2
         /// <summary>
         /// Gets all region names as string array.
         /// </summary>
-        public string[] RegionNames
+        public static string[] RegionNames
         {
             get { return regionNames; }
         }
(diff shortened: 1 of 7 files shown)


Allofich:
Fixed a bug from master, where all rumors were being processed as quest rumors on loading a DF Unity save in the check to remove orphaned quest rumors.

Allofich:
Just to clarify, the main thing this does, is that while in master branch non-quest rumors about new rulers, etc. are created, they are done randomly. With this PR, they are created as in classic and correspond to actual background events. Still mostly just a flavor thing.

Also, you'll get a lot of rumors about factions allying or becoming enemies. I'm not sure but these might only show in classic if the player is related to one of the factions (a member, or they are an ally of a faction the player is a member of, etc.) but you would almost never see them then, so this might be preferable.

Nystul-the-Magician:
cool stuff!
do I guess right that your implementation in dfunity won't do the same mistake as classic regarding the very rare update of rumors? ;)

Allofich:
That's right, I didn't replicate that mistake.

Interkarma:
Incredible work as always. Features like this help make up Daggerfall's background world, and we'd be missing it completely without you. Cheers! :)

PR #1165, opened 2019-01-26
Improve AI pursuit

Allofich:
1. Adjusted enemy controller height limit down to 1.65f. I had to limit it to this for the enemy thief in Privateer's Hold to really be able to get through the doorway every time without hitching up for a second .

2. Fixed problem with enemies refusing to follow player down staircase corridors because the obstacle check would see the sloping wall in front of them as an obstacle and they would stop moving downwards.

3. Enemies will only try to get to Occasionally I saw enemies wandering off down a wall looking for a way to get to me. Also they would often not follow the player well around corners due to the prediction logic. They now use the last place they saw their target if they don't have a clear path to the predicted spot, which addresses both of these.

### DIFF
--- a/Assets/Scripts/Game/EnemyMotor.cs
+++ b/Assets/Scripts/Game/EnemyMotor.cs
@@ -877,7 +877,8 @@ namespace DaggerfallWorkshop.Game
             tempMovePos = transform.position + testMove.normalized * 2;
             tempMovePos.y = transform.position.y;
 
-            avoidObstaclesTimer = 0.25f;
+            if (avoidObstaclesTimer == 0)
+                avoidObstaclesTimer = 0.25f;
             lastTimeWasStuck = Time.time;
             moveInForAttack = true;
         }
--- a/Assets/Scripts/Game/EnemyMotor.cs
+++ b/Assets/Scripts/Game/EnemyMotor.cs
@@ -360,7 +360,7 @@ namespace DaggerfallWorkshop.Game
                 tempMovePos = targetPos;
                 targetPosIsEnemyPos = true;
             }
-            else if (avoidObstaclesTimer == 0 && ClearPathToPosition(senses.LastKnownTargetPos))
+            else if (DaggerfallUnity.Settings.EnhancedCombatAI && avoidObstaclesTimer == 0 && ClearPathToPosition(senses.LastKnownTargetPos))
             {
                 targetPos = senses.LastKnownTargetPos + (senses.LastPositionDiff * 2);
                 tempMovePos = targetPos;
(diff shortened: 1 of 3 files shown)


ajrb:
Is this for advanced enemy AI, or classic? Or both?

Allofich:
Looks like I messed up classic AI's behavior when losing a direct path to the target. Will look into that.

Allofich:
Fixed another issue, where `avoidObstaclesTimer` could be repeatedly reset to full as AI is walking against a wall, and they would ignore their target. Now `avoidObstaclesTimer` must be allowed to return to 0 before being reset, which gives a chance for the AI to resume pursuit if a path to the target opens up.

ajrb:
> 1 and 2 apply to both. 3 also applies to both but won't have an affect on "classic AI" because it's what it was already doing. It's a little wasteful in classic AI mode because a spherecast check will be done to both the predicted spot and the last spot the target was seen, which are the same thing in classic AI mode. I could make one or the other only run if in enhanced AI mode, although it may look a little messy to do so. I'm sure the whole thing has optimization and reorganizing that should be done, so I should maybe do that all at once.
> 
That's perfectly reasonable. Sometimes it's better to keep code simpler and easy to read than shave an irrelevant few ms off that no one will ever notice.


> The classic AI mode is not a complete recreation of classic at the moment, though. Obstacle and fall avoidance are the same as the enhanced AI. Real classic AI has a simple obstacle avoidance routine. It barely does anything in the original game because the AI hits an obstacle and uses the avoidance, I think it tends to immediately bump into 3d geometry again, ending the avoidance movement, so it ends up not moving. I think it should work a little better if recreated in DF Unity. Also, to truly be like classic AI I think that AI should just make a beeline for the target even after they have gone behind a wall or something, not use the last spot the target was seen as both modes do in DF Unity now, although this makes them much worse at pursuit.
> 
I think this is a good compromise, if it feels like classic but AI diesn't get stuck or act moronic when getting to the player then that's perfect.

> Just for the fun of recreating it, I am kind of interesting in trying to make classic AI mode as close to the original as possible, including the simpler pursuit and obstacle avoidance, but I haven't done that yet.

That's up to you, but I would not bother. :-)

Allofich:
> Sometimes it's better to keep code simpler and easy to read than shave an irrelevant few ms off that no one will ever notice.

I did actually end up adding that optimization in. The whole AI targeting and tracking system probably ought to be redone at some point, but at least for now everything seems to be working decently enough and I'm okay for this to be merged.

> That's up to you, but I would not bother. :-)

I might not bother, I don't know yet.

Interkarma:
Great work, thank you!

PR #1170, opened 2019-01-26
Prevent wagon access from inside dungeons during identification

petchema:
Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=1756


Interkarma:
Thank you! :)

PR #1171, opened 2019-01-27
Fix sky blind spot

petchema:
Forums: https://forums.dfworkshop.net/viewtopic.php?p=19983#p19983
The root issue seems to be that transform.eulerAngles.y can become slightly negative (not by much, I've seen up to -0,000563718).


Interkarma:
Cheers for fix! :)

PR #1172, opened 2019-01-27
don't make sure the player has a spellbook

petchema:
Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=1738


Interkarma:
Cheers! Looks like I went overboard there making sure player has a spellbook.

Here, have a spellbook. :)

PR #1173, opened 2019-01-27
do not always count pacified NPCs toward enemies

petchema:
Allows for fast travel, sleeping,... when only pacified NPCs are around

Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=1752


petchema:
I haven't checked what's classic Daggerfall behavior

Interkarma:
I'm happy with that change, thank you. :)

PR #1174, opened 2019-01-27
Compact notebook messageboxes

petchema:
Current fast-travel from notebook message box is very large, and requires high resolution to fix in the screen (i think 1280x720 is already too small, eg https://www.youtube.com/watch?v=1k0G_VEv-DM&t=769 ):
- both location and region names are in the same text line;
- the explanation that it won't fast-travel immediately is also very lengthy.

This patch puts location name and region name on two separate lines, and replaces the explanation with a different wording which should be (I hope) sufficient.
I'm not sure text wrapping would be preferable. Wording can be discussed.

I also removed one blank line from notebook edition message boxen.


ajrb:
See #1175 which supercedes this PR.

PR #1175, opened 2019-01-28
Make journal notebook dialog boxes more compact (supercedes #1174)

ajrb:
@petchema Sorry I didn't like the changes you made, so this PR is the way I would like to change it instead and supercedes #1174. I hope you don't mind, I mean no disrespect - your contributions are very valuable.

Anyway, you definitely demonstrated that I'm bloody awful at getting around to stuff I don't personally care about... :embarrassed: I honestly had meant to get around to this.. at some point before the heat death of the universe.

Anyway I've tested at 320x200 and up, but your feedback is welcome.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallMessageBox.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallMessageBox.cs
@@ -304,6 +304,14 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             SetTextTokens(tokens, mcp);
         }
 
+        /// <summary>
+        /// Must be set before text otherwise layout has already occurred.
+        /// </summary>
+        public void SetHighlightColor(Color highlightColor)
+        {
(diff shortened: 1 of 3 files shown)


ajrb:
The reason for the titles is to allow quick visual parsing of what action the dialog is about for those in a hurry and are not reading the longer text for the 100th time. To help avoid accidental actions since the dialog boxes are consistent, which means they look the same if you don't read the words. People tend to quickly respond to frequent & familiar dialogs so I felt there should be a short version as well as long version of the action text.

petchema:
Yeah, I see the benefit of titles, in fact in GUIs they're usually part of the layout so they're both required and easy to locate.
Centering looks good. The only remaining problem is that we'd have to use titles consistently for all prompts...

ajrb:
They are, for the new prompts added in the Notebook / Journal upgrade. We are NOT going to start messing with classic prompts! Ok. Nope. Do not go there. LOL. :-)

However, if I have missed/forgotten about new DFU only prompts please remind me... I am awfully forgetful.

PR #1176, opened 2019-01-30
Update to Mod Builder Window

Lypyl:
Added compression option selection,  collect dependencies button and improved appearance.  Also removed legacy code no longer used by the mod system.


Interkarma:
Awesome, thank you! :)

PR #1178, opened 2019-01-31
 Fix temples selling spells to non-members 

ajrb:


Interkarma:
Awesome, thank you mate! :)

PR #1180, opened 2019-02-01
TextureReplacement: Support for TRAV0I00.IMG

TheLacus:
@Interkarma @Lypyl are you ok with this change? I moved the region overlay to its own panel to draw it above the map texture at any resolution. I think we will also need to allow high resolution overlays but the default one should be available as a fallback. It might even look fine as is with texture filter. Should i link it to `Settings.GUIFilterMode`?


Interkarma:
Yep, I'm happy with the change. Cheers. :)

PR #1183, opened 2019-02-04
Reset guild memberships and region reps upon becoming vamp

ajrb:
Also changed it so quests only get offered after completing initiation quest.

Sorry it took me so long to get around to this - penny hadn't dropped that it was ready.

I don't usually do this, but these changes are pretty simple and I'm tired so I've done zero testing. Need to figure out how to test this. Putting it up here in case someone else fancies testing it for me. (@petchema or @Interkarma) I will do it when I get a chance, but work has been quite intense recently so it's hard to predict when.

### DIFF
--- a/Assets/Scripts/Game/Guilds/Guild.cs
+++ b/Assets/Scripts/Game/Guilds/Guild.cs
@@ -264,6 +264,8 @@ namespace DaggerfallWorkshop.Game.Guilds
                 case GuildServices.Identify:
                     return true;
                 case GuildServices.BuySpells:
+                    return false;
+                case GuildServices.BuySpellsMages:
                     return true;
                 case GuildServices.Donate:
                     return true;
--- a/Assets/Scripts/Game/Guilds/MagesGuild.cs
+++ b/Assets/Scripts/Game/Guilds/MagesGuild.cs
@@ -140,7 +140,7 @@ namespace DaggerfallWorkshop.Game.Guilds
                     return true;
                 case GuildServices.Identify:
                     return true;
-                case GuildServices.BuySpells:
+                case GuildServices.BuySpellsMages:
                     return true;
                 case GuildServices.MakeSpells:
                     return IsMember();
--- a/Assets/Scripts/Game/Guilds/Services.cs
+++ b/Assets/Scripts/Game/Guilds/Services.cs
@@ -30,6 +30,7 @@ namespace DaggerfallWorkshop.Game.Guilds
         MakePotions,
 
         BuySpells,
+        BuySpellsMages,
         MakeSpells,
 
         BuyMagicItems,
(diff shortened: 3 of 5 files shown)


petchema:
The main problem I'd have to test such features is that I never played as a vampire in classic, so I'm not sure what to expect...

Allofich:
With this change will guild memberships be restored to what they were just before the player turned into a vampire when the vampirism is cured? You removed a comment about that so I'm guessing maybe not?

I wrote my understanding of what classic does a while ago in the comments here:
https://github.com/Interkarma/daggerfall-unity/commit/1dc5ebeb6024af9334f61ce592db243747eb91bd

I don't know that classic actually resets legal reputation when you become a vampire (it might, but I haven't tested) but it does seem consistent with losing guild memberships.

Allofich:
Although, if we restore guild memberships after curing vampirism like classic, it would also make sense to restore the legal reputations to pre-vampire.

While classic has an object type for backed up guild memberships, I'm not aware that it has any for backing up the legal reputations, which is part of what makes me suspect that maybe they aren't changed by vampirism in classic.

Interkarma:
Thanks for the feedback @Allofich. I'll check legal rep in classic when I can by infecting a character and observing what happens to their legal standing.

The way DF handles guild ranks here is interesting. With insta-promotion 28 days later to old rank (because guild faction rep is the same), it seems a bit redundant to even go through all that data manipulation for vampirism at all.

And considering player can permanently lose access to thieves and dark brotherhood, there's a case to made that guild reset when becoming a vampire is really just a broken feature in classic. It's still amazing how you work this stuff out though. :)

@ajrb I'll hold off merging this one for now until legal rep reset confirmed. Maybe we need a more robust debate around whether guild rank reset should even take place. It seems to add problems like permanently losing TG & DB without any real thematic benefit.

Allofich:
Do you not get TG and DB membership back when you are cured in classic? I suspect you do, as it seems to restore a backup snapshot of your memberships. But, if you first join as a vampire, you wont be able to redo initiation after being cured. You would have to become a vampire again to toggle back to the vampire memberships (I think). Though, I agree, seems like an incomplete feature.

ajrb:
> @ajrb I'll hold off merging this one for now until legal rep reset confirmed. Maybe we need a more robust debate around whether guild rank reset should even take place. It seems to add problems like permanently losing TG & DB without any real thematic benefit.

Yeah that's very sensible. Could we debate this here please: https://forums.dfworkshop.net/viewtopic.php?f=20&p=20719

That's why I created the thread anyway. I stupidly did this on my master branch so will have to sort that out so I can work on something else.

ajrb:
Just realised this will knacker this PR. I should have created a branch. Will close this PR probably

ajrb:
Just realised this will knacker this PR. I should have created a branch. Will close this PR

PR #1184, opened 2019-02-07
Clean up EnemyMotor

Allofich:
EnemyMotor, although generally working okay, had become somewhat of a mess as I've been adding and changing bits of it. I went back over the code execution and made several small cleanups, fixes and a few optimizations.

Fix up MakeEnemyHostileToAttacker()
Simplify rotating check
Use more accurate variable names
Don't do bow or spell attack checks when not available
Remove pointless variable assignments
Remove pointless normalizations
Remove pointless move() call
Move clockwiseCheck reset to FindDetour() so it's only done when needed
Correct assumed spell projectile speed
Use do-while loop in FindDetour so operations done in correct order
Fall check assumes obstacle check was just done, so do before fall check
Use wider angle for detecting doors to open so AI can open them more
easily

### DIFF
--- a/Assets/Scripts/Game/EnemyMotor.cs
+++ b/Assets/Scripts/Game/EnemyMotor.cs
@@ -59,10 +59,12 @@ namespace DaggerfallWorkshop.Game
         bool didClockwiseCheck;
         float lastTimeWasStuck;
         bool bashing;
+        bool hasBowAttack;
+        bool hasSpells;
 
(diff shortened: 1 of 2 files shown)


ajrb:
Nice cleanup, maybe not such a glamorous task but very good to do this!

Interkarma:
Just seconding Hazelnut's comment. Cheers. :)

PR #1185, opened 2019-02-08
Use a single timer for classic update simulation

Allofich:
Combines all the classic timer updating I had put into separate files into a single timer in GameManager, since only one is needed.

If GameManager isn't where you would want this, I can move it somewhere else.

Should merge cleanly with https://github.com/Interkarma/daggerfall-unity/pull/1184.


Interkarma:
Great idea, well done! I'm happy to have this in GameManager too. :)

PR #1187, opened 2019-02-11
Implement addArtifact command

Allofich:
Implements an "addArtifact" command that can be used at the console to add any of the game's artifacts to the player's inventory.
Ex. "addArtifact 0" adds the artifact of ID 0, the Masque of Clavicus Vile

The reason is to help in implementing and testing artifact effects.

Also fixed a few issues with the existing function to create artifacts.


Interkarma:
Great idea for testing!

PR #1188, opened 2019-02-12
warn the player before taking back an item not fully repaired

petchema:
Some people seem so used to instantaneous repairs that they get their items back before they're repaired; A warning seems appropriate.
Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&t=1765&p=21086#p21086

### DIFF
--- a/Assets/StreamingAssets/Text/DaggerfallUI.txt
+++ b/Assets/StreamingAssets/Text/DaggerfallUI.txt
@@ -38,6 +38,7 @@ wagonFullGold,      Your wagon could only hold {0} gold pieces.
 itemsIdentified,    Items identified.
 repairDone,         DONE
 repairDays,         %d days
+interruptRepair,    Take back that item before it's repaired?
 repairNote,         Left my {0} for repair at {1}.
 doesntNeedIdentify, This does not need to be identified.
 stealSuccess,       You are successful.
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTradeWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTradeWindow.cs
@@ -98,6 +98,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
         bool usingWagon = false;
         int cost = 0;
         bool usingIdentifySpell = false;
+        DaggerfallUnityItem itemBeingRepaired;
 
         static Dictionary<DFLocation.BuildingTypes, List<ItemGroups>> storeBuysItemType = new Dictionary<DFLocation.BuildingTypes, List<ItemGroups>>()
         {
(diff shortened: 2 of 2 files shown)


Assets/StreamingAssets/Text/DaggerfallUI.txt
```diff
@@ -38,6 +38,7 @@ wagonFullGold,      Your wagon could only hold {0} gold pieces.
 itemsIdentified,    Items identified.
 repairDone,         DONE
 repairDays,         %d days
+interruptRepair,    Take back that item before it's repaired?
```
petchema: Using commas in the message would break parsing :(

petchema:
> Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&t=1765&p=21086#p21086
There's a discussion there about whether it should be even allowed to take items back before they're finished repairing

Interkarma:
I've been thinking about this one for a few days. The classic intent seems to be for item to be unavailable during repair window. Within the game's fantasy, the hilt is likely removed while blade soaks in acid before repair and resharpening (I genuinely have no idea how these repairs would work, but you get the idea). It's probably not in a practical state to hand back to the customer.

But another part of me likes being able to take back item and forfeit repairs for their troubles (and maybe to put the hilt back on) so you can take your property back no worse for the shop time.

I personally prefer instant repairs, but I like having the choice of delayed repairs like classic. I also like having the choice of getting my gear back sooner if I want it. So I don't mind for this to merged.

But I'm curious how @ajrb would like to handle this as the original author of repair system? Are you happy with this, or would you prefer to handle a different way?

ajrb:
I thought I'd implemented as classic, but I've not re-checked. Happy for you to make the call on this one.

petchema:
Just to add to the confusion:

- Fantasy: repairing takes no time
- Low Fantasy: you can get your stuff back before it's ready, but forfeit the repair price
- Daggerfall: you can't take your stuff back until it's repaired. Unless you exploit an oversight
- Arena: you can pay more to get the repair done faster ("I'll be working on it right away, Sir"), or less if you don't mind waiting (or most likely can't afford full price). I assume you then cannot take your stuff back until repair (I haven't played Arena in a long, long time).
- Simulation: if it's a common steel sword, the blacksmith gives back a spare, there, instant repair. Otherwise announced repair time is just a (good) approximation. If you take back your broken weapon during the first hour or two, blacksmith grumbles but gives your stuff back. If time elapsed is below 50% or above 90% of repair time, you get back your stuff but repair price is forfeit. Otherwise the weapon is being worked on, and you can't take it back. If you insist, you're given your weapon back, but in a worse shape than before, and price is still forfeit. etc.

Maybe we could go for the simplest fix (don't allow stuff to be taken back before repair), and leave more involved implementations to mods? Arena way is very interesting, too.

Allofich:
> But some people seem to used to instantaneous repairs, that the "1 days" to repair was not interpreted and he took his weapon back without a second thought; and did not notice that the weapon was not actually repaired either.

If the problem is people taking back their item instantly because they don't know it's not repaired yet, maybe allow canceling and taking back as long as you're still in the inventory screen. (Or if you want to go for a little more realism, until an hour of game time has passed or something) After that, the player should know what's going on, so we could apply the rule that they can't take it back while it's being repaired.

petchema:
After watching a stream, I think that one thing missing is the display of repair times _before_ you commit to repair your items.
It doesn't seem unrealistic that you'd get that kind of information before accepting a repair

PR #1181, opened 2019-02-03
TextureReplacement: Support for FMAPXXXX.IMG

TheLacus:

PR #1189, opened 2019-02-13
Travel map update

Lypyl:
Minor bug fixes and functional improvements to the travel map & travel pop up window.

Travel popup window now shows travel option toggles correctly regardless of aspect ratio.  

Corrected a bug that was causing the find location routine from breaking the at button in the region view.  

The location list picker should now only displays locations known to the player.

Changed scrolling mode while zoomed in to be a toggle on/off.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTravelPopUp.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTravelPopUp.cs
@@ -35,12 +35,9 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
         TravelTimeCalculator travelTimeCalculator = new TravelTimeCalculator();
 
         Color32 toggleColor = new Color32(85, 117, 48, 255);
+        Color32 notToggledColor = new Color32(130, 22, 0, 255);
 
         Panel travelPanel;
-        Panel speedToggleColorPanel;
-        Panel transportToggleColorPanel;
-        Panel sleepToggleColorPanel;
-
         Button beginButton;
         Button exitButton;
         Button speedToggleButton;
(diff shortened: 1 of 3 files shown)


Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTravelPopUp.cs
```diff
@@ -213,25 +193,51 @@ public override void Update()
...
+                UpdateToggle(116, 67,4,4,notToggledColor);      //foot/horse
+            }
+            else{
+                UpdateToggle(116, 57, 4,4, notToggledColor);
+                UpdateToggle(116, 67,4,4,toggleColor);
+            }
+        }
+
+        //updates texture based on travel options selected
+        void UpdateToggle(int startX, int startY, int width, int height, Color32 color)
```
TheLacus: Unfortunately this breaks texture replacement. Imported textures can have higher resolution than original image so we can't change pixels colors in a reliable way. The workaround is to scale images individually and draw them in the correct order, which is what was already being done here.

TheLacus: I want to add that you can also draw all checkboxes on a single overlay texture with transparent background, what's important is that native texture is not edited because it can have any resolution and format if imported from mods. See #1180 and #1181.

Interkarma: We definitely want to keep the base image modable, thank you for picking this @TheLacus. 👍

Interkarma:
Thanks for continuing work on this one! :)

Interkarma:
Hi @Lypyl mate! Just a few conflicts in this one now due to other commits, and we should look at the modability of drawing toggles. Sorry to keep this one open so long without merging.

Interkarma:
Just closing for now, as no response to queries or cleaning up conflicts. Happy to reopen any time though when ready. :)

PR #1191, opened 2019-02-13
warn the player when paid with a letter of credit

petchema:
Players not familiar with the concept can be very confused
Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&p=21119#p21119


Interkarma:
I quite like this! Sensible and shouldn't be controversial at all. Cheers. :)

PR #1192, opened 2019-02-13
display letters of credit values in inventory lists

petchema:
This make letters of credit stand out compared to normal letters and recipes.
Since letters of credit management is automated, this is otherwise of little practical usefulness.

Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&p=21119#p21119


petchema:
http://ks369561.kimsufi.com/~petchema/gfx/Daggerfall/letters%20of%20credit.jpg

petchema:
On second thought it doesn't look very good in inventory...

Interkarma:
Agreed, it doesn't look great in the inventory like that. Perhaps as a compromise, it could just update the tooltip. Being a tooltip, it could easily show the non-abbreviated value (e.g. "Letter of credit (187184 gold)") without any extra clutter. This is possibly a bit redundant with the info panel already showing this information though - unless user prefers info panel disabled and likes tooltips.

PR #1193, opened 2019-02-14
popup text smooth scrolling

petchema:
Make popup text scroll up smoothly instead of line by line; The text stays easily readable while being removed from screen.

### DIFF
--- a/Assets/Scripts/Game/UserInterface/PopupText.cs
+++ b/Assets/Scripts/Game/UserInterface/PopupText.cs
@@ -27,7 +27,8 @@ namespace DaggerfallWorkshop.Game.UserInterface
         const int maxRows = 7;
 
         LinkedList<TextLabel> textRows = new LinkedList<TextLabel>();
-        float timer = 0;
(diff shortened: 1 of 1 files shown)


Interkarma:
This is really quite lovely! Having played with this for a while, I can say I much prefer it to the step-scrolling. I feel that I'm more likely to notice important information now thanks to the movement catching my eye.

It's possible we'll receive some requests for an option to restore step-scrolling. I've noticed several people prefer to play with classic fonts and even at lower resolutions. I'm happy to merge without that option for now, but it's worth keeping in mind for later. :)

petchema:
Well, classic doesn't scroll at all, it displays messages using some kind of circular buffer algorithm anyway.

On a different topic, I just noticed than in 4:3 resolutions the pop texts do not show up at the very top of the screen, so they don't scroll off-screen. I'm not sure if that was intentional, I think it clutters vision and should better be at the top of the screen always...

Interkarma:
I hadn't noticed the issue in 4:3 either, it's not intentional.

PR #1195, opened 2019-02-18
Make enemy height change only temporary

Allofich:
The changes I made to enemy collider height to allow them to fit through doors caused an issue where the upper part of the enemy would no longer respond to collisions such as arrows or clicking on them for the "You see a %s" message. The taller the enemy the more noticeable the problem would be. It was particularly noticeable with dreughs.

The only solution that came to mind is an imperfect one, but better than what we have currently. That is to only temporarily change collider height for half a second when an enemy brushes up against something. It will then return to normal if after the half-second the enemy is no longer touching anything.

This fixes the dreugh part of
https://forums.dfworkshop.net/viewtopic.php?f=24&t=1751

### DIFF
--- a/Assets/Scripts/Game/EnemyMotor.cs
+++ b/Assets/Scripts/Game/EnemyMotor.cs
@@ -59,6 +59,10 @@ namespace DaggerfallWorkshop.Game
         bool bashing;
         bool hasBowAttack;
         bool hasSpells;
+        float realHeight;
+        float centerChange;
+        bool resetHeight;
+        float heightChangeTimer;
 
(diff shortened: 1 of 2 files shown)


Allofich:
If you want to see visually what is happening, you can comment out the `controller.center = newCenter;` lines adjusting the center. Then the height-adjusted enemies will sink into the ground when their height is adjusted, and you can easily see as the adjustment is turned on and off.

Interkarma:
I like this solution! Will merge it and we can keep play testing and watch for feedback.

PR #1196, opened 2019-02-18
Make arrows use colliders for hit checks

Allofich:
Combined with https://github.com/Interkarma/daggerfall-unity/pull/1195, this fixes the other issue in https://forums.dfworkshop.net/viewtopic.php?f=24&t=1751 (about the bats).

1. Use colliders instead of ray casts for arrows.
- Enemies pushed around ("Jedi reflexes" of the bug report) less
- Shooting enemies at point-blank range will register hits better now (will still go through them if they are right next to you, but that doesn't bother me too much as it is unrealistic to shoot from that close anyway)
- Maybe a little better performance?, since instead of doing raycasts the collider that existed on the arrows is used.
- Maybe more reliable hit detection, although the ray-cast solution seemed to be working for me except at point-blank ranges.

2. Moved arrow disabling and damage to be called immediately on collision instead of in the update loop
- Together with the above, seems to have (possibly) fixed completely the issue of enemies being pushed by the arrow

3. Can no longer bash doors with arrows, as it didn't make much sense

### DIFF
--- a/Assets/Prefabs/Missiles/ArrowMissile.prefab
+++ b/Assets/Prefabs/Missiles/ArrowMissile.prefab
@@ -25,6 +25,7 @@ GameObject:
   - component: {fileID: 108732554255048532}
   - component: {fileID: 114624404217668130}
   - component: {fileID: 54102960107042576}
+  - component: {fileID: 64298782196861766}
   m_Layer: 0
   m_Name: ArrowMissile
   m_TagString: Untagged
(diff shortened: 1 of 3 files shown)


Allofich:
Not sure if it really mattered, but arrows' 3d objects are destroyed now, not just disabled.

Also, maybe this PR will help with this issue: https://forums.dfworkshop.net/viewtopic.php?f=24&t=1811. I've never had that issue happen before, but maybe using the collider will more reliably clean up the arrows.

Interkarma:
Thank you!

Allofich:
Should have mentioned, during testing this it seemed like the forced crouch from enemy arrows contacting with player may be exacerbated with this PR.

Interkarma:
I have a fix in pipeline for that, but probably won't be ready for this weekend's builds. I'll see if I can add something faster.

PR #1200, opened 2019-02-20
Prevent paying repair price several times

petchema:
cost only takes into account items that are not already being repaired
Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=1828

If they're only already repaired items in the remote list, test L721 does not prevent a display a "it will cost 0 gold" trade window in you click on REPAIR button. The same situation happens with identification, so this test could be improved, but I don't understand that code well enough.

Also, I don't understand how repair price seems correct in game (equal to "cost", as computed by UpdateCostAndGold()), when the deduced amount L765 is the value of GetTradePrice() that does not handle the case windowMode == WindowModes.Repair


petchema:
Ok,  that should fix the shortcomings mentioned above

Interkarma:
Awesome, thank you!

PR #1201, opened 2019-02-21
highlight recently increased skills

petchema:
When opening the character sheet, highlight all the skills that have
increased since the last time you opened the character sheet.

Exclude character sheet opening automatically triggered by level ups.

Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&t=1834

### DIFF
--- a/Assets/Scripts/Game/Entities/DaggerfallSkills.cs
+++ b/Assets/Scripts/Game/Entities/DaggerfallSkills.cs
@@ -72,9 +72,6 @@ namespace DaggerfallWorkshop.Game.Entity
         // This is designed so that effects are never operating on permanent skill values
         int[] mods = new int[Count];
 
-        // Skills to highlight
-        bool[] recentlyIncreased = new bool[Count];
-
         #endregion
 
         #region Constructors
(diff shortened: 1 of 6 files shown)


petchema:
This lacks serialization...

Allofich:
This data is also in classic saves, as `skillsRaisedThisLevel1` and `skillsRaisedThisLevel2`, if you want to import it.

Both are 32-bit `uint` values, but `skillsRaisedThisLevel2` only covers the last 3 skills, Blunt Weapon (32), Archery (33), and Critical Strike (34), so only the first 3 bits are used.
If a bit is 1, that means the skill will be highlighted when the skill sheets are opened.

`messageBox.SetHighlightColor(Color.white);`

The color used by classic is not white, it is a light blue color. My opinion is that for such artistic decisions we should offer the same as classic as default, and then offer ability to change it if we don't like classic's way. Though, it's just a minor feature, so if you and Interkarma want to use a different color like white, I'm okay with that.

Allofich:
By the way if you don't want to import the classic data I think that's okay for this PR and no one would mind. It's just a bit of polish that would be nice to have.

petchema:
> This data is also in classic saves, in `CharacterRecordData` and `CharacterDocument` as `skillsRaisedThisLevel1` and `skillsRaisedThisLevel2`, if you want to import it.

Okay, so DFU should probably support serialization too. I must admit I don't really know yet how it works :(

By just adding [SerializedField] annotation, it was serialized as a JSON array of booleans, but it was not restored correctly (or was reset after load). Also, would a list of skill names be more future proof if skills become moddable?

petchema:
> You used `messageBox.SetHighlightColor(Color.white);` but the color used by classic is not white, it is a light blue color.

I changed the color to the one used for buffed attributes, I assume that's the one you're talking about? I think it looks better now

Interkarma:
This works well, cheers!

PR #1202, opened 2019-02-21
notebook page to log onscreen messages (popup, middle screen)

petchema:
This is just a log of recent onscreen messages; The number of messages is bounded, and they are not persisted. The goal is just to provide a way to read messages that went by too fast.

Any way to get a more dense display?
Once text density is fixed, it would be great to set the maximum number of messages to exactly one screen, or to open the messages log to the last page.


Interkarma:
This is a very useful feature for those "wait, what did that say?" moments. Great work!

PR #1203, opened 2019-02-23
random Random.Range() fixes

petchema:
Upper bound is exclusive for Random.Range(int, int), leading to a few off-by-one bugs.
Solution is often to increase upper bound by one, in a few place Random.Range(float, float) is used instead.
There's an obvious but mild risk of introducing or uncovering bugs, even if I tried to be careful and started testing those changes.

Also refacto (separate commit): as was the case in some places already, when the code really means to use an inclusive upper bound, rewrite as Random.random(lowerbound, upperbound + 1) even if upperbound + 1 is a constant.
Any compiler worth its salt will optimize that away, and it conveys programmers intent better.

### DIFF
--- a/Assets/Scripts/Game/Items/LootTables.cs
+++ b/Assets/Scripts/Game/Items/LootTables.cs
@@ -166,7 +166,7 @@ namespace DaggerfallWorkshop.Game.Items
 
             // Random weapon
             chance = matrix.WP;
-            while (Random.Range(0, 100) < chance)
+            while (Random.Range(0, 100 + 1) < chance)
             {
                 items.Add(ItemBuilder.CreateRandomWeapon(playerEntity.Level));
                 chance *= 0.5f;
@@ -174,7 +174,7 @@ namespace DaggerfallWorkshop.Game.Items
 
             // Random armor
             chance = matrix.AM;
-            while (Random.Range(0, 100) < chance)
+            while (Random.Range(0, 100 + 1) < chance)
             {
                 items.Add(ItemBuilder.CreateRandomArmor(playerEntity.Level, playerEntity.Gender, playerEntity.Race));
                 chance *= 0.5f;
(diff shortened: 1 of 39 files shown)


Assets/Scripts/Game/Items/LootTables.cs
```diff
@@ -166,15 +166,15 @@ public static DaggerfallUnityItem[] GenerateRandomLoot(LootChanceMatrix matrix,
...
             {
                 items.Add(ItemBuilder.CreateRandomWeapon(playerEntity.Level));
                 chance *= 0.5f;
             }
 
             // Random armor
             chance = matrix.AM;
-            while (Random.Range(0, 100) < chance)
+            while (Random.Range(0, 100 + 1) < chance)
             {
```
Allofich: I think these should be as they are, in order for the percent chances of the items spawning to work as expected. For example, in order for 100% spawn chance to work 100% of the time.

Allofich:
Looks like there are both cases where we use a condition like:

`Random.Range(0, 100) < chance`

and cases where we use

`Random.Range(1, 101) <= chance`

Would be tidiest to use the same one in all cases, although not a big deal.

Assets/Scripts/Game/Items/LootTables.cs
petchema: You're right, if "chance" is a percentage of success, then a correct way to check success is
Random.Range(0, 100) < chance
or Random.Range(1, 101) <= chance

and consequently, to check for failure,
Random.Range(0, 100) >= chance
or Random.Range(1, 101) > chance

Only checks that do not match those cases are suspicious.
Ok, I need more time to work on that.

petchema:
Yes, that confused me

Allofich:
> What about wrapping that logic in a function?
Hmm, not sure. I'll wait for @Interkarma to give his opinion.

Interkarma:
This is a very nice review, thank you. I've read changes multiple times, they are fairly clear-cut and should not cause any issues. I've also play-tested everything I could and did not encounter any obvious problems. I have no problems merging this one.

PR #1204, opened 2019-02-23
Implement strafing for enhanced combat AI

Allofich:
Adds strafing to enhanced combat AI. I've tried to keep it somewhat subtle, so that it will look good and not too game-like. Used in both melee and ranged combat.

Also, this addresses an issue with enemies turning their back to their opponent too much in combat due to trying to path around an obstacle when backing up. Although it looked kind of good like they were "running away" after being backed against a wall it happened a bit too much and would only be appropriate for a weaker or more cowardly combatant, not for all cases. They can now get out of this situation by strafing, which allows them to keep facing their opponent.


Allofich:
Fixed merge conflicts.

Interkarma:
Have played with this one a bit and I like it very much! The strafing is perfectly minimal, just enough to keep me on my toes. You've done an excellent job here.

PR #1205, opened 2019-03-01
Prevent arrows from colliding more than once

Allofich:
With the latest code I noticed an exception when `AssignBowDamageToTarget` in `DaggerfallMissile.cs` tried to use the destroyed arrow gameObject in `attack.BowDamage(goModel.transform.forward);`.

I assumed that `DoCollision()` would only run once for arrows but it sometimes happens more than once. Now `DoCollision()` will only run once for arrows.

I don't know if spells are supposed to run `DoCollision()` more than once, but they do. Can you confirm, @Interkarma?


Interkarma:
I wasn't aware this was happening, thanks for letting me know. I'm happy for spell missiles to also return out if impactDetected is true, rather than just arrows.

PR #1210, opened 2019-03-03
fix shoplifting messages keys

petchema:
Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&t=1856&p=21577


petchema:
Pickpocketing uses HardStrings

PR #1211, opened 2019-03-04
Wabbajack

numidium:
Implements Wabbajack's special effect and also adds a helper method for implementing artifact effects in the future.

### DIFF
--- a/Assets/Scripts/Game/MagicAndEffects/Effects/Special/WabbajackEffect.cs
+++ b/Assets/Scripts/Game/MagicAndEffects/Effects/Special/WabbajackEffect.cs
@@ -10,6 +10,8 @@
 //
 using UnityEngine;
 using DaggerfallWorkshop.Game.Entity;
+using DaggerfallWorkshop.Utility;
+using DaggerfallWorkshop.Game.UserInterfaceWindows;
 
 namespace DaggerfallWorkshop.Game.MagicAndEffects.MagicEffects
 {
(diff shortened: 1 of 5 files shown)


Interkarma:
Nice! Looking forward to testing this one. Looks good on a quick read, well done. :)

Assets/Scripts/Game/MagicAndEffects/Effects/Special/WabbajackEffect.cs
```diff
@@ -64,7 +64,7 @@ public override void MagicRound()
                 DaggerfallEntityBehaviour newEnemyBehaviour = gameObject.GetComponent<DaggerfallEntityBehaviour>();
                 EnemyEntity newEnemy = (EnemyEntity)newEnemyBehaviour.Entity;
                 newEnemy.WabbajackActive = true;
-                newEnemy.CurrentHealth = enemy.MaxHealth - enemy.CurrentHealth; // carry over damage to new monster
+                newEnemy.CurrentHealth -= enemy.MaxHealth - enemy.CurrentHealth; // carry over damage to new monster
```
Interkarma: I wonder if this could create a situation where enemy dies instantly because the original enemy has lost more health than the transformed enemy has available? Another possible approach would be to get percent of health lost on original enemy then remove that percent of health from new enemy.

numidium: That happens pretty often in classic, actually. I was going for behavior that would mimic the original.

Interkarma: Sounds good, I honestly wasn't sure what classic's behaviour was here. :)

PR #1212, opened 2019-03-06
Address warnings

Allofich:
Addressed the following warnings on startup:

1. Unused variable warnings. A lot of these showed up recently. These are commented out now.
2. Inconsistent line-ending warnings (mixture of Windows and UNIX). Some files had a few Windows-style line endings, and one file (`DaggerfallAutomapWindow.cs`) had mostly Windows-style line endings. These were made to have UNIX-style endings as the rest of the project does.
3. `ExteriorAutomap` looked for non-existent layer "ExteriorAutomap", issued a warning that erroneously said it couldn't find layer "Automap", then defaulted to layer 10, which actually is "Automap". Fixed to just look for "Automap" from the start.

Another warning that appears on startup if you start from `DaggerfallUnityStartup.unity` is ```Could not locate TextManager GameObject instance in scene!``` However it causes no problem. If you look at the trace, DaggerfallUI.Awake creates a DaggerfallInventoryWindow, which then tries to access the TextManager instance, which doesn't exist yet. The warning message is issued, then a TextManager instance is created and returned, so execution continues normally.

Seems like either the warning is incorrect and should be removed, or the order of execution should be changed so that a TextManager instance is created before any attempt to access the instance is made (as is the case in `DaggerfallUnityGame.unity`, which shows no warning). I'll leave it to how you would like to handle, Interkarma.

Also, I think there in general may be too many log messages at startup, for things like potion recipes, etc. They could bury warning messages and make them go unnoticed. Seems to me like those types of messages should be removed for any game mechanic that isn't under active development, but I'll leave the call to you.


Allofich:
Updated `DaggerfallUnityGame` and `DaggerfallUnityStartup` to not have `Allow MSAA` checked on the `Camera` object. Having this checked created a warning in the editor that MSAA could not be used with the selected rendering method of `Deferred`, nor without turning on MSAA in the project quality settings.

Allofich:
@TheLacus, there are a couple things I wonder about that maybe you can answer or fix. I'm not familiar with mod settings so excuse me if I'm just doing something wrong.

1. When I open `Daggerfall Tools`, `Mod Settings` in the Unity editor, I immediately get a null exception on the line

`data.Version = EditorGUILayout.TextField("Version", data.Version);`

in `ModSettingsEditorWindow.cs`. It seems `data` is null.

I think this might happen because I have no mods, but do you know?

2. I'm using Visual Studio 2017, and various errors show up for missing definitions/type names etc. because the relevant pieces of code are within `#if UNITY_EDITOR` blocks. Are the blocks necessary?

TheLacus:
> 1. When I open `Daggerfall Tools`, `Mod Settings` in the Unity editor, I immediately get a null exception on the line
> 
> `data.Version = EditorGUILayout.TextField("Version", data.Version);`
> 
> in `ModSettingsEditorWindow.cs` because `data` is null.
> 
> I think this might happen because I have no mods, but do you know?

This happens when no mod data is loaded, thank you for pointing this out. I never experienced this issue because current mod is stored for next session and you can only replace it with another valid location. I'll push a PR with fix.

> 2. I'm using Visual Studio 2017, and various errors show up for missing definitions/type names etc. because the relevant pieces of code are within `#if UNITY_EDITOR` blocks. Is it necessary to condition them with `#if UNITY_EDITOR`?

Not sure about this one, i don't see any error with Visual Studio. Do they come from _ModSettingsTypes_? The directive is used to strip unnecessary code from release builds, there shouldn't be any issue with that.

Allofich:
> Not sure about this one, i don't see any error with Visual Studio. Do they come from ModSettingsTypes? The directive is used to strip unnecessary code from release builds, there shouldn't be any issue with that.

_ModSettingsTypes_, and also _ImportedComponentAttribute_ and _PostProcessingProfile_. The errors exist in Visual Studio but I can compile and run in the Unity editor fine.

I tried to look this up and found some suggestion about using "Build and Run" from Unity to create a Visual Studio solution and open that (rather than the "daggerfall-unity.sln" that appears in the local copy of the Daggerfall Unity repository), but I ran into a few obstacles when trying to use a solution build from Unity.

I have Visual Studio 2017 version 15.9.8, Visual Studio 2017 Tools for Unity, and Unity 2018.2.11f1. I open "daggerfall-unity.sln".

Also,
> I'll push a PR with fix.

Thanks for fixing that.

TheLacus:
I see three assemblies inside the project solution: _Assembly-CSharp_, _Assembly-CSharp-Editor_ and _Assembly-CSharp-firstpass_. Could the problem be that you have only the first one opened?
I tipically double-click any cs file from the Unity Editor to open Visual Studio. Alternatively there is also **Assets**->**Open C# Project**.

Allofich:
> I see three assemblies inside the project solution: Assembly-CSharp, Assembly-CSharp-Editor and Assembly-CSharp-firstpass. Could the problem be that you have only the first one opened?
I tipically double-click any cs file from the Unity Editor to open Visual Studio. Alternatively there is also Assets->Open C# Project.

Still no luck. I can also see all three of those assemblies in the solution, and I tried double-clicking from the Unity Editor and Assets->Open C# Project, but same result.

Well, I don't want to take any more of your time with this. I'll continue to experiment and look for solutions online.

Interkarma:
Thank you Allofich! I also agree there's too much log clutter at startup now. Happy for us to trim this down just to warnings/errors and environment information for troubleshooting end user problems.

PR #1217, opened 2019-03-08
Fix most remaining texture alignment problems

df-ferital:
I spent hours (if not days) on this one...
This fixes most of the texture UV coordinates problems uncovered on the forums:

https://forums.dfworkshop.net/viewtopic.php?f=24&t=328: I removed the special case @Interkarma put several years ago for the building window since it is not needed anymore. The weapon store used for The Count's Arsenal in Newtale (@petchema) is also fixed.
https://forums.dfworkshop.net/viewtopic.php?f=28&t=1588: the dungeon room is fixed but not the tavern pillars. @petchema, could you check that the dungeon corridor streched textures you saw are corrected?

To fix this, I used in particular a previously unused and unknown field from the mesh header. It is read as an integer but from my tests, the only values it can take are 0 and 65536, so I think it should actually be an unsigned int. But as it does not really matter since I do the check against 0, I didn't change the type.

The only known remaining issues are textures coordinates for tavern pillars like the ones in The King's Fairy, Daggerfall.

### DIFF
--- a/Assets/Scripts/API/Arch3dFile.cs
+++ b/Assets/Scripts/API/Arch3dFile.cs
@@ -164,7 +164,7 @@ namespace DaggerfallConnect.Arena2
             public Byte PlanePointCount;
             public Byte Unknown1;
             public UInt16 Texture;
-            public UInt32 Unknown2;
+            public int UVunpack;
         }
 
         /// <summary>
@@ -604,7 +604,7 @@ namespace DaggerfallConnect.Arena2
                 records[record].PureMesh.Planes[plane].Header.PlanePointCount = reader.ReadByte();
                 records[record].PureMesh.Planes[plane].Header.Unknown1 = reader.ReadByte();
                 records[record].PureMesh.Planes[plane].Header.Texture = reader.ReadUInt16();
-                records[record].PureMesh.Planes[plane].Header.Unknown2 = reader.ReadUInt32();
+                records[record].PureMesh.Planes[plane].Header.UVunpack = reader.ReadInt32();
 
                 // Read the normal data for this plane
                 Int32 nx = normalReader.ReadInt32();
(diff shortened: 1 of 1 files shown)


petchema:
> @petchema, could you check that the dungeon corridor streched textures you saw are corrected?

They're fixed

df-ferital:
Thanks! :)

petchema:
That can probably be simplified; Since the only values that are mapped are multiples of 0x0400, and you then add or substract from 0x2000 that is also a multiple of 0x0400, all partial results stay multiples of 0x0400; Dividing everything by 0x0400, if the input values are in -32768..32767 range, there's only 64 values that are mapped: -32 to 31

```
    -32..-25 =>  0  1 -6 -5 -4 -3 -2 -1
    -24..-17 =>  0  1 -6 -5 -4 -3 -2 -1
    -16.. -9 =>  0  1 -6 -5 -4 -3 -2 -1
     -8.. -1 =>  0  1 
     -8.. -1 =>       -6 -5 -4 -3 -2 -1
      0..  7 =>  0  1  2  3  4  5  6  
      0..  7 =>                       1
      8.. 15 =>  0 -1 -2 -3 -4 -5 -6  1
     16.. 23 =>  0 -1 -2 -3 -4 -5 -6  1
     24.. 31 =>  0 -1 -2 -3 -4 -5 -6  1
```

Arranging all the values in a table like that, 3 different mappings appear (*), u <= -7, u >= 7 and in between. After some tinkering, the different mappings can be expressed using modulo 8:

```
        private static int UVunpack(int u)
        {
            if (u >= 7)
                u = 1 - (u + 1) % 8;
            else if (u <= -7)
                u = 1 + (u - 1) % 8;
            return u;
        }
...
                    if (u % 0x0400 == 0 && flag == 0)
                    {
                        u = UVunpack(u / 0x0400) * 0x0400;
                    }

                    if (v % 0x0400 == 0 && flag == 0)
                    {
                        v = UVunpack(v / 0x0400) * 0x0400;
                    }
```

Maybe that range between -7 and 7 is wrong and we can do with just two mappings, say u positive or negative? I don't know. It didn't fix the King's Fairy pillar though :(

(*). actually values between -6 and 1 could be put in two different mapping ranges for the same result; I made the choice that looked the most symetric.

df-ferital:
You're right. Actually I spent most of my time figuring out how classic worked so I didn't try to optimize the original code. So great work! I'll update the PR.

For tavern pillars, I think it is related to the fact they're combined objects, I still have to dig into that. Maybe @Interkarma will have an idea?

df-ferital:
@petchema I added your optimization with a little rework on my side to avoid any division or multiplication.

I also renamed the `Unknown2` field to `UVunpack`.

petchema:
Mmmh, do you also experience texture issues inside the Direnni Tower?
http://ks369561.kimsufi.com/~petchema/gfx/Daggerfall/Direnni%20Tower%20textures.jpg

df-ferital:
I can't check now, do you experience with my changes?

petchema:
> I can't check for now, do you experience them only when applying the PR changes?

When applying any version of this PR, same result

df-ferital:
I fixed it, it seems coordinates equal to 8192 or -8192 are not to be processed.

Interkarma:
This is amazing work guys! I really appreciate it. I'm still testing this one, but will merge soon.

PR #1219, opened 2019-03-10
AI fixes

Allofich:
Distance used in decision for doing melee attack used the value
in EnemySenses, which could differ from the distance value used in
EnemyMotor. Noticeable when classic AI, which only moves as close
as necessary to attack, could stand motionless due to being close
enough to satisfy EnemyMotor to no longer advance on target, but not
satisfy EnemyAttack to due melee attack.

Also melee decision should have been to attack if <= melee range, to
match classic.

Also implemented classic's differing melee range for vs. player and vs.
AI. This was probably done because the range used against player wouldn't
look right for AI vs. AI, because the sprites wouldn't appear to be physically touching.
But I guess the longer range looked better when attacking the player (attacking the screen)

Fixes issue 2 in https://forums.dfworkshop.net/viewtopic.php?f=24&t=1859&sid=bed84458b14ebae4f57363cdf767816f.

### DIFF
--- a/Assets/Scripts/Game/EnemyAttack.cs
+++ b/Assets/Scripts/Game/EnemyAttack.cs
@@ -26,9 +26,9 @@ namespace DaggerfallWorkshop.Game
     [RequireComponent(typeof(EnemySenses))]
     public class EnemyAttack : MonoBehaviour
     {
-        public float MeleeAttackSpeed = 1.25f;      // Number of seconds between melee attacks
-        public float MeleeDistance = 2.25f;          // Maximum distance for melee attack
-        public float MeleeTimer = 0;                // Must be 0 for a melee attack or touch spell to be done
+        public float MeleeDistance = 2.25f;                // Maximum distance for melee attack
+        public float ClassicMeleeDistanceVsAI = 1.5f;      // Maximum distance for melee attack vs other AI in classic AI mode
+        public float MeleeTimer = 0;                       // Must be 0 for a melee attack or touch spell to be done
         public DaggerfallMissile ArrowMissilePrefab;
 
         EnemyMotor motor;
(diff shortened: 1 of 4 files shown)


Allofich:
Fixed issue of enemies standing still when player moves behind them. This was a regression caused by the strafing code.

Also I found and fixed another regression that was causing concealment spells (invisibility, shade, chameleon) to no longer be working.

Allofich:
Fixed enemies attacking through doors. I think this may also have been a regression.

Also disabled pacification for quest enemies. Noted as an issue in
https://forums.dfworkshop.net/viewtopic.php?f=5&t=1856

> Pacified the orc shaman that just told "prepare to die" using orcish skill

It's probably also possible to pacify quest enemies in classic, but I don't think any of the quests are really built with this in mind, so it seems like undesirable behavior/a bug.

Allofich:
Implemented smarter behavior for when enemies lose track of their target. Rather than going straight as they would before, they now look from the last place they saw their target, in the direction that they last saw the target moving. This makes, for example, the nightblade in the saved game in https://forums.dfworkshop.net/viewtopic.php?f=24&t=1859&sid=bed84458b14ebae4f57363cdf767816f more able to follow the player into and out of the room, and fixes issue 1 of that post.

Allofich:
Fixed issue 3. The effects of `MakeEnemyHostileToAttacker` to make the spawned guards aware of the player's location were canceled out by the `Start` function in `EnemySenses`, which ran afterwards, resetting the values. Changing the order fixed this.

One thing, the bug reporter was not correct that classic has this functionality. In classic guards are simply spawned nearby and don't get any special help to notice the player, so they can spawn standing there with their back turned, etc. (Which doesn't make much sense from an in-world point of view, since they are presumably responding to a crime, not just lurking around like a dungeon opponent) This behavior is therefore is an enhancement over classic rather than an emulation of classic behavior, which I noted in a comment.

Interkarma:
Thank you Allofich!

PR #1220, opened 2019-03-10
check that the advancing dagger is not too high or too low

petchema:
Very low AdvancementMultiplier in particular is game breaking

Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=1882


Interkarma:
Excellent catch, thank you for the fix!

PR #1222, opened 2019-03-10
fix "floating" house labels when free scale is enabled

petchema:
.x vs .y buglet

Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&t=1855


Nystul-the-Magician:
thx for fixing ;)

PR #1225, opened 2019-03-14
Use saving throws and implement Mehrunes Razor effect

numidium:
I noticed that saving throws weren't being used for most spells. If this is the desired behavior for now then I don't mind changing it back. Having saving throws active prevents MR from killing targets on the first hit every time.

I also altered the Wabbajack effect. It bypasses saving throws in classic so I added a flag for effects that ignore them. Added a check to prevent Wabbajack from disabling enemies that are in use by a current quest.

### DIFF
--- a/Assets/Scripts/Game/MagicAndEffects/EntityEffectManager.cs
+++ b/Assets/Scripts/Game/MagicAndEffects/EntityEffectManager.cs
@@ -464,24 +464,25 @@ namespace DaggerfallWorkshop.Game.MagicAndEffects
                     continue;
                 }
 
-                // Saving throws
-                if (!effect.BypassSavingThrows && FormulaHelper.SavingThrow(effect, entityBehaviour.Entity) == 0)
+                // Saving throws are performed if bundle is not self-cast or originate from another entity
+                // Self-cast spells such as a heals and buffs are not checked for saving throws
+                // But player can still catch themselves in their own AoE spells and receive a chance to save
+                if (sourceBundle.Settings.TargetType != TargetTypes.CasterOnly || effect.Caster != EntityBehaviour)
                 {
-                    if (IsPlayerEntity || showNonPlayerFailures)
+                    if (!effect.BypassSavingThrows && FormulaHelper.SavingThrow(effect, entityBehaviour.Entity) == 0)
                     {
-                        // Output "Save versus spell made." for external contact spells
-                        DaggerfallUI.AddHUDText(TextManager.Instance.GetText(textDatabase, "saveVersusSpellMade"));
+                        if (IsPlayerEntity || showNonPlayerFailures)
+                        {
+                            // Output "Save versus spell made." for external contact spells
+                            DaggerfallUI.AddHUDText(TextManager.Instance.GetText(textDatabase, "saveVersusSpellMade"));
+                        }
+                        continue;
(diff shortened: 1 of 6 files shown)


Interkarma:
Good catch on the saving throws. Yeah, this wasn't supposed to be formatted so that it only worked for paralyse effect. That's likely a mistake on my part during a tired code edit. Thanks for fixing.

And thanks for implementing yet another artifact effect. Very happy to get some help along these lines, I've been building effects for almost a year now and fatigue has set in. :)

Assets/Scripts/Game/MagicAndEffects/EntityEffectManager.cs
```diff
@@ -464,20 +464,20 @@ public void AssignBundle(EntityEffectBundle sourceBundle, bool showNonPlayerFail
                     continue;
                 }
 
-                // Special handling for paralysis
-                if (effect is Paralyze)
+                // Saving throws
```
Interkarma: Fixing this has exposed another bug. Player can now "save" against spells they cast on themselves (e.g. a simple healing spell). I'm sure I used to handle this, but it's been a while. If you don't mind, I'll send an edit to this PR to fix this one before merging.

Interkarma:
Actually, saving throws _were_ being handled for all spells with magnitude. This is done by EnityEffect.GetMagnitude() which uses saving throws to mitigate some or all of the magnitude from an effect. It also performs the self cast test that I was recalling earlier.

For effects without magnitude, they need special save handling. That's why paralysis was handled the way it was. With these recent changes, spells now get two spots where saving throw is called, which isn't right either.

I'll need to revert some of this change and clean up a little, but don't have the focus right now. I'll loop back soon. :)

Allofich:
Slightly off-topic, but since we're doing artifact effects now...

The Masque of Clavicus is a very simple effect, but I don't know if it will fit cleanly into the current effect model or not.

For the effect, what happens in classic is that when an NPC is spoken to, the player is checked for whether they have artifact 0 (the masque of clavicus) equipped, and if so, (currentPersonality / 5) is added to the NPC reaction. This is just a hardcoded case in the talk code, and I think the Masque of Clavicus might never lose enchantment points. Haven't confirmed in classic.

The check for the artifact is done by iterating through the equipment slots, looking at each item ( `DaggerfallUnityItem` in our case) and its enchantment data ( `legacyMagic`) for one with `type` of 26 ( `SpecialArtifactEffect` in `ItemsFile.cs`) and `param` of the artifact number sought, in this case `0`.

 The effect of the Masque of Clavicus could be re-created by adding the personality bonus to `reactionToPlayer` in function `GetReactionToPlayer` in `TalkManager.cs`. The effect is on this reaction, used for the initial roll to open dialog, not the other reaction used with asking questions.

There is no "effect" used by classic, but I imagine we want one to open it up to modding.

Interkarma:
Thanks for the great information Allofich. I'll have a think about how we might be able to deliver this via the effect system.

Something easy and direct that comes to mind is to create a `ReactionBonus` property in entity that's added to `reactiontoPlayer` as you mention. Then the Masque effect itself just has to raise `ReactionBonus` while effect is active. It would make for a fairly straight-forward setup and still allow the effect to be overridden by a mod to do something extra/different.

PR #1228, opened 2019-03-14
Fix some incorrect UV coordinates within the data files

df-ferital:
This provides a starting point to patch some wrong texture UV coordinates in ARCH3D.BSA at runtime.
For now, this fixes the following problems:
- Tavern pillars like in The King's Fairy, Daggerfall
- Compressed wall texture near the entrance of a weapon store like The Count's Arsenal, Newtale, Bhoriane (@petchema)
- Stretched dungeaon corridor floor (from model index 989)

### DIFF
--- a/Assets/Scripts/API/Arch3dFile.cs
+++ b/Assets/Scripts/API/Arch3dFile.cs
@@ -46,8 +46,11 @@ namespace DaggerfallConnect.Arena2
         private FaceUVTool.DFPurePoint[] calculatedUVBuffer = new FaceUVTool.DFPurePoint[calculatedUVBufferLength];
 
         // Special arrays containing UV coordinates fixes
-        private static readonly int[] uFirstTavernPillar = { 0, 0, 96 };
-        private static readonly int[] uSecondTavernPillar = { 1952, 0, 96 };
+        private static readonly int[] uFirstPillar = { 0, 0, 96 };
+        private static readonly int[] uSecondPillar = { 1952, 0, 96 };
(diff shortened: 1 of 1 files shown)


Nystul-the-Magician:
nice great work - pillars in mages guild are suffering from the same problems ;) those bugged me for a long time - so hopefully we will get them fixed :D

Allofich:
I'm also glad these texture issues are finally getting fixed. Thank you!

df-ferital:
I think all interior models containing pillars are now fixed. The problem is the switch statement I used is rather big now.

Not sure if this should be merged yet, there may be a cleaner way to handle that.

df-ferital:
Closing for now. Will open a new PR with a better algorithm.

Nystul-the-Magician:
Would have been ok for me with the switch statement - refinement later would have been fine ;)

petchema:
> Would have been ok for me with the switch statement - refinement later would have been fine ;)

I agree, it's fine to iterate; the first version doesn't need to be elegant if it does the job and is identified as something to improve

df-ferital:
Ok, I reopen it :).

I have started to work on a far better solution though, which will allow to patch every bit of a BSA file. This allowed me to change the wrong texture file used for the dungeon sewer floor @petchema mentioned on the forums.

It works pretty well, but I didn't transfer patches to the pillar textures yet.

So @Interkarma, if you want to merge the current solution in the mean time, I'm ok with it :)

Interkarma:
Happy to merge in the meantime. Thanks again for all your work on this. :)

PR #1229, opened 2019-03-14
Invert towns and homes colors on the map

df-ferital:
This inverts the color used for home and town dots on the travel map. Towns are now brown while homes are purple, as suggested by their respective button.  One other solution would have been to invert button colors but I think people are so used to them that I prefer the first solution.

IMO it is a way clearer that way.

This patch came to me after reading King of Worms mentioning this issue in classic here: https://forums.dfworkshop.net/viewtopic.php?f=12&t=1898#p22186

@Interkarma and others: I think this corrects a bug in classic but I would not mind if you don't agree.


Allofich:
Never noticed that. Sounds OK to me, but could you maybe attach a screenshot of before and after the change, so we can get a feel for if anything feels strange about it?

df-ferital:
Of course, here are the screenshots:

Before:
Towns before
Homes before

After:
Towns after
Homes after

Allofich:
Thanks, seems fine.

Interkarma:
That makes sense. Interesting I've never noticed that before, but now I can't help but see it.

PR #1231, opened 2019-03-18
Implement 2 new commands for testing AI

Allofich:
2 console commands that will be useful for working on the AI, as well as for just messing around and having fun with AI battles. Also lays some groundwork for implementing the artifact effects that involve summoning an ally.

"CreateMobile" = type "cm" followed by the numerical ID of the enemy type, then the number of the mobile team to spawn it in front of you.

Ex.) cm 0 1
Creates a rat (id 0) on the "player ally" team (team 1, won't attack player)

All teams are hostile to all other teams. Created a new "player enemy" (team 0) for general use in making enemies that are all allied with each other against the player.

Issues)
1: When created, mobiles are rotated to look in the direction the player was looking, but for some reason this seems a little finicky and they will at first briefly be looking at the player, or flicker back and forth a little before they settle on looking in the correct direction. Don't know why, as it's just a one-time "transform.LookAt" call, but the end result is correct, so I'm okay with it since it's just a test command. Might need to figure this out for the artifact effects, though.

2: Teams are not restored when loading a saved game. They get overwritten with the default teams. Since this is just a console command for testing I think that's okay for now. Will need to be solved when implementing the artifact effects.

"ToggleAI" = type "tai" to toggle between disabling and enabling all mobileEnemy AI. Doesn't affect townspeople.

P.S.
On the subject of AI (not related to this PR), the pursuit AI, such as when guards are pursuing you in town, is messed up at the moment, with them tending to go wandering in the wrong direction. I'm aware of it and working on a fix.

### DIFF
--- a/Assets/Game/Addons/UnityConsole/Console/Scripts/DefaultCommands.cs
+++ b/Assets/Game/Addons/UnityConsole/Console/Scripts/DefaultCommands.cs
@@ -33,6 +33,7 @@ namespace Wenzil.Console
             ConsoleCommandsDatabase.RegisterCommand(LoadCommand.name, LoadCommand.description, LoadCommand.usage, LoadCommand.Execute);
             ConsoleCommandsDatabase.RegisterCommand(GodCommand.name, GodCommand.description, GodCommand.usage, GodCommand.Execute);
             ConsoleCommandsDatabase.RegisterCommand(NoTargetCommand.name, NoTargetCommand.description, NoTargetCommand.usage, NoTargetCommand.Execute);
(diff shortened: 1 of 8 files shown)


Allofich:
> On the subject of AI (not related to this PR), the pursuit AI, such as when guards are pursuing you in town, is messed up at the moment, with them tending to go wandering in the wrong direction. I'm aware of it and working on a fix.

Added a fix for the pursuit AI in the following PR: 
https://github.com/Interkarma/daggerfall-unity/pull/1227

PR #1233, opened 2019-03-21
Implement Oghma Infinium and Azura's Star

Allofich:
Implements the Oghma Infinium. It was harder than I expected because there were bugs to chase down to get the correct texture and inventory behavior. (Being assigned by the data in MAGIC.DEF to the books group was interfering with these things).

I noticed problems with the leveling screen. I think they are the result of a recent commit.
https://github.com/Interkarma/daggerfall-unity/commit/b0f6f52bd3c7d99ff7e6873e18807f6c3f29be11#r32858800

Use `addArtifact 5` at the console for easy testing.

### DIFF
--- a/Assets/Game/Addons/UnityConsole/Console/Scripts/DefaultCommands.cs
+++ b/Assets/Game/Addons/UnityConsole/Console/Scripts/DefaultCommands.cs
@@ -1535,6 +1535,9 @@ namespace Wenzil.Console
                         case "torch":
                             newItem = ItemBuilder.CreateItem(ItemGroups.UselessItems2, (int)UselessItems2.Torch);
                             break;
+                        case "soultrap":
+                            newItem = ItemBuilder.CreateItem(ItemGroups.MiscItems, (int)MiscItems.Soul_trap);
+                            break;
                         default:
                             return "unrecognized keyword. see usage";
                     }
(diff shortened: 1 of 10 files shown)


Allofich:
Implemented Azura's Star. Use `addArtifact 9` and a Soul Trap spell to test.

Allofich:
This code chooses randomly from all available soul traps and Azura's Stars (just in case there is somehow more than one) to put a soul in.

But, will that cause problems with the sorting order on the inventory screen? We would probably want all soul traps with/without souls in them to be grouped together. Anyone know if this is already sorted for somehow? (I only tested this PR with Azura's Star, not with regular soul traps)

Allofich:
> But, will that cause problems with the sorting order on the inventory screen? We would probably want all soul traps with/without souls in them to be grouped together. Anyone know if this is already sorted for somehow? (I only tested this PR with Azura's Star, not with regular soul traps)

I tested and the soul traps were not automatically sorted. I decided to just use an empty Azura's Star first and then and empty soul traps, which is simpler code and probably expected behavior.

Also although it's not in the Azura's Star description, I recreated the classic behavior that Azura's Star enables 100% success rate for trapping souls while it's equipped, even when the soul goes to a regular soul trap. I'm not sure if that is supposed to happen or not, though.

Interkarma:
Nice one! Thanks for your work on these. :)

PR #1234, opened 2019-03-21
prevent f5 from closing level up screen

petchema:
In classic Daggerfall, one cannot exit the level up screen until all
ability bonus points have been distributed.

In Daggerfall Unity, this is true when trying to use the EXIT button or
the Escape key, but it was possible to exit the character sheet window
and distribute remaining bonus points later by pressing the character
sheet keyboard shortcut (f5 by default).

This patch closes this loophole, and factorizes some code between the
different ways to close the character sheet.

Bug tracker:
https://github.com/Interkarma/daggerfall-unity/commit/b0f6f52bd3c7d99ff7e6873e18807f6c3f29be11

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallCharacterSheetWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallCharacterSheetWindow.cs
@@ -203,7 +203,8 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
 
             // Toggle window closed with same hotkey used to open it
             if (Input.GetKeyUp(toggleClosedBinding))
-                CloseWindow();
+                if (CheckIfDoneLeveling())
+                    CloseWindow();
         }
 
         public override void OnPush()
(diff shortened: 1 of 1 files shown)


Allofich:
Hm, I was having the problem with the exit button, when testing the Oghma Infinium in my PR. Maybe my changes there somehow caused it.

petchema:
Oghma Infinium? Oh, now I see how you got there. I can't click EXIT with just your PR though, so there's still some mystery...

The two PRs stack correctly without conflict, can you still reproduce the issue with both PRs applied?

Allofich:
When I apply this PR, the problem is fixed. Also, I confirmed the problem I have is only with the Oghma Infinium in my PR, not with regular leveling.

Allofich:
I found the problem and pushed a fix to my PR.

The issue was that there was existing code to `PopToHud()` after using a magic item, and this would interfere with the `DaggerfallCharacterSheetWindow` I had just pushed as the effect of using the Oghma Infinium. The `OnPop()` method of the character sheet was instantly run, which included your recently added code to make `leveling` false. This is where I noticed that removing that line from there would fix the problem, but it wasn't the root issue. The root issue was that the window was being popped as soon as it was added, resulting in odd behavior. Sorry for thinking your code was causing the problem.

I fixed the issue for my PR by only popping the `DaggerfallInventoryWindow`, not all windows.

petchema:
Well, it could have been the case ;) Daggerfall UI is still a bit mysterious to me.
New code is cleaner, but I still wonder if there could be a better way to prevent character sheet from being closed or cancelled than call CheckIfDoneLeveling() in each path...

PR #1237, opened 2019-03-24
Prevent non-guard spawns while in water

Allofich:
Prevent random encounter spawns as mentioned here https://forums.dfworkshop.net/viewtopic.php?f=24&t=1918&sid=b161443d647dd3d6ddfe6e86d3b194ed.

They are also prevented in this case in classic.


Interkarma:
Thank you!

PR #1238, opened 2019-03-24
Apply screen weapon dye overrides for artifacts

numidium:
I altered a couple of them from classic so that they match their respective paper doll images better.  Let me know if I'm out of line in doing this.

### DIFF
--- a/Assets/Scripts/Game/Items/ItemHelper.cs
+++ b/Assets/Scripts/Game/Items/ItemHelper.cs
@@ -739,14 +739,12 @@ namespace DaggerfallWorkshop.Game.Items
                         {
                             switch (enchantment.param)
                             {
-                                case (int)ArtifactsSubTypes.Mehrunes_Razor: // Uses Mithril in classic but Elven matches its paper doll image
+                                case (int)ArtifactsSubTypes.Mehrunes_Razor: // Different from classic but Elven matches the paper doll more closely
                                     return MetalTypes.Elven;
                                 case (int)ArtifactsSubTypes.Mace_of_Molag_Bal:
                                     return MetalTypes.Ebony;
                                 case (int)ArtifactsSubTypes.Wabbajack:
                                     return MetalTypes.Steel;
-                                case (int)ArtifactsSubTypes.Volendrung: // Uses Steel in classic but Dwarven matches its paper doll image
-                                    return MetalTypes.Dwarven;
                                 default:
                                     break;
                             }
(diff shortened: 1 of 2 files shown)


Interkarma:
This sounds reasonable to me. I'll test and let you know if anything looks out of place.

Allofich:
At first I thought this was just changing dyes to make the on-screen image match the image in the inventory, which I would be fine with, but as far as I can see in classic magic weapons always seem to look like steel weapons with a blue glow around them, right? And this seems to be actually changing the materials so wouldn't it change the damage these artifacts do as weapons?

The material is a hidden attribute for artifacts as it can't be seen in the name like with non-magical/regular magic items. Also from testing in classic it seems that as long as an item is magical it can can hit regardless of the regular material restrictions (in classic, not sure we have that in DF Unity). This would mean that the only in-game effect of changing the materials would that it would change the damage values, doesn't it? If my understanding is right then I say the values should be left as they are.

Allofich:
OK, so, I guess in DF Unity the material shows when the item is magic. (Although the Mehrunes Razor looks green to me, suggesting Orcish, not Mithril like the comment in the PR code says.)

I support changing the colors to match the inventory icons. And if this PR only affects the colors, I'm fine with that. For classic, though, since it seems to show as steel-colored regardless, and the material is hidden from the user for artifacts, it would essentially be just a value meaning what the damage values are, nothing about appearance, so I think we should keep those as they are.

Allofich:
> case (int)ArtifactsSubTypes.Volendrung: // Uses Steel in classic but Dwarven matches its paper doll image

Volendrung seems to already be Dwarven in classic

Allofich:
OK, looks like this PR does indeed only affect the screen dyes, not the weapon damages. Sorry for the misunderstanding.

In that case the only thing I have to say is that Volendrung is fine as it is. (I haven't checked all of the artifacts, just Mehrunes Dagger and Volendrung)

numidium:
The Volendrung case was redundant since, as Allofich pointed out, it's dyed as Dwarven by default in DFU.  I know Mehrunes' Razor looks Orcish on the HUD in DFU by default but I don't recall it looking like that in classic. I've updated the comment so it doesn't make a claim one way or the other what it looked like in classic.

Nystul-the-Magician:
just a question out of curiosity: would it be hard to implement unique dyes for artifacts (different from the existing material dyes) - thinking of a mod here - would be interesting to see if that makes the artifacts feel even more unique

Interkarma:
@Nystul-the-Magician In theory should be possible to add new dyes. I experimented with this a little when building the system. Article below has screenshot with "blue boots".

https://www.dfworkshop.net/items-part-2-dyes/

Exposing this to mod system in a clean and sensible way is like a whole other ordeal, but the facility is there at least to tint items with custom colours.

Nystul-the-Magician:
nice thx for the link - didn't remember this one

PR #1239, opened 2019-03-25
fix regression in PlaySource.cs

petchema:
Fix regression I made in a3479c7c that made Lysandus silent :(


petchema:
[SAVE197_vengence.zip](https://github.com/Interkarma/daggerfall-unity/files/3005631/SAVE197_vengence.zip)

petchema:
With that fix I uncover another regression: now "Vengeance" plays inside buildings too.
git bisect points at 80b4f62c
[SAVE89_inside_vengeance.zip](https://github.com/Interkarma/daggerfall-unity/files/3005826/SAVE89_inside_vengeance.zip)

Interkarma:
>With that fix I uncover another regression: now "Vengeance" plays inside buildings too, or even exit the city.
Thank you. It appears my previous fix also stops tasks from shutting down correctly. I'll handle this a different way shortly.

Interkarma:
Thanks for spot on both of these @petchema :) I've just merged a fix for the WhenTask issue. I included this fix as well since I needed it for testing. Cheers!

PR #1240, opened 2019-03-28
remove a missed reference to old interface icons

petchema:
Seen in a stream

### DIFF
--- a/Assets/StreamingAssets/Quests/_TUTOR__.txt
+++ b/Assets/StreamingAssets/Quests/_TUTOR__.txt
@@ -214,8 +214,8 @@ QuestTimeLapse:  [1045]
 <ce>           walls. You can only rest in Taverns, if you have
 <ce>                        purchased a room there.
 <ce>                                    
-<ce>          The first thing you want to do is set your mode to
-<ce>         DIALOGUE: just press F4 (you can change the mode of
+<ce>         The first thing you want to is to press the F4 key to
+<ce>         set your mode to DIALOGUE (you can change the mode of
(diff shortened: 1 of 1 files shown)


Allofich:
Minor point, but, I think it would be better without "just", as I think that was in contrast to the removed "click until" alternative.

petchema:
I hope it's now good enough... until a total rewrite ;)

Allofich:
I think it's OK.

> until a total rewrite

Personally I like to keep the original text with minimal changes, or at least still have it available somehow.

petchema:
> Personally I like to keep the original text with minimal changes, or at least still have it available somehow.

That's pretty much the definition of what I've done with the tutorial so far: replace now irrelevant (hence confusing) informations with updated ones, keeping changes minimal. And indeed that tutorial may stay around forever as an option for purists.

It's clear however that this tutorial is very far from the state of the art when it comes to ingame help, and not only because of its content; So I was referring to projects like https://forums.dfworkshop.net/viewtopic.php?f=12&t=1372 or https://forums.dfworkshop.net/viewtopic.php?f=12&t=1930 even if what should be done is not precisely defined yet...

petchema:
"Quarter-screen HUD with icons and mouse cursor toggle" https://www.dfworkshop.net/projects/daggerfall-unity/roadmap/

Mmh! So the references to the "old" UI may still be useful later?

Interkarma:
Yeah, I've had a few requests and do plan to add it down the road. It's not that high priority though. I polled on twitter some time back, and the vast majority of users indicated they either don't care or actively dislike the chunky HUD.

PR #1242, opened 2019-03-29
Reworked the way to patch a BSA file after it has been loaded

df-ferital:
A simple patch system has been put in place which allows to modify any BSA file at runtime. For now, it is only used to fix models from ARCH3D.BSA.

The patch includes all previous modifications as well as a complete fix for the dungeon sewer junction mentioned here: https://forums.dfworkshop.net/viewtopic.php?f=5&t=1899#p22261

I also fixed the sewer dead end whose floor and roof textures were stretched.


petchema:
What is the risk of ARCH3D.BSA file variants out there?

df-ferital:
I don't think there are any variants of it. Even if there were, the probabiblity they would modify the same models is rather small.

PR #1243, opened 2019-03-30
Spell merchant improvements

petchema:
Sort spells by alphabetic order, display spells magicka costs.

I considered greying out spells from the merchant when the magicka cost is over your MaxMagicka; That is, you cannot casts them, even when fully rested.
But greying out usually means stuff you cannot do, and you can still buy those spells, so I finally decided against it (the code is ready for such change though).

Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&t=1916

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallSpellBookWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallSpellBookWindow.cs
@@ -215,7 +215,11 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
                 LoadSpellsForSale();
                 foreach (EffectBundleSettings bundle in offeredSpells)
                 {
-                    spellsListBox.AddItem(bundle.Name);
+                    // Show spell name and cost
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Game/UserInterfaceWindows/DaggerfallSpellBookWindow.cs
```diff
@@ -212,60 +212,17 @@ void RefreshSpellsList(bool preservePosition)
...
-                    if (!GameManager.Instance.EntityEffectBroker.ClassicSpellRecordDataToEffectBundleSettings(standardSpells[i], BundleTypes.Spell, out bundle))
-                        continue;
-
-                    // Store offered spell and add to list box
-                    offeredSpells.Add(bundle);
-                    spellsListBox.AddItem(standardSpells[i].spellName);
-                }
+                LoadSpellsForSale();
+                // I'm not sure GameManager.Instance.PlayerEntity.MaxMagicka would be a good idea here
+                PopulateSpellsList(offeredSpells, null);
```
Interkarma: Agreed, I think you made the right choice here.

PR #1244, opened 2019-03-31
Prevent rats being walked upon

Allofich:
https://forums.dfworkshop.net/viewtopic.php?f=24&t=1425

Should prevent both rats walking on each other and player walking over them.


Allofich:
This might also help with https://forums.dfworkshop.net/viewtopic.php?f=24&t=1942, as far as targeting small enemies like rats.

PR #1245, opened 2019-03-31
AI improvements

Allofich:
Make sure enemies don't get stuck with "avoidObstaclesTimer" > 0 if they
are somehow unable to point at the detour destination.
Raise avoidObstaclesTimer a bit now since it counts down regardless of
direction enemy is pointing, but set to 0 if got close enough to target
spot. 0.3 seems like a good distance for this check.
Make enemies aim at their own height for target when searching for
target, not just when approaching seen target.
Improve RayCheckForObstacle:
- Use a CapsuleCast. This does better at detecting things in the way.
- Adjust height of rays casted for checking for slope. Should still
detect climbable stairs while not treating large, unclimbable
differences in height as climbable.

### DIFF
--- a/Assets/Prefabs/Missiles/ArrowMissile.prefab
+++ b/Assets/Prefabs/Missiles/ArrowMissile.prefab
@@ -26,7 +26,7 @@ GameObject:
   - component: {fileID: 114624404217668130}
   - component: {fileID: 54102960107042576}
   - component: {fileID: 64298782196861766}
-  m_Layer: 0
+  m_Layer: 14
   m_Name: ArrowMissile
   m_TagString: Untagged
   m_Icon: {fileID: 0}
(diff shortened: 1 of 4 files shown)


Interkarma:
Hey Allofich! Thanks for PRs, I'll get these all merged soon. :)

And something else that seems related to enemy AI, someone on reddit encountered enemies "drifting" backwards after a knockback. It looks like the motor take a few moments to kick back in again. They posted a short video and some other details. I just wanted to link you the topic as I don't know if you follow the subreddit or not.

https://www.reddit.com/r/daggerfallunity/comments/bb5yu4/enemies_seem_floaty/

Allofich:
I watched the video, and I've never had that happen in my setup, but I think the poster is right that the "classicUpdate" is not firing reliably. I guess it worked well enough on my computer that I didn't notice an issue.

The poster's solution seems like it's on the right track. I'll try something like it.

Allofich:
The problem seems to have simply been that I was mixing the use of `Update` and `FixedUpdate`.

https://github.com/Interkarma/daggerfall-unity/pull/1253

PR #1247, opened 2019-04-05
Clean up Talkmanager

Allofich:
Not a complete cleanup, but just some tidying up that should hopefully not raise any objections.

Whitespace cleanup
Merge nested if-statements
Remove empty methods
Rename reactionToPlayer variable masking same-named variable
Remove unnecessary reactionToPlayer parameter
Inline variables where possible
Remove redundant parentheses and redundant else statement
Remove redundant "== true" for bools


Allofich:
Finished giving the whole file a clean.

PR #1248, opened 2019-04-05
BillboardBatch: Use IDs to access shader properties

TheLacus:


Interkarma:
A sensible change, thank you!

PR #1252, opened 2019-04-11
Deducts gold when buying spells from spellmaker

goodbadbugs:
The current state of spellmaking allows the player to make spells for free (for testing purposes). This was deemed no longer necessary and this commit will deduct gold from the player based on the cost of the spell.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallSpellMakerWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallSpellMakerWindow.cs
@@ -743,10 +743,9 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             //   Implement costs and remove gold from player and block if user does not have enough gold (message 1702)
             //   Spells will be free (both gold and spell points) during build-out of effect system.
 
-            var moneyAvailable = GameManager.Instance.PlayerEntity.GoldPieces;
-            var cost = totalGoldCost;
+            var moneyAvailable = GameManager.Instance.PlayerEntity.GetGoldAmount();
 
-            if (moneyAvailable < cost)
+            if (moneyAvailable < totalGoldCost)
             {
                 DaggerfallMessageBox mbNoMoney = DaggerfallUI.MessageBox(notEnoughGold);
                 mbNoMoney.ClickAnywhereToClose = true;
@@ -756,7 +755,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             }
             else
             {
-                GameManager.Instance.PlayerEntity.DeductGoldAmount(cost);
+                GameManager.Instance.PlayerEntity.DeductGoldAmount(totalGoldCost);
             }
 
             // NOTES:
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Game/UserInterfaceWindows/DaggerfallSpellMakerWindow.cs
```diff
@@ -743,6 +743,22 @@ private void BuyButton_OnMouseClick(BaseScreenComponent sender, Vector2 position
             //   Implement costs and remove gold from player and block if user does not have enough gold (message 1702)
             //   Spells will be free (both gold and spell points) during build-out of effect system.
 
+            var moneyAvailable = GameManager.Instance.PlayerEntity.GoldPieces;
+            var cost = totalGoldCost;
```
Interkarma: Probably no need to assign this to its own field here. You could just use

    if (moneyAvailable < totalGoldCost)
    ...
    GameManager.Instance.PlayerEntity.DeductGoldAmount(totalGoldCost);

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallSpellMakerWindow.cs
```diff
@@ -743,6 +743,22 @@ private void BuyButton_OnMouseClick(BaseScreenComponent sender, Vector2 position
             //   Implement costs and remove gold from player and block if user does not have enough gold (message 1702)
             //   Spells will be free (both gold and spell points) during build-out of effect system.
 
+            var moneyAvailable = GameManager.Instance.PlayerEntity.GoldPieces;
```
Interkarma: To get the total amount of money available, including letters of credits, use the following helper instead :)

    GameManager.Instance.PlayerEntity.GetGoldAmount()

goodbadbugs: Thanks, refactored :)

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallSpellMakerWindow.cs
```diff
@@ -743,6 +743,21 @@ private void BuyButton_OnMouseClick(BaseScreenComponent sender, Vector2 position
...
+                return;
+            }
+            else
+            {
+                GameManager.Instance.PlayerEntity.DeductGoldAmount(totalGoldCost);
+            }
+
             // NOTES:
             //  In classic, player must have a spellbook item in their inventory (message 1703).
             //  In Daggerfall Unity, the spellbook is a permanent item that all players own and cannot be removed.
```
petchema: I know this is unrelated to this patch, but I just realized that this comment is no longer true, and that the presence of a spellbook will need to be checked.
And when added, it will have to be checked before paying for the spell

petchema: Actually what is relevant to your patch is that they're other conditions that are checked and that could prevent the spell from being made (no effects, no spell name), so paying should happen after all those tests

goodbadbugs: Good call, thanks for your help on this.

PR #1253, opened 2019-04-11
Everything using ClassicUpdate uses FixedUpdate

Allofich:
Should fix https://www.reddit.com/r/daggerfallunity/comments/bb5yu4/enemies_seem_floaty/.

When I merged all the classicUpdateTimers into one, I put it into the GameManager, and stuck it in its `Update()` along with all the rest of its update work. However, since `FixedUpdate()` and `Update()` are not in sync with each other, accesses to this from `FixedUpdate()` functions were not reliable. Everything needs to be either in `FixedUpdate()` or in `Update()`. Since https://docs.unity3d.com/Manual/ExecutionOrder.html says "All physics calculations and updates occur immediately after FixedUpdate.", and enemy AI is closely related to movement, `FixedUpdate()` seems to be the right place. When frame rates are high `FixedUpdate()` runs quite a lot less than `Update()`. When frame rates are low, like when loading up a save in a city exterior, `FixedUpdate()` runs more frequently, but hopefully that won't be a problem.

~In addition to `EnemyMotor`, that poster on Reddit should have also been having problems with the melee timer in `EnemyAttack`, which should both be fixed by this.~

### DIFF
--- a/Assets/Scripts/Game/GameManager.cs
+++ b/Assets/Scripts/Game/GameManager.cs
@@ -448,7 +448,7 @@ namespace DaggerfallWorkshop.Game
             Debug.Log("Welcome to Daggerfall Unity " + VersionInfo.DaggerfallUnityVersion);
         }
 
-        void Update()
+        void FixedUpdate()
         {
             if (!IsPlayingGame())
             {
@@ -465,7 +465,10 @@ namespace DaggerfallWorkshop.Game
             }
             else
                 classicUpdate = false;
+        }
 
+        void Update()
+        {
             // Post message to open options dialog on escape during gameplay
             if (Input.GetKeyDown(KeyCode.Escape))
             {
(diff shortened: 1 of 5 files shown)


Allofich:
I got my example in the explanation mixed up. The melee timer would have been fine, because both it and ClassicUpdate were running in Update(). Anything referencing ClassicUpdate from FixedUpdate() would have been unreliable, meaning EnemyMotor, EnemySenses, and some randomly repeating sounds like animals in cities that use ClassicUpdate.

Interkarma:
Thanks for looking into it, and for the fix. It sounds like you've probably resolved it. We'll see if reddit poster or someone else flags it after next builds. Great work. :)

Assets/Scripts/Game/GameManager.cs
```diff
@@ -465,7 +465,10 @@ void Update()
             }
             else
                 classicUpdate = false;
+        }
 
+        void Update()
+        {
```
Interkarma: Just FYI this change breaks window popping behaviour by allowing pause menu to display when not intended. I didn't catch this one while testing. It's not a big deal though, I'll send a fix shortly. :)

Allofich: Thanks

PR #1257, opened 2019-04-13
adjust scrolling speed to limit buffering

petchema:
When all popuptext buffer can be displayed on screen, behave just like before;

When more lines are buffered, adjust scrolling speed so lines appear within a fixed time (currently 1.5s, adjustable) after their insertion. This should prevent unbounded lag when a lot of text is generated (spells spamming, info-click on buildings,...)

Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=1985


petchema:
Mmmh when there's a lot of buffering the text is skipped way faster than expected, that may require some more testing/tweaking...

PR #1258, opened 2019-04-14
Implement AI pathing to target position

Allofich:
Implements actual AI pathing to targets. With this, enemies are quite a bit better (still far from perfect) at navigating complex areas like building interiors or moving up or down land bridges or chasing the player from room to room.

In master branch the AI basically takes a "blindly feeling along the wall" approach, but this only works for simple obstacles and is pretty useless when up and down is involved. With this PR a grid of 1-unit spaced points is created and searched, using 8-directions and aiming for the direction of the target's predicted position. Attempts are made to recognize climbable slopes and falls. The AI then follows the path it created, whether it successfully reached the target position within the allowed number of tries or not. It follows the path point to point, short-cutting a point ahead if it has a clear path to it and the y-difference is small. 

I've been working on this for a while now and I finally got it to where my test cases seem to all be working OK, but I'm sure problem situations still exist.

Also includes a fix for magic users caused by my last AI PR. I noticed imps could get easily stuck vibrating back and forth. This fixes that.

If you want to test, turn on the "Gizmos" in the upper right corner of the Unity editor to see the pathing. The red lines are the grid of searched points, the blue line is the path result, and the green lines point to the position the enemy is currently trying to get to.

There's a lot of sphere-casting going on, but I haven't really noticed a performance drop in my testing. If it does cause performance drops, we could limit the searching or make it less frequent, randomly space them out a little so AI's created at the same time don't all search at the same time, etc.

### DIFF
--- a/Assets/Prefabs/Scene/DaggerfallActionDoor [Dungeon Serializable].prefab
+++ b/Assets/Prefabs/Scene/DaggerfallActionDoor [Dungeon Serializable].prefab
@@ -3,9 +3,9 @@
 --- !u!1 &122440
 GameObject:
   m_ObjectHideFlags: 0
-  m_PrefabParentObject: {fileID: 0}
+  m_CorrespondingSourceObject: {fileID: 0}
   m_PrefabInternal: {fileID: 100100000}
-  serializedVersion: 5
+  serializedVersion: 6
   m_Component:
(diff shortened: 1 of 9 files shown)


Allofich:
I gave this some testing but it still needs work.

PR #1260, opened 2019-04-14
set *BarGain vital indicators in SynchronizeImmediately

petchema:
Fixes a bug when the effect of some healing was still animated at the time of resetting (say, you where trying to heal yourself when you got slain and reloaded a game).

Also, cleanups

PositionIndicators() side effects are unlikely to change from one Update() to the next, is it worth optimizing?

Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&t=1880#p22163


Interkarma:
>PositionIndicators() side effects are unlikely to change from one Update() to the next, is it worth optimizing?

Possibly, although it would be a small gain. I'm not too worried right now unless it's something you want to tackle short term. Thank you for the fix. :)

PR #1262, opened 2019-04-16
Classic quest fixes

JayH2971:
Mostly removing the instant win glitch and fixing timers to exact values when needed
EDIT: And a lot more stuff. Feature creep at work.


Assets/StreamingAssets/Quests/K0C01Y00.txt
```diff
@@ -121,7 +121,7 @@ Message:  1013
  freelancers.
  
  
-<ce>                          -- The Thieves Guild
+<ce>                          The Thieves Guild
```
Allofich: Does the "--" cause some kind of problem? You've removed a few of these where a letter is being signed with a dash in front of a name.

Assets/StreamingAssets/Quests/A0C00Y00.txt
```diff
@@ -38,9 +39,7 @@ AcceptQuest:  [1002]
 <ce>                       bother trying to find me.
 
 QuestFail:  [1003]
-<ce>             Don't you have a letter to 'deliver' for me?
-                                     <--->
-<ce>         Come back when you have completed your mission, %pcf.
+<ce>             You had a letter to 'deliver' by now.
```
Allofich: Is it possible to make it so the original messages here show if you click the quest-giver before completing the mission? Looks like that is what they were intended for.

Allofich:
Left a couple comments, but I'm not familiar enough with the quests or the way they work to say much more.

I see a good bit more than fixing instant wins (I'm familiar with that issue) and fixing timers (not familiar with that issue), though. I know it would be work to write out, but I think it would be good to know what is being fixed for each quest, unless Interkarma is fine to just trust your work. It seems like it would be tough to review this many changes unless he's familiar enough with the quest file syntax and the way things work/don't work that he can just quickly skim through it all.

I

Allofich:
IMO it would be best to leave comments within each quest file saying what changes were made and why.

Allofich:
To be clear, I do trust that you know what you're doing, it's just that for me, anyway, I don't understand what a lot of the changes are as I'm not that familiar with the quest syntax, and others may be the same. Also I think it's good if we document when we're diverging from the original game for some reason.

Assets/StreamingAssets/Quests/K0C01Y00.txt
Interkarma: The dash indicates a comment in TEMPLATE. My parser will exclude a line when a dash is first non-whitespace character. I'm not 100% sure, as I'm not in front of my dev environment right now, but I think the lines that start with <ce> might be OK. But the left-aligned cases with only whitespace in front of them will be excluded.

Assets/StreamingAssets/Quests/A0C00Y00.txt
Interkarma: I was actually curious about this one too. This text block with a`<--->` separator should randomly display one of variants of flavour text. Was that not working in tests Jay?

Assets/StreamingAssets/Quests/K0C01Y00.txt
JayH2971: The -- edits I made were done blindly. Since this was my first pass I just went for the things that were most obvious without testing them one-by-one, and I've always been wary of using the -- at the start of a line. I wasn't watching whether a <ce> preceded them TBH.

Assets/StreamingAssets/Quests/A0C00Y00.txt
JayH2971: Yes, these 1003 comments are often written in first-person. There appears to have been some difficulty among the devs about how to use 1003 and 1009. To begin, Daggerfall's quest endings are instantaneous, so if you fulfill the failure condition there's never an opportunity to hear the failure statement from the questgiver, which is what 1003 appears to intend here. However, the devs never included a "say 1003" at any time, failure or not. 1003 is almost entirely abandoned. I think it's something they asked Ted Peterson to do and then never elaborated on it.

1009 is a separate case, because it tends to use <ce> flags. 1009 would appear when you engage the questgiver in ordinary conversation and would not execute during a quest. <ce> flags are not to be used for ordinary conversation; we use normal unflagged strings for those.

I added a third-person 1003 here because classic often lets time run out on a quest without any forewarning, and then people get confused and say, "I went to the town but the person wasn't there," and we don't know what went wrong until we open the quest parser. Some sort of failuyre information should be displayed, and the classic one was out-of-place.

JayH2971:
I completely agree. I'm so used to doing everything on my own that comments have been superfluous for me, but since this is a group project there's a need. I'll start adding them now, and I can go back to the more obscure decisions and add comments.

Assets/StreamingAssets/Quests/A0C00Y00.txt
bansheexyz: Allofich is saying that some quests (like this one) have messages that fire from the questgiver if you try to talk to them while the quest is unfinished. They are not responses to failure, and if they are fired that way, it is an error in the quest file. As an example of how they should work, see how 1015 fires in a0c00y08. These lines should be retained and fired off appropriately if possible.

1008 and 1009 I think show up in the conversation window when you talk to the person post-quest. They do not fire as click-through messages.

I don't think adding text to quest files is the right way to go about letting the player know when quest time expires. There should probably be a one-size-fits-all message that fires in the scrolling infotext area. Something like "Time has run out on your agreement with _qgiver_". The journal has an area for failed quests to be shown now.

Interkarma:
Jay, I've been considering not doing comment checks during text blocks. This will solve the problem of using a dash in comments, and comments can still be used outside of text blocks normally. Will that work in with you?

Assets/StreamingAssets/Quests/A0C00Y00.txt
JayH2971: I see now. I've looked through the 1003 messages for every classic quest, and universally they're a "get-out-of-here" message the game never implemented. I can do that if that's an improvement.

There are a few occasional instances when classic tells you you've failed a quest due to time-out, but I think that's a big defect. Whether your solution or something else gets added, there really should be some alert to the player about it.

JayH2971:
> Jay, I've been considering not doing comment checks during text blocks. This will solve the problem of using a dash in comments, and comments can still be used outside of text blocks normally. Will that work in with you?

I see no downside. Everything we write in the text section should be read, after all :) I don't think I've left any hyphenated comments in the text sections so this shouldn't open any Pandora's Boxes on us.

JayH2971:
A new issue I'm running into is what to do with "create npc _npc". This is used abundantly in classic to both generate and place the appropriate townsperson at the appropriate place. DFU uses much more liberal bounds to do so, and sometimes a townsperson that needs to be local is generated in a remote town. I'm sure this can be fixed engine-side at a proper point, but for now those quests are problematic and often not completable. I'm happy to replace them with "place npc _npc_ at _appropriateplace_", but I want to be sure everyone else agrees with doing so beforehand.

Allofich:
I prefer that we just try to fix that engine-side, especially if it is used abundantly.

Interkarma:
I'm with Allofich on this one. If it's an emulation problem, I'd rather fix engine side. Could you just list a couple of the quests experiencing this issue Jay? I add it to my internal tracker.

Assets/StreamingAssets/Quests/10C00Y00.txt
```diff
@@ -72,6 +72,7 @@ Message:  1013
 Meridia is a Daedra Regent and a collector of human specimens.
 <--->
 Meridia is a kind of acting Daedra Prince, and ruthless in her acquisitiveness.
+--unused string
```
Interkarma: Jay, if I stop comments being processed in text blocks as discussed, then this kind of thing will become part of the text block. A text block only ends once there's an empty line.

Assets/StreamingAssets/Quests/20C00Y00.txt
```diff
@@ -169,8 +170,11 @@ _1stparton_ task:
 	end quest 
 
 variable _2ndparton_
+--unused variable
 _clearclick_ task:
-	when _qgclicked_ and not _2ndparton_ 
+	when _qgclicked_ and not _mondead_ 
+	--erroneous variable use
```
Interkarma: I find this sort of commenting a bit unclear. Once the line has been fixed, the comment is no longer relevant. Once the file is merged, it will just seem to indicate a problem that's no longer there and will become confusing later.

Assets/StreamingAssets/Quests/10C00Y00.txt
JayH2971: True, I forgot that'll be changed soon. Thanks for the catch. IIRC that's the only one but I'll take another look.

Assets/StreamingAssets/Quests/20C00Y00.txt
JayH2971: I'm only using comments so far to justify the changes I'm making. Once everyone's satisfied I don't see a need for them, and I figured it'd be just as easy to remove them in the future. Once a quest is fixed and working, I don't see a need for artifact comments from the past to remain in them.

Interkarma:
I was thinking about putting this PR onto the roadmap and making it the focus of a mid-month update during 0.9 sometime. Are you comfortable with that? No pressure either way, I'm happy to let this one bubble away much until closer to 1.0 if you prefer. :)

JayH2971:
> I was thinking about putting this PR onto the roadmap and making it the focus of a mid-month update during 0.9 sometime. Are you comfortable with that? No pressure either way, I'm happy to let this one bubble away much until closer to 1.0 if you prefer. :)

Sure, that's fine. The testing process is going to be really long, so I can't put any official date or deadline to it. But the vast majority of fixes should be in place by now, and all the 1003 messages are now implemented.

JayH2971:
> I was thinking about putting this PR onto the roadmap and making it the focus of a mid-month update during 0.9 sometime. Are you comfortable with that? No pressure either way, I'm happy to let this one bubble away much until closer to 1.0 if you prefer. :)

On second thought, how about 1.0? My commitment right now is extremely sporadic and 0.9 is racing toward us. I don't think I'll be able to meet it.

Interkarma:
That's fine mate. Let's look at this sometime in 0.10, 0.11, etc. as a standalone update. That whole alpha/beta period is a great time to issue small themed updates.

JayH2971:
I'm making a google sheet to show the progress with these quests:

[(http://docs.google.com/spreadsheets/d/1W5R54I92BpisrldTuQj909wxqdwMneKQLG4VOI0x0sI/edit?usp=sharing)]

JayH2971:
In the R quests, there's something I'm fumbling around. I'm meeting multiple examples of Local_3.0, which forces the NPC to spawn in the same city as the questgiver, contrary to the quest's design. If I change it to Resident1-4, they spawn in another city but never female as the quest asks. I don't know what faction/group to resolve these NPCs to.

Interkarma:
I'd be interested in your take on how the Local_X.Y faction should generate an NPC (and where). In my comments, I took this to mean NPCs should be generated locally (rather than remotely) but that's obviously incorrect. I'd be happy to improve this if we can work out what these groups are supposed to do. Some stuff like this is a still largely unknown. :)

There are also easy ways to work around this in the quest script, but I'd like to fix the NPC generation first if possible. Your experience and observations spending so much time in the quests would be welcome here.

JayH2971:
I agree with fixing NPC generation. The problem is that I've spent so much time making my own quests and not trying classic ones, I don't have much info on how they should work. I'll have to slow down for this section and see how classic runs them.

JayH2971:
I added a new sheet to the google calc above to show the spawns for various Locals in quests. The results are not illuminating. The quests were all accepted from Lord Provlith in Castle Daggerfall.

Interkarma:
@JayH2971 - Hi Jay! Would you object if I merged this to a branch and then into master sometime during 0.9 cycle? Perhaps as a "classic quest fixes: stage 1"?

I know you still have more to do, but I've working through most of what you have so far, and you've fixed so many problems already. I feel it might also help the work to feel smaller for you if we merge what you've done so far. Then you can continue where you left off later feeling a bit refreshed. It also gives users a chance to play with a more stable quest base than they have in 0.8.

You've done some really solid work here. It would be amazing to let it go into the world, as frightening as that might feel sometimes. What do you think?

JayH2971:
By the Cliffs of Solitude! You should've sent me a PM or something about this! Yes, some fixes will still be problematic but I'm happy to double back as needed. Publish away.

A question came up on Discord: https://forums.dfworkshop.net/viewtopic.php?f=28&t=968&start=10#p11997 says you disabled the alternate Cyndassa werewolf path due to the incomplete magic system. Do you intend to bring the quest back to the Donald Tipton version?

Interkarma:
>By the Cliffs of Solitude! You should've sent me a PM or something about this! Yes, some fixes will still be problematic but I'm happy to double back as needed. Publish away.

Thanks Jay! Merging this to its own branch for now. If you want to make further changes to this lot, please push them to the new branch "classic-quest-fixes-part1". I had to resolve a few conflicts where we'd made the same fix.

>Do you intend to bring the quest back to the Donald Tipton version?

There were other problems with it as well from memory. This may have changed now that emulation compatibility has improved. I also somewhat feel this one belongs more with Tipton's Werefall mod than part of the default game. But if it's something you guys have a strong preference for, I'm happy to merge and make sure it works. :)

PR #1265, opened 2019-04-17
Added support for cursor override

TheLacus:
Allows mods to provide a custom cursor texture. This must use the 'Cursor' texture type and have Read/Write flag enabled. Import from loose files is not supported.


Interkarma:
This is awesome! Thanks for doing this. :)

PR #1266, opened 2019-04-18
Get a region vampire clan from factions.txt

df-ferital:
Classic uses the "vam" faction field of a region faction data to get its corresponding vampire clan. This makes DFU behave the same.

Also did a little refactoring of some methods I used there to avoid code duplication.


Interkarma:
Thank you. :)

PR #1267, opened 2019-04-20
Fix typo in AreEnemiesNearby

numidium:
@petchema brought this to my attention. Looks like I put an OR where an AND should have been.


petchema:
I noticed that issue, but I'm far from sure what are all the consequences of teams on IsHostile... 
Should PlayerAlly check be added each time?
For example, should GameManager:649 be changed from
`if (includingPacified || enemyMotor.IsHostile)`
to
`if (includingPacified || enemyMotor.IsHostile && entity.Team != MobileTeams.PlayerAlly)`
?

numidium:
I see your point, @petchema.  At the moment it likely won't have much of an effect since the only player-allied monsters in the game that ever appear are the Daedroths from the Sanguine Rose.  IsHostile is typically used as a catch-all for determining if the monster is hostile _towards the player_ but now that we have player allies I should take that into account.

PR #1268, opened 2019-04-21
Fixes to AI spell casting/shooting

Allofich:
1. Main fix here is that the prediction counter is no longer reset whenever the AI checks for whether it can shoot a spell. This was causing the imp and others with ranged spells to get stuck. (In master branch create an imp with "cm 1" and they should get stuck almost immediately. With this PR they are fixed. This problem surfaced after one of my recent PRs)

2. Enemies can now cast any spell as long as they have at least 1 spell point left. This matches classic and should allow them to cast the same number of times. It's also relevant to the classic AI's spell decision, as it checks for whether it has any spell points left when deciding if it can still cast. (Otherwise, we'd need to check whether it has at least the spell points of its cheapest ranged spell or something like that)

3. Cleaned up logic and treasure loot/other spells are ignored when the AI checks if they can shoot. They can now shoot projectile spells more often. (Tested with the mage on the pedestal in the entrance room of Woodborne Hall. In master they just wander around trying to find a way to you without a fall, without shooting)

4. Minor cleanups

5. Added a fix for when an AI gets stuck falling. I've only had this happen once so I think it's pretty rare, but the AI could get stuck while falling if the `controller.SimpleMove(Vector3.zero);` that applies gravity failed to move it downward, as it would return without updating `lastPosition`. Changed the order of code so that `lastPosition` will update and it won't get stuck.

### DIFF
--- a/Assets/Scripts/Game/EnemyMotor.cs
+++ b/Assets/Scripts/Game/EnemyMotor.cs
@@ -321,18 +321,6 @@ namespace DaggerfallWorkshop.Game
                 return;
             }
 
-            // Apply gravity
-            if (!flies && !swims && !isLevitating && !controller.isGrounded)
-            {
-                controller.SimpleMove(Vector3.zero);
-
-                // Only return if actually falling. Sometimes mobiles can get stuck where they are !isGrounded but SimpleMove(Vector3.zero) doesn't help.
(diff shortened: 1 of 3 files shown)


Interkarma:
Thank you for continuing to fix and refine the enemy AI. Without all your help, this would probably still be no better than it was in the early DFTFU days. The game is so much better for these changes - all of your input has been incredible really. :)

Allofich:
Thanks, I appreciate the kind words!

PR #1270, opened 2019-04-22
Paying spellmaker and spellbook check

petchema:
Supersedes #1252 (Deducts gold when buying spells from spellmaker)

Also, check for spellbook presence. In classic, when you have no spellbook:
 

- spell merchant does nothing, not explaining why;
- spell maker fails at the last minute ("you have no spellbook to inscribe this spell")

I think that in both cases the player, even without a spellbook, could want to peruse available spells, so I implemented a preliminary warning that does not block the access to the spell merchant and the spell maker.
A second check is done if you try to buy a spell though.
    
Maybe that's overkill and confusing, and access could simply be denied with an error message?
The simplification I'm considering:

- remove spell merchant warning. The player can peruse the spells, but will be denied to buy any. There's not much effort investment from the player so that should be ok;
- display a blocking error message when trying to access the spell maker


petchema:
Finally I prefer the simplified version, I think it will be less confusing

Interkarma:
Thank you so much @goodbadbugs and @petchema for both of your work on this recently. I've added @goodbadbugs to the DFU credits list, and I'll credit both of you in the change log update.

PR #1271, opened 2019-04-24
Make Texture Arrays with CopyTexture

TheLacus:
Terrain Texture Arrays are now created with Graphics.CopyTexture on supported platforms, which is much faster than using Get/SetPixels.

1. Get/SetPixels
2. CopyTexture (DXT1)
3. CopyTexture (RGB)

copyTexture

Beside the fact that the format is the one used by individual textures (which can be compressed), the improvement is that CopyTexture performs a GPU-side copy between textures, which is faster than uploading with Apply(). This also means that Read/Write flag can be unchecked, unless a mod wish to target older platforms with no support for this feature.

### DIFF
--- a/Assets/Scripts/Utility/TextureReader.cs
+++ b/Assets/Scripts/Utility/TextureReader.cs
@@ -10,6 +10,7 @@
 //
 
 using UnityEngine;
+using UnityEngine.Rendering;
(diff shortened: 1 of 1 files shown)


Interkarma:
A very nice saving, well done! :)

PR #1272, opened 2019-04-28
make sure wagon button is fully reset by buy window

petchema:
DaggerfallTradeWindow.OnPush() can be called before Setup() so wagonButton can be null in OnPush(), and its texture not settable there. 
So to cover all the bases the button background texture must also be set in Setup().

Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=2040

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTradeWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTradeWindow.cs
@@ -203,7 +203,6 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             SetupTargetIconPanels();
             SetupTabPageButtons();
             SetupActionButtons();
-            ResetWagonButton();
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTradeWindow.cs
```diff
@@ -613,6 +613,12 @@ protected override void FilterRemoteItems()
                 base.FilterRemoteItems();
         }
 
+        protected void ResetWagonButton()
```
petchema: would it have been more robust to just use ShowWagon(false)?

Interkarma: I think so yeah.

Interkarma:
Thank for looking at this. I greatly appreciate the help on all these issues. 👍

PR #1273, opened 2019-04-28
fix city gates regression

petchema:
Broken in 94a2c78b

Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=2002

### DIFF
--- a/Assets/Scripts/Utility/RMBLayout.cs
+++ b/Assets/Scripts/Utility/RMBLayout.cs
@@ -814,7 +814,7 @@ namespace DaggerfallWorkshop.Utility
             // Is this a city gate?
             if (IsCityGate(modelID))
             {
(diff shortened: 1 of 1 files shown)


Allofich:
Hm. Removing that comment causes

`Assets/Scripts/Utility/RMBLayout.cs(817,36): warning CS0219: The variable `gate' is assigned but its value is never used`

to appear in the Unity editor, but you're sure this fixes the issue? This warning is a false positive?

Interkarma:
The component needs to be added, but it doesn't need to be stored as "gate". Can just remove that and leave the component intact for gate handling.

Thanks for looking into this Pango. :)

Allofich:
Ah, I see. Sorry for not being more careful when I originally commented it out.

Interkarma:
Don't worry about it, I didn't catch it when I reviewed either. :)

PR #1274, opened 2019-04-28
fix mobile town NPCs spawning in front of player

petchema:
distanceToPlayer was updated too late, for recycled npcs that can make a lot of difference!

Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=1997


Interkarma:
Very simple and sensible. Thank you. :)

PR #1275, opened 2019-04-28
fix for building-to-building teleport ending in the Void

petchema:
In case current and anchor buildings have the same buildingKey, map pixels must also be compared to be sure it's not the same building.

Discovered data doesn't look good, that may not be the best possible fix.

Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=2018


Interkarma:
Good fix, thank you so much! :)

PR #1276, opened 2019-04-28
import enemies and loot on teleport in dungeon

petchema:
Everything's in the title, when teleporting into a dungeon from some
other place, import dungeon enemies (and loot).

Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=1661


Interkarma:
Thank you again! :)

petchema:
Just watched Lay Lilia video on her selection of mods, but beside that main topic she mentions an issue with Persistent Dungeons when teleporting back into a dungeon...
Looks like my patch implements classic behavior fine, but does not play well with persistent dungeons (namely, monsters and loot are unconditionally respawned)

petchema:
I'm not sure what issue she's talking about, I enabled Persistent Diungeons, did a partial "cleanup" of both Scourg Barrow and a normal dungeon, set an anchor, exited the dungeon, recalled back, and each time dungeons state seems to have been correctly restored (??)

PR #1279, opened 2019-04-30
scrolling speed adjustments

petchema:
Exponential speedups did not work as well as the math suggested, with large lags the acceleration made all lines disappear from screen in a blink, including the most recent ones; Or get stuck because speedup overflows floats capacity!
Replaced with simple rubberbanding (proportional speedup) for now. I don't like that acceleration profile as much, but it fixes those issues. That may be improved later.
Added a "addpopuptext" console command to push n lines in the buffer for testing.


Interkarma:
Thank you for continuing to refine. :)

PR #1280, opened 2019-05-01
Custom render mode for billboards

TheLacus:


TheLacus:
Added a quick fix to address https://github.com/Interkarma/daggerfall-unity/commit/6eac98936d8606be257737f86d58d321df079e34#commitcomment-33378240.

I'm going to rework it to make its own material with the render mode specified by xml file.

Interkarma:
Thank you for the quick fix as well. :)

PR #1281, opened 2019-05-01
Cleanup and fixes for EnemyMotor

Allofich:
-Cleanup-
Use auto properties
Divide long Move() function into smaller functions
Add/correct comments, renaming

-Fixes, improvements-
Allow backup for out-of-pitch target to move smoothly
Allow knock-back for paralyzed enemies
Knocked-back flyers will drop while being knocked back
Use direction, not movement, consistently with obstacle-check functions
Refined obstacle and fall detection
Smaller radius used when checking for arrow shot. Will shoot over tables
now.
Account for controller changing height when colliding with objects by
storing original height
Stop flyers from flying near floor during combat
Flyers/swimmers/levitating enemies check up and down when detouring

### DIFF
--- a/Assets/Scripts/Game/EnemyAttack.cs
+++ b/Assets/Scripts/Game/EnemyAttack.cs
@@ -304,7 +304,7 @@ namespace DaggerfallWorkshop.Game
                 }
 
                 // Knock back enemy based on damage and enemy weight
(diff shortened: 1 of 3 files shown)


Interkarma:
Thank you!

PR #1283, opened 2019-05-03
fix negative repair times when removing other repaired items

petchema:
**Background**

There's a base time for each item, proportional to the amount of damage it took. Then the time for the longest item to repair is increased by half the time taken by all the others (rationale, I assume, is that
while working on other items, work on this item is slowed down by half. On the other hand all other items are worked on in parallel, mmh). Anyway, it gives some incentive to repair several items at once, yet it's not totally for free.

Now what happens if you add items when some repairs are already ongoing? 
Implementation in DFU just reevaluates everything together; So old repair times may adjust if you add items to the mix (I've checked that classic does that too), and in the end you get the same times as if you put everything to repair at once, so it seems it's not possible to cheese the system that way.

**Issue**

Problems arise when some repairs complete, or worse if the repaired items are removed from the list: because if you keep reevaluating the repair times, suddenly the repair time for the longest repair goes down!

**Fix**

This patch just prevents repair times from decreasing; This should fix the problem at hand, should prevent most exploits (if you add your "lesser damaged items" in batches so that the same item always get the 50% penalty, you can still save some time), in some obscure cases it will be unfair (if you attempt the same trick and the 50% penalty goes to another item, some penalty may be accounted twice), but it stays relatively simple to implement...

Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=2053

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTradeWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTradeWindow.cs
@@ -433,8 +433,16 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             Debug.Log("UpdateRepairTimes called");
             int totalRepairTime = 0, longestRepairTime = 0;
             DaggerfallUnityItem itemLongestTime = null;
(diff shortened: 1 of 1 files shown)


Interkarma:
Thank you for fix!

PR #1285, opened 2019-05-04
Add Skull of Corruption effect and Tweak Sanguine Rose

numidium:
I made it so both the Skull and Rose only spawn allies if the monsters nearby are not allied with the player.  I'm aware that the Skull in classic only clones non-human enemies in classic but I didn't feel the need to limit it in that way (I'm willing to change it if there are objections).  The behavior otherwise matches classic to the best of my knowledge.  Please let me know if I'm missing anything.

On a side note, I'm impressed at how simplified the process of adding new artifact effects is now.  Well done, Interkarma!


Allofich:
> I'm aware that the Skull in classic only clones non-human enemies in classic but I didn't feel the need to limit it in that way

It works on all enemies in classic, so no problem there. I just confirmed by testing in classic.

Interkarma:
Thank you numidium, for the new effect and the compliment! :)

PR #1286, opened 2019-05-04
TextureReplacement: Initial support for billboard animations with TextureArray

TheLacus:


TheLacus:
Added initial support for drawing billboard animations with a texture array. I consider this to be experimental at the moment so it must be enabled with xml file, otherwise individual textures are swapped as before. I'd appreciate if you could give a quick look at the shader because i don't have any experience with this :)

Interkarma:
I'll need a little bit more time to review this one. I'm hoping to drop some bug-fix builds later today. Would it be possible to split the animated material fix and TextureArray changes? That way I can get the fix out sooner and still have time to review other changes later. :)

TheLacus:
Sure, see #1288

TheLacus:
Rebased on master.

TheLacus:
I'm reworking this to import texture arrays created from the Unity Editor and stored as assets.  This is the idea:

- Not animated: Import a single texture2D named `Archive_Record-0` to use with the standard shader.
- Animated
  - Mods: Import a single texture2Darray named `Archive_Record-TexArray` (`Archive_Record` is already in use for GameObjects that replace billboards).
  - LooseFiles: Import individual textures named `Archive_Record-Frame` and create a texture2Darray at runtime.

Pros:
- A texture array created inside the editor can be bundled and loaded from an assetbundle, no time is spent for creation and, most of all, no garbage is created at runtime.
- A texture array created from loose files would still be a better alternative to texture swaps on material.

Cons:
- An additional step for mod creators.

The ability to store a texture array as an asset is only available via Unity API, but i actually already wrote a basic editor script to ship with the modding tools. In alternative, there are some advanced tools in the Asset Store available for free.



Of course i'm open to any feedback before pushing this to core :)

TheLacus:
Need to check which archives are used with billboard batch and already use an atlas, to avoid creating duplicated resources. I believe these are only nature, lights and animals, while the misc atlas was disabled last year. So texture arrays can be used for static npcs and minor things like blood and missiles. @Interkarma  Is there something missing?

Once this is all working the next step is mobile npcs and enemies. I think they would benefit the most of texture arrays instead of current solution based on texture swaps.

The issue is that monsters often have different shapes for front and side so some empty space must be wasted as padding to respect array requirement of equal sizes. Mods should also provide a file (something like `Archive-Config.json`) with a list of rects for all records. Current solution only requires individual textures to be provided so i'm not sure if is worth it... I'm open to opinions on this.

Interkarma:
>I believe these are only nature, lights and animals, while the misc atlas was disabled last year. So texture arrays can be used for static npcs and minor things like blood and missiles. @Interkarma Is there something missing?

That sounds correct. I just had a quick scan through code, and can't see any other batches that we've forgotten about.

>The issue is that monsters often have different shapes for front and side so some empty space must be wasted as padding to respect array requirement of equal sizes. Mods should also provide a file (something like Archive-Config.json) with a list of rects for all records. Current solution only requires individual textures to be provided so i'm not sure if is worth it... I'm open to opinions on this.

While I think it's a worthwhile change to reduce memory overhead for enemies by requiring less empty space in padding, I'd hate to break mods like Dream or make it too difficult for them to transition to the new system. Mod authors are the primary "customer" of asset injection so they should probably have more input than myself. Maybe do a quick mockup of what you have in mind with some estimated memory savings to get feedback before committing all the way. Then we should be able to determine if the cost in time to both yourself and mod authors is worth the results. :)

TheLacus:
> While I think it's a worthwhile change to reduce memory overhead for enemies by requiring less empty space in padding, I'd hate to break mods like Dream or make it too difficult for them to transition to the new system. Mod authors are the primary "customer" of asset injection so they should probably have more input than myself. Maybe do a quick mockup of what you have in mind with some estimated memory savings to get feedback before committing all the way. Then we should be able to determine if the cost in time to both yourself and mod authors is worth the results. :)

Thank you for your answer. I wasn't thinking of memory saving but sharing an idea for a possible implementation of texture arrays. Npcs found inside buildings use a single record, whose frames all have the same size, so i think texture arrays are a perfect solution.

Mobile npcs use an entire archive to show different sides, animations etc. so my idea is to "blit" all records to the smallest size that can hold all of them. Rects are required to sample them back, but they should be created automatically, or at least with little effort for mod creators. My previous message might have suggested that i was thinking of manually created rects because i also talked of creating texture arrays from the editor, which means storing rects as metadata. 

My concern is actually that memory usage may be slightly increased, so we should decide if performance improvement is worth it. I'm going to make some test so i can bring actual results to the discussion. For the moment i'm also putting aside the idea of arrays stored as assets in editor so i can focus on one thing at a time. I'm with you about the involvement of mod authors; my idea is an opt-in option so we can get feedback before making any persistent change :)

Interkarma:
I've happy with the direction and work so far. Are you comfortable for me to merge for upcoming May builds?

TheLacus:
Thanks, i don't think is ready to be merged. I prefer to skip upcoming build to think more about it and discuss with mod creators. I'm closing for now, will reopen if we decide to go with it.

PR #1289, opened 2019-05-05
Implement Ring of Namira effect

numidium:


Assets/Scripts/Game/Formulas/FormulaHelper.cs
```diff
@@ -623,6 +621,21 @@ public static int CalculateAttackDamage(DaggerfallEntity attacker, DaggerfallEnt
 
             DamageEquipment(attacker, target, damage, weapon, struckBodyPart);
 
+            // Apply Ring of Namira effect
```
Interkarma: This is fine here for now. :) Just FYI though, I'm looking into an OwnerDamage enchantment callback when the owner of such enchanted items receives damage from another entity. This would allow your RingOfNamiraEffect  to simply request this callback and operate without needing to do a search here. Once I have a design I'm happy with, I'll port this to the new callback. I'll likely merge as-is for now though until I get to that.

Interkarma: Could wrap this in a `if (target == player) { }` so it doesn't go searching for Ring of Namira each time player strikes an enemy. And rather than search all equipped items, could just do it for ring slots. Check out how I did [something similar](https://github.com/Interkarma/daggerfall-unity/blob/master/Assets/Scripts/Game/MagicAndEffects/Effects/Special/LycanthropyEffect.cs#L349) for Hircine's Ring in LycanthropyEffect.

numidium: I had the notion that someone might want to equip artifacts on NPCs at some point but in hindsight that's probably not likely. The search is narrowed to just the player's ring slots now.

Interkarma: Awesome, cheers mate! :)

PR #1290, opened 2019-05-08
loans to be repaid one year later

petchema:
Loans were due immediately
Bug was not obvious because displayed due date does not show the year

Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=2072

### DIFF
--- a/Assets/Scripts/Game/DaggerfallBankManager.cs
+++ b/Assets/Scripts/Game/DaggerfallBankManager.cs
@@ -76,6 +76,7 @@ namespace DaggerfallWorkshop.Game.Banking
         public const float goldUnitWeightInKg = 0.0025f;
         private const float deedSellMult = 0.85f;
         private const float housePriceMult = 1280f;
+        private const uint loanRepayMinutes = DaggerfallDateTime.DaysPerYear * DaggerfallDateTime.MinutesPerDay;
 
         #region Ships:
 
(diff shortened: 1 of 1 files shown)


petchema:
As a general observation, this mechanism doesn't look fully fleshed out yet: Defaulting is only checked when banking window is opened. Also, there's no way to clear your banking status once defaulted.

Interkarma:
Thankyou for that fix. I'm surprised I never caught this earlier, obviously didn't spend enough time testing loans. Would you be interested in fleshing out the mechanism?

petchema:
Since it wasn't fleshed out in classic either, the first problem will be to decide what should be implemented.
Found that old thread of the topic: https://forums.dfworkshop.net/viewtopic.php?f=12&t=1388
It contains interesting ideas, maybe the discussion can be continued there...

Interkarma:
I look forward to seeing what you do with it. :)

Allofich:
There is a function in classic to spawn bounty hunters when you have defaulted. It's something that I have meant to recreate at one point, should be simple to do.

Allofich:
You probably know my opinion, but anyway... I haven't read everything in that forum thread but I think as default we should just keep it simple, spawn the bounty hunters of classic and have the bank refuse further loans if you defaulted (that message is already implemented I think), and anything beyond that be optional somehow.

There are also changes I've thought about that could be made to guard/crime behavior like:
- Being charged for every crime you commit, not just the latest (how to handle the courtroom messages, and will the sentence be too long?)
- Guards attacking monsters (how to handle quest-generated foes, and who should they attack between player and human or human-appearance enemy? Should they try to arrest or just kill?)

but they have issues to consider and I would be sure to make these optional if I ever add them.

Maybe a "more realistic world interactions" or some such name for these kinds of changes to shops/banks, guards/crime, etc. If the player wants one such change, I think they probably want all of them.

petchema:
Many things are out of my league though. I also think a lot of those loan features would be better implemented as a quest (from bank sending reminder letters, to handling bounty hunters behavior,...): if an instance of a quest can be started as you take a loan, and some quest conditions be injected from C#, I think the rest would be easier to maintain as a quest... except I don't know how to write one ;)

The other idea developed in the forums is collaterals; Both to cap how much you can borrow, and decide what bounty hunters can reclaim. It adds a lot of complexity though, and there's no sign of those being a plan in classic Daggerfall so it feels less legitimate. I think it could be interesting, other just bad game design. In short, at best it's optional material as you said.

PR #1295, opened 2019-05-12
Weaken shadows under scattered light

petchema:
When light is scattered, there should be less hard shadows (only ambient occlusion shadows); This is most obvious during heavy fog.

So I modified the shadows to be
- slightly lighter during overcast weathers;
- very light during snowing weather;
- almost inexistant during fog.

### DIFF
--- a/Assets/Scripts/Game/WeatherManager.cs
+++ b/Assets/Scripts/Game/WeatherManager.cs
@@ -34,6 +34,8 @@ namespace DaggerfallWorkshop.Game
         public AmbientEffectsPlayer WeatherEffects;
 
         [Range(0, 1)]
+        public float OvercastSunlightScale = 0.65f;
+        [Range(0, 1)]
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Game/WeatherManager.cs
```diff
@@ -283,22 +296,45 @@ void SetSunlightScale()
...
                 scale = SnowSunlightScale;
+                shadowStrength = SnowShadowStrength;
+            }
+            else if (IsOvercast)
+            {
+                scale = OvercastSunlightScale;
+                shadowStrength = OvercastShadowStrength;
+            }
+
+            shadowStrength *= Mathf.Exp(-50f * currentOutdoorFogSettings.density);
```
petchema: I could have added shadowStrength to FogSettings, but that would have broken the compatibility with Enhanced Sky...
So I used a formula instead

PR #1297, opened 2019-05-15
only dip head in front of static obstacles

petchema:
Do not automatically dip for moving obstacles, like missiles (including your own).

On the other hand, I wonder if enemies can't dodge missiles because of a similar mechanism (in EnemyMotor.HeightAdjst).
Maybe that's the missiles that should be handled differently? I don't know if that's even possible.

Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=2089


Interkarma:
Thank you!

PR #1298, opened 2019-05-16
Fix navigation in character creator

numidium:
Addresses issues described in https://forums.dfworkshop.net/viewtopic.php?f=24&t=2075 and https://forums.dfworkshop.net/viewtopic.php?f=24&t=2051.


Interkarma:
Thank you!

PR #1299, opened 2019-05-16
Reworked the UV unpacking method

df-ferital:
After extensive experiments, I came to a better UV unpacking method. Actually, only the U texture coordinate needs to be processed.

This fixes some side effects of the previous method, like some house roofs or the Daggerfall Castle treasury room, as mentioned here https://forums.dfworkshop.net/viewtopic.php?f=5&t=1899&start=30.


df-ferital:
Closing for now, needs more refinment.

PR #1300, opened 2019-05-16
Spatialized spell missile hit sounds

petchema:
tweak minDistance and maxDistance instead of spatialBlend and volume

volume is never effectively amplified, so setting volume > 1f has no effect.
Instead, one can increase minDistance, so the the sounds stays at nominal volume at a larger range.
maxDistance sets at what distance the sound stops becoming quieter with distance. This is actually better than using spatialBlend, because the sound stays directional, which is important to know where are enemies fighting.

Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=2089&p=24412#p24412


Interkarma:
I like this a lot! And such a clever solution too. Cheers. :)

PR #1301, opened 2019-05-18
Make random spymaster NPCs act like non-guild

ajrb:
There are some NPCs marked as TG spymasters in various houses, I had suppressed ones in House2 buildings already. I've changed this so all misplaced spymaster NPCs now have special case handling so they act just like regular NPCs instead of just making them disappear. Basically the same as I did for holy order NPCs that are not in a temple. (viewtopic.php?f=28&t=1238)

I also fixed it so the responses are now general ones for spymasters, left it alone for the priests since they already use a parent faction with no guild group.

Additionally the guild responses were getting random messages using strings when the msgs being used have formatting tags, so I fixed that too.


Interkarma:
Thank you for the fixes mate! :)

PR #1303, opened 2019-05-19
Loan improvements

petchema:
Loans are checked every night, so a default can occur while you're not interacting with the bank.
As a side effect, you can repay your loan anytime during the due day (well, before 15:00 at least ;) )

Reminders are sent to the HUD when due date becomes less than 6, 3, and 1 months away. It would be even better if banks could send letters someday...
Banking status (accounts, loans) for all active regions can be checked from the character sheet, by clicking on GOLD status.

When due date is reached, the bank will first try to pay the loan using your account money. If that works, the loan will just be cleared.
Otherwise, default occurs with the bank, and your regional legal rep will decrease. The bank will also reject any later loan request.

Currently, the legal rep loss is not so huge that the city guards will go after you just for that crime; However I wanted to keep the legal rep loss consistent with other crimes.
It seems to not have been the case back in the day either https://groups.google.com/forum/#!searchin/alt.games.daggerfall/loans%7Csort:date/alt.games.daggerfall/JjCGQXFyuew/Pf9HMApBxSoJ


ajrb:
I thought the banks were supposed to send bounty hunters after the player, although that seems to not have actually been implemented in classic DF. I think that we were planning to have that in DFU, but I can't be sure and TBH I don't fancy trawling through forum posts right now. Anyone else recall?

Assets/StreamingAssets/Text/DaggerfallUI.txt
```diff
@@ -73,6 +73,16 @@ potionMixed,        Your potion has been mixed.
 potionFailed,       Those ingredients did not concoct an effective potion.
 reqIngredients,     You do not have the ingredients required.
 
+- Banking statuc:
```
ajrb: Static. :)

Assets/Scripts/Game/DaggerfallUI.cs
```diff
@@ -1309,6 +1311,68 @@ public DaggerfallMessageBox CreateHealthStatusBox(IUserInterfaceWindow previous
             return healthBox;
         }
 
+        public DaggerfallMessageBox CreateBankingStatusBox(IUserInterfaceWindow previous = null)
```
ajrb: Why is this in here and not in a banking class?

petchema: There's nothing related to GUI in banking class, would you really prefer that method to be there?

petchema:
> I thought the banks were supposed to send bounty hunters after the player, although that seems to not have actually been implemented in classic DF. I think that we were planning to have that in DFU, but I can't be sure and TBH I don't fancy trawling through forum posts right now. Anyone else recall?

I'm not sure what you're referring to, is it this thread?

https://forums.dfworkshop.net/viewtopic.php?f=12&t=1388

Anyway, bounty hunters looks too complex for my understanding of Daggerfall Unity. I think this patch will be compatible with such change, but I'm probably not the one that will implement it.

Interkarma:
>Anyway, bounty hunters looks too complex for my understanding of Daggerfall Unity. I think this patch will be compatible with such change, but I'm probably not the one that will implement it.

The quest system would be perfect to deliver this, it can do almost everything required. Perhaps Jay_H or Kamer would be interested in writing this quest? You'd just need to start and stop quest:

* When loan first defaults, start bounty quest and cache quest UID as part of the serialized banking data.
* When loan repayed, stop quest based on cached UID.

You could even start/stop quest when player is the region where default exists. This would allow quest to use macros for current region in flavour text (i.e. stuff the hunters say, or letters they carry).

Assets/Scripts/Game/DaggerfallUI.cs
ajrb: There's an entire UI banking class, could go in there. I'm asking more than saying it's wrong.. was curious is all.

ajrb:
Yeah, I think you're right - so the code can just start up the quest, like the main quests. It can even go in that DFU folder I created for DFU custom quests. (I knew we'd find a use eventually!)

@petchema  Great to see this area getting some love mate!

JayH2971:
I'm happy to write a quest for it. I'll need to know the general framework and what it's expected to accomplish.

Assets/Scripts/Game/DaggerfallUI.cs
petchema: That's an option... Their lifetimes do not match, but I think this method can be made static, so that's an option

petchema: To be perfectly honest, in this case I used CreateHealthStatusBox() as an example.
But I agree that it adds bloat to DaggerfallUI; maybe CreateHealthStatusBox should also be put somewhere else...

petchema: I moved the code as static methods into DaggerfallBankingWindow.
Not great, but most banking UI code is now in one place.

Interkarma:
I'm looking at dropping new builds on Friday. Are you generally happy with how this one is for now? I can leave open if you'd like to keep working on it. :)

petchema:
I think it can be released as-is, new features can be added as later patches

Assets/Scripts/Game/LoanChecker.cs
```diff
@@ -0,0 +1,72 @@
...
+using System.Collections.Generic;
+using DaggerfallConnect.Arena2;
+using DaggerfallWorkshop.Game.Banking;
+using DaggerfallWorkshop.Game.Entity;
+using DaggerfallWorkshop.Utility;
+using UnityEngine;
+
+namespace DaggerfallWorkshop.Game
+{
+    public class LoanChecker : MonoBehaviour
```
Interkarma: Does this class need to be a MonoBehaviour? It doesn't appear to be added to any GameObjects, and you're just calling the static methods inside of it. Could drop the MonoBehaviour and perhaps make the whole class static.

petchema: Done

Interkarma:
This is really lovely and a very worthwhile update to the banking system. Looking forward to also featuring this in May update. Well done!

petchema:
I just noticed (after watching a stream actually) that there's a classic quest, K0C30Y03, about a banker asking you to find a noble that defaulted a loan.
I don't know all the details of the quest, but it could give some ideas about what could happen if you're on the other side of things...

Interkarma:
Nice catch. Haha, imagine if DF was an MMO of some kind - the game could hand out bounties for players to "collect" from each other for a reward. :)

PR #1304, opened 2019-05-20
Rework the UV unpacking method

df-ferital:
This fixes the UV processing method. Previous changes I made in PR #1299 were not as robust as they are now.

It now seems DFU even does a better job at processing some of classic UV coordinates. See Bargain Gear Store in Ephesus for an example.

I reworked the method after the following considerations:
- Until proven otherwise, models which require some processing on there UV coords are always exterior buildings, so we can easily filter these since buildings model ids are below 1000.
- Only coordinates above 14335 or below -14335 need to be 'unpacked' (as was already guessed by Interkarma years ago). The only noticeable exception seems to be -7168, which I found on several building planes: this one has always to be processed as well. 
- The correct texture coordinate is given by subtracting the nearest multiple of 8192.

So in the end, not only this new method fixes all regressions issued by pango and others on the forums (house roof, Daggerfall Castle treasury room), but also fixes new models without the need to alter original game data (Bargain Gear Store in Ephesus, Mages Guild in Dragontail).

### DIFF
--- a/Assets/Scripts/API/Arch3dFile.cs
+++ b/Assets/Scripts/API/Arch3dFile.cs
@@ -4,7 +4,7 @@
 // License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
 // Source Code:     https://github.com/Interkarma/daggerfall-unity
 // Original Author: Gavin Clayton (interkarma@dfworkshop.net)
-// Contributors:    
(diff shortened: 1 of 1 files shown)


Interkarma:
Well done! I won't pretend to have a deep understanding of the new method, but I trust your work completely. :) I'll test what I can then merge ahead of next builds.

PR #1305, opened 2019-05-24
implement smarter location finder using edit distance

petchema:
This is a bit experimental; It gives interesting results, but there's no guarantee I didn't break anything (say, that all the places that could be found before can still be accessed?).
Also, costs have been set by hand, and are likely not optimal.

EditDistance contains a quite flexible edit distance function:

```
public EditDistance(Func<string, string> canonize_string,
                    Func<char, float> insert_cost,
                    Func<char, float> delete_cost,
                    Func<char, char, float> replace_cost,
                    Func<string, int, float> seek_cost,
                    Func<string, int, float> trim_cost)
```
- canonize_string(s) is the function that will be used to preprocess dictionary and search strings
- insert_cost(c) is the cost function for adding a specific character
- delete_cost(c) is the cost function for removing a specific character
- replace_cost(c1, c2) is the cost function for replacing a specific character with another
- seek_cost(s, start) is the cost function for seeking in the dictionary string s to the start-th character
- trim_cost(s, stop) is the cost function for seeking in the dictionary string s from the stop-th character to the end

Lowering the latter 2 costs (compared to insert_cost) allows to approximate a substring search.

```
float distance(string s1, string s2, float upperBound = float.PositiveInfinity)
```
This computes the edit distance to transform string s1 into string s2.
If an upperBound is provided, the function short-circuits and returns float.PositiveInfinity if it detects during computation that the final cost will be above upperBound. This requires all costs to be positive.

`void SetDictionary(List<String>)`
Set the dictionary that will be used to next searches. It canonizes all the dictionary strings, so it could be useful to only call it when the dictionary changes.

`MatchResult[] FindBestMatches(string needle, int ntop)`
Finds at most ntop dictionary entries that have the smallest distance to needle string.
MatchResult is a structure that contains the (non-canonized) matching string, its distance to needle, and a relevance. The results with the best relevance (hence smallest distances) are returned at the beginning of the list.

DaggerfallEditDistance is a static class that provides an EditDistance object pre-configured with useful parameters:
- trimmed/lowercased strings canonization;
- smaller cost for adding or removing separators (space and dash) compared to letters;
- smaller cost for replacing a separator with another;
- smaller cost for substring prefix search, and to a lesser extend beginning of word search;
- smaller cost for substring suffix search, and to a lesser extend ending of word search.

DaggerfallTravelMapWindow has been modified to use findBestMatches() of this object; In case of approximate matches, a location picker window will appear with name suggestions.
DaggerfallListPickerWindow was slightly modified to take keyboard focus in this case (it did not always take the focus on first use).

Search still feels instantaneous with the normal lists of locations of a Daggerfall region, and DaggerfallTravelMapWindow doesn't even optimize calls to SetDictionary() so it could still be made a bit faster.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallListPickerWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallListPickerWindow.cs
@@ -100,12 +100,6 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             pickerPanel.Components.Add(scrollBar);
         }
 
-        public override void OnPush()
-        {
-            base.OnPush();
-            pickerPanel.SetFocus();
-        }
-
         public override void Update()
         {
             base.Update();
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallListPickerWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallListPickerWindow.cs
@@ -100,6 +100,12 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             pickerPanel.Components.Add(scrollBar);
         }
 
+        public override void OnPush()
+        {
+            base.OnPush();
+            pickerPanel.SetFocus();
+        }
+
         public override void Update()
         {
             base.Update();
--- a/Assets/Scripts/Game/Entities/PlayerEntity.cs
+++ b/Assets/Scripts/Game/Entities/PlayerEntity.cs
@@ -364,7 +364,7 @@ namespace DaggerfallWorkshop.Game.Entity
                     int amount = (int)(DefaultFatigueLoss * fatigueLossMultiplier);
                     if (climbingMotor != null && climbingMotor.IsClimbing)
                         amount = (int)(ClimbingFatigueLoss * fatigueLossMultiplier);
-                    else if (playerMotor.IsRunning)
+                    else if (playerMotor.IsRunning && !playerMotor.IsStandingStill)
                         amount = (int)(RunningFatigueLoss * fatigueLossMultiplier);
                     else if (GameManager.Instance.PlayerEnterExit.IsPlayerSwimming)
                     {
--- a/Assets/Scripts/Game/Formulas/FormulaHelper.cs
+++ b/Assets/Scripts/Game/Formulas/FormulaHelper.cs
@@ -797,10 +797,10 @@ namespace DaggerfallWorkshop.Game.Formulas
                     random = UnityEngine.Random.Range(0f, 100f);
                     if (random <= 0.6f)
                     {
-                        // TODO: Werewolf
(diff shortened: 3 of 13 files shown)


Assets/Scripts/Game/UserInterfaceWindows/DaggerfallListPickerWindow.cs
```diff
@@ -100,6 +100,12 @@ protected override void Setup()
             pickerPanel.Components.Add(scrollBar);
         }
 
+        public override void OnPush()
+        {
+            base.OnPush();
+            pickerPanel.SetFocus();
```
petchema: doesn't seem very reliable :(

Interkarma:
This looks like some clever code and a nice QoL feature. Give me some time to unpack all this. :)

petchema:
I checked that searching for each location in Daggerfall and Wrothgarian Mountains regions results each time in a single exact match (see "selfcheck" commit). Speed on my setup (Intel i5-6400 @ 2.70GHz):
time (ms)
Mean	20.6890891535243
Standard Error	0.161458618373232
Mode	19
Median	20
First Quartile 	13
Third Quartile	28
Variance	80.9960270838069
Standard Deviation	8.99977927972719
Kurtosis	-0.852429684886404
Skewness	0.232291481255751
Range	44
Minimum	4
Maximum	48
Sum	64281
Count	3107

Interkarma:
I straight-up love this change. It's one of the best QoL improvements to the game and fits perfectly with the classic style. I'm more than happy to integrate this. Is there anything you want to do with it before I merge?

petchema:
I'm not sure I want to keep the latest commit in (the autotest one), specially since I had some issue porting it to Tedious Travel, compiling the mod fails with

    Error (584): Internal compiler error: Unexpected error when loading type
    `System.Collections.Generic.Dictionary<TKey,TValue>.KeyCollection.Enumerator'

So I already simply reverted that commit in the port: https://github.com/petchema/TediousTravel/commits/location-finder-edit-distance
but that makes one more difference to keep track of between the DFU travel map class and the Tedious Travel travel map class.


Other than that, I think the UI could be improved; Right now you type your search string, validate, and if they're several matches their list is displayed. But you have no way to go back to your input string to fix a typo.

At very least it would be an improvement to be able to go back to your search string; and ideally the list of suggestions could be displayed next to the input field so you can easily switch between the two. And to be really fancy, given that the search engine is quite fast it could display suggestions as you type or at least in near real time (say any time you stop typing for 100ms).

But I'm not sure that can be pulled off with just UI classes as they are. Thinking of it the save game form has input field and list, and you can switch focus between the two, so maybe it's already doable.

In any case I'd take any help to improve the UI while staying compatible with classic look&feel... But that could be done after this PR is merged

petchema:
Done, can be merged

Interkarma:
Thank you! :)

PR #1306, opened 2019-05-25
Fixes

ajrb:
* Ensure i'm at shows current location after searching 
* Implement name function for %n macro. (requires checking by @Nystul-the-Magician )
* Implement %vcn macro. (needs checking by @Interkarma)
* Implement %rn macro.
* Fix bug when loading a game when holding a bow that displayed no arrows message after loading.
* Fix unhandled null exception when no building discovered.
* Fix negative reaction message having black background.
* Fix potion recipes added to loot and shelves, they were all showing as unknown until saved.
* No materal should be applied to arrows in loot

### DIFF
--- a/Assets/Scripts/Game/Questing/QuestMCP.cs
+++ b/Assets/Scripts/Game/Questing/QuestMCP.cs
@@ -32,6 +32,13 @@ namespace DaggerfallWorkshop.Game.Questing
                 this.parent = parent;
             }
 
+            public override string Name()
+            {
+                // Set seed to the quest UID before falling through to random name generation. (See t=2108)
+                DFRandom.srand((int) parent.UID);
+                return null;
+            }
+
             public override string FactionOrderName()
             {
                 // Only used for knightly order quests, %kno macro. (removing 'The ' prefix from name for readability)
--- a/Assets/Scripts/Game/Questing/QuestMCP.cs
+++ b/Assets/Scripts/Game/Questing/QuestMCP.cs
@@ -105,6 +105,16 @@ namespace DaggerfallWorkshop.Game.Questing
                 }
             }
 
+            public override string VampireNpcClan()
+            {   // %vcn
+                if (parent.LastResourceReferenced == null && !(parent.LastResourceReferenced is Person))
+                    return null;
+
+                // TODO: How to find vampire clan of an NPC? Is faction data as used here the right method that covers all cases?
+                Person person = (Person) parent.LastResourceReferenced;
+                return (person.FactionData.type == (int)FactionFile.FactionTypes.VampireClan) ? person.FactionData.name : null;
+            }
+
             public override string QuestDate()
             {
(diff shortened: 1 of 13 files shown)


Nystul-the-Magician:
I think this might break dialog texts, since many dialogs use the %n macro to resolve the name of the dialog partner, what your change does (as far as I understood) is that it just generates a random name now which would give pointless random names instead of the name of the dialog partner - what do you think?

Interkarma:
Makes sense thank you Nystul. :)

ajrb:
@Nystul-the-Magician Okay thanks, all I could find was uses where %n was referring to a third party. I'll take another look and add conditions.

ajrb:
So, all other usage of %n is refering to a third person except for those 3 greetings. I've added contextual handlers for quests and talking.

Interkarma:
Thanks for all the additional love and care on this one, @ajrb. :) Will look at this in more detail and give you feedback when I can.

Nystul-the-Magician:
looks good to me ;)

Assets/Scripts/Game/Questing/QuestMCP.cs
```diff
@@ -105,6 +112,16 @@ public override string Pronoun3()
                 }
             }
 
+            public override string VampireNpcClan()
+            {   // %vcn
+                if (parent.LastResourceReferenced == null && !(parent.LastResourceReferenced is Person))
+                    return null;
+
+                // TODO: How to find vampire clan of an NPC? Is faction data as used here the right method that covers all cases?
```
Interkarma: I'm certain you can just use the region NPC is in to determine their vampire clan name. The following should do it:

    VampireClans vampireClan = FormulaHelper.GetVampireClan(regionIndex);
    string clanName = TextManager.Instance.GetText("Races", vampireClan.ToString().ToLower());

Interkarma:
@ajrb Hey mate! Are you happy to stick in the vampire stuff and remove the TalkManager changes? Fertial has also fixed the same issue in #1325.

If you're busy, I can just quickly make these changes before merging. I'm trying to get all the PRs processed this weekend/weekend so I still have some time to knock out a couple of new enchantment effects next week.

df-ferital:
@ajrb Actually not all your TalkManager changes, only the line change you made in TalkToNpc to get the correct popup background.

df-ferital:
@ajrb Actually not all your TalkManager changes, only the change you made in TalkToNpc to get the correct popup background.

df-ferital:
@ajrb Actually not all your TalkManager changes, but only the change you made in TalkToNpc to get the correct popup background.

Interkarma:
Thanks for all your awesome work on this one @ajrb! Sorry it took so long to merge.

ajrb:
No problem, I'm in Romania on business, but was planning to make the changes to this PR tomorrow. I assume you've made the vamp clan change yourself and just overridden my TalkToNpc change with Ferital's PR?

PR #1308, opened 2019-05-27
Dungeon wagon prompt setting

ajrb:
Wagons can now be accessed when player is within 10m of a dungeon exit.
The access wagon prompt can be disabled in settings.
Added messages informing player why the wagon button does nothing in inventory window.


Interkarma:
I love this change! :)

PR #1309, opened 2019-05-27
Add level progress dialog

ajrb:
Note: will not be correct if CalculatePlayerLevel() formula is changed by mods.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallCharacterSheetWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallCharacterSheetWindow.cs
@@ -28,6 +28,8 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
     {
         #region Fields
         const string nativeImgName = "INFO00I0.IMG";
+
+        protected const string textDatabase = "DaggerfallUI";
         private const int noAffiliationsMsgId = 19;
 
         StatsRollout statsRollout;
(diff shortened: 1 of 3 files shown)


petchema:
Yes, I was thinking about adding more infos about skills advancement too, but I had the same fear that it could go out of sync with mods modifying leveling logic.
I guess that, worse case, they could also mod that dialog...

zaklaus:
Wouldn't it be possible to split the calculation into two methods, so that we could use `CalculatePlayerLevelProgression()` in the UI logic?

```cs
// Calculate player level progression.
public static float CalculatePlayerLevelProgression(int startingLevelUpSkillsSum, int currentLevelUpSkillsSum)
{
    Formula_2i del;
    if (formula_2i.TryGetValue("CalculatePlayerLevelProgression", out del))
        return del(startingLevelUpSkillsSum, currentLevelUpSkillsSum);

    return (float)(currentLevelUpSkillsSum - startingLevelUpSkillsSum + 28f) / 15f;
}

// Calculate player level.
public static int CalculatePlayerLevel(int startingLevelUpSkillsSum, int currentLevelUpSkillsSum)
{
    return (int)Mathf.Floor(CalculatePlayerLevelProgression(startingLevelUpSkillsSum, currentLevelUpSkillsSum));
}
```

Maybe I'm missing something out, but I think this could work.

Interkarma:
Nice one, thank you!

PR #1310, opened 2019-05-28
fix background behind some messageboxen

petchema:
Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=2126

Git bisect points at commit 445a9a89

There's a lot of ways to create DaggerfallMessageBox objects, maybe there's another, existing,way that I missed...

### DIFF
--- a/Assets/Scripts/Game/DaggerfallUI.cs
+++ b/Assets/Scripts/Game/DaggerfallUI.cs
@@ -1141,6 +1141,15 @@ namespace DaggerfallWorkshop.Game
             return messageBox;
         }
 
+        public static DaggerfallMessageBox MessageBox(TextFile.Token[] textTokens, IMacroContextProvider mcp = null)
+        {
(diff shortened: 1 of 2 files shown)


ajrb:
@Interkarma FYI I have added a fix for this into my fixes PR. Up to you if you'd prefer a new shortcut method for tokens as @petchema has done here.

Baljak:
@petchema Sorry I just saw your PR and I've also fixed this in the same way in my NPC greetings update.
Since I've also done a lot of work in TalkManager as well and reworked the way this part of the code is called, I think it would be best to close the PR to avoid any merge conflicts.
@ajrb Could you also update your own PR to remove this part of the fix as well?

df-ferital:
@petchema Sorry I just saw your PR and I've also fixed this in the same way in my NPC greetings update.
Since I've also done a lot of work in TalkManager as well and reworked the way this part of the code is called, I think it would be best to close the PR to avoid any merge conflicts.
@ajrb Could you also update your own PR to remove this part of the fix as well?

petchema:
@Interkarma I'm closing this one then

PR #1311, opened 2019-05-28
Do not show the same resolution with multiple refresh rates

TheLacus:
Fix issue reported [here](https://www.reddit.com/r/daggerfallunity/comments/bkb1hv/a_question_about_the_resolution_settings_from_the/):  Daggerfall Unity always uses the highest supported so there is no need to show multiple values on UI.


Interkarma:
Thank you! :)

PR #1312, opened 2019-05-28
Implemented detail slider in pause window

TheLacus:


TheLacus:
I decided to save to settings only when window is closed, do you agree with this choice? There are already some settings saved every time but i don't this is really necessary.

Do you also want the fullscreen button to be functional?

TheLacus:
I added fullscreen button only for builds. If you prefer to keep it disabled i'll remove the last commit :)

PR #1315, opened 2019-05-30
Implement bow drawback option

ajrb:
Adds a 2 stage process for shooting arrows with a bow. Hold down right mouse to draw and release to fire. Uses 3 frames of animation which were previously unused due to single stage attack animation process.

1) Draw the bow. frames 0-3
2) Release the arrow. frames 4-6

Set BowDrawback true to enable. Left click releases without firing arrow, this will automatically happen after 10s anyway.

This came out of the discussion here https://forums.dfworkshop.net/viewtopic.php?f=12&t=1748. I had intended the whole thing to be in a mod, but after looking at the weapon handling and animation code, it feels like it would need a massive amount of re-factoring to achieve this. So, I've put this into the core DFU codebase as an option. 

I think it has as much value as click to attack, and is a nice updating of archery interaction for DFU. In fact I would quite like it to be the default, but that's up to the dev community to decide. I do intend to add some mod hooks / formulas so that this 2 stage process can be integrated with damage and hold limits in a mod. Would like to be sure you're okay with these changes first though @Interkarma .

### DIFF
--- a/Assets/Resources/defaults.ini.txt
+++ b/Assets/Resources/defaults.ini.txt
@@ -69,6 +69,7 @@ SoundVolume = 1.0
 MusicVolume = 1.0
 InstantRepairs = False
 AllowMagicRepairs = False
+BowDrawback = False
 
 [Map]
 AutomapNumberOfDungeons=5
--- a/Assets/Scripts/Game/EnemyAttack.cs
+++ b/Assets/Scripts/Game/EnemyAttack.cs
@@ -229,7 +229,7 @@ namespace DaggerfallWorkshop.Game
             PlayerEntity playerEntity = GameManager.Instance.PlayerEntity;
 
             // Calculate damage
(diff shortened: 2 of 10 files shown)


petchema:
I like it!
The "resting" frame also takes much less screen estate, and feel less aggressive.

There may be some "material palette" issues, the highlights on my mithril bow don't have the same material tint (I checked without mods too):

mithril bow

ajrb:
This change hasn't affected material palette. I went and checked classic and it doesn't appear to do the material colouring for bows at all.

Glad you like it anyway.

petchema:
Yes, that's something that was added in Daggerfall Unity (was it @Allofich doing? I'm not sure).
Anyway, you're right it's unrelated to this PR, just something I happened to notice while testing it...

Interkarma:
This is nicely done! There was a lot of fiddly work to inject this into the animation setup and I appreciate the time taken on it. :)

PR #1316, opened 2019-05-31
Implement Masque of Clavicus effect

numidium:
I made the reputation boost +20.  Please let me know if this should be higher for gameplay purposes or if there is a value that is more Classic-accurate.

### DIFF
--- a/Assets/Scripts/Game/MagicAndEffects/Effects/Special/MasqueOfClavicusEffect.cs
+++ b/Assets/Scripts/Game/MagicAndEffects/Effects/Special/MasqueOfClavicusEffect.cs
@@ -45,6 +45,10 @@ namespace DaggerfallWorkshop.Game.MagicAndEffects.MagicEffects
             }
         }
 
+        public override void Start(EntityEffectManager manager, DaggerfallEntityBehaviour caster = null)
+        {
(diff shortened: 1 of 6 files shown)


Assets/Scripts/Game/MagicAndEffects/Effects/Special/MasqueOfClavicusEffect.cs
```diff
@@ -0,0 +1,54 @@
...
+            // Improve reactions with all social groups
+            foreach (FactionFile.SocialGroups group in Enum.GetValues(typeof(FactionFile.SocialGroups)))
+            {
+                playerEntity.ChangeReactionMod(group, adjustmentAmount);
+            }
+        }
+
+        public override void Start(EntityEffectManager manager, DaggerfallEntityBehaviour caster = null)
+        {
+            RoundsRemaining = 1; // permanently stay at >0
```
Interkarma: This shouldn't be necessary, item-based effects are permanently held open until the related item is removed.

For spell-based effects, RoundsRemaining is decremented each magic round. So just setting to 1 in Start wouldn't have the intended outcome. You can safely remove this one. :)

Interkarma:
Thanks for all your work on this. :)

PR #1317, opened 2019-05-31
Enable texture override for paperdoll

TheLacus:
Enabled injection of custom textures to new paperdoll renderer. Additional textures named `textureName_Mask` can provide masking defined on alpha channel. Maks provided via assetbundles should use texture type _Single Channel_ to automatically pick format with minimum amount of memory used.

### DIFF
--- a/Assets/Scripts/Game/Items/ItemHelper.cs
+++ b/Assets/Scripts/Game/Items/ItemHelper.cs
@@ -331,14 +331,16 @@ namespace DaggerfallWorkshop.Game.Items
             }
 
             // Get mask texture where alpha 0 is umasked areas of image and alpha 1 are masked areas of image
-            // Note: Texture replacement will need to support import of an optional mask texture for each replacement
-            ImageReader.UpdateMaskTexture(ref data);
+            Texture2D maskTexture;
+            if (TextureReplacement.TryImportTexture(archive, record, 0, item.dyeColor, TextureMap.Mask, out maskTexture))
+                data.maskTexture = maskTexture;
+            else
+                ImageReader.UpdateMaskTexture(ref data);
 
             Texture2D tex;
-            if (!forPaperDoll && TextureReplacement.TryImportTexture(archive, record, 0, item.dyeColor, out tex))
+            if (TextureReplacement.TryImportTexture(archive, record, 0, item.dyeColor, TextureMap.Albedo, out tex))
             {
                 // Assign imported texture
-                // Paperdoll is disabled for now
                 data.texture = tex;
             }
             else
--- a/Assets/Scripts/Game/UserInterface/PaperDoll.cs
+++ b/Assets/Scripts/Game/UserInterface/PaperDoll.cs
@@ -30,26 +30,23 @@ namespace DaggerfallWorkshop.Game.UserInterface
         const int paperDollHeight = 184;
         const int waistHeight = 40;
 
-        DFSize backgroundFullSize = new DFSize(125, 198);
-        Rect backgroundSubRect = new Rect(8, 7, paperDollWidth, paperDollHeight);
+        readonly DFSize backgroundFullSize = new DFSize(125, 198);
+        readonly Rect backgroundSubRect = new Rect(8, 7, paperDollWidth, paperDollHeight);
 
         #endregion
 
         #region Fields
 
-        static Color32 maskColor = new Color(255, 0, 200, 0);   // Special mask colour used on helmets, cloaks, etc.
-        DFPosition paperDollOrigin = new DFPosition(200, 8);    // Used to translate hard-coded IMG file offsets back to origin
+        static readonly Color32 maskColor = new Color(255, 0, 200, 0);   // Special mask colour used on helmets, cloaks, etc.
 
-        bool showBackgroundLayer = true;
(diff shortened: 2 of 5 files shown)


TheLacus:
An example as shown in editor:

251_32-0_Ebony with Mask

Interkarma:
Woohoo! Great to see some progress on this already. :)

TheLacus:
Examples of valid xml files:

```
<?xml version="1.0"?>
<info> 
    <scaleX>1</scaleX>
    <scaleY>1</scaleY>
    <relOffsetX>0</relOffsetX>
    <relOffsetY>0</relOffsetY>
</info>
```

```
<?xml version="1.0"?>
<info> 
    <Redguard>
        <scaleX>1</scaleX>
    </Redguard>
</info>
```

```
<?xml version="1.0"?>
<info> 
    <Redguard>
        <Female>
            <scaleX>1</scaleX>
        </Female>
    </Redguard>
</info>
```

```
<?xml version="1.0"?>
<info> 
    <Male>
        <scaleX>1</scaleX>
    </Male>
</info>
```

`scale` is self-explanatory, while `relOffset ` is an offset from default position. If you want i can add `offset` to override the actual offset defined in source image. Is there something else that should be editable from xml?

TheLacus:
On second thought i'm not sure this is really necessary. Aren't paperdoll items already available in different variants for body shapes? Each of them gets its own xml file so it shouldn't be necessary to provide subsections like i've shown above.

I think i'll wait for a precisation on what kind of data you expect xml files to provide, maybe things are simpler than i thought.

Interkarma:
>Aren't paperdoll items already available in different variants for body shapes?

Yep, 8 in total. One for each of the 4 body morphologies (argonians, elves, humans, khajiit) which then come in two variants for male/female.

I don't think we'll need much here. If the item is an exact upscale and repaint, it should fit automatically. But I can see a situation where custom art (e.g. a clothing pack) might need some minor tweaks to align correctly because items have a different general shape to original art.

Something to consider is the targetRect being passed is in classic pixel grid coordinates. This is scaled into a screenRect (render texture target is our screen in this case) in 0-1 domain. Artists might want more control over positioning HD item images than classic resolution would allow, so it's probably better to allow modders to override screenRect for fine control over the position at any scale.

If that seems a bit abstract (and fiddling with raw UV values can be) we could instead express the mod value as an offset from the automatic position derived from classic data. The 8x paper doll is rendered to a fixed 880x1472 surface, so it could even be a +/- adjustment for horz/vert on the 8x pixel grid. That would give the artist fine control over final position and they only need to think in terms of "up one pixel, right two pixels from where it is now". That of course breaks if the scale goes up higher than 8x, but I genuinely believe that 880x1472 should be enough pixel density for just the HD paper doll.

I'm also open to ideas on all this. I mainly just wanted to get the rendering working and then we could tinker a bit to see what worked best when it came to replacements. Thanks again for making a start on this so soon. :)

TheLacus:
> 
> 
> > Aren't paperdoll items already available in different variants for body shapes?
> 
> Yep, 8 in total. One for each of the 4 body morphologies (argonians, elves, humans, khajiit) which then come in two variants for male/female.
> 
Thanks for the clarification, I misunderstood description of 0b3fa7c97b3cfe4f0ed946fe6eba147ee9e68eb4 where you stated XML file should be used to finetune based on race and gender. If each texture is for a single specific body morphology, i guess there is no need to provide more than one rect. I'll simplify this PR appropiately :)

> Something to consider is the targetRect being passed is in classic pixel grid coordinates. This is scaled into a screenRect (render texture target is our screen in this case) in 0-1 domain. Artists might want more control over positioning HD item images than classic resolution would allow, so it's probably better to allow modders to override screenRect for fine control over the position at any scale.
> 
> If that seems a bit abstract (and fiddling with raw UV values can be) we could instead express the mod value as an offset from the automatic position derived from classic data. The 8x paper doll is rendered to a fixed 880x1472 surface, so it could even be a +/- adjustment for horz/vert on the 8x pixel grid. That would give the artist fine control over final position and they only need to think in terms of "up one pixel, right two pixels from where it is now". That of course breaks if the scale goes up higher than 8x, but I genuinely believe that 880x1472 should be enough pixel density for just the HD paper doll.

I implemented the latter, but you raise a good point on breakability with scale changes. Maybe i'll ask for opinions on forum.

TheLacus:
Ok, this what i came up with:

```xml
<?xml version="1.0"?>
<info>
    <rect scale="1">
        <x>31</x>
        <y>0</y>
        <width>50</width>
        <height>43</height>
    </rect>
</info>
```

```xml
<?xml version="1.0"?>
<info>
    <rect scale="8">
        <x>248</x>
        <y>0</y>
        <width>400</width>
        <height>344</height>
    </rect>
</info>
```

Rect is expressed in pixels with origin on the top-left corner. The `scale` attribute defines the target resolution on which the rect is based. If different than paperdoll resolution, rect values are scaled appropriately. This should be easier to work with than raw 0-1 float values and at the same time it won't break if paperdoll resolution is increased in the future.

Interkarma:
That's a brilliant compromise, well done!

Interkarma:
Merging. Looking forward to watching this one progress. :)

PR #1318, opened 2019-06-01
generate OnEncounter event that mods can subscribe to

petchema:
First goal is to make Tedious Travel work more reliably at high time compressions, but future mods could also benefit from this event.

Also, make rest window subscribe to this event to remove some specific code.


petchema:
Submitted a PR to Tedious Travel to use this mechanism:

https://github.com/TheNewBob/TediousTravel/pull/15

Interkarma:
Works well, thank you!

PR #1319, opened 2019-06-01
Implement IEnumerable<BaseScreenComponent>

pcbeard:
This allows foreach(var component in components) to properly infer the type.

### DIFF
--- a/Assets/Scripts/Game/UserInterface/Panel.cs
+++ b/Assets/Scripts/Game/UserInterface/Panel.cs
@@ -99,12 +99,12 @@ namespace DaggerfallWorkshop.Game.UserInterface
             base.Update();
 
             // Update child components
-            for (int i = 0; i < components.Count; i++)
+            foreach(var component in components)
             {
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Game/UserInterface/Panel.cs
```diff
@@ -99,12 +99,12 @@ public override void Update()
             base.Update();
 
             // Update child components
-            for (int i = 0; i < components.Count; i++)
```
Interkarma: Hi, thanks for your improvement. I'm happy with the IEnumerable improvement to ScreenComponentCollection, however I would prefer these loops in Panel not to use foreach. Thank you. :)

pcbeard: Thanks. Very interesting project. I am a fan of all of the TES games and this is my first experience with Daggerfall. I'm also on my way to learning Unity, so this seemed like a good way to jump in with both feet.

PR #1320, opened 2019-06-06
Refactor TransportManager

zaklaus:
Implementation of a horse mount (or dismount when already on a horse/cart) for G key. This required some changes to the `TransportManager` to make sure transportation change conditions are kept in sync via a public API. `ToggleMount()` can easily mount or dismount a horse if conditions are met.

I can optionally get rid of the G key shortcut, if wanted.

NOTE:
Call to `ToggleMount()` works even inside of buildings or dungeons. This behaviour was kept intentionally to allow mods to use mounts even inside interiors. Of course, the front-end logic defaults to classic behaviour and only allows you to toggle mount in the exterior.

Introduced an option in the Enhancements section called `HorseMountShortcut`.

### DIFF
--- a/Assets/Scripts/Game/TransportManager.cs
+++ b/Assets/Scripts/Game/TransportManager.cs
@@ -101,7 +101,7 @@ namespace DaggerfallWorkshop.Game
         /// <summary>
         /// Mounts a horse or cart if available if on foot, otherwise gets on foot
         /// </summary>
-        public void ToggleMount()
+        public void ToggleHorse()
         {
             if (TransportMode == TransportModes.Horse || TransportMode == TransportModes.Cart)
             {
@@ -248,7 +248,7 @@ namespace DaggerfallWorkshop.Game
                 {
                     if (GameManager.Instance.PlayerController.isGrounded)
                     {
-                        ToggleMount();
+                        ToggleHorse();
                     }
                 }
             }
--- a/Assets/Scripts/Game/TransportManager.cs
+++ b/Assets/Scripts/Game/TransportManager.cs
@@ -12,7 +12,6 @@ using DaggerfallConnect.Arena2;
 using DaggerfallConnect.Utility;
 using DaggerfallWorkshop.Game.Banking;
 using DaggerfallWorkshop.Game.Items;
-using DaggerfallWorkshop.Game.UserInterfaceWindows;
 
 namespace DaggerfallWorkshop.Game
 {
@@ -36,6 +35,7 @@ namespace DaggerfallWorkshop.Game
         public float RidingVolumeScale = 1.0f;  // TODO: Should this be the same setting as PlayerFootsteps.FootstepVolumeScale?
 
         #endregion
+
         #region Properties
 
         public TransportModes TransportMode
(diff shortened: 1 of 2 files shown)


petchema:
They're forward and backward KeyBind.txt compatibility issues with this patch, which can be a problem because it's a settings file (hence shared among all instances, and not cleaned by reinstalling DFU):
- If you had something already bound to G key, bindings play musical chairs, each one getting its default binding (I think) until everything is assigned a separate key;
- Either way, in keybindings setup screen the G binding appears in front of "Steal mode", right below "Transport". All bindings after that are shifted one place;
- If a KeyBind.txt with the new horse mount keybind is read by a version of DFU that doesn't have that feature, all bindings are lost (no bindings in KeyBind.txt, everything bound to "None" in setup).

zaklaus:
Thanks for the report, I will adjust the changes accordingly.

ajrb:
I think this is probably mod territory since there are no plans to expand the keybindings screen in DFU.

zaklaus:
I've edited the InputManager to support these issues better in the future. Unknown actions will be ignored (and saved on exit). As for the `MountHorse` particularly, I believe it could be moved onto a mod territory, in that case I'd still recommend cherry-picking TransportManager changes, so that these 
 related actions will be kept in sync across the codebase.

zaklaus:
I've separated input changes into 2 branches. This PR now only contains changes made to the TransportManager and relevant facilities. zaklaus:keybindings contains `InputManager` changes as well as the G keybind for mount toggle.

zaklaus:
Once I return home, I would share the toggle mount mod, which would make use of this PR specifically.

Interkarma:
Hello @zaklaus! :)

Thank you for your work here, but I'm not sure why the changes are necessary. It seems possible to create all of this functionality in a mod without refactoring TransportManager. Could you please outline the core issue that requires these changes?

zaklaus:
Hi,

I originally wanted to create a simple toggle horse mod on a hotkey, however I've noticed that there wasn't an API for checking whether the horse is available. I've noticed the UI logic directly accesses the player's inventory to find out if he owns a special item (horse), which in turn would enable the horse button in the transport menu.

Since I needed to do the same check on the mod side, I found it redundant and easier to just present an API for easily checking whether the player owns a horse/cart with a simple method to easily change to either of those modes (`ToggleMount()`).

It is possible this isn't really required, since the core will probably never deal with transportation mode changes outside of the UI logic (T menu), but having an API to check for requirements instead of accessing the inventory directly to figure out if I can saddle a horse just felt more clean.

Interkarma:
Thank you, I understand where you're coming from now. I'm happy for TransportManager to expose the HasHorse, etc. helpers. And while ToggleMount() is rather specific to your mod, it's a simple helper and I don't mind it being there. I also really appreciate that you trimmed this one down earlier after feedback from the others. Will test and merge this one in next few days.

zaklaus:
Thank you, it was a nice experience exploring the codebase.

bansheexyz:
Just wanted to add that I personally like your idea here, zak. Toggling between foot and horse with "T" instead of bringing up a menu seems like it would work well.  You could force the cart-drawn horse if a cart is owned, and then obviously do ship when standing in ocean water. I wonder why classic thought a menu was necessary? Anyway, fairly simple mod that saves a click, not up to me but seems like a good stock option.

PR #1321, opened 2019-06-06
TravelPopup: Added OnPreFastTravel event

TheLacus:


TheLacus:
Will be used by Loading Screen mod for fast travel transitions.

PR #1322, opened 2019-06-06
Improve KeyBind data handling

zaklaus:
This PR tries to tackle issues mentioned by https://github.com/Interkarma/daggerfall-unity/pull/1320#issuecomment-499596177 by adding simple error resolution measurements to ensure backward and forward compatibility of the game.

Unknown actions will now be ignored by the `InputManager`, but kept in memory, so we can store them back to the settings file once we rebuild it.


Interkarma:
I'm looking into adding a new binding UI a bit later in project, and may overhaul how bindings work to solve this same problem in a different manner. At this time though, I'm not wanting to change available keybinds or adjust how this is handled. I'm sorry for any inconvenience caused.

zaklaus:
Hi,

no problem with that, I'm looking forward to the changes.

PR #1323, opened 2019-06-10
[SaveLoadManager] Delete Notebook File

jbrod22:
PR adds the notebook to deleted files.

Noticed in unity that save folders were not being deleted because they weren't empty. Found they all had one NotebookData.txt file. Adding this line removed the folders and logging errors.


ajrb:
Oops my bad. Thanks for fixing. :-)

PR #1324, opened 2019-06-10
Spell merchant consistency

petchema:
Thinking again about an old patch (#1063), I think it makes more sense to make spell merchant window behavior consistent with all others, and not close on double-click.
On the other hand, display available money since that's an important information for the player. It lands perfectly at the right of cost, just like in standard merchants interface.


Interkarma:
This is very good, thank you! :)

PR #1325, opened 2019-06-10
Make NPC greetings selection match classic

df-ferital:
Make the greeting text selection process the same as in classic, with some improvements and a few bug fixes. To do this, I reverse engineered the 3 classic functions used to get the greeting text.

Now to explain the changes I made: in classic, a faction related (mostly guild related in fact) greeting is selected each time the player talks to an NPC which is not a member of a faction of type 14 or 15, i.e. "Court of" or "People of" current region. Then, the correct greeting text is given by checking all player guild affiliations, especially guild enemies and allies. I did the same in DFU, fixing two bugs of the original function.

I also improved the overall selection process in two ways:
- I decided to exclude unique NPCs whose first parent is a province, as this gave weird greetings in classic. Because these unique NPCs are not members of a guild or guild-like faction, in classic, they could greet the player with sentences beginning with "As a member of Sentinel..." and other inadequate text. Now in DFU, all NPCs belonging to province children factions (this includes children of children), apart from Knightly Orders, will greet the player like all common people would do.
- One other improvement over classic is also to filter the usual "Well met, stranger" greeting (record 8570), which is always fired when the faction of an NPC is unrelated to all player current guilds. This, even if player current reputation with the NPC faction (e.g. The Merchants or The Bards) is very good. So, I restricted this boring greeting to NPCs for which player reputation with their respective faction is below or equal to 5. So, when the player can really be considered a stranger to the NPC he is talking to. In other cases, NPCs of such factions will use common people greetings. I tested it in game and this is much better in my opinion.

### DIFF
--- a/Assets/Scripts/API/FactionFile.cs
+++ b/Assets/Scripts/API/FactionFile.cs
@@ -883,6 +883,30 @@ namespace DaggerfallConnect.Arena2
             }
         }
 
+        /// <summary>
+        /// Check if a faction is another faction ally.
+        /// </summary>
+        /// <param name="firstFaction">The faction to check the allies of.</param>
+        /// <param name="secondFaction">The potential allied faction.</param>
+        /// <returns>True if factions are allied, otherwise false.</returns>
+        public static bool IsAlly(ref FactionData firstFaction, ref FactionData secondFaction)
+        {
+            return firstFaction.ally1 == secondFaction.id || firstFaction.ally2 == secondFaction.id ||
+                   firstFaction.ally3 == secondFaction.id;
+        }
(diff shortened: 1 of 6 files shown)


Interkarma:
This is really cool. I'll make sure I test as much as possible and get it merged in time for June builds.

Interkarma:
Ready to merge this one now. Just have a couple of conflicts to fix with MacroHelper.cs to fix when you can. Let me know if any assistance needed. :)

df-ferital:
Hi Interkarma :)

I rebased on last the master revsion, conflicts are now solved.

PR #1328, opened 2019-06-17
Spawn guards after rifling through private property

numidium:
This triggers the guard spawner after the player says "yes" to the "This looks like private property" dialog box.  Shop interiors are relatively small so the spawner doesn't always find a minimum distance, especially in one-room shops.

I'm thinking of doing an overhaul of interior guard spawns so they are handled in a more reliable way that also makes more gameplay sense.  I'll do that in a separate PR unless you would like to have that in this one.

### DIFF
--- a/Assets/Scripts/Game/PlayerActivate.cs
+++ b/Assets/Scripts/Game/PlayerActivate.cs
@@ -694,6 +694,11 @@ namespace DaggerfallWorkshop.Game
                 GameManager.Instance.PlayerEntity.TallyCrimeGuildRequirements(true, 1);
                 Debug.Log("Player crime detected: rifling through private property!!");
 
+                // Send the guards
+                PlayerEntity playerEntity = GameManager.Instance.PlayerEntity;
+                playerEntity.CrimeCommitted = PlayerEntity.Crimes.Theft;
+                playerEntity.SpawnCityGuards(true);
+
                 // Open inventory window with activated private container as remote target (pre-set)
                 DaggerfallUI.PostMessage(DaggerfallUIMessages.dfuiOpenInventoryWindow);
             }


ajrb:
The only reason I didn't add this in at the time was that the guard spawning still needed work. I believe @Allofich has this on his todo list, so would be good to coordinate completing this aspect of guard spawning. Unless it already did get completed.

Not sure it's worth doing this change until guards are done, although I did it for stealing IIRC. But then shop spaces are normally more predictable.

Interkarma:
Thank you for helping with a start on the remaining crime features, I really do appreciate it!

My only issue right now is that stealing from private property does not yet perform any skill checks to determine successful theft, and does not increment any skill tallies. There's only placeholder hold when player is accessing private property.

It doesn't seem appropriate yet to enable guards (punishment) until the player at least has a chance (skill check) and appropriate reward (skill tally) as well. I'd do this the other way around and start spawning guard after the other two parts of the process are done.

For interior guard spawning specifically, using the front door closest to ground level is probably a good choice. I was thinking about this myself when mulling around developing the crime systems. Let me know if we're thinking alike here. :)

numidium:
> My only issue right now is that stealing from private property does not yet perform any skill checks to determine successful theft, and does not increment any skill tallies. There's only placeholder hold when player is accessing private property.

I never knew there were skill checks/tallies for private property containers.  I agree that it should be included in this PR if that is the case.

> For interior guard spawning specifically, using the front door closest to ground level is probably a good choice. I was thinking about this myself when mulling around developing the crime systems. Let me know if we're thinking alike here. :)

Yep, that's pretty much exactly what I had in mind.  I can include that too.

Interkarma:
>I never knew there were skill checks/tallies for private property containers. I agree that it should be included in this PR if that is the case.

I did think there was a stealing check similar to shoplifting when accessing private containers, but happy to be wrong here (it means less complexity after all). I just haven't gotten to the point of diving into that yet. If someone know for sure (@Allofich maybe) please let me know. Otherwise we can discover through testing in classic.

>Yep, that's pretty much exactly what I had in mind. I can include that too.

Perfect! :)

Interkarma:
Thanks for your work on this. I'll just merge for now and we can work out skill checks a bit later in 0.8.

Allofich:
There's a skill check done in classic, yes. I can add it in when I have time.

Interkarma:
>There's a skill check done in classic, yes. I can add it in when I have time.

Lovely, thanks so much @Allofich :)

PR #1329, opened 2019-06-23
adjust TilemapTextureArray mipmap levels to remove most aliasing

petchema:
I did perceptual tests on modded terrain textures, then some analysis, using DREAM RC4 and no post processing mod on my screen (1080p).

First I checked that a mipmap level of 8.0 looked smooth in the distance: that works.
Then I checked how far the transition between level 7.0 and 8.0 could be pushed away before aliasing appears. In my tests, 50.0 was about right.
Then I checked how far the transition between level 0.0 and 7.0 could be pushed away before aliasing appears. In my tests, 4.5 was close to the max.

But what's the scale between the two? I assume the goal is to keep the angular diameter of pixels about constant. Pixel widths varies according to 1/d, but their height decreases faster because the surface also appears more and more tilted (from camera point of view).
I spare you the details, but height seems to vary in 1/d².

Mipmap scale is logarithmic (vs pixel size), so you end up with level = const - 2*log2(d)

In other words, you should gain one mipmap level each time the distance increases by a sqrt(2) factor.
And it matches almost perfectly with the two bounds found perceptually!

However I don't know how my perceptual tests are portable to other systems...

Forums: https://forums.dfworkshop.net/viewtopic.php?f=27&t=1168&p=25573#p25573

### DIFF
--- a/Assets/Shaders/DaggerfallTilemapTextureArray.shader
+++ b/Assets/Shaders/DaggerfallTilemapTextureArray.shader
@@ -99,28 +99,11 @@ Shader "Daggerfall/TilemapTextureArray" {
 			//half4 c = UNITY_SAMPLE_TEX2DARRAY_GRAD(_TileTexArr, uv3, ddx(uv3), ddy(uv3)); // (see https://forum.unity3d.com/threads/texture2d-array-mipmap-troubles.416799/)
 			// since there is currently a bug with seams when using the UNITY_SAMPLE_TEX2DARRAY_GRAD function in unity, this is used as workaround
 			// mip map level is selected manually dependent on fragment's distance from camera
-			float scale = sqrt(_TileTexArr_TexelSize.x * 64.0f);
-			float scaledDist = distance(IN.worldPos.xyz, _WorldSpaceCameraPos.xyz) / scale;
+			float scale = sqrt(_TileTexArr_TexelSize.x * 64.0f) * 12.38;
+			float relativeHeight = max(_WorldSpaceCameraPos.y - IN.worldPos.y, 1.8);
+			float dist = distance(IN.worldPos.xyz, _WorldSpaceCameraPos.xyz);
 
-			float mipMapLevel;
-			if (scaledDist < 17.5f)
-				mipMapLevel = 0.0;
-			else if (scaledDist < 25.0f)
-				mipMapLevel = 1.0;
-			else if (scaledDist < 35.0f)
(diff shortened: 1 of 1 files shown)


petchema:
It gives terrible blurriness with unmodded terrain textures. I guess it must be adjusted depending on textures resolution?
Either that, or separate shaders should be used for modded and unmodded textures...

Nystul-the-Magician:
we might be able to modify the shader a bit and use the texture size property there - see https://docs.unity3d.com/Manual/SL-PropertiesInPrograms.html

petchema:
Any way to optimize the computation of scale (or even 1.0f / scale)? It should never change for a given texture...

petchema:
I found another issue...
The terrain may look good on foot or on horse, but if you look at it from high above, the hypothesis that you're looking tangentially doesn't hold, and it looks terribly blurry again :(
mipmap-dream-nopostprocessing

petchema:
Fixed!
heightfix-dream-nopostprocessing

Interkarma:
This looks great! Looking forward to playing with it. Well done. :)

petchema:
It still makes some approximations about the terrain (mainly, that the texture is horizontal) so in some cases, like if you're looking upward at a hill you're climbing, or you're simply looking at the front of a bump, the selected mipmap level is too high.
A surface shader that computes mipmap level from ddx and ddy would give better results, but I don't quite understand the math :(

Interkarma:
My original tilemap shader in 2014 used ddx/ddy to manually calculate mipmap level. Not sure if that will be helpful as a reference or not. Here's the link to the article from way, way back.

https://www.dfworkshop.net/improved-terrain-tiling/

Interkarma:
Very nice improvement, thank you! This makes the new tilemap terrain look every bit as nice in the air as my old terrain shader. Awesome! :)

Nystul-the-Magician:
> 
> 
> My original tilemap shader in 2014 used ddx/ddy to manually calculate mipmap level. Not sure if that will be helpful as a reference or not. Here's the link to the article from way, way back.
> 
> https://www.dfworkshop.net/improved-terrain-tiling/

I am not 100% sure if I remember correctly, but I think back then when I wrote the texture array tilemap shader I pretty much started from your shader with the ddx/ddy computations but as far as I remember I had problems with seams on the edges of tiles so I decided against the ddx/ddy.
It is likely that I did something wrong or it was a unity bug back then, so trying it out again would be a good thing to do!

PR #1330, opened 2019-06-25
Refine indoor guard spawns and private property theft

numidium:
1. When in a shop, guards will now spawn at the lowest entrance.
2. Stealing private property now tallies Pickpocket and does a check against it.  If the check fails, guards are spawned and PC is accused of stealing.


Interkarma:
Nice one! Looking forward to review. :)

Interkarma:
It plays well, feels true to classic, and you seem to have covered everything. A very nice PR, thank you so much for your work on this. :)

PR #1332, opened 2019-06-29
Tilemap ddx ddy mipmap

petchema:
Hopefully I fixed the seams problem. The trick was to make ddx and ddy derive a function that varies similarly to UV * texturesize, but without its discontinuities: the function at the origin of UV value, but before tile index/tile UV coords mapping (I called it unwrappedUV in the code).

(it seems there has been a long story of trying similar stuff, so I hope I'm not hitting the same shaders bug as everyone else...)

### DIFF
--- a/Assets/Shaders/DaggerfallTilemapTextureArray.shader
+++ b/Assets/Shaders/DaggerfallTilemapTextureArray.shader
@@ -63,7 +63,6 @@ Shader "Daggerfall/TilemapTextureArray" {
 		{
 			float2 uv_MainTex;
 			//float2 uv_BumpMap;
-			float3 worldPos;
 		};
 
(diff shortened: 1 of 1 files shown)


petchema:
No seams on my GTX960:
seamless

They're huge seams on hardware that only use half floats, but then it happens with mipmaping no matter what (0.7.x shader, 0.8.2 one, or this one).

Also, the 0.5 of 0.5 * log2(d) in the code may be a little low, you get some aliasing if you use point filtering. Then again, you want crunchy pixels, you get them. I'm not sure if it should be adjusted.

Nystul-the-Magician:
Great that you take a look! Have to try it out and give feedback :D

Interkarma:
This look great! Amazing work sorting all that out. :)

PR #1334, opened 2019-07-01
Retain commoners social group

df-ferital:
Matched to classic: NPCs whose faction is "People of" or "Court of" current region retain their original social group even if they use their parent region faction data to interract with the player.

This fixes the bug mentioned by @petchema here: https://forums.dfworkshop.net/viewtopic.php?f=24&p=25816#p25816


Interkarma:
Thank you for fix! :)

PR #1335, opened 2019-07-02
fixed bugs with user note markers and teleporter markers:

Nystul-the-Magician:
- markers are now correctly restored on dungeon exit and reentering
- markers are now correctly removed when exiting dungeons (no left-over artifacts from other dungeons)

I am sorry for this stupid oversights on my side
If it is possible to create new test builds with this fix that would be awesome ;)


Interkarma:
Thanks for fixes! I'll drop some new builds on Friday to include these and other changes.

PR #1336, opened 2019-07-04
Fix macro %n gender generation in quests

df-ferital:
Gender generated when expanding %n in quests was not kept. Fixed it by using DFRandom.


Interkarma:
Can you tell me more about the problem? Both are just randomly selecting a gender from a 0-1 roll. The only difference will be the seed state.

df-ferital:
Yes sorry I should have been more specific. The seed state is the problem there. When used in a quest, the name generated by %n macro expansion used to change everytime the player reads a letter using the macro, since it is randomly generated. Hazelnut fixed it using the quest uid as seed, but not for gender generation. So, to also keep the gender, DFRandom has to be used there.

Interkarma:
Thank you! Sorry, I didn't look at the change in context. Appreciate the help. :)

PR #1337, opened 2019-07-05
 fixed town map positioning (player marker positioning) for custom 1x1 locations

Nystul-the-Magician:
fixed this in exterior automap code - against my initial intuition it turned out to be the right place to make those changes


Interkarma:
Thank you for fixing. :)

PR #1338, opened 2019-07-08
fixed bug with named residence not being resolved correctly on info click

Nystul-the-Magician:
solves bug report mentioned in https://forums.dfworkshop.net/viewtopic.php?f=24&t=2227

moved player house name handling into function GetBuildingDiscoveryData
so all residence name resolving is done in this function now

### DIFF
--- a/Assets/Scripts/Internal/PlayerGPS.cs
+++ b/Assets/Scripts/Internal/PlayerGPS.cs
@@ -1010,12 +1010,7 @@ namespace DaggerfallWorkshop
                 return false;
 
             // Get discovery data for building
-            discoveredBuildingOut = dl.discoveredBuildings[buildingKey];
-
-            // Check if name should be overridden (owned house / quest site)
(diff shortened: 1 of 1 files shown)


df-ferital:
@Nystul-the-Magician @Interkarma Those changes broke the town map when loading a savegame which has been saved in an interior building. Curisouly enough, the issue does not arise if you opened the town map from another working savegame (i.e. saved in exterior) before loading the interior savegame.

Attached is a save you can check to reproduce the issue.

[SAVE1.zip](https://github.com/Interkarma/daggerfall-unity/files/3383860/SAVE1.zip)

Interkarma:
Thanks for the catch Ferital! I didn't test this case before merging. Really appreciate the second set of eyes.

Nystul-the-Magician:
These issue seems to be unrelated to the changes of this pr - strange why this happens now...
will fix

PR #1339, opened 2019-07-08
Import spell icon packs from mods

TheLacus:
Spell icon packs can be imported from mods when files are added to a folder name **SpellIcons**. During mod building, directory of selected files are matched and results, if any, are automatically added to `SpellIcons` array of `Contributes` property inside mod manifest file.

This process allows to provide icon packs (and possibly other assets in the future) in the same way for both mods and loose files, doing all the work at build-time rather than run-time where it could affect performance.

This is an example of a result manifest file.

```json
{
    "ModTitle": "Example Spell Icons pack",
    "ModVersion": "",
    "ModAuthor": "",
    "ContactInfo": "",
    "DFUnity_Version": "0.8",
    "ModDescription": "Adds a spell icons pack.",
    "GUID": "2f923d7b-b8e5-4db6-a535-6655896b87fc",
    "Files": [
        "Assets/Game/Addons/TestMod/SpellIcons/vmblast-test.png",
        "Assets/Game/Addons/TestMod/SpellIcons/vmblast-test.json"
    ],
    "Contributes": {
        "SpellIcons": [
            "vmblast-test"
        ]
    }
}
```

### DIFF
--- a/Assets/Scripts/Game/UserInterface/SpellIconCollection.cs
+++ b/Assets/Scripts/Game/UserInterface/SpellIconCollection.cs
@@ -12,11 +12,13 @@
 using System;
 using System.IO;
 using System.Collections.Generic;
+using System.Linq;
 using UnityEngine;
 using DaggerfallConnect.Utility;
 using DaggerfallWorkshop.Utility;
 using DaggerfallWorkshop.Game.MagicAndEffects;
 using DaggerfallWorkshop.Game.Serialization;
+using DaggerfallWorkshop.Game.Utility.ModSupport;
 
 namespace DaggerfallWorkshop.Game.UserInterface
 {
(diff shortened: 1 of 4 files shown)


Assets/Scripts/Game/UserInterface/SpellIconCollection.cs
```diff
@@ -228,42 +230,67 @@ void LoadSpellIconPacks()
...
-                    pack = ReadMetadata(metadataPath);
-                    LoadPackIcons(pack, atlasTexture, metadataPath);
-                }
-                else
+                // Read each atlas found and its metadata
+                foreach (string path in atlasPaths)
                 {
-                    // Create empty metadata file if none found
-                    pack = CreateIconPackEmptyMetadata(atlasTexture, metadataPath);
-                    return;
```
TheLacus: I kept this return but i believe it shouldn't be there. Can you confirm what is the appropriate behaviour here?

Interkarma: You can likely remove because safety check on following code won't do anything with 0 row count or icon count. But it's also a perfectly valid place to return as no further work needed if spell pack is empty.

TheLacus: If meta data is missing for a pack, all the following packs are also skipped. Actually i think i found a bigger issue: a default file is created during startup scene, but then this one is loaded after transition to game scene, raising an exception due to bad operation `atlas.width / pack.rowCount` with zero rows.

This happens because DaggerfallUI.Start() is called again; Spell packs initialization should be moved to Awake or, even better, when requested the first time i think.

I also saw this when testing with two copies of test pack.
iconPacks

I can work on these issues with a new PR if you want.

Interkarma: Of course, thank you for the help. :)

PR #1340, opened 2019-07-08
Ambient sounds spatialization

petchema:
use AudioSource's minDistance to play distant ambient sounds at correct volume

- Simplified thunder sound: instead of enforcing source direction using a coroutine, put the sound source much further and compensate with high minDistance.
- Sounds are not always located within minDistance, so there's some volume randomization that wasn't there before; That can be adjusted if you feel that volume so too often too low;
- Increase ambient AudioSource's 3D spread, so that the hear opposite to the sound source gets more input; Specially in dungeons, to simulate environment reverberation and increase perceived volume.

This is a minimal patch, for example only one ambient sound can be played at any moment (not that PlayOneShot cannot be used several times simultaneously on an AudioSource, but all sounds will share some parameters, like the 3D location). 
Unity has lots of untapped sound features, AudioMixerGroups, ReverbZones,...

### DIFF
--- a/Assets/Scripts/Game/AmbientEffectsPlayer.cs
+++ b/Assets/Scripts/Game/AmbientEffectsPlayer.cs
@@ -174,20 +174,26 @@ namespace DaggerfallWorkshop.Game
 
         private void AmbientPlayOneShot(SoundClips clip, float volumeScale)
         {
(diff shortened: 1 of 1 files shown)


Interkarma:
Thank you for your help. :)

PR #1341, opened 2019-07-09
Fix dungeon exit wagon access upon first inventory use

petchema:
Forums;
https://forums.dfworkshop.net/viewtopic.php?f=12&t=2098&p=26161#p26161


petchema:
Found a regression, when accessing inventory near a dungeon entrance, upon first try the wagon is automatically selected; On later attempt, it's not selected by default, like before.

petchema:
> Found a regression, when accessing inventory near a dungeon entrance, upon first try the wagon is automatically selected; On later attempt, it's not selected by default, like before.

Fixed

PR #1342, opened 2019-07-10
Only allow wagon access on sight

petchema:
Patch on top of #1341

There's probably still some bug, but I need to find a dungeon to test it:
Some dungeons have more than one exit door, but DungeonWagonAccessProximityCheck() checks the proximity to the StartMarker, not from a dungeon exit. If my hypothesis is correct, when DungeonExitWagonPrompt is disabled you cannot use those secondary exit doors to access your wagon.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
@@ -554,6 +554,9 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             }
             else
                 SelectActionMode(ActionModes.Equip);
+                
+            if (GameManager.Instance.PlayerEnterExit.IsPlayerInsideDungeon && !allowDungeonWagonAccess)
+                DungeonWagonAccessProximityCheck();
         }
 
         public override void OnPush()
@@ -642,9 +645,6 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
 
             // Update tracked weapons for setting equip delay
             SetEquipDelayTime(false);
-
-            if (GameManager.Instance.PlayerEnterExit.IsPlayerInsideDungeon && !allowDungeonWagonAccess)
-                DungeonWagonAccessProximityCheck();
 
             // Refresh window
             Refresh();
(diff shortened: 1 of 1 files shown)


Interkarma:
I'm not really aware of any dungeon with multiple exit quads. These are only enabled in the entrance block, and blocks should only have a single exit quad. It's possible, but it must be exceedingly rare as I can't think of any encounters with this case.

petchema:
Yes, it is quite rare, I have clear memories of finding some in classic Daggerfall (say, one at the back of a square room similar to the one where the secret room with the Great Unicorn Horn is attached to), sadly I didn't record the name of the dungeons :(
The UESP only mentions Palace Sentinel, I think I saw it before on the map, but ran out of time to find it again...

petchema:
Secondary dungeon exit door in Gaerwing's Guard, Wayrest
Thanks jayhova for the hint!
[SAVE311.zip](https://github.com/Interkarma/daggerfall-unity/files/3414518/SAVE311.zip)

petchema:
Same location in Daggerfall classic
[SAVE4.zip](https://github.com/Interkarma/daggerfall-unity/files/3414533/SAVE4.zip)

Interkarma:
You found your white whale, nicely done. :)

Something we could do to support this other than using a raycast to exit quad, is a distance check from player to closest exit quad (rather than start marker distance check). This would be fairly inexpensive and cover all exits in the way it works now without the need for player line of sight to exit. What are your thoughts?

petchema:
Anything that works :)
And you seem to have a clear idea how to proceed... I started looking at PlayerActivate.GetDoors() but how the different objects relate to one another is still obscure for me...

Interkarma:
I wonder if @ajrb would like to make this refinement as the creator of this feature? All that's required is to pull the entrance model from scene and check distance to those instead of start marker. That way it will work uniformly for the rare dungeons with multiple exits.

Otherwise, I'm happy to roll this in when I have the chance. :)

ajrb:
Nope, happy for you or @petchema to make the change. I think is fine as is.. in that wagon is accessible from the main door you come in through.

Interkarma:
> I think is fine as is.. in that wagon is accessible from the main door you come in through.

The issue is if player has disabled wagon prompt, they're now unable to access wagon at all from secondary door (which they could do in both classic and DFU prior to wagon prompt option). This refinement simply solves that edge case.

It's not really a big deal of course, and these secondary exits are exclusively rare. As promised, I'm happy to roll this in when I have time. :)

Interkarma:
@petchema I think we can close this one now that any dungeon exit door can be used for proximity check. Is that OK with you?

petchema:
Yup, sounds fine

PR #1343, opened 2019-07-10
Faster noprompt wagon access

petchema:
Patch on top of #1341 and #1342 

When opening the inventory by keypress near dungeon exit in no DungeonExitWagonPrompt mode, do not only allow access to the wagon but also auto-select it.
This makes this mode actually more convenient than prompt mode.

Forums: https://forums.dfworkshop.net/viewtopic.php?f=12&t=2098&start=10#p26151

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
@@ -1046,9 +1046,20 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             // Set allow wagon access if close enough (10m) to exit.
             GameObject playerAdvancedGO = GameObject.Find("PlayerAdvanced");
             DaggerfallDungeon dungeon = GameManager.Instance.DungeonParent.GetComponentInChildren<DaggerfallDungeon>();
-            Vector3 exitVector = dungeon.StartMarker.transform.position - playerAdvancedGO.transform.position;
-            if (exitVector.magnitude < 10)
+            Vector3 exitVector = playerAdvancedGO.transform.position - dungeon.StartMarker.transform.position;
+            if (exitVector.magnitude < 2)
+                // fix bad case when you've just been spawned on StartMarker
                 allowDungeonWagonAccess = true;
+            else if (exitVector.magnitude < 10f)
+            {
+                RaycastHit hit;
+                Ray ray = new Ray(dungeon.StartMarker.transform.position, exitVector);
+                if (Physics.Raycast(ray, out hit))
+                {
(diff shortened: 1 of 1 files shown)


ajrb:
I intentionally didn't do this when I made the original change as the consequence is that players that know nothing of the proximity check (i.e. non-forumites) will find that the inventory opens the wagon unexpectedly. Although they probably wont even notice and not sure many people drop loot in a pile next to the exit when they have a wagon so is prob not a big deal. Personally I tend to prefer consistency.

I'm also curious why you feel the need for a raycast, after checking the distance. I felt the gain from this is hardly worth the trouble.

petchema:
> [...] Although they probably wont even notice and not sure many people drop loot in a pile next to the exit when they have a wagon so is prob not a big deal. 
Wagon is only autoselected if you open your inventory using keypress; If you open it by clicking on some loot, the loot will take precedence over the wagon.
> I'm also curious why you feel the need for a raycast, after checking the distance. I felt the gain from this is hardly worth the trouble.
Now that the wagon is autoselected whenever possible, it's even more obvious that you can access it even thru walls. So I made the check a little more strict.

ajrb:
Thanks for the explanations, makes total sense now. Regarding multi exit dungeons, I didn't even know that was a thing! Anyway thinking about it, your wagon will only be at one of them - unless it teleports psychically..

petchema:
> thinking about it, your wagon will only be at one of them - unless it teleports psychically..

Well, dungeon doors are a kind of topological conundrum. There can be several entrance doors, but they all lead to the dungeon StartMarker. Conversely, there can be several exit doors, but from what I remember you always exit in front of a single specific door too.

So yes, I thought about the solution you mention, but whether it's ok is rather arbitrary. I'm for being able to access your wagon from any of the exit doors, but I admit that's arbitrary too. Maybe we should find one example of such secondary exit door in classic to study it in depth.

Holy crap, I just understood why the UESP mentions Palace Sentinel as an example of dungeon with several exit doors: they're side by side! I'm not sure it's a good example then ;(

Interkarma:
@petchema I think this one can be closed also now that #1342 is no longer required.

petchema:
Do you mean that the UI changes of this pull request should be reimplemented on top of new proximity changes?

Interkarma:
My mistake, sorry. Yes the UI changes are fine. :)

petchema:
Should be okay now

Interkarma:
I've just merged #1341 and have a conflict in this PR. Let me know if any help needed. I'll merge this one once conflicts resolved. :)

petchema:
Re-merging master worked automatically on my box for some reason... I don't have rerere enabled so standard merge algorithm alone must have done the trick...
Anyway, the conflict is solved, and from quick testing everything seems still working fine

PR #1347, opened 2019-07-12
Experimental support for custom books

TheLacus:
Preliminary support for import of custom books for bookshelves. 

## How it works
1. A book file is positioned inside _StreamingAssets\Books_. For example `StreamingAssets\Books\example-book.TXT`.
2. A map table is positioned inside _StreamingAssets\Books\Mapping_. For example:

```json
[
    {
        "Name": "example-book",
        "Title": "An Example Book"
    }
]
```

This table has two purposes: distinguish classic books replacements from additional custom books (could also have done this with a separate folder) and announces book titles before importing the actual resources. Custom subfolders can potentially be supported in the future.

3.  Each book is associated to a numeric id (only valid for the session) starting from 112 and added to **bookIDNameMapping** in ItemHelper.

## How to test

1. Add files as explained above.
2. Edit DaggerfallBookshelf to skew chances of selecting custom books (or add a lot of custom books...).
3. Find a book shelf (or add component to any test prop). The title should be visible and book resource readable.

This feature is disabled by default for the moment, and should not affect core in any way if disabled.

### DIFF
--- a/Assets/Game/Addons/ModSupport/ModManager.cs
+++ b/Assets/Game/Addons/ModSupport/ModManager.cs
@@ -746,26 +746,34 @@ namespace DaggerfallWorkshop.Game.Utility.ModSupport
         public static void SeekModContributes(ModInfo modInfo)
         {
             List<string> spellIcons = null;
+            List<string> booksMapping = null;
 
             foreach (string file in modInfo.Files)
             {
                 string directory = Path.GetDirectoryName(file);
-                if (directory.EndsWith("SpellIcons"))
-                {
-                    if (spellIcons == null)
-                        spellIcons = new List<string>();
 
-                    string name = Path.GetFileNameWithoutExtension(file);
-                    if (!spellIcons.Contains(name))
-                        spellIcons.Add(name);
-                }
+                if (directory.EndsWith("SpellIcons"))
+                    AddNameToList(ref spellIcons, file);
+                else if (directory.EndsWith("Books/Mapping"))
+                    AddNameToList(ref booksMapping, file);
             }
 
-            if (spellIcons != null)
+            if (spellIcons != null || booksMapping != null)
             {
                 var contributes = modInfo.Contributes ?? (modInfo.Contributes = new ModContributes());
-                contributes.SpellIcons = spellIcons.ToArray();
+                contributes.SpellIcons = spellIcons != null ? spellIcons.ToArray() : null;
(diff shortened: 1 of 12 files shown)


TheLacus:
@Interkarma can i ask you why the message is masked here? My understanding is that item message for books is only used as book id (from books.txt), am i wrong?

```cs
public static string messageToBookFilename(int message)
{
    int decodedValue = message & 0xFF;
    return "BOK" + decodedValue.ToString("D5") + ".TXT";
}
```

Interkarma:
>can i ask you why the message is masked here?

@TheLacus This method was created by InconsolableCellist almost 3 years ago. My understanding matches yours, and I don't think it needs to be masked looking at it again now. The message value should just be the book ID.

TheLacus:
Thanks for confirmation.

TheLacus:
Ok, right now custom books can be read from bookshelves as well as found in random loots. 

For the purpose of serialization, a file in Books folder stores link between resource file name and a item id, assigned at runtime. This file should not be deleted or links are lost. A recent discussion with UV about new models made me realize that it might also be better for books to request mod authors to directly provide a random int id, given that chances of collision can be considered low enough. I don't have a strong opinion of this at the moment.

I think current implementation is ready to be made available as an experimental feature, disabled by default. After initial feedback, i'd like to work on the following features:
- Unique books: not found inside loots, must be made available by mod (specific location, quest reward. etc.).
- "dynamic" books: can be edited by mod to make book-like logs...

Interkarma:
@TheLacus Are you happy for this to go into this month's builds in time for Alpha, or would you like to let it cook a bit longer before I merge?

TheLacus:
I think is worth to include it in build so we can receive early feedback from mod authors.

It must be enabled from ini settings under _Experimental_ section to be functional, so i think it's clear that it could be subjected to changes. I'll write a post in mod forums to explain how to use it with more details.

Interkarma:
Cheers mate. Merging now! :)

Interkarma:
Just FYI, this feature might really be put to use soon. None other than Ted Peterson is offering to write custom books for DFU to expand on the in-game fiction. Follow progress here on Twitter.

https://twitter.com/gav_clayton/status/1163205333705084928

He's even raised the idea of certain books being available only after MQ is completed. We could tie that to a simple global var check and only offer book once specific quest globals are met.

TheLacus:
Hey, that's great! I'm happy to expand system to allow to specify filters for book availability, such as global variables.

PR #1350, opened 2019-07-14
apply a -0.5 mipmap bias to DaggerfallTilemapTextureArray shader

petchema:
Seems still almost aliasing free according to visual tests.

Implementing mipmap bias by hand, does standard Texture.mipmapBias apply to custom surface shaders? And if so, where should it be set?

Forums: https://forums.dfworkshop.net/viewtopic.php?f=4&t=37&p=26322#p26322


Interkarma:
>Implementing mipmap bias by hand, does standard Texture.mipmapBias apply to custom surface shaders? And if so, where should it be set?

My understanding is that mipmapBias is set per texture, so any materials using that texture will have the set bias against the mip chain for that texture. I've only used this in limited ways however, so my understanding is likely to be limited.

PR #1351, opened 2019-07-15
Update enumerated saves when a new savegame is created

df-ferital:
Enumerated saves were not updated when a new savegame was created. This made QuickLoad not working immediately after using QuickSave for the first time (QuickLoad worked, however, when restarting DFU).


Interkarma:
Thank you. :)

PR #1355, opened 2019-07-19
Classic quests, second part

JayH2971:
I hope I'm doing this right? Never merged into anything other than Master. Feedback welcome if I'm ruining things.

Credit to pango for adding an edit I failed to notice.


Interkarma:
Thank you so much @JayH2971. I'm going to roll all of these into 0.9 at end of month. Amazing work on all of this. Take your time with any remaining quests you want to work on. Only do what you can when you feel up to it. :)

JayH2971:
I have some bad news about the spawning groups. You can see my research here, in the lower half:

https://docs.google.com/spreadsheets/d/1W5R54I92BpisrldTuQj909wxqdwMneKQLG4VOI0x0sI/edit#gid=332645013

There appears to be no rhyme or reason about how "Local_#" makes a spawn local or remote. In A0C00Y14, Local_3.0 spawns in a local city, but in R0C11Y16 Local_3.0 is remote.

Interkarma:
>There appears to be no rhyme or reason about how "Local_#" makes a spawn local or remote. In A0C00Y14, Local_3.0 spawns in a local city, but in R0C11Y16 Local_3.0 is remote.

Oh well, we do the best we can with the limited information available. It would have been great to have a design document for all this stuff, but we've managed to struggle through without one all the same. :)

I'm happy for behaviour to remain undefined until we either gain a better understanding or quest needs to be adjusted to better match intended behaviour.

Interkarma:
@JayH2971 BTW, I'm going to generate 0.9 builds over weekend. I'll merge all of these fixes prior. If there are any last-minute changes you want to squeeze in, you have a couple of a days still. After 0.9 drops, just start sending fixes to master again as before.

Interkarma:
@JayH2971 I've fixed the issue with Person details macro (Bug 25 on forums) now in commit https://github.com/Interkarma/daggerfall-unity/commit/a1b69575b9dc7e760d699d973860dce25abf3168

JayH2971:
Cool! I don't have anything left on the cutting board. The remaining quests need more testing and that'll take me some time.

PR #1356, opened 2019-07-20
Import Texture Arrays as assets

TheLacus:
Terrain texture arrays can now also be created inside the Unity Editor and
provided via mods as assets. This brings the following benefits:

* Minor improvements for memory management and initialization times. The individual source
textures don't need to exist at runtime, only the final texture array.
* Failure due to resolution, format or mipmaps mismatches are not possible at
runtime. A texture array is only rejected if slice count is incorrect.
* Runtime creation of texture arrays requires CopyTexture support or Read/Write flag
(with increased memory usage). Texture arrays shipped as assets works on any platform
that support this feature alone.

The disadvantage is the introduction of an additional step in mod creation
workflow. For this reason is introduced as an optional feature.

This commit adds a simple texture array creator but there are also more advanced
free alternatives on the Unity Asset Store, like [Texture Array Inspector](https://assetstore.unity.com/packages/tools/utilities/texture-array-inspector-109547).

### DIFF
--- a/Assets/Scripts/Editor/TextureArrayWizard.cs
+++ b/Assets/Scripts/Editor/TextureArrayWizard.cs
@@ -1,0 +1,120 @@
+// Project:         Daggerfall Tools For Unity
+// Copyright:       Copyright (C) 2009-2019 Daggerfall Workshop
+// Web Site:        http://www.dfworkshop.net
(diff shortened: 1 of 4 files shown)


Interkarma:
Excellent work! :)

PR #1359, opened 2019-07-22
Fast travel to ground level, always

petchema:
After some fast travel, player position could only be adjusted upward; When terrain was very steep downward it could result in fall damage (and death). 

(Maybe there's a use case I'm not aware of for this behavior?)

Make adjustment unconditional; Height formula adjusted to (slightly higher) PlayerMotor.FixStanding formula.

Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=2307


Interkarma:
Gosh, I think this bit of code dates back to around 2014. I can't recall if there was a specific reason I did things that way. Your fix appears sensible and I can't think of any reason not to do it that way now.

petchema:
With this PR reverted, original issue is back...
Either we decide that it's a bug in Mountains & Hills that should make sure spawnmarkers stay at ground level, or RepositionPlayer() should be told when to ground player.
Had a version with an extra "bool grounded" parameter, but position can also be given a very negative height to ground the player:
https://github.com/Interkarma/daggerfall-unity/compare/master...petchema:fix-fix-fast-travel-to-town
Will require more testing this time :/

PR #1360, opened 2019-07-22
DaggerfallTilemapTextureArray.shader cleanups

petchema:
(patch on top of #1350 )

Trying (unsuccessfully) to fix seams problem, I did lot of small cleanups on shader code:
    
- Remove commented code (that's what versioning is for);
- Added TEXCOORDn annotations on Input
- used UNITY_SAMPLE_TEX2DARRAY_SAMPLER_LOD to share SamplerState between albedo and normal texture lookups
- added inline annotation to GetMipLevel (but I suppose the compiler must have guessed that much already)
- replaced fmod(x, 1.0) with frac(x)
- replaced tile rotation branching with matrices. I could have replaced matrix product and vector addition with just a matrix product, but I would have to promote matrix to float3x3 because OpenGL ES 2.0 doesn't support non-square matrices:
    
      static float3x3 rotations[4] = {
          float3x3(1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0),
          float3x3(0.0, 1.0, 0.0, -1.0, 0.0, 1.0, 0.0, 0.0, 0.0),
          float3x3(-1.0, 0.0, 1.0, 0.0, -1.0, 1.0, 0.0, 0.0, 0.0),
          float3x3(0.0, -1.0, 1.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0)
      };
    
      float2 uv = mul(rotations[index % 4], float3(frac(unwrappedUV), 1.0));
    
With all those modifications I have seen no change in FPS on my rig, maybe they're better tools to measure shaders performance? Maybe it's simply irrrelevant

### DIFF
--- a/Assets/Shaders/DaggerfallTilemapTextureArray.shader
+++ b/Assets/Shaders/DaggerfallTilemapTextureArray.shader
@@ -23,7 +23,7 @@ Shader "Daggerfall/TilemapTextureArray" {
 		_TileTexArr("Tile Texture Array", 2DArray) = "" {}
 		_TileNormalMapTexArr("Tileset NormalMap Texture Array (RGBA)", 2DArray) = "" {}
 		_TileMetallicGlossMapTexArr ("Tileset MetallicGlossMap Texture Array (RGBA)", 2DArray) = "" {}
-		_TilemapTex("Tilemap (R)", 2D) = "red" {}		
+		_TilemapTex("Tilemap (R)", 2D) = "red" {}
 		_TilemapDim("Tilemap Dimension (in tiles)", Int) = 128
 		_MaxIndex("Max Tileset Index", Int) = 255
 	}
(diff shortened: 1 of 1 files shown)


ajrb:
Any chance these changes will remove the lines seen with non-nvidia gfx hardware?

petchema:
I don't think this PR will change shader behavior; Stuff like TEXCOORDn annotations could potentially help with compatibility, but I doubt it'll do anything in this specific case. I should mention I have no hardware to test this hypothesis; All I have is a very old AMD card that features seams, but I don't know how representative it is.
I already mentioned the best lead I have in the forums here: https://forums.dfworkshop.net/viewtopic.php?f=5&t=1464&p=26906#p26906
That is, an issue with the use of fmod: https://stackoverflow.com/questions/14737910/cg-repeating-a-texture-using-fmod-causes-aliasing
and a workaround that only works with the card I mentioned above...

Interkarma:
Thank you for cleanup! :)

ajrb:
That's fair enough - TBH I'm just glad someone is looking and understanding them.

Nystul-the-Magician:
> With all those modifications I have seen no change in FPS on my rig, maybe they're better tools to measure shaders performance? Maybe it's simply irrrelevant
most likely shader compiler optimized those code parts anyway - those compilers are pretty good these days

petchema:
> other possibility is that this shader is not computation bound but memory bound

Yes, I read stuff like that. This shader is so relatively simple that it's probably memory bound.
That's what my "Maybe it's simply irrrelevant" covered...

Nystul-the-Magician:
nevertheless your changes make the code cleaner imo, so thumbs up

PR #1361, opened 2019-07-26
Handle %bn (biography name) macro expansion

df-ferital:
Macro %bn is used only for Nightblade biography history. It expands to a random name using player character race and a random gender.

But as the occupation of the corresponding individual is given by the %q8 macro which expands to a neutral or (mostly) masculine noun (priest, wizard, sorceror, revolutionary, anarchist, warlord, robber baron, assassin), I chose to differ from classic and force the gender to male.

I also fixed MacroHelper.GetNameBank to match classic so it now returns a Khajiit name for Khajiits and an Imperial name for Argonians.


Interkarma:
Works well for me. Thank you. :)

PR #1363, opened 2019-07-26
Fix for non-temple guild greeting messages & provide reference to calling window to custom services

ajrb:
As discussed here:L

https://forums.dfworkshop.net/viewtopic.php?f=23&t=2336&p=27331#p27331


ajrb:
Pretty sure only the Archeologists mod uses custom services at the moment. Added a calling window reference so that the yes/no message box doesn't blank the background. See https://github.com/ajrb/dfunity-mods/blob/master/Archaeologists/Scripts/ArchaeologistsGuild.cs line 370 for where this gets used.

ajrb:
Also added header data to dumped file output. Originally part of UV's closed PR for an editor but this change is good one and removes need for these values to be manually added.

PR #1365, opened 2019-07-27
Restrict building interior loot piles to TG, DB and pubs

ajrb:
This is not exactly as classic but as close as I can get without reverse engineering why they appear is desert FGs but no where else.

I think this is close enough now having spent a few hours testing various buildings in classic.


Interkarma:
Works for me! :)

PR #1370, opened 2019-07-30
NPC reactions enhancement

df-ferital:
Improve the way NPCs react to player questions.

I decided to give Streetwise and Etiquette more weight by setting reaction modifiers related to NPC social groups. For example, talking bluntly to a Noble or Scholar will now give a penalty, while doing the same with a Commoner or a member of the Underworld (Thieves Guild or Assassin) will give a bonus. Of course, using Etiquette through a polite tone will trigger the exact opposite. Merchants, as a sort of middle class, will be unaffected by this change. I gave them a default modifier of +5 when using both skills which makes them behave the same as before as I also changed skill check result: failing will give a penalty of -10 while succeeding will give +5.

I also lowered the required value to get the best NPC reaction, though it's still difficult to get, especially with a low personality and low Etiquette and Streetwise skills.

Finally, I also set default modifiers for "Tell me about" kind of questions the same as there "Where is" counterparts.

In my opinion, this is quite an improvement over classic.

While doing so, I also noticed and fixed a problem with the Talk Window: tone checkbox was not reset to Normal after modifying the tone, then closing the window and reopening it.

### DIFF
--- a/Assets/Scripts/Game/TalkManager.cs
+++ b/Assets/Scripts/Game/TalkManager.cs
@@ -94,7 +94,13 @@ namespace DaggerfallWorkshop.Game
         const int minLikeReaction = 10;
         const int minVeryLikeReaction = 30;
 
-        readonly short[] questionTypeReactionMods = { 10, -5, 0, 0, 0, -10, -5, -5 };
+        // Changed from classic (classis uses { 10, -5, 0, 0, 0, -10, -5, -5 } )
+        readonly short[] questionTypeReactionMods = { 10, -5, 0, 0, 10, -5, 0, 0 };
+
+        /// Improvement over classic, add reaction modifiers when using Etiquette
+        /// and Streetwise in relation to social groups.
+        readonly short[] etiquetteReactionMods = { -10, 5, 10, 15, -15 };
+        readonly short[] streetwiseReactionMods = { 10, 5, -10, -15, 15 };
 
         // From FALL.EXE. In classic the answers for sgroup 0 (commoners) and sgroup 1 (merchants) are reversed, and classic flips reference to these sgroups
         // during dialogue to compensate, but it's cleaner to just fix the data. It is fixed to the correct order here.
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Game/TalkManager.cs
```diff
@@ -533,24 +543,20 @@ int GetReactionToPlayer_0_1_2(QuestionType qt)
...
 
-            // Make roll result be the same every time for a given NPC.
-            if (currentNPCType == NPCType.Mobile)
-                DFRandom.Seed = (uint)lastTargetMobileNPC.GetHashCode();
-            else if (currentNPCType == NPCType.Static)
-                DFRandom.Seed = (uint)lastTargetStaticNPC.GetHashCode();
-
             if (toneIndex != 1)
-                toneModifier = DFRandom.random_range_inclusive(1, 100) >= skillValue ? -5 : 10;
+                toneModifier += DFRandom.random_range_inclusive(1, 100) >= skillValue ? -10 : 5;
```
petchema: You can use Dice100.FailedRoll(skillValue) ? -10 : 5 here

df-ferital: Good catch 👍 Done.

Interkarma:
Cheers! Running new builds today, will include this one.

Interkarma:
@df-ferital This one might need a little more tuning (when you have time of course). I've had multiple reports from player's receiving cold shoulders from NPCs. It doesn't seem consistent though.

For an example, I have a basic Warrior character in save below with a PER of 50, using neutral tone, and a 0 legal rep. He's your average common adventurer hanging out at the Fighter's Guild. When I talk to the quest-giver NPC, he rejects me every time (but still hands out quest service of course). But the NPC by the front door seems happy to chat. I haven't dived into the code yet to try and isolate why one reacts so hostilely and the other does not. Just wanted to check this matches expectations on your end. Cheers mate!

[SAVE686.zip](https://github.com/Interkarma/daggerfall-unity/files/3479439/SAVE686.zip)

Nystul-the-Magician:
> While doing so, I also noticed and fixed a problem with the Talk Window: tone checkbox was not reset to Normal after modifying the tone, then closing the window and reopening it.
this was done on prupose - I considered it an improvement not to have to always switch away from normal tone if you e.g. want to train the skills - how do you guys think?

PR #1371, opened 2019-07-30
Fix N0B00Y16 "open chest"

JayH2971:
S.07 failed to fire, wasn't recognizing the condition. One-line fix. Quest has another problem regarding failure line: if you talk to the merchant without opening the chest, you'll never get a success message afterward. Have tried to fix that problem and can't.

### DIFF
--- a/Assets/StreamingAssets/Quests/N0B00Y16.txt
+++ b/Assets/StreamingAssets/Quests/N0B00Y16.txt
@@ -217,7 +217,9 @@ _a6_ task:
 	say 1043 
 
 _S.07_ task:
-	have _crn_ set _S.08_ 
+	start task _S.08_
+--	have _crn_ set _S.08_ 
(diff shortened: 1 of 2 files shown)


Interkarma:
Thanks for looking at this Jay. The quest action "have anItem set aTask" has been implemented, I'll look into why it doesn't trigger here.

Will also see if I can solve issue of speaking to merchant without opening chest.

JayH2971:
For some reason github for Linux stacks up all my fixes into the same pull request. Not intended, I just don't know how to separate them.

TheLacus:
> 
> 
> For some reason github for Linux stacks up all my fixes into the same pull request. Not intended, I just don't know how to separate them.

You need to create a separate branch ;)

Interkarma:
I'll merge this one for now just so we have a working version of this quest. I'd still like to take a look at why "have item" isn't firing properly here, but don't have the time right now.

PR #1376, opened 2019-07-31
Handle %nrn macro expansion

df-ferital:
Only used in Thieves Guild quest O0B00Y01. Return the name of current region ruler.

Edit: actually it's also used in some non quest rumors.


Interkarma:
Thank you! :)

PR #1382, opened 2019-08-01
Fix %loc macro expansion in castles

df-ferital:
Fix %loc expansion failures as reported here: https://forums.dfworkshop.net/viewtopic.php?f=5&p=27843#p27843

I also filtered NPC names in topics so you shouldn't be able to ask people anymore about themselves both in "Where is people" and "Tell me about" question categories. The filter is a little ugly at the moment but at least it's working.

### DIFF
--- a/Assets/Scripts/Game/TalkManager.cs
+++ b/Assets/Scripts/Game/TalkManager.cs
@@ -510,6 +510,7 @@ namespace DaggerfallWorkshop.Game
             ResetNPCKnowledgeInTopicListRecursively(listTopicPerson);
             ResetNPCKnowledgeInTopicListRecursively(listTopicThing);
             ResetNPCKnowledgeInTopicListRecursively(listTopicTellMeAbout);
+            rebuildTopicLists = true;
         }
 
         public int GetReactionToPlayer(FactionFile.SocialGroups socialGroup)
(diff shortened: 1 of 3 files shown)


Assets/StreamingAssets/Text/ConversationText.txt
```diff
@@ -1,45 +1,45 @@
-- Text database for conversations used in talk window.
```
TheLacus: Entire file is edited due to line break conversion, without any meaningful change. Can you amend commit to remove this file?

Assets/Scripts/Game/TalkManager.cs
```diff
@@ -2778,16 +2771,27 @@ private void AssembleTopiclistTellMeAbout()
                             break;
                         case QuestInfoResourceType.Location:
                             itemQuestTopic.questionType = QuestionType.QuestLocation;
-                            Place place = (Place)questResourceInfo.Value.questResource;
+                            Place place = (Place) questResourceInfo.Value.questResource;
```
TheLacus: There shouldn't be a space here.

Assets/Scripts/Game/TalkManager.cs
```diff
@@ -2778,16 +2771,27 @@ private void AssembleTopiclistTellMeAbout()
...
                             if (place.SiteDetails.buildingName != null)
                                 captionString = place.SiteDetails.buildingName;
                             else
                                 captionString = place.SiteDetails.locationName;
                             break;
                         case QuestInfoResourceType.Person:
                             itemQuestTopic.questionType = QuestionType.QuestPerson;
-                            Person person = (Person)questResourceInfo.Value.questResource;
+                            Person person = (Person) questResourceInfo.Value.questResource;
```
TheLacus: Ditto.

Assets/Scripts/Game/TalkManager.cs
df-ferital: Visual Studio automatically put this space, no pb I'll remove it. :)

Assets/Scripts/Game/TalkManager.cs
df-ferital: Same here...

Assets/StreamingAssets/Text/ConversationText.txt
df-ferital: I know, I initially changed something to this file then reverted it. However, original file has messed up line endings with a mix of both Windows and Unix eol so I let it here. I'll push this to a separate commit if you prefer.

Nystul-the-Magician: I always have problems with line endings warnings on my windows/visual studio systems. I installed git/tortoise git with all standard/recommended settings but I somehow end up with those warnings all the time. Who knows how to get rid of these messages/problems?

df-ferital: I think the eol option has to be set to "check out as is commit Unix style "

TheLacus: That's fine then, feel free to make a commit for it. thank you :)

df-ferital: I removed the file for now :)

Nystul-the-Magician: > 
> 
> I think the eol option has to be set to "check out as is commit Unix style "

so the (tortoise) git default settings are wrong then if I understand correctly?

df-ferital: Yes, since DFU files use Unix-style line endings, Git for Windows default settings ("checkout as-is, commit as-is") can lead to files having a mix of Windows and Unix line endings. This, because Visual Studio does not respect current file end of lines format when pasting code in it. However, changing the global Git option is usually not recommended, especially if you're working on several projects. This is the case for me, but my difftool (Beyond Compare, not free but by far the best one :)) clearly shows if line endings are messed up.

The best solution there is to force Visual Studio to use LF line endings with DFU. There is a plugin for that: https://marketplace.visualstudio.com/items?itemName=JakubBielawa.LineEndingsUnifier.

Though I didn't check it yet.

Issue #1381, opened 2019-08-01
Thieves Guild no longer shows on map for members


Interkarma:
**Describe the bug**
Thieves Guild (and assuming Dark Brotherhood) no longer appear on map when character is a member.

Link to topic on forums.
https://forums.dfworkshop.net/viewtopic.php?f=5&t=2386

Possibly another regression from https://github.com/Interkarma/daggerfall-unity/commit/2b219142ec822904d0c94736c96a49f6a1e0095c as with lockpicking state.

@Nystul-the-Magician Is this something you would like any help with? Please just let me know and I will do what I can to assist. It might be best to just revert that change for now and then tackle from a different angle.

ajrb:
This could be simply because when I implemented the guilds and this function, I integrated with how the discovery code worked back then. Possibly the guild functions just need adjusting, but I think Nystul will need to judge that. Happy to rework them if needed and someone can explain what changed conceptually.

Nystul-the-Magician:
happy to take a look, but will not have time before start of next week :)

ajrb:
I've taken a look and this change 2b21914 has broken the returning of discovery data and now calls GetBuildingDiscoveryData() which seems to build it from the location building directory and completely ignores the overridden names in the discovery data. Since this is where the guilds set the display name override, that's why the guild nameing has broken.

The code removed has not been replaced just by calling GetBuildingDiscoveryData() and I cannot work out why this is being called now. When a building is discovered in a location and added to the dict it has all the data in a DiscoveredBuilding data structure, including name overrides for quests and guilds etc. Why regenerate all that every time, if that's the way forward then the DiscoveredBuilding struct is pretty pointless or that's my reading of it at least.

The player house handling was done slightly differently and you preserved that since the code was explicitly there, but anything that overrides the discovered building name will have broken with this change I am afraid.

It may be because I don't really understand how this fix was intended to correct the bug reported, but I think it should be reverted and the original bug and approach to fixing it analysed again.

Nystul-the-Magician:
> It may be because I don't really understand how this fix was intended to correct the bug reported, but I think it should be reverted and the original bug & approach to fixing it analysed again. I think the temp discovery data cached, and the ability to override names using it, needs to remain. However, if someone has a better idea for how to handle name overriding then I am all ears.
[...]

Issue #1373, opened 2019-07-30
Lockpicking Regression


Interkarma:
**Describe the bug**
Lockpicking should only allow a single failed attempt, then player must level up skill before they can try again. This is how skill was implemented, but broken in 0.10.2.

**To Reproduce**
Attempt to pick a shop's lock outside of normal hours. It's possible to spam the lock until a successful attempt is reached.

petchema:
Bisection points at commit 2b219142ec822904d0c94736c96a49f6a1e0095c

Interkarma:
Thank you @petchema. I didn't consider this implication of change. Will see if I can rework logic to fix.

PR #1384, opened 2019-08-01
fix for github issue 1381

Nystul-the-Magician:
(https://github.com/Interkarma/daggerfall-unity/issues/1381)

public function now only calls private function
all resolving is done there
if there is a building override name this name is taken otherwise further building name resolving is done

### DIFF
--- a/Assets/Scripts/Game/ExteriorAutomap.cs
+++ b/Assets/Scripts/Game/ExteriorAutomap.cs
@@ -693,7 +693,7 @@ namespace DaggerfallWorkshop.Game
                                             string overrideBuildingName = string.Empty;
                                             if (place.SiteDetails.buildingKey == discoveredBuilding.buildingKey && GameManager.Instance.TalkManager.IsBuildingQuestResource(buildingSummary.buildingKey, ref overrideBuildingName, ref pcLearnedAboutExistence, ref receivedDirectionalHints, ref locationWasMarkedOnMapByNPC))
                                             {
-                                                if (pcLearnedAboutExistence)
+                                                if (locationWasMarkedOnMapByNPC)
                                                     buildingQuestName = place.SiteDetails.buildingName; // get buildingName if involved in active quest (same as overrideBuildingName)
                                             }
                                         }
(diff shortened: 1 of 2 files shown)


Nystul-the-Magician:
need to fix a bit more (residence names don't show up right now)

Nystul-the-Magician:
I will have to make some changes to building discovery code to support "reveal on info click without pc having learned from townsfolk" and getting permanently marked on map as well
for now I decided to info click to behave like this:
- "residence" if residence was not marked on map by npcs
- named residence if residence was marked on map by npcs
so pc can not reveal for now by spam clicking all buildings

this is because the "get discovery code" will return a flag on finding building discovery but this needs to be different for info clicks than when automap uses the code, otherwise automap would always reveal all residences as named residences
I will put some time and effort into this issues next week
for now it should do as described above

Interkarma:
>I will have to make some changes to building discovery code to support "reveal on info click without pc having learned from townsfolk" and getting permanently marked on map as well but not accidently revealed when building map nameplates

Is this related to buildings involved in a quest?

Interkarma:
Just FYI, this also fixes the lockpicking regression issue #1373 that was also caused by the earlier change to discovery. My apologies for not spotting these two issues before merging, I was very time poor leading up to alpha. Happy to spend bit more time with you on this one. Between you, me, and @ajrb I'm sure we'll smash it. :)

Nystul-the-Magician:
> 
> 
> > I will have to make some changes to building discovery code to support "reveal on info click without pc having learned from townsfolk" and getting permanently marked on map as well but not accidently revealed when building map nameplates
> 
> Is this related to buildings involved in a quest?

yes, one aspect of it is about residences and if pc got them marked on map by npcs (override name is alone is not enough for this since it is available from quest start, so additional handling code that calls into talk manager checks if a npc marked the residence on the map or not for now).
all those "problems" and issues with nameplates sometimes not showing up on map is related to the different systems using the building discovery code (but need them to behave slightly different, e.g. automap nameplate generation needs to get building names but not reveal them but instead info clicks should reveal them) - the goal from my side is to have one function to handle all those cases but maybe with flags to control behavior (so we don't have duplicate code for building discovery name handling) - in the past we had 2 functions and e.g. player house naming would only be returned by one of the 2...

Interkarma:
Thank you, take your time. This end of the project is all about making things incrementally better. The mad rush to implement systems is over now.

Nystul-the-Magician:
working on this now - looking good - will make a new pull request when done - so no hurry to merge this one ;) will close this one when I am finished ;)

PR #1388, opened 2019-08-03
FindClosestDoorToPlayer() to return false if all doors are filtered out

petchema:
DaggerfallStaticDoors.FindClosestDoorToPlayer() was returning true if all doors were filtered out by requiredDoorType.
When this happened, doorPosOut would still have its default value of playerPos, making DaggerfallInventoryWindow believe that indeed a door was found close to the player.

This only happens in some dungeons. The issue was first seen in a stream, in the Citadel of Hawking, Daggerfall:

https://www.twitch.tv/videos/461269349?t=01h50m27s


petchema:
Gamesave in the Citadel of Hawking, Daggerfall
[SAVE324.zip](https://github.com/Interkarma/daggerfall-unity/files/3464410/SAVE324.zip)

Interkarma:
Thanks for submitting PR. I'd like to take a closer look at behaviour before merging. :)

Interkarma:
Thank you for fix! Very sensible, I just had to understand the problem a bit more for myself.

PR #1391, opened 2019-08-05
building discovery rework

Nystul-the-Magician:
First of all: PLEASE help me to test this out before merging - I tested quite a bit and it seems to work ok, but the devil does not sleep ;)

now regarding the changes:
reworked the building discovery system a bit - no fear - nothing too heavy.
just tried to split stuff a bit better between functions:
- DiscoverBuilding is now the only function that really changes discovery state (except one small exception about player house name which I plan to fix as well)
- GetDiscoveredBuilding only uses this data and does not compute any building names on its own (except player house name)
- GetBuildingDiscoveryData gets initial building information from current location (no building name expansion like quest-named residences, no thieves guild name, etc.) - this function is not used directly by other systems but used by DiscoverBuilding and GetAnyBuilding - so no code duplication anymore and thus less bugs hopefully
- GetAnyBuilding now uses the other two functions (GetDiscoveredBuilding and GetBuildingDiscoveryData) to get any building discovery data (undiscovered and discovered) - this is much cleaner now and fixes some incorrect answers to the "Where am I?" question, I tested to make sure this does not break dialog linked topics in quests (and my hardcore testing quest (the non-member mages guild quest with the missing book passed my tests)).

I would like to rename GetBuildingDiscoveryData into something that better reflects the fact that this function computes intermediate results without expanding quest named residences, special guildhalls etc. Would like to hear your suggestions, something like GetInitialBuildingDiscoveryData or GetRawBuildingDiscoveryData were my best ideas but they don't seem too good to me

some of the bugs that should be fixed now are:
- special guild halls should show up again
- named residence showing up on map but not telling the right name on info click should be fixed
- named residences not showing up on map in some cases should be fixed
- potential problem with the same residence used by 2 quests simultaneously should behave like this: info click when undiscovered will give one of the two, otherwise the name first marked on map (update: we can change this so it shows the last name that was marked by npcs on map - what do you think how it should behave?) 
- talk question "Where am I?" should now always give resolved names of residences etc.

### DIFF
--- a/Assets/Scripts/Game/ExteriorAutomap.cs
+++ b/Assets/Scripts/Game/ExteriorAutomap.cs
@@ -693,7 +693,7 @@ namespace DaggerfallWorkshop.Game
                                             string overrideBuildingName = string.Empty;
                                             if (place.SiteDetails.buildingKey == discoveredBuilding.buildingKey && GameManager.Instance.TalkManager.IsBuildingQuestResource(buildingSummary.buildingKey, ref overrideBuildingName, ref pcLearnedAboutExistence, ref receivedDirectionalHints, ref locationWasMarkedOnMapByNPC))
                                             {
-                                                if (pcLearnedAboutExistence)
+                                                if (locationWasMarkedOnMapByNPC)
                                                     buildingQuestName = place.SiteDetails.buildingName; // get buildingName if involved in active quest (same as overrideBuildingName)
                                             }
                                         }
--- a/Assets/Scripts/Game/Questing/Quest.cs
+++ b/Assets/Scripts/Game/Questing/Quest.cs
@@ -603,6 +603,14 @@ namespace DaggerfallWorkshop.Game.Questing
             // remove all quest topics for this quest from talk manager
             GameManager.Instance.TalkManager.RemoveQuestInfoTopicsForSpecificQuest(this.UID);
 
+            // undiscover all quest residences used by this quest
+            QuestResource[] allQuestResources = this.GetAllResources(typeof(Place)); // Get list of place quest resources
+            for (int i = 0; i < allQuestResources.Length; i++)
+            {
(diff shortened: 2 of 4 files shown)


Interkarma:
@Nystul-the-Magician I think you've done some great work here. I have tested as thoroughly as I can. This change has solved all the problems that I'm aware of and hasn't broken anything else that I've been able to find. The code changes look sensible and I understand why you had to make those changes.

I think this one is OK to merge. Unless someone else (paging @petchema, @ajrb) finds an issue, I'll put this into builds this weekend so that people can find TG and DB guild houses again.

Thanks for all your work on this.

ajrb:
Seems fine to me, although not sure what you mean by "except one small exception about player house name which I plan to fix as well".

As for naming, GetBuildingDiscoveryData -> GetBaseBuildingDiscoveryData ?

Nystul-the-Magician:
> 
> 
> Seems fine to me, although not sure what you mean by "except one small exception about player house name which I plan to fix as well".

I think I was unclear about this - I meant player house name resolving, not sure if the place where it does add the player house name is the optimal place.

Do you guys have a savegame with a character who owns a house so I can test stuff a bit. This was one of the things I could not test when reworking the code - since I lack a suitable savegame

> As for naming, GetBuildingDiscoveryData -> GetBaseBuildingDiscoveryData ?
this suggestion is good! thanks

ajrb:
@Nystul-the-Magician add money using console, go to bank, buy a house... :-)

Nystul-the-Magician:
you are right, never thought of this - lol ;)

Interkarma:
Are you happy with this on now @Nystul-the-Magician? If so, I'll put into builds this weekend. :)

Nystul-the-Magician:
yeah happy with this - just renamed that one function ;)

PR #1392, opened 2019-08-06
make sure you're grounded after fast travel

petchema:
Replacement for PR #1359 
It adds a "grounded" boolean parameter tto RepositionPlayer() to force new position to be on the ground; That parameter is used when the player is relocated to a StartMarker using a cardinal direction, which include fast travel case but should always be a safe thing to do.

I'll keep testing it, but I'm releasing the PR so that more people can test it.

Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=2307

### DIFF
--- a/Assets/Scripts/Terrain/StreamingWorld.cs
+++ b/Assets/Scripts/Terrain/StreamingWorld.cs
@@ -1267,8 +1267,9 @@ namespace DaggerfallWorkshop
 
         // Repositions player in world after a teleport.
         // If position.y is less than terrain height then player will be raised to sit on terrain.
+        // If grounded is true, position.y will be unconditionally set to terrain height.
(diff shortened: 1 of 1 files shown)


Interkarma:
I much prefer this approach, thank you! :)

Interkarma:
I much prefer this approach, thank you! :)

petchema:
Well, it seems even this generates issues, this time with travel to ship :(
Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=2521

I'm not sure if that can be fixed, or if this will have to be declared a Mountains & Hills mod issue at this point

Interkarma:
We could add additional exclusions when fast travelling to one of the two ship cells. I can't think of any other fast-travel target where the marker isn't positioned at ground level. Your new approach isn't a bad one though. I forgot about ships as well.

PR #1396, opened 2019-08-08
 Add loiter limit setting

ajrb:
Also refined layout a little, moved both repair options to be adjacent and made layout more balanced


Interkarma:
You love to sneak in the extra stuff haha. :)

Looks good, thank you mate! :)

PR #1401, opened 2019-08-09
Quest dispense lists

ajrb:
See thread: https://forums.dfworkshop.net/viewtopic.php?f=22&p=28902
Thanks to Firebrand for doing the quest names!

The message box with OK button is to prepare player for a potentially long wait while all the availiable quests are parsed. This is to get the names and ensure no broken quests get offered. Seems like a lot of space below the button. This is the best I could get it to look, but happy if someone can improve it.

### DIFF
--- a/Assets/Resources/defaults.ini.txt
+++ b/Assets/Resources/defaults.ini.txt
@@ -114,3 +114,4 @@ PlayerTorchFromItems=False
 CombatVoices=True
 EnemyInfighting=True
 EnhancedCombatAI=True
+GuildQuestListBox=False
--- a/Assets/Scripts/Game/Questing/QuestListsManager.cs
+++ b/Assets/Scripts/Game/Questing/QuestListsManager.cs
@@ -31,7 +31,7 @@ namespace DaggerfallWorkshop.Game.Questing
         Zenithar = 'Z',
     }
 
-    struct QuestData
+    public struct QuestData
     {
         public string name;
         public string path;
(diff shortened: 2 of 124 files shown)


Interkarma:
I've lookef into speeding up quest offer time in the past, and it's a real challenge. The compile step is quick, rather it's all the staging checks required to make sure quest deploys that hit hardest. Because the offer text usually displays stuff like target town/dungeon, travel time, NPC names, etc. all that work actually has to be done before the quest can even hit offer stage.

Something I've noticed though is that quest setup times are much faster in builds. If you create a test build on your end, does the "please wait" button still feel necessary? It's worth a comparison to see what I mean.. :)

ajrb:
I think it's worth having in there, there can be a lot of quests sometimes. I considered implementing a cut down quest parsing flow, but that's premature optimisation. Let's see how people react and take it from there?

Interkarma:
I wouldn't cut down parsing flow, I've already trodden well over this ground in 2017. The short delay isn't caused by parsing anyway, rather all the required staging to correctly present offer text.

I don't mind if you leave message in or not, just give a go in builds to see the difference in staging times compared to editor. :)

ajrb:
I should have been clearer, I mean cut down parsing for populating the list box. Nothing really needs staging at that point, just need to do that for the one that gets selected by the player.

Interkarma:
I see! Yeah that would be fine. :) I guess the the only trouble is when quest failed to stage and player sees the "gave the job to some spellsword" message. It kind of still works anyway, I think.

Interkarma:
This works really well! No problems encountered in several tests. Will merge to master and we can keep testing ahead of next builds.

PR #1402, opened 2019-08-09
Bugfixes

ajrb:
Add check to prevent errors with modded items. Not all magic items will use legacy enchantments. e.g. locators in Archs guild mod.
Allow wagon ingredient access consistently.
Stop magic items being sold at normal stores.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallGuildServicePopupWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallGuildServicePopupWindow.cs
@@ -736,12 +736,15 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
         {
             // Open potion mixer window if player has some ingredients
             CloseWindow();
-            for (int i = 0; i < playerEntity.Items.Count; i++)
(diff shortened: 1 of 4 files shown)


ajrb:
Also noticed that some checks for ingredients were not looking at wagon items

Interkarma:
@ajrb The magic item change in this PR looks like it might also prevent repairing items.
https://forums.dfworkshop.net/viewtopic.php?f=5&t=2494

ajrb:
I didn't add the repair magic items option, that was done while I was having a break from DFU.

Interkarma:
Nah, just tagged you looking for a bit of help and thought it might be related to the magic item change in this PR.

Once I had a look though, it was clearly not related, check out my edit above. :)

petchema:
Firebrand wonders if preventing magic items from being sold in normal shops is ok.
Forums: https://forums.dfworkshop.net/viewtopic.php?f=7&t=2491&start=10#p29492

I just tried selling enchanted clothes in a general store in classic, enchanted gems in a jewelry, enchanted longsword and shield at a weapon smith and each time it was possible

ajrb:
Oh, it was reported as a bug on forums so I fixed it. Since TG has a fence specifically for selling magic items, I assumed that implied that they couldn't be sold in normal stores.  Looks like I will need to do some thorough testing of classic and make sure it all matches when I get back.

PR #1407, opened 2019-08-10
add console command to open the potion maker

petchema:
Having to go see a potion maker makes potions brewing very cumbersome, if not useless.
Players that don't mind bending classic rules can use that command to brew potions from any place.

(Requested by Shinino on Discord)


ajrb:
No, the potion makers are perks of working hard to progress in guilds. A command would bypass this and is essentially a cheat. I would prefer to see a mod with a mobile cauldron that could be purchased when potion brewing access is reached. I'm much more a fan on in-game stuff and console is really for debug & development only.

petchema:
I agree that a mod would be better...
Then some mechanics could be added, say you have to learn the procedure to brew each new recipe from a potion maker.

But it will take time to implement, and right now the fact you need to go to a potion maker is one of the reasons potion brewing is not worth the trouble and barely anybody does it

ajrb:
Well I suppose it's consistent with all the other UI pop up commands. I don't like it though, although that might be because I spent a long time programming the guilds and services system.

PR #1410, opened 2019-08-11
Fix arrow blood splash position

JorisVanEijden:
Fix for bug https://forums.dfworkshop.net/viewtopic.php?f=24&t=2219

I stepped through it with a debugger and the splash for arrows against enemies was being spawned at 0,0,0.
So I copied the NPC arrow hitting logic and now it works again.


Interkarma:
Clear and works well, thank you! :)

PR #1415, opened 2019-08-11
Fix fast travel for characters with Damage from Sunlight disadvantage

EPNW:
I tested the behaviour in classic and it seems that fast travel for characters with `Damage from Sunlight` disadvantage behave just like fast travel for vampires (in terms of arrival).


Interkarma:
Thank you for the fix!

PR #1416, opened 2019-08-11
always select EQUIP when opening inventory explicitly

petchema:
However, keep selecting the WAGON (instead of nothing) when opening the inventory explicitly near a dungeon entrance when DungeonExitWagonPrompt is off.

This is a trade-off @ajrb and I want to try concerning proximity wagon access,
Forums: https://forums.dfworkshop.net/viewtopic.php?f=4&t=2490


Interkarma:
Sounds good. Let's see what kind of feedback we get!

Issue #1383, opened 2019-08-01
Macro %di expansion can fail and crash the game


df-ferital:
**Bug description**
Macro %di expansion can fail and crash the game.

**To Reproduce**
1. Load the attached save (courtesy of Interkarma, save provided on the forum thread regarding %loc issue)
2. Using the console, activate the NPC know everything flag to easily reproduce the issue.
3. Ask any NPC about Evelyna Kingsmith in **Tell me about only** (not in Where is people)
4. At some point, when the anwer uses %di, the game should crash.

The problem lies at line 230 in QuestMCP.cs in `LocationDirection` method: `LastPlaceReferenced ` is not supposed to be null there but it is with this particular quest.

**Expected behavior**
The game should never crash :)

**Desktop**
 - OS: Windows 10
 - Version 1803

**Additional context**
Discovered while testing my fix for %loc expansion...

[SAVE675.zip](https://github.com/Interkarma/daggerfall-unity/files/3458173/SAVE675.zip)

Nystul-the-Magician:
will take a look but after fixing the discovery issues - so please be patient

JorisVanEijden:
I noticed 2 problems with the %di functionality.
1. It doesn't check if there is actually a location referenced in the quest. (crash)
2. If the location from the quest is the same location you're in it doesn't work. (text like "it's ... never mind ... of here")

I created a PR (#1418) to solve these.
1. by checking if there is a referenced location.
and by adding a reference to a persons location when the quest mentions a person.
2. by getting the direction to the building inside the location.

Nystul-the-Magician:
thanks for taking a look!

JorisVanEijden:
Note, the offending quest text is `_contact_, eh? A mysterious =contact_ who operates out of __contact_, %di of here.`

As per InterKarma request I'm splitting this into multiple PRs:
#1425 for the crash
Supplied savefile will no longer crash but will say `Evelyna Kingsmith, eh? A mysterious blue-cloaked woman who operates out of The Ashwing Residence, ... never mind ... of here.`
#1426  for adding the contact's location to the quest
NPC will say `Evelyna Kingsmith, eh? A mysterious blue-cloaked woman who operates out of The Ashwing Residence, never mind ... of here.` Note the triple dots that are gone. This is because the location is now known, but the same as where the player is so it can't calculate a direction.

[...]

Issue #1408, opened 2019-08-11
0.10.4 Dark Brotherhood contact building not being marked on map again


Interkarma:
Reported by BansheeXYZ in below forum topic. Might be something we missed @Nystul-the-Magician 

https://forums.dfworkshop.net/viewtopic.php?f=24&t=2493&p=29006#p29006

Nystul-the-Magician:
whats interesting about this is:
i deleted the named residences from the discovery savegame data for the location. Then nothing is revealed on map as expected. When talking to the npcs about the residence (Kingsley Residence) and getting it marked on the map it gets marked as Wickston Residence - so it might be a problem with either:
- different residence name resolving in quests and discovery system
- or, a savegame from an older build where the residence name was resolved differently back then when quest started
@Interkarma: would you like to assist me with helping to track this down since I am a bit unsure how to proceed here

Interkarma:
Thanks for taking a look, Nystul! I'm starting to suspect second case, but will try to clarify when I can. Will take a closer look when I can. I appreciate your help. :)

Interkarma:
Have confirmed with Banshee this is a fresh save. As you're more familiar with talk system, are you able to let me know where NPC fails to mark on map in talk UI, and what info it's failing to find? I suspect now this might be related to automatic NPCs not being correctly placed by quest system. If you're at least able to confirm where things fail in talk UI, I can backtrack from there.

Thank you for any help you can give. :)

Nystul-the-Magician:
nothing fails building is added to map as expected just the name does not match the name that appears in log - so I suspect discovery system resolves to different name than quest

you can easily check by first removing the discovery data from savegame folder and then let an npc mark the residence on the map - it will appear but the name will be different

Interkarma:
Interesting, thank you for reinforcing. I will see what I can do. :)

Interkarma:
This appears have been resolved. Closing.

PR #1418, opened 2019-08-12
1383 macro %di expansion can fail and crash the game

JorisVanEijden:
Suggested fix for #1383

### DIFF
--- a/Assets/Scripts/Game/Questing/Person.cs
+++ b/Assets/Scripts/Game/Questing/Person.cs
@@ -4,8 +4,8 @@
 // License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
 // Source Code:     https://github.com/Interkarma/daggerfall-unity
 // Original Author: Gavin Clayton (interkarma@dfworkshop.net)
-// Contributors:    
-// 
+// Contributors:
+//
 // Notes:
 //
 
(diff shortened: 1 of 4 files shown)


Interkarma:
Hi there! Thank you for the PR, but I can see lots of changes across unrelated files and part of the code. Could you please tidy up PR just to the minimum required number of line changes. Thank you. :)

JorisVanEijden:
The whitespace changes are caused by the project's .editorconfig
The change in Questing/Person.cs is to fill the location reference when someone's home/location is mentioned.
The change in TalkManager.cs does look weird. Maybe something went wrong in merging.
The change in Text/ConversationText.txt is to provide a better fitting "error message" when the direction is unknown because you're at the same location.
".. located in the vicinity of here" looks better than the current ".. located never mind... of here"

I will look into forcing my IDE to ignoring the project's .editorconfig if whitespace changes are not acceptable (you may want to consider changing the .editorconfig in that case)
I will look at the changed made in TalkManager.

Do you want me to split the PR into 2?
1 for fixing the bug (only adding a null-check) and 1 for improving the way the %di macro works?

Interkarma:
Interesting, which IDE are you using? It's easy enough to change .editorconfig, I just haven't experienced this myself or from contributors. I'm using Visual Studio on this end.

Two PRs would be ideal, thank you! I can then fast-track the crash fix into next builds and give the %di changes some extended testing time.

BTW, I'm also taking a look at issue #1408 - there's a possibility I might also make edits to Person.cs. I won't be sure until I dive into it some more. I'm sure we can work any conflicts easily enough, just wanted to give you a heads-up.

JorisVanEijden:
I'm using Rider and it applies the settings from .editorconfig to every file committed. I've disabled .editorconfig support now for this project.

#1425 is the PR for the nullreference

Interkarma:
I've merged the two fixes #1425 and #1426 now. Thank you for splitting these out and working with me on the formatting side of things. It was much easier to review. Very much appreciated!

I'll close this one now as the edits are covered by those two PRs.

PR #1425, opened 2019-08-13
1383 Fix crash with %di getting direction without location.

JorisVanEijden:
Handle `DaggerfallWorkshop.Game.Questing.Quest.LastPlaceReferenced` returning null.


Interkarma:
Fix is clear and works well, thank you!

Just one minor thing. When adding text that is to be presented in game, I would prefer it go into the one of the text databases. These will be migrated into the global translation database when ready, whereas text in code won't be. I'll merge fix as-is then add to text database on my end.

PR #1426, opened 2019-08-13
1383 Add quest location reference from Person

JorisVanEijden:
Fixes messages like `Some weird-acting =contact_ in __contact_, %di of here.`
__contact_ would be filled by Person but not added as Quest.LastPlaceReferenced, which %di uses.

### DIFF
--- a/Assets/Scripts/Game/Questing/Person.cs
+++ b/Assets/Scripts/Game/Questing/Person.cs
@@ -293,6 +293,9 @@ namespace DaggerfallWorkshop.Game.Questing
             // Store this person in quest as last Person encountered
             // This will be used for subsequent pronoun macros, etc.
             ParentQuest.LastResourceReferenced = this;
+            Place homePlace = GetHomePlace();
+            if (homePlace != null)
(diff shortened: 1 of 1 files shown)


Nystul-the-Magician:
i am not sure if homeplace is correct to be used here, if I remember correctly this can differ from the current npc location (maybe @Interkarma can remember - I think there was something about it)

JorisVanEijden:
Yes, it is the home or start location of the npc and this can differ from
the current npc location.
But this is what is used by the double underscore and triple underscore
"macros". And those are often the locations referred to by %di.
As in `_person_ comes from __person_ which is to the %di of here`

Nystul-the-Magician:
alright then

Interkarma:
Thank you for the fix!

Interkarma:
Have discovered this plays havoc with the "travel from quest journal" QoL feature. Quest journal will now try to send player to NPC Home location rather than the Place mentioned in the journal entry. I'll see if I can unify the two requirements. @ajrb Just letting you know that I'm looking into this now.

PR #1428, opened 2019-08-14
fix for answer rumors (answers to tell me about topic) being incorrectly added to "any news" rumors

Nystul-the-Magician:
fix for answer rumors (answers to tell me about topic) being incorrectly added to "any news" rumors

this issue leaded to some quests being easier to solve than intended (or in worst case trivial to solve)
tested with non-member mages guild quest with missing book (N0C00Y10)

answer rumors (given as answers to "tell me about" topics) defined in the quest source in the line a quest resource is defined, usually after "anyInfo" with the command "rumors"
answer rumors should not be confused with "any news" rumors (can be populated with RumorsDuringQuest and RumorMill quest action)
answer rumors must not appear as answer to "any news" questions.
usually "any news" answers are the less informative ones, resource related anyInfo and rumors are more specific but anyInfo are meant to be correct and rumors may be wrong (that's why the spymaster always gives the anyInfo answers)

due to a mistake on my side in commit ec37210e12e9690d3eb5da731d83dfa133f7b424 those answer rumors got added to the "any news" answers

in quest N0C00Y10this issue resulted in npcs giving "any news" ansers like "_thief_ lives at _house_ in __house_" which made no sense: why would a random npc 1) know about the specific person, 2) why would he/she tell the pc as random rumor

after this change the correct way to get to the thief is: ask about the missing book name in tell me about and learn about the contact person, who then will give the name of the thief, this will make the name of the thief appear under "tell me about" - which will give the anyInfo answers (message 1014) - npcs of certain type like scholars will give these - or rumors answers (message 1015) by common passengers.
"any news" will only give the unspecific answers from RumorsDuringQuest as intended (message 1005)

this pull request fixes this issue

### DIFF
--- a/Assets/Scripts/Game/TalkManager.cs
+++ b/Assets/Scripts/Game/TalkManager.cs
@@ -1891,10 +1891,6 @@ namespace DaggerfallWorkshop.Game
 
             // update topic lists
             rebuildTopicLists = true;
-
-            // Update rumor mill
-            if (questResourceInfo.rumorsAnswers != null)
-                AddQuestRumorToRumorMill(questID, questResourceInfo.rumorsAnswers);
         }
 
         public void DialogLinkForQuestInfoResource(ulong questID, string resourceName, QuestInfoResourceType resourceType, string linkedResourceName = null, QuestInfoResourceType linkedResourceType = QuestInfoResourceType.NotSet)


Nystul-the-Magician:
if any custom quest (thinking of quests by https://github.com/JayH2971) want to add rumors to both the "tell me about" rumors answers as well as to the "any news" rumors the correct way to do it is to add it to the quest resource (for the "tell me about" rumors answers) and then use rumor mill command on quest startup to add it also to the "any news" rumors

Interkarma:
Thank you!

Issue #1427, opened 2019-08-14
0.10.4 The dagger icon in character creation can go off scale.


DSeyka:
When creating a character, adding enough advantages or disadvantages can cause the dagger icon on difficulty scale to go off it, and even off-screen.
![image](https://user-images.githubusercontent.com/34726992/62995782-d91edc80-be6a-11e9-85e9-afe70fc42ed6.png)
OS is Windows 10, game version is 0.10.4, unmodded.
It also happens in 0.10.3, and I *think* it didn't use to prior to alpha releases.

DSeyka:
Note: the issue is cosmetic only, since the game still won't allow you to create a character when the dagger is out of bounds.
The relevant code is [url=https://github.com/Interkarma/daggerfall-unity/blob/e7d221473581d800f91c8cb351f757af2458a8e2/Assets/Scripts/Game/UserInterfaceWindows/CreateCharCustomClass.cs#L466]here[/url], it should be trivial to just clamp its y coordinate.

PR #1430, opened 2019-08-14
visually clamp difficulty dagger to gauge

petchema:
Fixes issue #1427


Interkarma:
Thank you for the fix! :)

PR #1434, opened 2019-08-16
no fog on automap

petchema:
Make sure markers are not foggy when looking at the automap underwater

I added a script to the camera's gameObject to temporarily disable RenderSettings' fog, I don't know if that's the simplest way to fix that.

Forums:
https://forums.dfworkshop.net/viewtopic.php?f=4&t=163&p=29038#p29038


petchema:
Some underwater save for testing:
[SAVE42.zip](https://github.com/Interkarma/daggerfall-unity/files/3515002/SAVE42.zip)

Nystul-the-Magician:
adding the script is ok I think, but it might be easier to just change RenderSettings' fog when automap is opened (function UpdateAutomapStateOnWindowPush) and backup old settings when it is closed (function UpdateAutomapStateOnWindowPop)
It is a matter of taste and design philosophy imo. I would prevent adding a script just for this one purpose tbh ;)

petchema:
The benefit of switching fog on the camera is that the water keeps being rendered correctly in the background:
automap-fog

Nystul-the-Magician:
that is a good point - from my side this justifies your approach as superior ;)

Interkarma:
I'm sorry @petchema this one now has some conflicts in automap after merging Nystul's recent fixes. Might need to quickly rebase or add changes again. My apologies for the hassle.

petchema:
Conflict was coming from change of line-return style, easy fix :)

Issue #1369, opened 2019-07-30
Quest N0B00Y16 Hand-in Regression


Interkarma:
**Describe the bug**
Quest N0B00Y16 results in failure even after successfully opening box and returning to quest-giver. Quest-giver is then locked out as quest unable to terminate. Calling in @JayH2971 for some help with this one. :)

**To Reproduce**
Load below save and attempt to hand in quest.

[SAVE671.zip](https://github.com/Interkarma/daggerfall-unity/files/3445351/SAVE671.zip)

**Expected behavior**
Quest should hand in normally for reward if successful.

Interkarma:
Apologies for tagging Ferital & Nystul earlier. I thought this was an NPC talk issue until I realised it was actually the quest itself.

JayH2971:
Acknowledged.

Interkarma:
In his PR to workaround this, Jay mentions the below doesn't fire in this quest.
```
_S.07_ task:
	have _crn_ set _S.08_
```
Leaving this issue open to remind to look at why this is.

JorisVanEijden:
While running this in the debugger the HaveItem action in S.07 is checking each Update if the small chest is in the players inventory (it is) and it triggers S.08. Which does nothing because it's isComplete is already true.
So that part is working.

I could also hand in the chest at the merchant, but in one attempt I got message 1050 and 1040 at the same time and an Imp spawned. In another instance I got message 1050 and 1051 at the same time.
So something is not quite what it should be.
I'd like to continue debugging this tomorrow to learn more about the quest system.

Interkarma:
Thank you for the clarity @JorisVanEijden. Makes sense and I can see that I missed this when setting up action.

PR #1437, opened 2019-08-17
1369 have item should not clear target

JorisVanEijden:
Fix for (among others probably) #1369 
`have _item_ set _state_` is only supposed to _enable_ `_state_`, not disable it.

As per 
https://en.uesp.net/wiki/Daggerfall:Quest_pseudo-codes#Pseudo-code_Type_0x46_.2F_70_.28PC_has_Items.29_.5B5_sub-records.5D
and 
https://www.dfworkshop.net/static_files/questing-source-docs.html#usingitems


Interkarma:
Thank you!

PR #1441, opened 2019-08-19
Add journal message when getting a house

ajrb:
So players can easily keep track of where their owned properties are.


Interkarma:
Nice! Thank you for adding this.

Issue #1421, opened 2019-08-13
Quest S0000011 Secret door unable to be opened


yoboseo:
**Describe the bug**
A secret door on a cave wall in Scourg Barrow does not open. This door leads to a hand wheel which opens a nearby portcullis leading to one of the possible quest objectives, leaving the quest impossible to complete should you take this path during the quest.

**Expected behavior**
The door should open when clicked on.

**Screenshots**
[Location of the door and attached savefile standing in front of the door.](https://i.imgur.com/j8jQFgl.png)
[SAVE73.zip](https://github.com/Interkarma/daggerfall-unity/files/3495971/SAVE73.zip)

JayH2971:
What quest are you referring to? I'm not aware of any quest that requires this path, and you don't have one active in your save file.

petchema:
I'm not aware of any quest related to that part of Scourg Barrow either; But I checked in classic and indeed there's a secret passage there, you just need to click it so it opens.

yoboseo:
S0000011 - Barenziah's Book. The objective is to find a book in one of two dungeons. If you act fast the book is in the Stronghold of Orsinium. If you take too long the necromancers steal the book and put it in Scourg Barrow. Its location in Scourg Barrow is north of the cave door on a pedestal in a room behind a portcullis, which is opened by activating a hand wheel behind the cave door.
I didn't have the quest active in that test file, I just wanted to get to the door quickly.

petchema:
I just learned something then :)

JayH2971:
There isn't anything more I can do for this issue. This goes all the way back to 0.1, object activation. For some reason the object here won't open the door.

JorisVanEijden:
Seems to missing an action on Model 55033 (description "CAV") in S0000204.RDB
Does anyone have an example of a secret door that does work?

JorisVanEijden:
Never mind, solved it in referenced pull request. 
And apparently one other location got fixed that used the same door model.

petchema:
one other location? This is really intriguing :smile:

PR #1453, opened 2019-08-21
Add "CAV" to ActionDoor check for hidden cave doors

JorisVanEijden:
Tested and confirmed to solve issue #1421


Interkarma:
Cheers! I thought it would be something like that but haven't had the time to look yet. Happy it was a simple solve.

PR #1455, opened 2019-08-21
Enchanted items should not be considered ingredients

ajrb:


Interkarma:
Good catch! I forgot to consider that case after setting up enchanting. Thank you.

PR #1457, opened 2019-08-22
adjust changed default keybindings in tutorial

petchema:
Not as good as having macros to use current keybindings instead, but I didn't manage to implement that.

Forums: https://forums.dfworkshop.net/viewtopic.php?f=4&t=429&p=27238#p27238


Interkarma:
Thank you! Even that small change is a good one.

Issue #1449, opened 2019-08-20
Direction to same location fails.


JorisVanEijden:
**Bug description**
When a "Tell me about..." conversation reply mentions a direction to a location (through the `%di` token) it will fail when you are already at that location.

**To Reproduce**
1. Load the attached save [SAVE675.zip](https://github.com/Interkarma/daggerfall-unity/files/3520818/SAVE675.zip)
2. Open the console, use `talk_NpcsKnowEverything` and close the console.
3. Click on an npc.
4. Use "Tell me about.." to ask about "Evelyna Kingsmith".
5. Note the response: "Evelyna Kingsmith, eh?  A mysterious blue-cloaked woman who operates out of The Ashwing Residence, **...never mind...** of here. There may be more to it."
(You may have to ask about her a couple of times to get the location response)

**Expected behavior**
I expect the response to state "North of here" in stead of "...never mind... of here", since the Ashwing Residence building is to the North of the player's current position.

**Desktop:**
 - OS: Windows
 - Version: git master, commit 5ca7b52a1eb1c4d8e943df53c3fe1a196aecef79

**Additional context**
If you use "Where is..." to ask about Evelyna Kingsmith the reply will tell you she's to the north.

The "about" code at `Questing.Quest.QuestMacroDataSource.LocationDirection()` is only accurate to Location granularity.
The "where" code at `TalkManager.TalkManagerDataSource.LocationDirection` is accurate to building granularity but is coupled to `TalkManager.currentKeySubjectBuildingKey`.

We could copy the code from the talkmanager to the questdatasource, or rewrite the talkmanager code to split up the functionality to create a method that both can use.
But maybe it is possible to create something more generic (in the PlayerGps?) that can always give you the direction from the player to any Location, Building, Region, Npc, etc?

Nystul-the-Magician:
do you want to work on this or should I take a look?

JorisVanEijden:
I'm missing the experience and background with DFU to make an informed decision on the optimal way forward.
It's easy enough to make building granularity work in the QuestMacroDataSource. But having identical functionality in two very different parts of the code smells like a need for refactoring.
If you say to go for the localized minimal-change solution then I can work it out.
If you share the feeling this might be extracted and generalized to a different location then I'm a bit out of my depth as to where and how this should go.

Nystul-the-Magician:
hmm not sure, I wrote the TalkManager variant and tested it back then
[...]

PR #1462, opened 2019-08-23
 %di macro improvements and bugfixes, code rework

Nystul-the-Magician:
fixes issue https://github.com/Interkarma/daggerfall-unity/issues/1449

- tell me about answers for persons with %di should work now (both indoors (buildings and castles) and outdoors), directions given should be correct
- "where is" answers for residences fixed (named residence instead of just "residence")
- tell me about entries are available everywhere, but only npcs in same region of place or person resource will know answers (TalkManager function CheckNPCcanNowAboutItem is used to check if npcs can know) - this will even apply to spymaster and console command talk_npcsKnowEverything
- %di in tell me about answers for persons in same region but different location will now correctly refer to the location name followed by compass direction to the remote location
- code improvements, cleanup and function renaming
- fix for npcs sometimes not giving the correct building name if it was undiscovered
- fixed %di resolving error when in same building as target npc
- npcs will now always know tell me about and where is answers if the person is in the same building
- functions renamed to better fit their purpose - name correctly reflects now if the functions is used for direction for locations or buildings (TalkManager has new function GetLocationCompassDirection (not to be confused with function GetBuildingCompassDirection) that resolves to compass direction to location of place resource)
- QuestMCP direction resolving now uses TalkManager functionality

tested with quests: R0C10Y12, S0000001, N0C00Y10


Interkarma:
I've been playing with this on and off for a while now, and have not experienced any obvious problems or regressions that I've noticed. I'll merge this one now. Thank you Nystul!

PR #1464, opened 2019-08-25
Add exclusive fullscreen checkbox to advanced video options

FrozenFish24:
`-window-mode exclusive` arg wasn't working, so I added it as a proper option.

exclusive

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallPauseOptionsWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallPauseOptionsWindow.cs
@@ -293,6 +293,15 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
         private void FullScreenButton_OnMouseClick(BaseScreenComponent sender, Vector2 position)
         {
             fullScreenTick.Enabled = DaggerfallUnity.Settings.Fullscreen = Screen.fullScreen = !Screen.fullScreen;
+
+            if (DaggerfallUnity.Settings.Fullscreen && DaggerfallUnity.Settings.ExclusiveFullscreen)
+            {
+                Screen.SetResolution(
+                    DaggerfallUnity.Settings.ResolutionWidth,
+                    DaggerfallUnity.Settings.ResolutionHeight,
+                    FullScreenMode.ExclusiveFullScreen);
+            }
+
             if (!saveSettings)
                 saveSettings = true;
         }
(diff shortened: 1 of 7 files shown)


TheLacus:
Could you _fixup_ the second commit into the first?

Interkarma:
Hey there, thanks for the PR! I'm happy to have an exclusive fullscreen option. If nothing else it will help people using gsync/freesync.

One thing, could you please change the option from "Exclusive" to "ExclusiveFullscreen"? This makes it a bit clearer from the setting along what it's for. When existing users sync this setting for the first time, it will appear at the bottom of the group list and won't be as apparent that it's related to fullscreen.

I'll test changes when I can and let you know if any feedback.

FrozenFish24:
Renamed to `ExclusiveFullscreen` and squashed commits. However, I think I've run into a Unity bug.

Switching between Borderless Fullscreen and Exclusive Fullscreen seems to ignore the resolution and forces native res until you `Screen.SetResolution` a second time. For example if I set my res to 1280x720, check `Fullscreen` and hit apply then check `Exclusive Fullscreen` and apply, the res changes to 1920x1080 (my native) despite appearing correct in the options and ini.

Starting the game does another `Screen.SetResolution` so maybe this isn't such a problem.

I think this is the same issue:
https://issuetracker.unity3d.com/issues/screen-dot-setresolution-width-height-fullscreenmode-disregards-resolution-when-changing-to-exclusivefullscreen

FrozenFish24:
Good thinking, I've commented out the UI option.

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallPauseOptionsWindow.cs
```diff
@@ -293,6 +293,15 @@ private void ControlsButton_OnMouseClick(BaseScreenComponent sender, Vector2 pos
         private void FullScreenButton_OnMouseClick(BaseScreenComponent sender, Vector2 position)
         {
             fullScreenTick.Enabled = DaggerfallUnity.Settings.Fullscreen = Screen.fullScreen = !Screen.fullScreen;
+
+            if (DaggerfallUnity.Settings.Fullscreen && DaggerfallUnity.Settings.ExclusiveFullscreen)
```
Interkarma: Just FYI, I'll merge this as-is but the Fullscreen button in DF will soon be reverted to toggle large HUD on and off. It had a different meaning in classic than the modern fullscreen/windowed.

Interkarma:
Thank you, merging this one now!

PR #1465, opened 2019-08-25
Fix NPC talk reactions

df-ferital:
Fix NPC reactions issues mentioned on the forums, especially here: https://forums.dfworkshop.net/viewtopic.php?f=24&t=2410&start=10#p29493

This restores random seed based on current NPC guid so each NPC in the same town will now react differently as this was already the case before PR #1370.

I also lowered the difficulty level required to get a successfull answer when asking for people location with low personality characters, balancing this change by also lowering the bonus given to find a location (+5 instead of +10). Still, it's much more difficult than classic to find people, where we nearly always get a successful answer even with a very low personality, because of an incorrectly applied modifier (classic seems to always apply +10, regardless of the kind of question).

As with my previous PR #1370, Streetwise and Etiquette are more useful than in classic where one would always use Normal tone since it almost always gives better results. As an example, in DFU, I managed to get a positive answer from the same NPC by switching from Normal to Blunt tone (NPC was a commoner, for a Noble I would have used Etiquette of course).

And as requested on the forums (https://forums.dfworkshop.net/viewtopic.php?f=5&t=2404), tone now remains checked as it was during the last talk session. Since this does not match classic, I let you decide (@Interkarma and others) if you prefer to keep it that way or not.

### DIFF
--- a/Assets/Scripts/Game/TalkManager.cs
+++ b/Assets/Scripts/Game/TalkManager.cs
@@ -95,7 +95,7 @@ namespace DaggerfallWorkshop.Game
         const int minVeryLikeReaction = 30;
 
         // Changed from classic (classis uses { 10, -5, 0, 0, 0, -10, -5, -5 } )
-        readonly short[] questionTypeReactionMods = { 10, -5, 0, 0, 10, -5, 0, 0 };
+        readonly short[] questionTypeReactionMods = { 5, 0, 0, 0, 5, 0, 0, 0 };
 
         /// Improvement over classic, add reaction modifiers when using Etiquette
         /// and Streetwise in relation to social groups.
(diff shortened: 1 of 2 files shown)


Interkarma:
Ah, I see what was happening now. Great work figuring that out!

Very happy for tone to remain on last selection. Definitely a click-saver for skilling up. I know this one will make @Nystul-the-Magician happy. :)

bansheexyz:
From a design standpoint, is there any reason to make tone selection a manual process? The player is always going to want the best reaction, so the only reason he wouldn't pick the best tone would be if he didn't understand what it did. It's not like there's some strategic decision here where the info gotten is different based on tone, it's get info or don't get it.

df-ferital:
Using Blunt or Polite tone requires a skill check where Normal does not. Failing this check gives an additional penalty instead of bonus. So no, it is not preferable to automatically select tone based on NPC social class as characters with very low Streetwise or Etiquette skills might always fail.

ajrb:
I didn't know that actually, another one for the player manual!

Interkarma:
Looking forward to pushing this into new builds. Thank you! :)

PR #1466, opened 2019-08-26
Update npc presence when rest dialog dismissed

ajrb:
I get annoyed when resting in a guild for npcs to arrive and have to exit and reenter the building for them to show up. Added a check when rest dialog is dismissed.


petchema:
Will they also be able to disappear as you sleep?
It's certainly more logic, but some people may whine (say, rest in fighters guild to (ab)use trainer...)

ajrb:
You think it should only work to make NPCs appear. Yeah, I think I agree actually, since I've done that myself. So PR updated!

ajrb:
When I originally implemented NPC schedules and closed the classic loophole allowing save/load to have the shopkeep vanish for easy stealing, the possibility of ejection was discussed and discarded for core DFU.

Interkarma:
Love it, thank you! :)

PR #1469, opened 2019-08-28
Travel Map Refactor

Interkarma:
Reworks travel map to fix memory leaks and excessive texture creation. Now textures are created only when needed and cached for next use.
Animated elements such as identifying region and location crosshair are done with a single persistent pixel buffer and texture rather than rebuilding several new textures almost every frame.
Panning and zooming are now done with GPU cropping in Panel rather than creating new textures almost every frame.
Functionality and behaviour seems to match previous travel map interface (minus the memory problems). No obvious regressions found at this time. Tested both with classic and DREAM UI textures.
Merging to master so other developers can start playing with changes. Please let me know if any feedback @petchema, @ajrb, @Nystul-the-Magician, @df-ferital, @TheLacus. I'm going to delay new builds to first weekend of September so this has some time to break in. Thanks all!


petchema:
@TheNewBob Since Tedious Travel uses its own copy of the travel map code, I assume it won't break (to be tested), but it won't benefit from those improvements until it is updated.

PR #1473, opened 2019-08-29
Fix issues when activating Sheogorath

### LINKED ISSUE #1456: Info mode on Sheogorath gives out a commoners name
Couldn't find this in the commits to be already fixed, so here we go - Bug reported by pango [here](https://forums.dfworkshop.net/viewtopic.php?f=24&t=2068).

**Describe the bug**
By clicking via Info mode on Sheogorath he gets a random(?) name and is handled as a normal NPC in this matter

**To Reproduce**
1. Go to Sheogorath
2. Click via Info mode on Sheogorath
3. A name appears above the screen.

**Expected behavior**
According to Jay: "It turns out Sheogorath has no name. If you click on him in any way, including F3, he'll give his speech. If you're out of speech range, nothing will show up; it'll just say, "You're too far away."

**Screenshots/Save**
See first post in thread above.

**Desktop 
 - OS: Windows, Linux Universal

df-ferital:
Sheogorath should not be activated like a normal NPC, neither in information nor any other mode. As this is also the case for all static NPCs attached to "ShowText" or "ShowTextWithInput" action, I just moved the already available check from TalkManager to PlayerActivate and included "ShowText" action flag into it.

In information mode, this matches classic behavior where we never get any display name for this kind of NPC.

In other modes (steal, grab and talk), this fixes an exception which popups for Sheogorath due to the fact both his instances have an invalid faction ID.

Fix #1456

### DIFF
--- a/Assets/Scripts/Game/PlayerActivate.cs
+++ b/Assets/Scripts/Game/PlayerActivate.cs
@@ -485,6 +485,15 @@ namespace DaggerfallWorkshop.Game
 
         void ActivateStaticNPC(RaycastHit hit, StaticNPC npc)
         {
+            // Do not activate static NPCs carrying specific non-dialog actions as these usually have some bespoke task to perform
+            // Note: currently only ShowText and ShowTextWithInput NPCs are excluded
+            // Examples are guard at entrance of Daggerfall Castle and Benefactor and Sheogorath in Mantellan Crux
+            DaggerfallAction action = npc.GetComponent<DaggerfallAction>();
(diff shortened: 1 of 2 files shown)


df-ferital:
This also fixes issues with guards at entrances of Daggerfall and Sentinel Castles. Therefore, I removed previous checks using billboards.

ajrb:
Great stuff. That id check was smelly as hell.

Interkarma:
Clear and matches classic. Thank you!

PR #1474, opened 2019-08-29
Fix for coin anim on inventory window

ajrb:
It was only being set when window was setup, and once set the anim was either there or not for the rest of the session.

See: https://forums.dfworkshop.net/viewtopic.php?f=5&t=2610


Interkarma:
Cheers! :)

PR #1475, opened 2019-08-31
Allow magic item subclass definition in quests

JorisVanEijden:
Fix for https://forums.dfworkshop.net/viewtopic.php?f=25&t=1098&start=30#p28634

Implements subclasses for magic items (class 4) in quests.

Testcase:
`startquest S0000501`
That quest uses 
`Item _reward1_ item class 4 subclass 26`
which should generate a random "Unrestrainable" item as \_reward1\_


Interkarma:
This is an excellent change and works well. Thank you! :)

PR #1476, opened 2019-08-31
Missing book noises

petchema:
Complete the sonorisation of journal/notebook, spellbook and book reader:

Journal/notebook:
- OpenBook on open (like classic)
- PageTurn on page flip (like classic)
- OpenBook when switching between active quests, finished quests, etc. (like classic)
- PageTurn when editing the notebook (similar to classic spellbook editing)

Spellbook:
- PageTurn when editing the spellbook (like classic; except that in classic when you rename a spell the sound is played before you enter the name, and when the spell is renamed, it is immediately cast!)

Book reader:
- PageTurn on page flip (like classic)
Classic also plays ButtonClick on EXIT, but it doesn't seem to match Daggerfall Unity practice.

PageTurn are throttled, because they were easy to spam, specially with mouse wheel.


Interkarma:
I like the "sound" of this PR. :)

One thing though - I've changed the books from page-turning to scrolling recently. I know you've throttled it, but it still feels out of place and too rapid to me in a scrolling UI where only a few lines are changing at a time.

petchema:
I added some logic to only play PageTurn even page worth of text.
Sounds better, but feels a bit overkill?

Interkarma:
You're starting to make me think we should change the behaviour back to flipping pages. If we're calculating the page span, might as well make the page-flip operation closer to classic again. What are you're thoughts?

You don't need to do that work if you don't want to btw, just inviting opinion. :)

petchema:
Well, I find scrolling text easier to navigate, so I think it should be the last option to consider.
I really like the approach you used for the BookReader, paginate the whole text once, then display what is inside a sliding window. If anything, I'd use a similar approach for the journal/notebook too, because it lends itself better to smooth scrolling and having a scrollbar over the whole text, say in a modded UI.
With that in mind, static pages could either be determined during rendering, or the idea of pages totally forgotten and PageTurn sounds called vestigial and dropped with it.

Either way, what I did with this PR, which is not computing static pages exactly, just checking when there's no text overlap with the last time I called what's on screen "a page", like tumbling windows, would become obsolete, on top of being overly complex. I'll remove it immediately, so the other changes can be merged.

petchema:
Done

petchema:
> With that in mind, static pages could either be determined during rendering

In fact, when the whole text is already rendered and just scrolled over, adding PageTurns is dead simple: https://github.com/Interkarma/daggerfall-unity/compare/master...petchema:pageturn-sounds

Interkarma:
That's cool! I'm glad you prefer the scrolling though, I thought it made books easier to read as well.

Let's see what kind of feedback we get, and go back to page-turning if that's the overwhelming response.

SivaMachina:
Logically the page turning for books makes sense. Because they are books not scrolls. you could use the mouse wheel to turn pages.

petchema:
Yes, I think books deserve page turn sounds.
Will see if/how page turns are added to journal later, based on users feedback.

PR #1478, opened 2019-08-31
Make more npc types local residents.

JorisVanEijden:
Testcase: quest K0C00Y04 has 2 Npcs in Group_7
Story and timer seem to indicate these have to be local and the quest itself doesn't place them anywhere.
PR makes not only -3 and -4 local, but also -5 and lower. (-8 is individual ans is already always local)

Original report: https://forums.dfworkshop.net/viewtopic.php?f=29&t=2331


Interkarma:
Individuals are definitely not always placed locally. A good example is Brisienna, who is an individual NPC always placed to a remote tavern.

I'd like to investigate the test case a little bit more before making a decision on this one.

PR #1481, opened 2019-09-01
Don't end quest before revealing location if needed.

JorisVanEijden:
Fix for https://forums.dfworkshop.net/viewtopic.php?f=29&t=2411&start=10


Interkarma:
Very good! That decompiler of yours is looking like a very valuable tool. :)

PR #1482, opened 2019-09-01
Prevent vampire sun damage in prison.

JorisVanEijden:
Minimal code changes approach to prevent vampires burning up in prison.

A larger change would be to tie make the court window count as IsPlayerInside, but since the court window explicitly transitions to exterior and the player inside/outside is tightly coupled with the transitions I decided to go another route.

Source: https://www.reddit.com/r/daggerfallunity/comments/cwcsrb/is_going_to_prison_supposed_to_kill_vampires/
https://forums.dfworkshop.net/viewtopic.php?f=24&t=2605


Interkarma:
That's quite a clever and minimal solution, thank you. I like that it will also address the other effects that cause damage in sunlight like class disadvantages or enchanted items.

I was thinking about how to solve this one, but this is quite a bit neater than what I'd come up with so far. :)

PR #1484, opened 2019-09-01
workaround for "water everywhere" texture issue on low-end hardware

petchema:
It does not fix shadows issues or anything else, but still that's a noticeable improvement for people that have to run DFU on that hardware.

That looks like a very inexpensive workaround (most likely lost in the noise), and it seems to work fine on my regular hardware (from mathematical standpoint it also feels right to use rounding instead of floor() to make sure int -> float -> int conversions are robust), so maybe it can be added directly into the standard shader?

Otherwise, say if the risk of regression is deemed too high, it could be added to a variant of the shader, so it could be enabled at runtime; But I don't know how to do such things...

Tested on Intel HD-3000 by Siva Machina on Discord;
Tested on nVidia GTX-960 by myself (for lack of regression)

Forums:
https://forums.dfworkshop.net/viewtopic.php?f=5&t=2519&p=30610#p30610

### DIFF
--- a/Assets/Shaders/DaggerfallTilemapTextureArray.shader
+++ b/Assets/Shaders/DaggerfallTilemapTextureArray.shader
@@ -85,7 +85,7 @@ Shader "Daggerfall/TilemapTextureArray" {
 		void surf (Input IN, inout SurfaceOutputStandard o)
 		{
 			// Get offset to tile in atlas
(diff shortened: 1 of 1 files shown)


SivaMachina:
If more information of my laptop is a Dell Latitude E6420 with 8 GB ram running Kubuntu 18.04.

Interkarma:
No issues experienced with this change on my hardware. Seems like it's a safe change to make for everyone. Well done! :)

PR #1485, opened 2019-09-01
Harden received armor

petchema:
3 modifications:
- wait for the weapon smith MessageBox to be closed before opening the inventory;
- make sure an item is picked from the list of proposed items (and, say, not from the wagon) before closing the inventory;
- close all transient windows before closing the inventory.

Use case for the latter: the player selects to EQUIP some item he cannot equip (forbidden material, for example). EquipItem() opens a MessageBox to tell it's not possible, chooseOne closes to top window which is the MessageBox, and the inventory stays open.

Another implementation would be to set a flag, and close the inventory in Update() when all transients have been closed by the player. The benefit is that the player would see that the item couldn't be equipped.


petchema:
Game save: Aspirant knight of the dragon that cannot wear iron armor, in front of the weaponsmith of the guild
[SAVE371.zip](https://github.com/Interkarma/daggerfall-unity/files/3564511/SAVE371.zip)

ajrb:
Code looks good to me.

ajrb:
No I don't think that's necessary. If player really wants a specific item of armour, let them try again. :-)

PR #1487, opened 2019-09-05
Quest item dropping fixes

JorisVanEijden:
- Picking up previously dropped quest items would not unset the dropped state because they come from Loot, not Dropped.
- Moving quest Items to Loot piles or Merchants would not set the dropped state.
- Picking up quest items without AllowDrop would not unset the dropped state.
- Bitwise or in stead of logical or could cause null reference.

Test quest:
```
- Test quest for dropping and picking up quest items
Quest: __TEST001
QRC:
QBN:
Place _house_ local house2
Item _treasure_ portrait

	place item _treasure_ at _house_

_pickedup_ task:
	clicked item _treasure_
	clear _dropped_

_dropped_ task:
	dropped _treasure_ at _house_
	clear _pickedup_
```


Interkarma:
Yeah, that logical or typo is embarrassing. Thanks for the fixes Joris! :)

PR #1488, opened 2019-09-05
Fixed unsupported usage of "until" idiom

JorisVanEijden:
Fixed unsupported usages of "until" idiom where a condition should unset a variable.

See https://forums.dfworkshop.net/viewtopic.php?f=29&t=2628#p30745 for background.


Interkarma:
Thank you!

PR #1489, opened 2019-09-05
Implement classic tweak regarding some Order of the Raven buildings

df-ferital:
Found while reverse engineering FALL.EXE to understand map buildings to blocks building mapping process.

Classic uses this exact same runtime patch for some Order of the Raven buildings. These can be found  in Dwynnen or Whitewold for example, and aren't colored in blue on the map.

This is really hacky stuff but it works!


Interkarma:
Great stuff, thank you!

PR #1490, opened 2019-09-05
A closed guild now display "Guild is closed." message.

df-ferital:
A little cosmetic change. This does not match classic but guilds are not stores...

### DIFF
--- a/Assets/Scripts/Game/PlayerActivate.cs
+++ b/Assets/Scripts/Game/PlayerActivate.cs
@@ -291,10 +291,10 @@ namespace DaggerfallWorkshop.Game
                     if (!buildingUnlocked && buildingType < DFLocation.BuildingTypes.Temple
                         && buildingType != DFLocation.BuildingTypes.HouseForSale)
                     {
-                        string storeClosedMessage = HardStrings.storeClosed;
-                        storeClosedMessage = storeClosedMessage.Replace("%d1", openHours[(int)buildingType].ToString());
-                        storeClosedMessage = storeClosedMessage.Replace("%d2", closeHours[(int)buildingType].ToString());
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Game/PlayerActivate.cs
```diff
@@ -291,10 +291,15 @@ void Update()
                     if (!buildingUnlocked && buildingType < DFLocation.BuildingTypes.Temple
                         && buildingType != DFLocation.BuildingTypes.HouseForSale)
                     {
-                        string storeClosedMessage = HardStrings.storeClosed;
-                        storeClosedMessage = storeClosedMessage.Replace("%d1", openHours[(int)buildingType].ToString());
-                        storeClosedMessage = storeClosedMessage.Replace("%d2", closeHours[(int)buildingType].ToString());
-                        DaggerfallUI.Instance.PopupMessage(storeClosedMessage);
+                        string buildingClosedMessage;
```
ajrb: For brevity you could write this like:

`
string buildingClosedMessage = (buildingType == DFLocation.BuildingTypes.GuildHall) ? HardStrings.guildClosed : HardStrings.storeClosed;
`

df-ferital: Of course I'll change it. I was never fond of the ternary operator syntax, not very readable in my opinion, that's why I tend to never use it :)

ajrb: I find it a lot more readable when instantiating a var to one of two different values, but it's my personal preference so that's why it was a suggestion only. :)

ajrb:
I like this, a nice little refinement.

Issue #1374, opened 2019-07-31
Lycanthropy health degradation when not sated doesn't match classic


Interkarma:
If a werebeast doesn't feed, maximum health should gradually reduce until it bottoms out at 4. In the current implementation, health simply goes to 4 once the requisite time has passed. This needs refinement to better match classic.

JorisVanEijden:
Possible fix in #1491

PR #1491, opened 2019-09-05
Fixes for #1374 and lack of max health restore on cure and feed.

JorisVanEijden:
I set it to reduce max health by 24 points per day, please adjust to taste ;)

I also added removal of the max health limit when feeding or being cured.


Interkarma:
Cheers! That will do for now, although I suspect classic spreads the max health degrade several percentage points per day rather than a fixed amount. A level 2 or 3 werewolf would still bottom out in a couple of days rather than be spread over the next few weeks.

JorisVanEijden:
> I suspect classic spreads the max health degrade several percentage points per day rather than a fixed amount.
That's what I built initially. But then I thought "Why bother putting points and effort into having high health if you're affected by stuff like this just a fast as if you didn't?"

PR #1493, opened 2019-09-05
Fix classic bug and clean up readability

JorisVanEijden:
Messages out-of-order bug fixed.
See https://en.uesp.net/wiki/Daggerfall:Elysana%27s_Betrayal#Bugs
and https://forums.dfworkshop.net/viewtopic.php?f=29&t=2192

Also fixed typo from classic.

Renamed some variables along the way to make working with the quest easier.


Interkarma:
Thanks so much. That one has always annoyed me. Good to have it solved in DFU.

PR #1494, opened 2019-09-05
Adjust direct stat, skill and reaction mod mechanism

ajrb:
Allows the direct stat mods to be able to be used outside the effect bundle system. Without this fix any application of direct stat mod value will accumulate until the player has at least 1 effect bundle applied.

I looked at the prototype vampire class VampiricFortifyEffect and copied it's way of adjusting stats. Unfortunately it didn't work without this fix since the direct stat array is not cleared unless the entity has a bundle associated.

See https://github.com/ajrb/dfunity-mods/blob/encumbrance/RoleplayRealism/Scripts/_startupMod.cs and look at the EncumbranceEffects_OnNewMagicRound() method at the bottom. 

@Interkarma I separated this from the other mod support changes in case this is problematic. I can't see any issues. I did consider creating an effect bundle but since the effect was calculated and permanent I decided that the dynamic effect system was not right. This is not based on a spell or item, just whether that module of the mod is active.

### DIFF
--- a/Assets/Scripts/Game/MagicAndEffects/EntityEffectManager.cs
+++ b/Assets/Scripts/Game/MagicAndEffects/EntityEffectManager.cs
@@ -1494,15 +1494,15 @@ namespace DaggerfallWorkshop.Game.MagicAndEffects
         /// </summary>
         void DoMagicRound()
         {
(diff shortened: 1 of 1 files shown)


Interkarma:
This is a sensible fix. Thank you!

PR #1495, opened 2019-09-06
Fix bug where no action mode selected first time inventory accessed

ajrb:
Only occurred the first time inventory used if not in a dungeon.

I also simplified the code for wagon access because I simply could not follow it with my tired old brain. I literally spent over an hour getting confused so I had to change the logic for my own sanity. Sorry Pango, my brain gets confused easily sometimes. I think it behaves the same unless I missed something in testing. Could you check you're happy with it please?

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
@@ -315,7 +315,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
 
             // Setup initial state
             SelectTabPage(TabPages.WeaponsAndArmor);
-            SetupDefaultActionMode();
+            SelectActionMode((lootTarget != null) ? ActionModes.Remove : ActionModes.Equip);
 
             // Setup initial display
             FilterLocalItems();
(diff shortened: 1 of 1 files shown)


Interkarma:
I noticed that bug as well outside of dungeons, thank you for fix. This changes resolves that and proximity wagon access still works correctly as far as I can see. Also agree the the code is easier to follow.

I was just about to run those builds, I'll quickly merge this one first though.

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
```diff
@@ -1065,6 +1036,22 @@ void ShowWagon(bool show)
...
+            {
+                ShowWagon(true);
+                SelectActionMode(ActionModes.Remove);
+            }
+            else if (GameManager.Instance.PlayerEnterExit.IsPlayerInsideDungeon &&
+                     PlayerEntity.Items.Contains(ItemGroups.Transportation, (int)Transportation.Small_cart) &&
+                     DungeonWagonAccessProximityCheck())
+            {
+                allowDungeonWagonAccess = true;
+                ShowWagon(true);
```
petchema: It seems it should be
                 allowDungeonWagonAccess = true;
                if (lootTarget == null)
                    ShowWagon(true);
to allow wagon access by proximity but not autoselect it when you loot a corpose while near the entrance
Forums:
  https://forums.dfworkshop.net/viewtopic.php?f=5&t=2728

ajrb: Yeah, I already noticed and fixed it in #1502. Unfortunately it missed the build. Sorry about that.

petchema: Ah right, it's still not merged
Okay!

PR #1496, opened 2019-09-07
Enemies shooting point-blank

JorisVanEijden:
Restored point-blank range in which enemies don't use ranged attacks but made it 6m rather than the original 9m.

Also changed the random checks which were using classic random for non-deterministic purposes.


Interkarma:
Thank you @JorisVanEijden. That's much easier to read.

PR #1497, opened 2019-09-07
Expand houses offered for sale

ajrb:
Doesn't match classic exactly.
Fixed house ownership to work with new discovery system, at present if players buy a house it's not shown on the map.

### DIFF
--- a/Assets/Scripts/Game/BuildingDirectory.cs
+++ b/Assets/Scripts/Game/BuildingDirectory.cs
@@ -9,8 +9,6 @@
 // Notes:
 //
 
-using System;
-using System.Collections;
 using System.Collections.Generic;
 using UnityEngine;
 using DaggerfallConnect;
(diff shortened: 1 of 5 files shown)


Assets/Scripts/Game/BuildingDirectory.cs
```diff
@@ -157,11 +155,31 @@ public List<BuildingSummary> GetBuildingsOfFaction(int factionId)
...
+            int maxForSale = Mathf.Min(buildingDict.Count / 10, 20);
             List<BuildingSummary> forSale = new List<BuildingSummary>();
+            List<BuildingSummary> candidates = new List<BuildingSummary>();
             foreach (BuildingSummary building in buildingDict.Values)
+            {
                 if (building.BuildingType == DFLocation.BuildingTypes.HouseForSale)
+                {
                     forSale.Add(building);
+                }
+                else if (building.BuildingType > DFLocation.BuildingTypes.House1 &&
```
df-ferital: Should be >= to include House1 type

Assets/Scripts/Game/BuildingDirectory.cs
```diff
@@ -157,11 +155,31 @@ public List<BuildingSummary> GetBuildingsOfFaction(int factionId)
...
+                         building.BuildingType < DFLocation.BuildingTypes.House4 &&
```
df-ferital: Should be <= to include House4 type

ajrb: Good catch. Copied from an exclusion conditional and brain farted that switch around.

Assets/Scripts/Game/BuildingDirectory.cs
ajrb: Good catch. Copied from an exclusion conditional and brain farted that switch around.

PR #1498, opened 2019-09-07
Fix interiors to use correct block data

df-ferital:
Interior were using raw block data, not block data processed through RMBLayout.GetLocationBuildingData.

This fixes this issue I discovered after entering the Order of the Raven building in Whitewold after the patch I made for it in PR #1489.


ajrb:
Do you use the Archaeologists mod? If so can you just check the world data overrides still work okay please? Looking at the code I think they'll be just fine but would like to check. Just find a guild and check it's not become empty or anything.

df-ferital:
I just tested the Archaeologists mod and it works with my changes.
But while doing so I noticed a small issue in your mod, the Archaeologist trainer seems to have the Thieves Guild faction ID as she greeted my character as a member of The Thieves Guild. All other Archaeologists faction IDs seem correct though.

ajrb:
Thanks for checking. Yes I re-used the faction id, didn't think of the consequences for talking. Will address this in next build.

Interkarma:
Thank you!

PR #1501, opened 2019-09-08
Fix old ruler name being identical to new one

df-ferital:
In rumors involving a new ruler, when the old one was mentioned, he used the same name. Fixed this by matching classic: random old name uses the HIWORD part of the ruler name seed while new name uses the LOWORD.

FACTION.TXT was tweaked so the following factions always have the same ruler (as intended) and are therefore excluded from "new ruler chance":
Anticlere: Lady Doryanna Flyte
Isle of Balfiera: Medora
Lainlyn: Baron Shrike
Orsinium: Gortwog


Interkarma:
Really appreciate all your work on this stuff. :)

PR #1504, opened 2019-09-09
Fix Malacath quest name

df-ferital:
Malacath quest name was wrong as mentioned on the forums: https://forums.dfworkshop.net/viewtopic.php?f=25&t=2684


ajrb:
Thanks, not sure how I got that wrong. Just be aware that this will not stop the quest from running if someone has adult content switched off.

df-ferital:
@arj Thanks for the info.
@Interkarma Should we also filter this particular when ChildGuard is on?

Interkarma:
@df-ferital There's no global ChildGuard enable as yet. Just player paper doll nudity on/off. I've let mods do most of the heavy lifting here in terms of replacing the pixel nudity of sprites, etc. It's something we should look at adding in a more global way in future. For now, I'm happy for the quest just to be allowed, unless there's any objections.

df-ferital:
Thanks. I asked because of Hazelnut message and I saw this in `QuestListManager`:
```C#
if (!quest.adult || DaggerfallUnity.Settings.PlayerNudity)
   pool.Add(quest);
```
 and thought it was suppose to filter quests.

ajrb:
It does, yes. Interkarma may have missed that. :-)

However it doesn't work for the daedric quests, and what could we do if a player summons and then the quest isn't started. Only solution would be to develop a safe quest version or use one of the other daedric quests instead but I really don't feel it's important.

Interkarma:
I haven't missed it - PlayerNudity flag was intended for the paper doll only. It has just been co-opted for other purposes for now. I'd rather have a more comprehensive setup if it's something we're going to do properly later.

ajrb:
Yeah, I also feel quite ambivalent about it as well. I used the nudity flag because it was the only thing available to filter the adult quests on. Possibly it should be better communicated in the settings but beyond that I don't think it's worth any effort. DF is so tame over 2 decades later.

df-ferital:
Ok, thank you both, I'll leave this very small PR as it is then :)

PR #1506, opened 2019-09-10
Prevent game load from restoring players breath underwater

JorisVanEijden:
Forum topic: https://forums.dfworkshop.net/viewtopic.php?f=24&t=2681

This is one approach to fixing the reported issue.
It seems to work fine on the saves I have, but this definitely needs thorough testing for side-effects.


Interkarma:
This is clean and works well. Thank you. :)

Issue #1505, opened 2019-09-09
Travel map reticle not updated when scrolling region around with arrow buttons


petchema:
**Describe the bug**
The travel region map reticle keeps blinking at the same position when the region is scrolled around using arrow buttons, pointing at an incorrect position on the map.
I noticed the bug because while in the Wrothgarian Mountains I tried to locate a town that was not in the current screen, and using arrow buttons didn't help me see where it was.
Remark: smooth scrolling (shift key + mouse movement) works fine, the reticle follows map movements.

**To Reproduce**
1. Open a large region in the travel map (one with scrolling arrow buttons)
2. Make the reticle blink either by clicking on "I'M AT" button or by using the "FIND" function
3. Quickly click on a scrolling arrow button
4. Notice that the map scrolled, but the reticle is at the same place on screen

**Expected behavior**
Reticle should be scrolled by the same amount as the underlying region map.

**Desktop**
Tested with
 - OS: Linux
 - Version 0.10.7

petchema:
Fixed by #1509

PR #1509, opened 2019-09-12
Update crosshair on travelmap

JorisVanEijden:
Fix for #1505 

Crosshair would stay at same position when moving the map below it with the on-screen arrows.


petchema:
works for me :+1:

PR #1517, opened 2019-09-17
Adds possibility of quest NPC infighting

JayH2971:
Requires a custom quest mod, which is not implemented in DFU core, to take advantage

### DIFF
--- a/Assets/Scripts/Game/EnemySenses.cs
+++ b/Assets/Scripts/Game/EnemySenses.cs
@@ -728,8 +728,8 @@ namespace DaggerfallWorkshop.Game
                     if (targetBehaviour.EntityType == EntityTypes.EnemyMonster || targetBehaviour.EntityType == EntityTypes.EnemyClass)
                         targetSenses = targetBehaviour.GetComponent<EnemySenses>();
 
-                    // For now, quest AI can't be targeted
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Game/EnemySenses.cs
```diff
@@ -728,8 +728,8 @@ void GetTargets()
                     if (targetBehaviour.EntityType == EntityTypes.EnemyMonster || targetBehaviour.EntityType == EntityTypes.EnemyClass)
                         targetSenses = targetBehaviour.GetComponent<EnemySenses>();
 
-                    // For now, quest AI can't be targeted
-                    if (targetSenses && targetSenses.QuestBehaviour)
+                    // Allows infighting via custom quest action
+                    if (targetSenses && targetSenses.QuestBehaviour && !targetSenses.QuestBehaviour.IsAttackableByAI)
```
Interkarma: IsAttackableByAI isn't a member of QuestResourceBehaviour. The target field will also need to be added before it can be referenced.

I'm happy to add this down the track when I merge this one though. :)

JayH2971: Sorry, I'm just the intermediary... Thanks for understanding Interkarma.

PR #1518, opened 2019-09-18
Grammar and typo fixes

Arcane21:
Fixed many improperly spelled items, capitalized many proper nouns, some consistency fixes

### DIFF
--- a/Assets/Game/Addons/ModSupport/ModSettings/PresetPicker.cs
+++ b/Assets/Game/Addons/ModSupport/ModSettings/PresetPicker.cs
@@ -53,12 +53,8 @@ namespace DaggerfallWorkshop.Game.Utility.ModSupport.ModSettings
                 mainPanel.Outline.Enabled = true;
                 NativePanel.Components.Add(mainPanel);
 
-                var titleLabel = new TextLabel();
-                titleLabel.Size = new Vector2(60, 12);
-                titleLabel.Position = new Vector2(0, 5);
-                titleLabel.HorizontalAlignment = HorizontalAlignment.Center;
-                titleLabel.Text = ModManager.GetText("presetCreator");
-                mainPanel.Components.Add(titleLabel);
+                mainPanel.Components.Add(MakeCancelButton(uiManager, "presets"));
+                mainPanel.Components.Add(MakeTitleLabel("presetCreator"));
 
                 var creatorPanel = new Panel();
                 creatorPanel.Outline.Enabled = false;
@@ -71,15 +67,7 @@ namespace DaggerfallWorkshop.Game.Utility.ModSupport.ModSettings
                 creatorPanel.Components.Add(MakeLabelledTextBox(descriptionTextBox, ModManager.GetText("description"), VerticalAlignment.Middle));
                 creatorPanel.Components.Add(MakeLabelledTextBox(authorTextBox, ModManager.GetText("author"), VerticalAlignment.Bottom));
 
-                var confirmButton = new Button();
-                confirmButton.Size = new Vector2(40, 12);
-                confirmButton.Position = new Vector2(0, 110);
-                confirmButton.HorizontalAlignment = HorizontalAlignment.Center;
-                confirmButton.Outline.Enabled = true;
-                confirmButton.BackgroundColor = new Color(0.0f, 0.5f, 0.0f, 0.4f);
-                confirmButton.Label.Text = ModManager.GetText("ok");
-                confirmButton.OnMouseClick += ConfirmButton_OnMouseClick;
-                mainPanel.Components.Add(confirmButton);
+                mainPanel.Components.Add(MakeSimpleButton("ok", new Vector2(0, 110), ConfirmButton_OnMouseClick));
             }
 
             private Panel MakeLabelledTextBox(TextBox textBox, string text, VerticalAlignment verticalAlignment)
@@ -112,8 +100,19 @@ namespace DaggerfallWorkshop.Game.Utility.ModSupport.ModSettings
 
             private void ConfirmButton_OnMouseClick(BaseScreenComponent sender, Vector2 position)
             {
-                createPresetCallback(titleTextBox.ResultText, descriptionTextBox.ResultText, authorTextBox.ResultText);
-                uiManager.PopWindow();
+                if (string.IsNullOrEmpty(titleTextBox.ResultText) || string.IsNullOrEmpty(descriptionTextBox.ResultText))
+                {
+                    var messageBox = new DaggerfallMessageBox(uiManager, this);
+                    messageBox.AllowCancel = true;
+                    messageBox.ClickAnywhereToClose = true;
+                    messageBox.SetText(ModManager.GetText("titleOrDescriptionMissing"));
+                    uiManager.PushWindow(messageBox);
+                }
+                else
+                {
+                    createPresetCallback(titleTextBox.ResultText, descriptionTextBox.ResultText, authorTextBox.ResultText);
+                    uiManager.PopWindow();
+                }
             }
         }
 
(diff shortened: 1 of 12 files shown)


deepfighter:
The "Ark'ay" fix in the books.txt shouldn't be changed, as it's supposedly the intended title "Ark'ay". That the books contradicts this is in typing in the last line "Arkay" is imho no typo, as nobody knows how it was to be intended (his mortal name Ark'ay and his god name Arkay, maybe?). This book got some revisions during the development process and this could be just be forgotten to be changed and therefore - lore-wise - both names are correct and thus it should be kept like in classic.

Arcane21:
> The "Ark'ay" fix in the books.txt shouldn't be changed, as it's supposedly the intended title "Ark'ay". That the books contradicts this is in typing in the last line "Arkay" is imho no typo, as nobody knows how it was to be intended (his mortal name Ark'ay and his god name Arkay, maybe?). This book got some revisions during the development process and this could be just be forgotten to be changed and therefore - lore-wise - both names are correct and thus it should be kept like in classic.

Good point. Vaermina and Vaernima also seem to both be correct according to lore, so I'll revert that.

bansheexyz:
Some of your changes are correct, but some are not. I'd study the language more before attempting to make sweeping changes like this:

-The articles before unique items are not really an error, totally preferential change here.
-"Final" should not be "Finale", the error here is the missing word "battle" after final. Finale is used for performances and it's clearly not a metaphor here. See line 84/85 for this "battle" to be referenced again.
-Emperor should be capitalized there, it is a title and titles only get capitalized before names. You can google this rule.
-The changes to "Mortalcreature" aren't necessary, it's basically a nickname being given to you. There's no way these two words would've been accidentally combined and capitalized in three separate places.

Arcane21:
> Some of your changes are correct, but some are not. I'd study the language more before attempting to make sweeping changes like this:
> 
> -The articles before unique items are not really an error, totally preferential change here.
> -"Final" should not be "Finale", the error here is the missing word "battle" after final. Finale is used for performances and it's clearly not a metaphor here. See line 84/85 for this "battle" to be referenced again.
> -Emperor should be capitalized there, it is a title and titles only get capitalized before names. You can google this rule.
> -The changes to "Mortalcreature" aren't necessary, it's basically a nickname being given to you. There's no way these two words would've been accidentally combined and capitalized in three separate places.

As for the first one, good point, I'll tweak this to be correct.

The second one is also a good point, I capitalized it in a fww other places it was lowercase, can;t believe I missed that.

As to the third, I'm aware it was likely intentional, but it still doesn't make much sense. I get the dialouge is supposed to reflect a broken form of English, but it still comes off as grammatically bizzare, even for writing with a broken speech pattern. I left the rest alone because it made perfect sense to be rendered as broken English, but I cannot recall any instance of spoken broken English rendered in script format with words joined together like that, especially two words that form natural conclusions to their syllables.

I'll revert it, but it just looks wrong, even for writing that is supposed to resemble broken English.

bansheexyz:
I forgot a word myself and edited my comment. I meant to say "emperor" should *not* be capitalized there. For example: "Bow to your emperor" is correct and "Bow to Emperor Uriel" is correct. It's only capitalized when it's before a name.

https://grammarist.com/style/president/

Arcane21:
> I forgot a word myself and edited my comment. I meant to say "emperor" should _not_ be capitalized there. For example: "Bow to your emperor" is correct and "Bow to Emperor Uriel" is correct. It's only capitalized when it's before a name.
> 
> https://grammarist.com/style/president/

I see. I'll revert that to be correct as well.

tolbell:
> As for the Articles before certain artifacts, not only was it not consistent with later games,

That doesn't make it a typo.

> it wasn't even consistent within Daggerfall's nomenclature for artifact names.

It could also be said that the names without "The" are inconsistent with those that have it. If some are prefixed with "The" and others are not, it seems arbitrary to toss out all the ones that have "The."

> Further, the names would indicate false or lesser copies exist due to the "The" article, and since they are one of a kind, they made no sense at all.

I disagree that this is implied by the article. My understanding is that "The" implies that there is only one of a given object.

There is still one instance of "Mortal creature" left to be reverted.

Arcane21:
> > As for the Articles before certain artifacts, not only was it not consistent with later games,
> 
> That doesn't make it a typo.
> 
> > it wasn't even consistent within Daggerfall's nomenclature for artifact names.
> 
> It could also be said that the names without "The" are inconsistent with those that have it. If some are prefixed with "The" and others are not, it seems arbitrary to toss out all the ones that have "The."
> 
> > Further, the names would indicate false or lesser copies exist due to the "The" article, and since they are one of a kind, they made no sense at all.
> 
> I disagree that this is implied by the article. My understanding is that "The" implies that there is only one of a given object.
> 
> There is still one instance of "Mortal creature" left to be reverted.

I've fixed the remain instance mentioned for Dagon's quest.

As to the Artifacts, I looked up the names for all the games on the UESP (a source even the official developer have admitted to consulting to keep the lore straight), and aside from Daggerfall, all the other games have remained consistent with the name changes I made to the artifacts. I understand it may irritate some who are obsessed with the original text remaining inviolate, but for reasons of consistency, since all the artifacts and the Daedric Princes they are associated with are supposed to the same artifact for all games in the series, just used by different characters throughout the series timeline, then I must fall on the side of making sure the text remains consistent with the overall series canon.

Also, granted, Molag's Mace appears in Skyrim as the "Rusty Mace" before Molag Bal restores it to full power and gives it to you as a reward, but it's still the same mace, just drained of it's power before the quest end powers it up and hands you the full powered version, the "Mace of Molag Bal"

Ergo, keeping the "The" is just adhering to a name change that does not mesh with the rest of the series canon

tolbell:
In that case, I suggest you change the title. In my mind, this is more controversial than simply fixing typos and grammar mistakes.

Arcane21:
> In that case, I suggest you change the title. In my mind, this is more controversial than simply fixing typos and grammar mistakes.

Noted. I'd like to have input from the project lead on this before I do anything one way or another.

Interkarma:
I'd prefer the artifact names remain as classic. I actually changed many of these early in the project as you have here, and received some push-back. I now see this as a uniquely Daggerfall quirk that has been present for decades. And the classic names are also referenced elsewhere in relation to Daggerfall. For example, the UESP artifacts page for Daggerfall below. While it's still easy enough to tell what is what, this is something I'd prefer to keep like classic.

https://en.uesp.net/wiki/Daggerfall:Artifacts

I think the other changes are fine though. I'll check through everything else before merging and let you know if any other comments. Thank you. :)

Arcane21:
> I'd prefer the artifact names remain as classic. I actually changed many of these early in the project as you have here, and received some push-back. I now see this as a uniquely Daggerfall quirk that has been present for decades. And the classic names are also referenced elsewhere in relation to Daggerfall. For example, the UESP artifacts page for Daggerfall below. While it's still easy enough to tell what is what, this is something I'd prefer to keep like classic.
> 
> https://en.uesp.net/wiki/Daggerfall:Artifacts
> 
> I think the other changes are fine though. I'll check through everything else before merging and let you know if any other comments. Thank you. :)

Understood. I will comply ASAP.

Interkarma:
Looks good, thank you. :)

tolbell:
I know I'm late, but I have one more objection to this. I understand why you wanted to change the name of "The Hircine Ring" (although the argument can be made that the word "hircine," while being the name of a daedric prince in TES, is also a regular English adjective). Regardless, "The Hircine's Ring" certainly has... an odd ring to it. As a native speaker, I don't think one would usually put "The" before a possessive noun - "Kevin's house" sounds fine, but not "The Kevin's house."

If it must be changed, I suggest "The Ring of Hircine" (which is consistent with other Daggerfall artifacts such as "The Ring of Namira") or simply "Hircine's Ring" (as in Morrowind).

Interkarma:
That's fair, and I agree. Have reverted that name change now.

PR #1520, opened 2019-09-19
Allow access to hide billboards for mods.

concarne000:
Here is a solution I've found to give public access to force hide a billboard, specifically with the intention of implementing 3D models, without impacting the engine or other mods, while also being able to access the engines concealment for magic effects.

### DIFF
--- a/Assets/Scripts/Game/Entities/EntityConcealmentBehaviour.cs
+++ b/Assets/Scripts/Game/Entities/EntityConcealmentBehaviour.cs
@@ -26,6 +26,17 @@ namespace DaggerfallWorkshop.Game.Entity
         protected EntityEffectManager entityEffectManager;
         protected MeshRenderer meshRenderer;
 
+
+        private bool isConcealed = false;
+
+        public bool IsConcealed
+        {
+            get { return (isConcealed); }
+        }
+
+        // Override any concealment to force billboard to always be hidden
+        public bool forceConcealment { get; set; }
+        
         #region Unity
 
         void Awake()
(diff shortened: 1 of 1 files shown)


ajrb:
I think this is the wrong approach to take for this. @TheLacus if you disagree and want me to explain why and how I think this should be done, please say.

TheLacus:
Hi! I appreciate the effort but a mod that replace mobile billboards with mobile models should replace the actual asset, not just hide it. Other mods are either unaffected by this change or strictly require a billboard, so the game shouldn't tell them that one is used while is actually not true.

Mobile Npcs are instantiated from `DaggerfallMobilePerson [Standalone]` prefab, which holds everything related to npc behaviour and management of entity itself, while its child `MobilePersonBillboard` only has `MobilePersonBillboard` instance and the `MeshRenderer`.

What we need is an interface, let's call it `IMobilePersonBillboard` that defines non-private properties and methods of `MobilePersonBillboard` and is implemented by this class as well as any custom component used by mods for models.

```cs
public interface IMobilePersonBillboard
{
    bool IsIdle { get; set; }
    bool IsUsingGuardTexture { get; }
    void SetPerson(Races race, Genders gender, int personVariant, bool isGuard);
    Vector2 GetSize(); // Renamed from GetBillboardSize()
}
```

TheLacus:
@Allofich Can we add public property `IsGuard` to `MobilePersonNPC` and remove `IsUsingGuardTexture` from here?

Allofich:
@TheLacus, I think that should be fine. I think I did it that way because the original game was identifying guard MobilePersonNPCs by them using the index for the guard graphic, and not with a separate isGuard-type property.

Interkarma:
@TheLacus what are your thoughts on this one? Should it be closed, as you had another direction in mind?

TheLacus:
> 
> 
> @TheLacus what are your thoughts on this one? Should it be closed, as you had another direction in mind?

Yes, i was waiting for review of #1598 but i think i can do it now.

PR #1522, opened 2019-09-21
player smoother movements

petchema:
- smooth follower could be reset very often at high FPS because speed threshold was computed based on framerate instead of fixed updates

- unstick handling could trigger during movements tangential to the (world) origin

Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&t=2435&p=31737#p31728


Interkarma:
This works well, a nice fix and improvement. Thank you! :)

PR #1524, opened 2019-09-22
Change nature 3d asset replacement to limit performance hit

ajrb:
Change 3d asset replacement for nature flats to only replace when terrain distance is <=1 for performance. At larger terrain distances, the standard DFU billboard system is still used since it's highly efficient and billboards are more than sufficient at this distance.

The nature flats used with standard DFU billboarding can still be replaced with high-res/different images.

- Update natures on terrains moving in/out of dist 1.
- Only do extra updates once any nature mesh replacement has been found.
- Ensure nature mesh replacement doesn't change nature placement. (bugfix)

### DIFF
--- a/Assets/Scripts/Terrain/StreamingWorld.cs
+++ b/Assets/Scripts/Terrain/StreamingWorld.cs
@@ -10,7 +10,7 @@
 //
 
 #define SHOW_LAYOUT_TIMES
-#define SHOW_LAYOUT_TIMES_NATURE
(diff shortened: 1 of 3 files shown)


ajrb:
This has been tested by VMBlast while I was away for past 2 weeks.

Interkarma:
Looks good on a read, and I'll trust the testing is OK from the people involved. Merging now. :)

PR #1525, opened 2019-09-23
make sure videos play under best conditions

petchema:
- stop digitalized music (only MIDI synthesis was muted);
- stop ambient loop sounds
- videos played by quests can only be interrupted by Escape key. This should prevent mouse clicks from accidentally closing the video.
- implement HideCursor. This requires some cooperation with PlayerMouseLook that constantly updates mouse cursor visibility. I'm not sure how to best interact with it though


petchema:
Gamesave near the Mantella for testing
[SAVE353.zip](https://github.com/Interkarma/daggerfall-unity/files/3640546/SAVE353.zip)

PR #1528, opened 2019-09-25
fix for incorrect residence name resolving for quest person resources

Nystul-the-Magician:
(fixes issue https://forums.dfworkshop.net/viewtopic.php?f=24&t=2730)


df-ferital:
This seems to work with quest K0C00Y06, at least it corrects the culprit location mismatch issue identified here: https://forums.dfworkshop.net/viewtopic.php?f=5&t=2747#p31832 (second point). Great work!

Interkarma:
Awesome work, thank you Nystul! :)

PR #1530, opened 2019-09-26
Guard against array index out of bounds.

JorisVanEijden:
And that's all this does.
It does not increase the label limits, it just doesn't show any labels above the limit.

Fix for https://forums.dfworkshop.net/viewtopic.php?f=25&t=2780


Interkarma:
Thank you, I should have considered that myself.

PR #1533, opened 2019-09-28
Fix mismatch between questPool and questPicker.ListBox.

JorisVanEijden:
Also prevent loading quests twice.
https://forums.dfworkshop.net/viewtopic.php?f=5&t=2790&sid=46b0dbc40ac30f112d73be0fefb16002

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallGuildServicePopupWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallGuildServicePopupWindow.cs
@@ -78,6 +78,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
         int curingCost = 0;
 
         List<QuestData> questPool;
+        private List<Quest> loadedQuests = new List<Quest>();
 
         #endregion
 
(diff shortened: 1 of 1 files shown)


ajrb:
This approach is better and doesn't require the removeAt's which is better. Two observations though. Firstly, maybe clear the list once one is selected so the parsed quest objects can be GC'ed. Secondly, does the list of quest data (questPool) still need to be there at all?

JorisVanEijden:
Yeah, there's a few issues we're working around here to avoid having to refactor the whole quest loading.
Issue 1 is the fact that the QuestListsManager has no way to get the Displaynames of quests.
Issue 2 is the fact that LoadQuest instantiates some stuff in the TalkManager that needs to be cleaned up for quests that are loaded but will not be started.
The whole distinction between Loading and Starting quests has become a bit vague at points.

For this issue I think the scope of the fix can be limited.
Let's work in another branch on refactoring the loading a bit.

ajrb:
I considered altering the quest parsing to allow a shortcut where the file was only parsed enough to get the display name and to ensure it wasn't malformed in some way. I decided to keep the change less intrusive. I hadn't realised that parsed quests inits stuff in talk manager just by parsing the quest... so that means there might be issues without this listbox presentation as players can decline a quest and ask for another.

JorisVanEijden:
That is taken care of in UserInterfaceWindows.DaggerfallQuestPopupWindow.OfferQuest_OnButtonClick() when you decline the quest.

ajrb:
This looks good to me from reading the code.

Interkarma:
Merging, thank you for the fix.

PR #1541, opened 2019-09-29
Optimize location loading

df-ferital:
Improve location loading by keeping the list of blocks related to a location inside `DFLocation`. I made this after I noticed `RMBLayout.GetLocationBuildingData` was called several times: when entering a town, when entering a building, when talking to a NPC, etc. Unfortunately, calling this method can be relatively heavy in large cities (around 0.150s on my computer). Furthermore, when loading a savegame or when travelling to a new location, the method can be called 4 or 5 times.

So, this PR improves performance while making a location buildings list accessible immediately, reducing code bloat and improving overall clarity. The only drawback comes from `Clock.GetTravelTimeInSeconds` which may become a little slower, since it does not require location buildings list, only location coordinates. However, this isn't really noticeable in game.

### DIFF
--- a/Assets/Scripts/API/DFLocation.cs
+++ b/Assets/Scripts/API/DFLocation.cs
@@ -481,6 +481,9 @@ namespace DaggerfallConnect
 
             /// <summary>Exterior map data including layout information.</summary>
             public ExteriorData ExteriorData;
+
+            /// <summary>List of RMB blocks.</summary>
+            public DFBlock[] Blocks;
         }
 
         /// <summary>
--- a/Assets/Scripts/API/MapsFile.cs
+++ b/Assets/Scripts/API/MapsFile.cs
@@ -14,6 +14,7 @@ using System;
 using System.IO;
 using System.Collections.Generic;
 using DaggerfallConnect.Utility;
+using DaggerfallWorkshop.Utility;
 #endregion
 
 namespace DaggerfallConnect.Arena2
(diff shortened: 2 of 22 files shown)


petchema:
For some reason I still need
`using DaggerfallWorkshop.Utility.AssetInjection;`
in Assets/Scripts/API/MapsFile.cs, replaced in last commit, to compile this PR

df-ferital:
This must be because I opened this quite some time ago, so it might need to be adjusted a little bit when working with last master revision. I will rebased it when I can.

Interkarma:
I'm a bit worried about how this one will play with @ajrb's block replacement. This can switch environments around even at runtime (e.g. from a quest action), and there seems to be some danger of additional caching causing things to get out sync here.

Let me come back to this one after worldDataDev branch has been merged. I'll be looking at all these PRs soon with weekly builds. :)

df-ferital:
No problem, as this is only an optimization, I will rebased it on last master revision when potential conflicting other PR will be merged. I'll also test it again :)

ajrb:
How does this improve location loading performance? I don't see the GetLocationBuildingData being a significant amount of time when I profile location loading with the Unity profiler. Looks like most of the time is taken constructing the base block game objects, adding components and applying materials.

When I log the time taken by the SetLocation() call in StreamingWorld.UpdateLocation() using a stopwatch, it's mostly less than 5ms, with the longest time being 39ms for a large city (Daggerfall) and when compared to the total time to construct Daggerfall location being 2167ms I'm not convinced this is where we need to optimise things.

It may be that it helps in other places and I am taking your title too literally, and it may also be affected by using an SSD and modern CPU (Ryzen)

df-ferital:
I don't have much time to work on this right now and I'm not sure it's still need. Also, I suppose I changed too many things in this PR, beyond its original scope, which was only to prevent several useless calls to `RMBLayout.GetLocationBuildingData`.

PR #1551, opened 2019-10-03
Add IPlayerActivable interface for custom activations

TheLacus:
Added an interface which can be used to subscribe to custom player activations for any object. I'm personally going to use it for my harvesting mod where i need to detect when a crop is picked.

### DIFF
--- a/Assets/Scripts/Game/PlayerActivate.cs
+++ b/Assets/Scripts/Game/PlayerActivate.cs
@@ -4,7 +4,7 @@
 // License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
 // Source Code:     https://github.com/Interkarma/daggerfall-unity
 // Original Author: Gavin Clayton (interkarma@dfworkshop.net)
-// Contributors:    Allofich, Numidium
+// Contributors:    Allofich, Numidium, TheLacus
 // 
 // Notes:
 //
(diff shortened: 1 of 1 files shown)


JayH2971:
Does this include enemy entities?

petchema:
is there some overlap with Hazelnut's ModelActivation, also in PlayerActivate?

TheLacus:
> 
> 
> Does this include enemy entities?

It triggers a target on player action. It's the same thing used for doors, loot, npc talk functionalities etc.

> 
> 
> is there some overlap with Hazelnut's ModelActivation, also in PlayerActivate?

No, it has a different purpose, i think. It allows any object to detect player click, is not linked to model id. For example crops are billboard and not models so i can't use ModelActivation.

TheLacus:
Actually i think it wouldn't be a bad idea to use an interface or abstract class for all core activations. Right now everything has its own check (with GetComponent) and this comes with a performance cost.

ajrb:
> 
> 
> > Does this include enemy entities?
> 
> It triggers a target on player action. It's the same thing used for doors, loot, npc talk functionalities etc.
> 
> > is there some overlap with Hazelnut's ModelActivation, also in PlayerActivate?
> 
> No, it has a different purpose, i think. It allows any object to detect player click, is not linked to model id or any classic model data. For example crops are billboards and not models so i can't use ModelActivation.

How do you add this to your objects? The model activation works on existing DF objects from the game data using the model id, not sure how I would be able to use this interface instead without going though every scene after it's initialised and adding a component implementing this interface to every model with a certain id. Does your mod already do this or does it create new flats?

Very happy to move to a more general solution for custom activations if it's easy to do. My model activation regsitration was very specific to my needs.

TheLacus:
I add a component that implements this interface to my prefabs. Existing plants are replaced by asset-injection framework; i wouldn't be able to make default crops harvestable anyway because they are batched.

I'm not suggesting to remove `ModelActivation`, i only added a more generic way to make an object interactable which is needed for new models added by mods.

ajrb:
Looks good to me from reading the code.

Interkarma:
This is a clever idea! Merging.

PR #1553, opened 2019-10-05
Don't display a message when entering a TG or DB

df-ferital:
Actually don't display the message if player is a member of the guild, as classic does. I wasn't able to test the case where player is not a member as in DFU, it seems Thieves Guild and Dark Brotherhood hideouts are always locked in this case.

Edit. @ajrb confirmed that contrary to what classic does, DFU behavior is to always lock Thieves Guild and Dark Brotherhood hideouts when player is not a guild member, so I simplified the check.

### DIFF
--- a/Assets/Scripts/Game/PlayerActivate.cs
+++ b/Assets/Scripts/Game/PlayerActivate.cs
@@ -386,9 +386,14 @@ namespace DaggerfallWorkshop.Game
 
                         DaggerfallMessageBox mb;
 
+                        PlayerGPS.DiscoveredBuilding buildingData;
+                        GameManager.Instance.PlayerGPS.GetDiscoveredBuilding(building.buildingKey, out buildingData);
+
                         if (buildingUnlocked &&
                             buildingType >= DFLocation.BuildingTypes.House1 &&
                             buildingType <= DFLocation.BuildingTypes.House4 &&
+                            buildingData.factionID != (int)FactionFile.FactionIDs.The_Thieves_Guild &&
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Game/PlayerActivate.cs
```diff
@@ -386,9 +386,16 @@ void Update()
...
                         DaggerfallMessageBox mb;
 
+                        PlayerGPS.DiscoveredBuilding buildingData;
+                        GameManager.Instance.PlayerGPS.GetDiscoveredBuilding(building.buildingKey, out buildingData);
+
                         if (buildingUnlocked &&
                             buildingType >= DFLocation.BuildingTypes.House1 &&
                             buildingType <= DFLocation.BuildingTypes.House4 &&
+                            (buildingData.factionID != (int)FactionFile.FactionIDs.The_Thieves_Guild ||
+                             !GameManager.Instance.GuildManager.HasJoined(FactionFile.GuildGroups.GeneralPopulace)) &&
```
ajrb: Checking whether player has joined TG DB is not needed here, as if they are member, then buildingUnlocked will equal false. Could you simplify you additions to the conditional here please:
`buildingData.factionID != (int)FactionFile.FactionIDs.The_Thieves_Guild &&
buildingData.factionID != (int)FactionFile.FactionIDs.The_Dark_Brotherhood &&`

df-ferital: Thanks for the tip, updated.

PR #1554, opened 2019-10-05
Expose horse animations frame index for mods

ajrb:
Needed for next version of enhanced horse riding to allow hi-res graphics

### DIFF
--- a/Assets/Scripts/Game/TransportManager.cs
+++ b/Assets/Scripts/Game/TransportManager.cs
@@ -60,7 +60,7 @@ namespace DaggerfallWorkshop.Game
         }
 
         public ImageData RidingTexture { get { return ridingTexture; } }
-        public int FrameIdx { get { return frameIdx; } }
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Game/TransportManager.cs
```diff
@@ -60,6 +60,7 @@ public PlayerPositionData_v1 BoardShipPosition
         }
 
         public ImageData RidingTexture { get { return ridingTexture; } }
+        public int FrameIdx { get { return frameIdx; } }
```
TheLacus: I'm not a fan of how `index` was abbreviated to `idx` (only two chars saved and is somewhat less readable) but this is just personal preference.

ajrb: Oh, okay. I've abbreviated Index to Idx so much I find it just as readable. Happy to change it here as readability is important.

PR #1556, opened 2019-10-06
Add gender filtering for social quests

### LINKED ISSUE #1542: Enable gender filtering for quests
Some classic quests are supposed to be offered only by male questors. In fact, all quests in which the quest giver is set as `group Questor male`. Verified in classic to be sure of what I'm saying here.

The easiest way to correct this is to add a new `gender` field to `QuestData` structure with the corresponding column in QuestList-Classic.txt and QuestList-DFU.txt.

ajrb:
Applied M (male) to all quests with "group Questor male" in quest list

Should fix #1542 to match classic. Needs some checking by ferital.


Interkarma:
A good change, thank you!

PR #1557, opened 2019-10-06
Interior replacement core changes

UncannyValley21:
A clean commit with all changes to the existing core system

-3D props can now be rotated in X and Z. Default is set to 0 so it doesn't affect vanilla Assets
-Ladders will only be set to interactive if they are set uo as objectType 3 (propModel), all vanilla ladders are set up this way
-When replacing buildings, always use the original NameSeed, LocationID and Sector values


ajrb:
I've tested this with the building replacement world data override functionality and it works fine. It's also perfectly backward and forward compatible. Quite amusing to set up a shop shelf rotated wierdly.. :-)

Interkarma:
Looks good, merging! :)

Issue #1543, opened 2019-10-01
Console command to add all inventory items


TheLacus:
Mod authors have requested the ability to add all weapons and paperdoll items to inventory for testing. This feature request is open to everyone, including new contributors.

Console commands are defined at https://github.com/Interkarma/daggerfall-unity/blob/master/Assets/Game/Addons/UnityConsole/Console/Scripts/DefaultCommands.cs. See `AddWeapon` for an example of how to add items to inventory.

If nobody is interested I'll eventually do it myself.

ajrb:
This is done and released in 0.10.8

PR #1558, opened 2019-10-06
Add all equippable items command

ajrb:
Enable testing of paperdoll images as requested by #1543.

Adds all types of equippable items for the current character gender/race, but doesn't add variants.

Also extended dumpblock command to be able to dump dungeon blocks, and added a dumplocation command to dump location records.


ajrb:
I'd like to merge this today, unless anyone has any reservations.

PR #1559, opened 2019-10-07
Fix some circular dependencies so TG and DB are proplerly GC'ed

df-ferital:
Previously, Thieves Guild or Dark Brotherhood instances were never garbage collected because `PlayerGPS` and `StreamingWorld` kept strong references to them through events `OnEnterLocationRect` and `OnAvailableLocationGameObject`. This, because unsubscribing to these events took place in both class destructors, so, to summarize, it never happened.

This resulted in the following behavior: if you loaded a savegame with a character who was a member of any of these guilds, then loaded another savegame from another character who wasn't, you would still discover the guild hideouts using the automap.

Unfortunately, any game saved prior to this fix may still have the guilds available on the automap, this because all discovered buildings are saved.

### DIFF
--- a/Assets/Scripts/Game/Guilds/DarkBrotherhood.cs
+++ b/Assets/Scripts/Game/Guilds/DarkBrotherhood.cs
@@ -12,6 +12,7 @@ using DaggerfallConnect.Arena2;
 using DaggerfallWorkshop.Game.Entity;
 using System;
 using DaggerfallWorkshop.Utility;
+using DaggerfallWorkshop.Game.Serialization;
 
 namespace DaggerfallWorkshop.Game.Guilds
 {
@@ -77,20 +78,6 @@ namespace DaggerfallWorkshop.Game.Guilds
         #endregion
 
         #region Guild Membership and Faction
-
-        public DarkBrotherhood()
-        {
-            // Register for location entry events so can auto discover guild houses.
-            PlayerGPS.OnEnterLocationRect += PlayerGPS_OnEnterLocationRect;
-            StreamingWorld.OnAvailableLocationGameObject += StreamingWorld_OnAvailableLocationGameObject;
-        }
-
-        ~DarkBrotherhood()
-        {
-            // Unregister events
-            PlayerGPS.OnEnterLocationRect -= PlayerGPS_OnEnterLocationRect;
-            StreamingWorld.OnAvailableLocationGameObject -= StreamingWorld_OnAvailableLocationGameObject;
-        }
 
         public static int FactionId { get { return factionId; } }
 
@@ -188,7 +175,13 @@ namespace DaggerfallWorkshop.Game.Guilds
         override public void Join()
         {
             base.Join();
+            RegisterEvents();
             RevealGuildHallOnMap();
+        }
+
+        override public void Leave()
+        {
+            UnregisterEvents();
         }
 
         public override TextFile.Token[] TokensIneligible(PlayerEntity playerEntity)
@@ -207,6 +200,20 @@ namespace DaggerfallWorkshop.Game.Guilds
         #endregion
 
         #region Event handlers
+
+        private void RegisterEvents()
+        {
+            // Register for location entry events so can auto discover guild houses
+            PlayerGPS.OnEnterLocationRect += PlayerGPS_OnEnterLocationRect;
+            StreamingWorld.OnAvailableLocationGameObject += StreamingWorld_OnAvailableLocationGameObject;
+        }
+
+        private void UnregisterEvents()
+        {
+            // Unregister events
+            PlayerGPS.OnEnterLocationRect -= PlayerGPS_OnEnterLocationRect;
+            StreamingWorld.OnAvailableLocationGameObject -= StreamingWorld_OnAvailableLocationGameObject;
+        }
 
         private void PlayerGPS_OnEnterLocationRect(DFLocation location)
         {
@@ -228,6 +235,16 @@ namespace DaggerfallWorkshop.Game.Guilds
             if (buildingDirectory)
                 foreach (BuildingSummary building in buildingDirectory.GetBuildingsOfFaction(factionId))
                     GameManager.Instance.PlayerGPS.DiscoverBuilding(building.buildingKey, GetGuildName());
+        }
+
+        #endregion
+
+        #region Serialization
+
+        internal override void RestoreGuildData(GuildMembership_v1 data)
+        {
+            base.RestoreGuildData(data);
+            RegisterEvents();
         }
 
         #endregion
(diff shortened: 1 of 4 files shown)


ajrb:
Great catch on the bug there, and you correctly tracked down what was causing the issue which is great. However, I don't like your approach to fixing the issue. I hope you don't mind but I've just fixed this the way I think it should be and submitted PR #1560 with my fix.

df-ferital:
I'm not sure your approach is better, as being expelled from the guild won't make the events being unregistered so the guild would not be deleted. I'm pretty sure we need a Leave method there.

ajrb:
You cannot get expelled from TG or DB since there's no way to rejoin, unlike the other guilds. Seemed like it would be a bit harsh to a new joiner who messes up their first quest or two.

See the override of CalculateNewRank() in both those guild classes.

df-ferital:
If your character becomes a vampire, he's expelled from all guilds, including TG and DB. With your method, events are not unregistered so Thieves Guild and Dark Brotherhood instances remain active, leading to the same issue as when loading different saves.

ajrb:
Ah I think I see now, we have different perspectives on what the issue is. I see the issue as simply being one character play session affecting a different character session. i.e. a player not in TG see's the TG guild halls because the previous session was a member. My view on seeing the guild halls if expelled/leave is that the character won't forget where they are just because they're no longer a member. It's a moot point for TG and DG expulsion as I pointed out above.

That's why I saw the addition of a leave function as unnecessary, but I can see your argument for adding it and while I personally don't see much value in hiding the halls for vamp/were-s, you also have a very good point about the potential value for guild mods of being able to define leave behaviour. Maybe some shady guild where you get kicked out after being framed by another member and have to try to clear your name to re-join.

I am happy to go with your version that adds the leave method. I don't think it really matters whether the load/new events are used for this - that was just a way of avoiding needing the leave method. So if you feel this is the right approach, could you just collect the event add/removes into methods like I did? Then I will commit your PR and close mine.

Baljak:
Yes no problem I admit I was a little lazy to I duplicated code, I agree it's better top have them in dedicated methods.

Outside this PR, was it intentional to prevent DB and TG exclusion in DFU with the exception of being a vampire? I'm asking this because this differs from what classic does. I would have asked on the forums but only moderators such as you or Jay_H can answer there: https://forums.dfworkshop.net/viewtopic.php?f=28&p=32595&sid=06ab2bedfca9108de045dfa1adda0b87#p32595

df-ferital:
Yes, no problem, I admit I was a little lazy so I duplicated some code, I agree it's better top have events registering/unregistering in dedicated methods, I'll update my PR.

Outside this PR, was it intentional to prevent DB and TG exclusion in DFU with the exception of being turning into a vampire? I'm asking this because this differs from what classic does. I would have asked on the forums but only moderators such as you or Jay_H can answer there: https://forums.dfworkshop.net/viewtopic.php?f=28&p=32595&sid=06ab2bedfca9108de045dfa1adda0b87#p32595

ajrb:
Yes it was intentional, I didn't override those methods accidentally. ;-)  I've explained why in that thread.

df-ferital:
OK, not sure I completely agree with that, since it somewhat breaks roleplay to be able to remain in a guild while having a negative rep with it. Even if I understand your motivations. However, as this is outside the scope of this PR, we may discuss this elsewhere :)

I updated my PR by the way.

ajrb:
Looks good to me, will check it and merge this evening.

df-ferital:
Thanks for reviewing :)

I see your point regarding guild expulsion. One issue in classic was you could be recruited by DB even if your reputation with them was negative. Then, if your reputation with them remain below 0 after completing the initiation quest, you would be expelled immediately when talking to guild a member. From my point of view, the DB or TG should not try to recruit you unless your reputation is at least -5 with them. This also gives a chance to fail the first normal quest after the initiation quest, without being expelled from the guild.

ajrb:
I found a problem when I tested this prior to merging. Seems if you load an overworld (outside) save after a char with TG/DB membership, the sequence of execution means that the guild instance is only free'd for GC later in the loading process meaning that if you pop map up the label will appear unfortunately.

What I've done is combine both approaches and reopened my PR #1560. So the leave method is there and will ensure if a player leaves TG/DB they no longer see the labels as you wanted, but the save/new events are being used to unregister the event listeners as early as possible.

df-ferital:
This is not an issue, because we unregister events in `Leave`, so immediately when calling `GuildManager.Clear` or `GuildManager.RemoveMembership`. Then, we don't need to care about when the instances are actually deleted by the GC. Since the old instances are not longer registered, TG/DB hideouts of the previous savegame will not appear on the map. You can verify it in game.

ajrb:
Well I'm surprised that you don't see it, but I have saves that reproduce the issue reliably and I was verifying it in-game. I didn't track it down because using the save/new events avoids any timing issues and works with all combos I've tried. I'll link to a zip with the saves in after I've finished testing another PR.

df-ferital:
Maybe you mean when loading a save outside town? Indeed I didn't check this case, sorry...

ajrb:
No in town, but not in a building. Maybe I am loading saves with no memberships... let me check.

ajrb:
Nope, I'm stumped why I see this behaviour and you don't. I'd appreciated it if you could test these saves and see what happens your end with them. Load save 104, then load 153 and hit M, scroll down and see the DB label.

https://drive.google.com/file/d/1TQCJt-OjUwcLdr0CViD4fvoy46jCAgtj/view?usp=sharing

df-ferital:
Good catch! Hope you don't mind that I used a little part of your PR but put if in GuildManager instead. I think it's a little better this way, in particular for potential modded guilds which might also require events cleaning.

Hope we're done with this issue now...

ajrb:
Nope, I don't mind at all. This is the most long distance pair-programming ever! lol  I'll take a look again tomorrow. I like that the save/new events are in the manager rather than each of the relevant guild classes though.

df-ferital:
Yes that was pretty long but I'm quite confident we're done with it now :)

ajrb:
When I looked at actually overriding the Leave() method in a mod I found a problem with it being called from ClearMembershipData() because that is used by the game mechanics to just clear memberships and if a mod implements leaving behaviour, then it's executed during load etc. This means that Leave() can't be used by GuildManager to unregister the events, and guild implementations have to handle unregistering themselves.

Just wanted to let you know that I've had to change it for that reason in a PR I'm preparing that will allow vanilla guilds to be overridden by mods. I'm actually implementing TG/DB expulsion with extreme prejudice for the R&R mod... the idea was too tempting to ignore. Thanks for the suggestion of the Leave() function for modding purposes!

PR #1560, opened 2019-10-07
Unregister events for TG and DB guild instances

ajrb:
So they are correctly garbage collected and guild halls are not erroneously revealed after load/new game.

Supercedes PR #1559.

-----
Previously, Thieves Guild or Dark Brotherhood instances were never garbage collected because PlayerGPS and StreamingWorld kept strong references to them through events OnEnterLocationRect and OnAvailableLocationGameObject. This, because unsubscribing to these events took place in both class destructors, so, to summarize, it never happened.

This resulted in the following behavior: if you loaded a savegame with a character who was a member of any of these guilds, then loaded another savegame from another character who wasn't, you would still discover the guild hideouts using the automap.

Unfortunately, any game saved prior to this fix may still have the guilds available on the automap, this because all discovered buildings are saved.

### DIFF
--- a/Assets/Scripts/Game/Guilds/DarkBrotherhood.cs
+++ b/Assets/Scripts/Game/Guilds/DarkBrotherhood.cs
@@ -189,6 +189,11 @@ namespace DaggerfallWorkshop.Game.Guilds
             RevealGuildHallOnMap();
         }
 
+        override public void Leave()
+        {
+            UnregisterEvents();
+        }
+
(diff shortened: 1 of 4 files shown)


ajrb:
@df-ferital argues that adding leave function to guilds has value, so closing my approach in favour of his.

ajrb:
Seems after testing that a combination of both approaches is needed, so updated and reopened this PR.

df-ferital:
I don't think this is needed, see my comment on my own PR.

ajrb:
Combined into #1559, so closing again.

PR #1561, opened 2019-10-09
Don't attempt to create textures for non existant images with zero width/height

ajrb:
This allows virtual assets (i.e. don't exist in vanilla files) to be replaced by mods. I am using it to allow extra images for the horses neck to be contributed by other mods and used by R&R mod, even though none exist in vanilla DF.

### DIFF
--- a/Assets/Scripts/Utility/ImageReader.cs
+++ b/Assets/Scripts/Utility/ImageReader.cs
@@ -378,7 +378,7 @@ namespace DaggerfallWorkshop.Utility
             imageData.height = dfBitmap.Height;
 
             // Create Texture2D
-            if (createTexture)
+            if (createTexture && imageData.width > 0 && imageData.height > 0)
             {
                 // Get colors array
                 Color32[] colors = GetColors(imageData);


TheLacus:
`ImageReader.GetImageData` first accesses file from classic assets path

```
TextureFile textureFile = new TextureFile(Path.Combine(dfUnity.Arena2Path, filename), FileUsage.UseMemory, true);
textureFile.LoadPalette(Path.Combine(dfUnity.Arena2Path, textureFile.PaletteName));
dfBitmap = textureFile.GetDFBitmap(record, frame);
```

and only then create a texture from it. A bitmap with resolution 0x0 is the result of a failure to validate asset file rather than an intended situation to be found in.

Can you provide more context on why this is needed? Custom textures can already be imported directly with the `TextureReplacement` class.

ajrb:
Yes - it's the accessing of the classic assets path bit if the texture was not actually a classic asset. It still tries, and fails, to create a texture when no replacement is found because createTexture flag is not set to false. Since dfBitmap is empty as there was never a classic asset with this name, the texture creation logs an error. This change makes it so that it doesn't try to create a texture from an empty dfBitmap and so not log the error.

The reason I want to do this is to support virtual assets that mods can optionally use, which can then be replaced in other mods. The specific case is for the extra neck for enhanced riding, by default my mod reuses part of the existing asset, but if a mod replaces using certain filenames, it uses those instead. It does this by trying to load the image and then uses it if the texture is not null.

Maybe there's a better way to achieve this, that's why I wanted you to review it.

ajrb:
Well, I wrote an explanation. Then I re-read what you said and facepalmed! I'm a bone headed idiot. Got so wrapped up with what I was doing and simply never stepped back and thought why don't I just get the replacement image.

Thanks TheLacus. God I feel stupid now...

PR #1562, opened 2019-10-09
Enable guild modding

ajrb:
* **Extract IGuild interface and use it instead of Guild abstract class** - This will allow mods to be more easily created with guild implementations that deviate from default vanilla guild behaviour. Note there are still some constraints, like 9 rank based etc.

* **Allow vanilla guilds to be overridden** - by registering a new implementation extended from core or completely new implementations.

* **Move save/new event unregistering into guild classes** - This is so that GuildManager.ClearMembershipData() does not call Leave() to unregister the guild events. This is because the clear method is for clearing memberships while loading etc (happens twice) and is not for when the player is actually leaving the guild. If mods add behaviour for leaving, it should not be initiated when just saving/loading/new game etc. Now only RemoveMembership calls Leave() so this is the case.

### DIFF
--- a/Assets/Scripts/Game/Guilds/DarkBrotherhood.cs
+++ b/Assets/Scripts/Game/Guilds/DarkBrotherhood.cs
@@ -80,16 +80,6 @@ namespace DaggerfallWorkshop.Game.Guilds
 
         #region Guild Membership and Faction
 
-        public DarkBrotherhood()
-        {
-            RegisterEvents();
-        }
-
-        ~DarkBrotherhood()
-        {
-            UnregisterEvents();
-        }
-
         public static int FactionId { get { return factionId; } }
 
         public override int GetFactionId()
--- a/Assets/Scripts/Game/Guilds/DarkBrotherhood.cs
+++ b/Assets/Scripts/Game/Guilds/DarkBrotherhood.cs
@@ -241,7 +241,7 @@ namespace DaggerfallWorkshop.Game.Guilds
 
         #region Serialization
 
-        internal override void RestoreGuildData(GuildMembership_v1 data)
+        public override void RestoreGuildData(GuildMembership_v1 data)
         {
             base.RestoreGuildData(data);
             RegisterEvents();
--- a/Assets/Scripts/Game/Guilds/DarkBrotherhood.cs
+++ b/Assets/Scripts/Game/Guilds/DarkBrotherhood.cs
@@ -13,6 +13,7 @@ using DaggerfallWorkshop.Game.Entity;
 using System;
 using DaggerfallWorkshop.Utility;
 using DaggerfallWorkshop.Game.Serialization;
+using DaggerfallWorkshop.Game.Utility;
 
 namespace DaggerfallWorkshop.Game.Guilds
 {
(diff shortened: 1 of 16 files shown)


ajrb:
See https://github.com/ajrb/dfunity-mods/blob/master/RoleplayRealism/Scripts/DarkBrotherhoodRR.cs for an implementation of DB expulsion with prejudice.

@df-ferital Could you take a look over this and test it a bit. I think we proved yesterday that different eyes spot different things.

Assets/Scripts/Game/Guilds/DarkBrotherhood.cs
```diff
@@ -79,6 +80,16 @@ public class DarkBrotherhood : Guild
...
         #region Guild Membership and Faction
 
+        public DarkBrotherhood()
+        {
+            RegisterEvents();
+        }
+
+        ~DarkBrotherhood()
+        {
+            UnregisterEvents();
```
df-ferital: Not needed because when the destructor is called, it means events are already unregistered.

This is the case because the events themselves keep a strong reference on the instance so as long as they are registered, the instance can't be garbage collected.

Assets/Scripts/Game/Guilds/ThievesGuild.cs
```diff
@@ -76,6 +77,16 @@ public class ThievesGuild : Guild
...
         #region Guild Membership and Faction
 
+        public ThievesGuild()
+        {
+            RegisterEvents();
+        }
+
+        ~ThievesGuild()
+        {
+            UnregisterEvents();
```
df-ferital: Same as for the DarkBrotherhood destructor: this is not needed (see my comment above).

Assets/Scripts/Game/Guilds/DarkBrotherhood.cs
```diff
@@ -79,6 +80,16 @@ public class DarkBrotherhood : Guild
 
         #region Guild Membership and Faction
 
+        public DarkBrotherhood()
+        {
+            RegisterEvents();
```
df-ferital: This is not needed since events are registered when joining the guild or when loading the game.

Assets/Scripts/Game/Guilds/ThievesGuild.cs
```diff
@@ -76,6 +77,16 @@ public class ThievesGuild : Guild
 
         #region Guild Membership and Faction
 
+        public ThievesGuild()
+        {
+            RegisterEvents();
```
df-ferital: Same as for the DarkBrotherhood class, not needed since events are registered when joining the guild or when loading the game.

ajrb:
> I'm not entirely satisfied with the fact that guild instances like DB and TG are "knowledgeable" about events such as starting or loading a game. In my opinion, this should clearly stay in the scope the `GuildManager` itself. I understand why you decided to remove events subscription from this class, but I think they should stay there. What should need to be changed is to remove the call to `Guild.Leave` in `ClearMembershipData` and replace it with `Guild.UnregisterEvents`. 

That's an interesting suggestion that the guild mgr could orchestrate this via methods in the interface to avoid having the load/new event handling within the guild implementations. I actually tried it out, and unfortunately found that the loading system can abort and restart it's coroutine so there's no way to ensure that everything is done just once or even in order. I'm not about to start rewriting Interkarma's loading code at this stage, so we will live with the guild classes needing to manage their own event registration in a self contained manner to ensure each instance is tidied correctly.

I have no idea if this is why I didn't do this back when I originally implemented guilds, or if it just didn't occur to me. Anyway you were right about the constructor/destructor, they are not required but I just copied them in regardless. Have you tried this PR in-game?

df-ferital:
I just tested your PR in-game, everything seems ok. 👍

PR #1565, opened 2019-10-10
Enemies light effects [RFC]

### LINKED ISSUE #1544: Fire Daedra and Fire Atronach don't emit light around them
After I enquired this feature to be added on the forums, I was told that this is actually a feature of vanilla Daggerfall which is missing the the DFU.  It would be really great to have these mobs emit light around them in DFU as well. 

Im not sure if any other mobs should have this property, maybe frost daedra as well?

Ive made these mobs emissive in the DREAM mod, but thats not really it.

fire

petchema:
This is my attempt to fix #1544, and also remove vampires and ghosts shadows.
Limitations:

- When using "cm" command, spawn enemies do not have light effects, that could mean I'm adding light effects at a too "high" in the code?
- Some parameters like light position is hardcoded, I think it looks good but maybe not compatible with future enemies modding? ( https://forums.dfworkshop.net/viewtopic.php?f=12&t=1983 )

So I'm mainly pushing it here to increase the visibility of this patch


petchema:
Emissive textures:
emissive textures

Interkarma:
Love this change! Merging to master for this weekend's builds. :)

PR #1567, opened 2019-10-13
(de)buff icons sizes and positions

petchema:
Allow to select between different (de)buff icons layouts

### DIFF
--- a/Assets/Resources/defaults.ini.txt
+++ b/Assets/Resources/defaults.ini.txt
@@ -50,6 +50,7 @@ EnableInventoryInfoPanel=True
 EnableEnhancedItemLists=False
 EnableVitalsIndicators=True
 EnableModernConversationStyleInTalkWindow=False
+IconsPositioningScheme=Classic
 HelmAndShieldMaterialDisplay=0
 InteractionModeIcon = minimal
 EnableArrowCounter=True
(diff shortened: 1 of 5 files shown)


petchema:
classic buffs
medium buffs
small buffs
small deck left buffs
small deck right buffs
small vert left buffs
small vert right buffs

HybOj:
This is lovely! Maybe consider adding the variants with icons at the bottom row of the monitor, between the health bars and compass? Anyway, this is really good and needed for ppl who prefer as clean HUD as possible.

petchema:
Bottom of the screen is used for debuffs inflicted by enemies, that doesn't show up on screenshots (as I don't know a simple way to trigger them at will), but debuffs positions are roughly horizontal mirrored positions of buffs, I just kept that classic layout trait.
Horizontal mirroring is not enforced by the code, one could say decide to put buffs to the left and debuffs to the right; But you always need to define two zones.

HybOj:
Oh you are right, Ive forgot about that, sorry

bansheexyz:
Well, here's a scalecap of 2x on the bottom edge. I think it looks good. Too bad DF didn't color-code buff vs debuff so we have to put them on different rows.

buffsbottom5c

Worst case scenario

buffsbottom5

Interkarma:
This is excellent work Pango. If you don't mind, I'll keep this one in the list until the large HUD is further along (going to wrap this up soon now) so that everything can be done in one release.

Interkarma:
Since I've deprioritised the large HUD for the moment, let's just merge this one. It's too good to sit in the queue. We can work out debuffs and bottom alignment once the large HUD comes online early next year.

PR #1568, opened 2019-10-13
Do not delay quest create foe by one spawnInterval

petchema:
Forums:
https://forums.dfworkshop.net/viewtopic.php?f=25&t=2556&p=32825#p32825


Interkarma:
A good change, thank you. :)

PR #1569, opened 2019-10-14
Boost Jumping for lycanthrope characters

df-ferital:
Add the missing bonus to Jumping skill for werewolf or wereboar characters. I also fixed the UESP page in the meantime, which lacked this information (https://en.uesp.net/wiki/Daggerfall:Lycanthropy#Advantages).


Interkarma:
Thank you!

PR #1574, opened 2019-10-17
raise event on teleporting out of dungeons

petchema:
raise transition event on teleporting out of dungeons so it can be handled by Persistent Dungeons mod.

/!\ It seems to work, but I won't pretend I understand all what's going  on.

Forums:
https://forums.dfworkshop.net/viewtopic.php?f=27&t=1667&p=29398#p29398


Interkarma:
That should be fine, it was totally missing that event. Thanks so much! :)

PR #1575, opened 2019-10-17
Add UIWindowFactory

ajrb:
Add a UIWindowFactory to allow mods to provide new implementations of UI windows


ajrb:
I think I've got all the new's. Left a few of the default commands that might be useful to have the original UI window remain availiable. Can change them if requested to.

Could use some more testing. I have not tested every single window construction path.

ajrb:
The conflicts can now be resolved by simply remove the original lines

Interkarma:
Thank you, will merge this one now.

PR #1576, opened 2019-10-18
Fix PC greeting text for the first question

df-ferital:
Fix the the first question greeting which was not updated when changing talking tone and also match classic behavior in using specific greeting names for NPCs (friend, mate, bud...) when NPC reaction is negative, see below:

Before:
before

After:
after


Interkarma:
Good catch.

PR #1580, opened 2019-10-19
Fixes for the upcoming build

ajrb:
Instantiate persistent windows only when needed and allow re-instantiation. Allows persistent windows to be replaced.

Made members protected on rest and pause windows during testing. All windows will require protections to be modified, but I suggest we do them as required. I certainly don't want to do them all en masse.

Fix loading new locations from dfmod's. Have to wait until mod manager is initialised before checking for new locations and filling cache

Move mod manager singleton creation.


ajrb:
Had to move singleton creation into start to ensure world data replacement can rely on instance indicating that mods are loaded. Without this, the cache is filled with indicators that no replacement locations have been found.

ajrb:
I've spent a few hours play testing the latest code with this PR and found no issues.

Interkarma:
Thank you @ajrb :)

PR #1581, opened 2019-10-20
fix minor typo

Arcane21:
Fixes a stray typo missed in earlier pull request


Interkarma:
I read through that PR twice and didn't spot this one either. Thanks for fix! :)

PR #1582, opened 2019-10-21
Fix %qdt macro expansion

df-ferital:
Correct macro %qdt expansion so each quest log entry now has its own time as in classic. Before that, it always resolved to quest start time.

Also fix quest K0C01Y00 message 1017 which always used the current time.

Note that this doesn't break current savegames as I added a check to ensure `LogEntry.dateTime` is not `null`. Else, it resolves to quest start time as before this fix.


Interkarma:
Looks good, thank you!

PR #1583, opened 2019-10-22
Set main quest NPCs as questors

df-ferital:
Main quest questor NPCs were never flagged as questors, so they never said their post success or post failure greetings to the player.

Also fix quest S0000005 where Aubk-i post success and post failure greetings were inverted. And removed the word "dead" from post failure greeting as neither her husband nor her husband's grandmother are dead.

Edit: I'm not a native English speaker, but it seems to me that "The %ra that refused visit my husband's grandmother" should be changed to "The %ra who refused to visit my husband's grandmother". Could someone confirm?

### DIFF
--- a/Assets/Scripts/Game/Questing/Person.cs
+++ b/Assets/Scripts/Game/Questing/Person.cs
@@ -84,6 +84,7 @@ namespace DaggerfallWorkshop.Game.Questing
         public bool IsQuestor
         {
             get { return isQuestor; }
+            set { SetQuestorStatus(value); }
         }
 
         public bool IsIndividualNPC
@@ -484,6 +485,17 @@ namespace DaggerfallWorkshop.Game.Questing
             isDestroyed = true;
         }
 
+        /// <summary>
+        /// Set Person as questor and assign questor data if needed.
+        /// Uses last clicked NPC data so has to be called after talking to the actual questor.
+        /// </summary>
+        private void SetQuestorStatus(bool status)
+        {
+            isQuestor = status;
+            if (isQuestor)
+                questorData = QuestMachine.Instance.LastNPCClicked.Data;
+        }
+
         #endregion
 
         #region Private Methods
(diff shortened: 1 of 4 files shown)


JayH2971:
"that" is acceptable but "who" is grammatically correct.

ajrb:
I think it should stay as it is. Not only is 'that' acceptable since this is something someone is saying, in context it gives a different nuance. It immediately follows the race of the person in question, so using 'that' instead of 'who' indicates the speakers' opinion that the person is unworthy of being fully considered a person simply due to race. Hence that/object instead of who/person.

I could be reading too much into it, but I'm not steeped in lore to know if the DF authors would have put in subtle racism like this. Either way I don't think the change can be justified on the basis of bad english.

df-ferital:
Thank you both, so I left 'that' and added 'to'.

Interkarma:
Quests can also drop questor status as part of the script. I think the core confusion here is that I used "questor" in two different ways:

1. Person.IsQuestor property is used for random quest-givers from guilds, etc.
2. Quest.GetQuestors() is used for individual questors added by scripts (e.g. "add Barenziah as questor" and "drop Barenziah as questor" in Sx011.

If we're going to raise the Person.IsQuestor flag when adding a questor from script, then PR will also need to lower flag again if that questor is dropped later.

These two uses happened at different stages of quest system evolution, and I clearly didn't quite unify them well enough. I'll leave this one open for a while, as I'd like to take another at how I've handled this.

df-ferital:
I understand, thanks for your review :). Actually, the only real issue I tried to solve here was the post success or failure greeting which wasn't used for main quests NPCs. I thought this was the best solution but I'm not familiar enough with DFU quest system.

Interkarma:
I'm happy with the solution. The only thing I think it's missing is lowering the Person.IsQuestor flag again when Quest.DropQuestor is called.

I just want to take another look at how I've done this first. I might end up going with your solution anyway, as it's a good low impact change that achieves the result. :)

df-ferital:
You're right, I modified my PR a little to lower `Person.IsQuestor` when `Quest.DropQuestor` is called. 
However, since post success and post failure greetings are said after `DropQuestor` has been called, I also changed `TalkManager.GetNPCQuestGreeting` to only check `Person.QuestorData`. This should not change anything regarding previous behavior.

PR #1586, opened 2019-10-24
Small copy/paste mistake fix.

JorisVanEijden:
The "named" version of anyInfo is not used in any current quests but if we support it it should work correctly.


Interkarma:
Can you expand on why you believe this this a copy-paste error? "anyInfo" and "used" are already being parsed separately and have different purposes. I don't see any issue with the code as it stands.

PR #1587, opened 2019-10-25
World data override system development (WIP, not ready for merge)

ajrb:
This is my continued development of the world data systems. Doesn't seem to break anything that I've noticed during my testing, but I will understand if you would rather hold off releasing.

- RMB blocks can now be replaced, have added a custom serializer for the ground data
- Added world data variants system so different overrides can be used based on triggers - works for locations, blocks and buildings.
- Variant blocks/buildings can be triggered just for a specific location or for every instance the block used
- Variants triggered are saved & loaded from a new file
- Adds a quest action 'worldupdate' so quests can trigger variants
- Fixed some issues with dump commands


Interkarma:
>Adds a quest action 'worldupdate' so quests can trigger variants

OMG! I never would have thought of this. The possibilities here are very exciting.

ajrb:
Going to close this and re-submit when it's complete.

PR #1590, opened 2019-10-28
Gold infopanel update

petchema:
Show amount of gold and gold weight in infopanel when the mouse hovers over the inventory GOLD button

I reify the gold pieces into a temporary DaggerfallUnityItem for the purpose of this feature; That may not be optimal, but this is not in a performance critical path and allows good code reuse...

Forums: https://forums.dfworkshop.net/viewtopic.php?f=23&t=2717


jimfcarroll:
@Interkarma , what's needed to get this in the next release build? I ran with it for a while a couple months ago. I can put it back in now and run with it to help verify there's no other issues if that will make a difference.

Interkarma:
I'd rather not be generating a new item and burning a new UID every time the mouse enters a panel. While the UID number space is huge, it is still finite. There must be another way to do this, such as adding an overload to UpdateItemInfoPanel.

jimfcarroll:
Well, it looks like your UID mechanism is a ulong. I'm surprised you didn't use the C# native one.

As a side, just looking at the uid code generation, does Unity handle threading for you? It looks like in a multi-threaded application you'd have problems with the way that's implemented.

petchema:
I went quite far with macros (to anyone interested, https://github.com/petchema/daggerfall-unity/compare/gold-infopanel-update...petchema:gold-infopanel-update-using-macros ) then stepped back to something much simpler...

PR #1591, opened 2019-10-28
Revert change from PR#1557 to building matching

ajrb:
This change meant that replacement buildings e.g. new guilds could take a guild hall entry out of the pool, messing up the matching algorithm.
This does mean that replacement classic guild buildings probably won't work without futher changes, but this fixes an issue where classic guilds are broken with the Archaeologists mod in DFU 0.10.8 - 0.10.10


Interkarma:
Thank you Hazelnut!

ajrb:
No problem, it was my mess to sort. Sorry about that. :)

PR #1592, opened 2019-10-28
Use WoodlandHills flats in MountainWoods climate, set Swamp and SubTropical people to Redguard

jammerns:
This fixes the issue with WoodlandHills flats being unused in MountainWoods climate regions as discussed in this thread: https://forums.dfworkshop.net/viewtopic.php?f=4&t=2927

It also sets the People faction to Redguard in Swamp and SubTropical regions as these are entirely within Hammerfell.


Interkarma:
I'll verify that MountainWoods actually uses those terrain flats when I can. Thank you for this. :)

But populations are correct as they stand. Please visit places like Sentinel in classic that fall with Subtropical yet have a Breton population. I also don't believe populations are determined this way any longer. From memory, @df-ferital reimplemented this to work like classic some time back using another method.

jammerns:
Interesting, it did change the appearance of the citizens, it might be the names that are generated from a different table based on region. I opened a different pull request containing only the climate change: https://github.com/Interkarma/daggerfall-unity/pull/1593

PR #1593, opened 2019-10-29
Use WoodlandHills flats in MountainWoods climate

jammerns:
Change only the climate from https://github.com/Interkarma/daggerfall-unity/pull/1592


jammerns:
Just checked by running the DosBOX version and it looks like the woodland hills flats are *not* in classic, so I'll close this for now.

PR #1594, opened 2019-10-29
Fix NPCs factions and talk reactions

df-ferital:
This assigns correct factions to NPCs inside buildings, both to get the list of NPCs having work and to get a proper reaction when talking to such people. This also fixes the reaction value of all NPCs, including mobile ones.

Before, when talking to any NPC, his/her reaction value was always computed using the current region faction reputation. This corrects it by using the NPC faction or its parent one in case he's a member of a sub-faction (this matches what classic does).

Also, since I almost entirely decompiled classic talk system in C code, I saw how classic original code is a real mess in this case. For example, when talking to an NPC, its top parent faction of kind "Group", "Province" or "Temple" is used to get his reaction value. But then, the original NPC faction is also used to get the greeting number when the NPC belongs to "People of" or "Court of" current region... So I simplified this to always use the NPC parent group reputation in both cases, including "People of" and "Court of" in what I'm calling "parent groups". Well, except if the NPC is an Individual: if you are loved by Court of Daggerfall but hated by King Gothryd, there is absolutely no reason the latter would greet you warmly.

I also changed one last thing: NPCs belonging to Random Ruler, Random Noble and Random Knight factions now always use the "Court of" their current region faction. The above generic factions are useless and this makes much more sense.

### DIFF
--- a/Assets/Scripts/Game/Player/PersistentFactionData.cs
+++ b/Assets/Scripts/Game/Player/PersistentFactionData.cs
@@ -809,5 +809,26 @@ namespace DaggerfallWorkshop.Game.Player
 
         #endregion
 
+        #region Parent Group
+
+        /// <summary>
+        /// Find the top-level parent group of a given faction. This parent can be a Group, a Province or a Temple.
+        /// </summary>
+        /// <param name="faction">The faction to get the parent of.</param>
+        /// <param name="parentFaction">The parent group faction.</param>
+        public void GetParentGroupFaction(FactionFile.FactionData faction, out FactionFile.FactionData parentFaction)
+        {
+            parentFaction = faction;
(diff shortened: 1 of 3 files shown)


df-ferital:
No, Merchants are unaffected since their faction ID is always correct. The only affected static NPCs are those whose faction ID is 0 and NPCs belonging to the Random Ruler faction or any of its sub-factions.

This means that all people in Palaces should now be able to offer Nobility quests while people from other builidings should continue to offer Merchants/Commoners quests as they did before. This mostly matches what classic does.

Interkarma:
Thanks so much! I'll get this merged in next weekend's builds. I've just run out of time this week to review any more PRs. At least I'm whittling down the backlog now. :)

PR #1595, opened 2019-10-29
Add VS Code workspace settings and Json schemas

TheLacus:
Add _settings.json_ file for Visual Studio Code with the following options:

- `files.exclude` created automatically by Unity when the folder is opened for the first time.
- `files.associations` for TEMPLATE quest files (fall back to plaintext if TEMPLATE extension is not installed).
- `csharp` editor settings for space indent, with the value that is already used by all C# scripts.

This file can also be used to associate [json schemas](https://code.visualstudio.com/docs/languages/json#_json-schemas-and-settings), simplifying the creation of json files contributed by mods.


TheLacus:
book_mapping_json

PR #1596, opened 2019-10-30
Increase accessibility of Trade Window for modding purposes

AsesinoBlade:
Hi all,
I am taking advantage of The UIFactory to add my inventory filter as a mod.  I need to have some variables and a method switched to protected in order to extend the window.

Thank you for your consideration.
A


df-ferital:
With small changes like that, my advise would be to squash all your commits down into a single one.

ajrb:
Yeah I agree. 9 commits for as many lines is OTT. Other than that it looks fine to me.

petchema:
PRs can be squashed at merge time, so that should be okay
MPR options

AsesinoBlade:
Thanks all, I am unsure how I would squash it. I am happy to do so if someone can point me to an instruction guide. 

Thanks again all,
A

TheLacus:
I made some changes. For example `usingWagon` is toggled by method `SelectWagon()` so setter shouldn't be private. There was also a field that was already exposed by a public property, so i restored private modifier.

Please tell me if you have everything to make your mod and i'll merge :)

AsesinoBlade:
Hi @TheLacus, I tried out the changes ; it appears to work fine except for the get set WindowMode and Guild.  Maybe I am building my constructor incorrectly; I just copied the constructor for DaggerfallTradeWindow.

Here is my constructor:
`      public AsesinoTradeWindow(IUserInterfaceManager uiManager, DaggerfallBaseWindow previous = null, WindowModes windowMode = WindowModes.Sell, IGuild guild = null)
            : base(uiManager, previous)
        {
            this.WindowMode = windowMode;
            this.Guild = guild;
        }`
With the way I did it, i need to have access to set in order for the constructor to compile.

Other than that, it seems to work fine; I tested a few things out and it is working correctly.

Thanks again for all the help @TheLacus, I really appreciate it.
a

ajrb:
`
public AsesinoTradeWindow(IUserInterfaceManager uiManager, DaggerfallBaseWindow previous = null, WindowModes windowMode = WindowModes.Sell, IGuild guild = null) : base(uiManager, previous, windowMode, guild) {
}`

You need to pass the constructor params into the base class.

AsesinoBlade:
Thanks Arjb. I will try that when I get home. 
I appreciate the help. 
Thanks
A

PR #1598, opened 2019-10-31
Support providers of mobile person graphics

TheLacus:
Allows mods to provide a gameobject that replaces mobile person prefab. An abstract component class is used to interact with the game, specifically to handle actual npc assignment and animation state.

This is a work in progress born from #1520

A basic example that replaces billboards with cubes:

```cs
using UnityEngine;
using DaggerfallWorkshop;
using DaggerfallWorkshop.Game.Entity;
using DaggerfallWorkshop.Game.Utility.ModSupport;

namespace MobilePersonCubeMod
{
    [ImportedComponent]
    public class MobilePersonCube : MobilePersonAsset
    {
        MeshRenderer meshRenderer;

        private void Awake()
        {
            meshRenderer = GetComponent<MeshRenderer>();
        }

        public override bool IsIdle
        {
            get; set;
        }

        public override void SetPerson(Races race, Genders gender, int personVariant, bool isGuard)
        {
            Vector3 size = GetSize();
            Trigger.height = size.y * 1.2f;
            Trigger.radius = size.x * 1.2f;
        }

        public override Vector3 GetSize()
        {
            return meshRenderer.bounds.size;
        }
    }
}
```

### DIFF
--- a/Assets/Scripts/Editor/SetupMobilePersonEditor.cs
+++ b/Assets/Scripts/Editor/SetupMobilePersonEditor.cs
@@ -42,10 +42,10 @@ namespace DaggerfallWorkshop
             }
             if (GUILayout.Button("Align To Ground"))
             {
-                MobilePersonBillboard mobilePerson = setupMobilePerson.GetComponentInChildren<MobilePersonBillboard>();
+                var mobilePerson = setupMobilePerson.GetComponentInChildren<MobilePersonAsset>();
                 if (mobilePerson)
                 {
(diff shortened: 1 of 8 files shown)


Interkarma:
I see what you're doing here and think it's an awesome idea. I don't have any issues with this and I'm happy to merge now if you're comfortable having in live.

TheLacus:
Thanks, I'm ok with this to be merged when you can.

PR #1600, opened 2019-11-04
Show warning for incompatible mods when game is launched

TheLacus:
A warning popup is shown if a mod is enabled and specifies a required game version that is higher than running build. It is possible to dismiss popup and launch the game anyway.


Interkarma:
Thanks for all these fixes. I'm sorry to not have merged by now, I've been a bit overwhelmed by the number of recent PRs.

I'm finally getting on top of everything and will merge this in next weekend's builds. :)

PR #1601, opened 2019-11-05
Add support for mod dependencies

TheLacus:
Allows mods to define other mods as dependencies to enforce a minimum version to be installed and/or to be positioned above in the load order. Dependencies are also used for the creation of an automatic load order.

```json
"Dependencies": [
    {
        "Name": "dungeonexit",
        "IsOptional": false,
        "IsPeer": false,
        "Version": "1.0.2"
    }
]
```

mods_dependencies


JayH2971:
Could there also be an "incompatibilities" field, for contraindications?

Interkarma:
This one now has some conflicts to resolve. Sorry for the inconvenience Lacus!

I've almost run out of time this weekend again. I'll loop this into next weekend's builds if you're ready for it.

PR #1602, opened 2019-11-06
Fix 1gp reward bug, properly this time!

ajrb:
Please someone check this before it goes out in a build. Ta.

Sorry for taking so long @Interkarma.


Interkarma:
Thanks so much @ajrb! :)

PR #1603, opened 2019-11-07
Implement option to partially parse a quest

### LINKED ISSUE #1536: Quest Tracking - Unknown Quests in Dialogue
**Describe the bug**
When looking through the dialogue options after having played the game and completing a few small quests (For example, the "A Delivery" quest from the House of Dibella) a series of unknown people and objects appears in the "Tell Me About..." dialogue section, as well as unknown Residential locations not relating to any previous quests.

**To Reproduce**
1: Accept quests from any Quest Giver, and complete the quests as indicated. (Make sure to accept/complete multiple quests in a single day for the most immediate effect.)
2: After completing said quests, go talk to any NPC and attempt to look at the "Tell Me About..." or "General" locations.
3: You will see a number of unusual names/items/locations have appeared that do not have relevance to any previous quests you have completed.

**Expected behavior**
I should not be seeing random talking points showing up in my dialogue options; they should either relate to currently-running quests, or be something relating to the locale I am currently in.

**Screenshots**
DagUnityBug5

 - OS: Windows 10 Home Edition
 - Version/Build: 10.0.17763

**Additional context**
- The residential locations only seem to show up when a quest is currently active; the locations show up along with the actual location you would need during a quest. (For example, I might get the Moorsley Residence as the location I need to go to, however 3-4 other residences will also show up as talking point options.)
- Clicking on the extra talking points results in no reply from the NPC. Further, if you attempt to talk to the NPC about anything thereafter, they no longer respond. 
- I have found a workaround that rebooting the game clears out these invalid locations/talking points. As far as issues with no response from the NPC, you can start getting responses from them again if you ask them about a "Regional" option, such as "Any Temple of Kynareth."

ajrb:
This is what I considered doing when I originally implementing the quest lists option, but I was concerned about altering the quest parsing stuff so instead I decided to do full parse of every quest and have a 'please wait' message instead.

Since this is the root cause of the bugs we've seen with this feature, I've gone back to my original idea and added a boolean param that allows a partial quest parse to be requested. This doesn't parse the QRC or QBN sections, so the quest object returned is not startable.

This should fix #1536 and also speed up the list preparation. I've left the please wait message in there for now. It could be removed as it's quick enough with this change, but I would like to know that having this partial parse is acceptable by @Interkarma first. Also it's is nice flavor in a way, so I would be happy to keep it. Maybe just have a random delay as the quest giver goes to 'look' at availiable quests? :-)

### DIFF
--- a/Assets/Scripts/Game/Questing/Parser.cs
+++ b/Assets/Scripts/Game/Questing/Parser.cs
@@ -51,7 +51,8 @@ namespace DaggerfallWorkshop.Game.Questing
         /// </summary>
         /// <param name="source">Array of text lines from quest source.</param>
         /// <param name="factionId">Faction id of quest giver for guilds.</param>
-        public Quest Parse(string[] source, int factionId)
+        /// <param name="partialParse">If true the QRC & QBN sections will not be parsed</param>
+        public Quest Parse(string[] source, int factionId, bool partialParse = false)
         {
             Quest quest = new Quest();
             quest.FactionId = factionId;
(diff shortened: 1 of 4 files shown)


Interkarma:
This works well and is a nice low-impact refinement for quest lists. I'll roll this one into today's builds.

PR #1605, opened 2019-11-09
fix for bug: "opening quest log can clear marked residences from map"

Nystul-the-Magician:
fix for bug described here: https://forums.dfworkshop.net/viewtopic.php?f=24&t=2994

info:
the line removed was the culprit causing the bug described, it was necessary before the discovery rework but is no longer needed (since named residences get correctly undiscovered once quest is handed in (I checked this) and when I reworked discovery system last time I changed it so that it deletes the discovery list of orphaned entries on load and save (function PlayerGPS.RemoveUnnamedResidencesFromDiscoveryData) - so no need to handle orphaned entries here


Interkarma:
Thank you so much, Nystul. Great work! :)

PR #1609, opened 2019-11-10
Fix issues with replacement buildings

ajrb:
When altering named buildings, set factionId to zero and the building data pool items (from location) will be used as normal. Set factionId to >0 and pool items will not be used as it will be assumed this is a new named building.

This was the best way I could think of to fix the issues raised here https://forums.dfworkshop.net/viewtopic.php?f=5&p=34619#p34608

While also retaining the fix made by reverting in #1591 

@Interkarma I changed the pool items from a struct to a class so a ref is passed back and the pool item usage can be marked as unused if the replacement data is a new named building with a non zero faction id. Can you confirm this wont screw something else up? I couldn't see any potential issues with it.


Interkarma:
Thank you! This all looks fine on a read, but it's always the simplest changes that turn out the most deadly. :)

I'll start working through PRs again within the next week.

Interkarma:
Looks good. I'll pop this in a bug-fix build today.

Issue #1550, opened 2019-10-03
Support videos as loose files on Linux


TheLacus:
Video replacements can be provided from mods (with any supported format) or as loose files (with expected **.mp4** extension and format).

Unfortunately mp4 is not supported on Linux so loose files mods are not viable on this platform. This is what the Unity documentation suggests:

> For most plaftforms, H.264 (typically in a .mp4, .m4v, or .mov format) is the optimal supported video codec because it offers the best cross-platform compatibility. Linux is the notable exception.
> 
> For Linux, the optimal encoding is usually a .webm container with video tracks encoded using the VP8 codec, and audio encoded using the Vorbis codec. 

Support for loose files on Linux should be added, for example with a `#if` compiler directive to use **.webm** when Linux is targeted. It's important that the new format is tested on Linux both inside the Unity Editor and with a compiled build. We might also switch to a different common format is there are no drawbacks for currently supported platforms (Windows and OSx).

TheLacus:
If i have confirmation that webm works on Linux i'll proceed with this change.

petchema:
Yup, it works fine with just
```                 // Seek from loose files
-                string path = Path.Combine(moviePath, name + ".mp4");
+                string path = Path.Combine(moviePath, name + ".webm");
```
Tested with the FMV converted using
`ffmpeg -i .../DAG2.mp4 -c:v libvpx -c:a libvorbis -b:v 6M .../DAG2.webm`

First 10s for testing:
[DAG2-webm-10s.zip](https://github.com/Interkarma/daggerfall-unity/files/3780377/DAG2-webm-10s.zip)

TheLacus:
Thank you!

petchema:
Does something need to be done for packaged videos too?

TheLacus:
> 
> 
> Does something need to be done for packaged videos too?

No.

PR #1611, opened 2019-11-13
use .webm video format on Linux

petchema:
.mp4 format is not supported by VideoPlayer on Linux:

https://docs.unity3d.com/Manual/VideoSources-FileCompatibility.html

Fix for #1550


Interkarma:
Nice one! Cheers Pango. :)

PR #1613, opened 2019-11-19
prevent flying enemies from flying too low

petchema:
In frame 0 wings are in high position, assume body is the lower half

Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&t=2898

### DIFF
--- a/Assets/Scripts/Game/SetupDemoEnemy.cs
+++ b/Assets/Scripts/Game/SetupDemoEnemy.cs
@@ -40,6 +40,35 @@ namespace DaggerfallWorkshop.Game
                 this.gameObject.SetActive(false);
         }
 
+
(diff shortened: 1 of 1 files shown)


petchema:
We may have to replace formula with enemies properties at some point; Say, to add flying daedra seducers (unless implemented as two separate enemies?)

Interkarma:
This is a sensible fix. Can't believe I didn't consider this originally.

>We may have to replace formula with enemies properties at some point; Say, to add flying daedra seducers (unless implemented as two separate enemies?)

Definitely something to look at in the near future. I'd rather keep it the one enemy type/entity for a variety of reasons, but I think we can update the animation states to provide support for the transformation and transition to flying state.

PR #1614, opened 2019-11-19
Quest fix Nov 2019

JayH2971:
Fix text error C0C00Y10


df-ferital:
There is a problem with your commit, it only adds an empty line.

Interkarma:
> There is a problem with your commit, it only adds an empty line.

This is actually the fix. :) There needs to be whitespace between text blocks and comments or the comment will be included in text block.

Interkarma:
@JayH2971 from the subject of the PR, should I assume that more fixes are incoming?

JayH2971:
I'm sort of leaving it open-ended right now. No guarantees if more come in anytime soon, so don't wait for me.

df-ferital:
> > There is a problem with your commit, it only adds an empty line.
> 
> This is actually the fix. :) There needs to be whitespace between text blocks and comments or the comment will be included in text block.

Sorry I didn't know that :)

PR #1618, opened 2019-11-20
Adds several missing UI sounds

D0ct0r3d:
A large majority of all the sounds are now restored. There are still a few (approx 10%) buttons throughout that are still missing the sound to match classic. I'll add those as time permits and as I discover them. 

Everything i've added should be a 1:1 match for how classic did the sound. There are some buttons where you would think a clicksound would be appropriate for the feedback, but it wasn't in classic - so I didn't do it in the changes. I tried to replicate it exactly.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallExteriorAutomapWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallExteriorAutomapWindow.cs
@@ -1587,6 +1587,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
         {
             if (InDragMode())
                 return;
+                DaggerfallUI.Instance.PlayOneShot(SoundClips.ButtonClick);
 
             ActionSwitchToNextExteriorAutomapViewMode();
         }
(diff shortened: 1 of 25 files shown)


D0ct0r3d:
I believe I have most of them done now. I've been playing it like this since I started and I haven't found anymore missing ones in quite a while. I'm sure there's possibly a few here and there that I haven't caught yet. But, I think it's at a point now to where I'm gonna declare this one ready for review. 

The goal was to just add sounds where they originally were in classic and not try to add any that weren't there to begin with. I believe I did that. Though, it's possible I might have added a clicksound or two in certain UI areas where there originally wasn't one. I had the DOS version on one screen and the Unity on the other and was clicking back and forth on each UI and making notes. 

Overall though, I think the game sounds better now - and the buttons give you feedback when you click them and it's much closer to classic sound behavior now.

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallExteriorAutomapWindow.cs
```diff
@@ -1587,6 +1587,7 @@ private void GridButton_OnMouseClick(BaseScreenComponent sender, Vector2 positio
         {
             if (InDragMode())
                 return;
+                DaggerfallUI.Instance.PlayOneShot(SoundClips.ButtonClick);
```
petchema: Minor remark, I'm not sure what editor you're using, but some indentations are not consistent with the rest of the code, and in a few place like this one can be somewhat confusing (with the since instruction "if" just before).

petchema:
I expected this work to be time consuming, but not to that level, good job!

Interkarma:
Hey there! :)

Just so you know, I intentionally left out a lot of the clicking sounds, as I felt this was a bit excessive in classic. I haven't received any complaints about the quieter experience. But on the other hand, nobody seems to have complained either as more clicks have been added back into the game. It's probably fine either way, and matching classic is usually the safer option. I'll take this for an extended test drive later to see how I feel before merging.

When you get a chance, please take a look at the formatting issue Pango mentions above. There's also a conflict with live code to fix up. Thank you for all your work on this. :)

PR #1619, opened 2019-11-21
Crosshair interaction icon options

ajrb:
Adds two options for the interaction icon to settings that replace the cursor with the icon, except for the general grab mode. This has been requested a few times to reduce the potential for mis-activations.


Interkarma:
I like this one a lot!

PR #1620, opened 2019-11-22
Make previous position of difficulty dagger slowly fade away

petchema:
It can be difficult to assess by how much the difficulty dagger went up or down after adding an advantage or a disadvantage, specially since panels are being displayed on top of the dagger while it changes.

This adds a subtle animation that makes the previous position of the dagger fade away for a second.

It also gives a nice trail as you tweak your HPs per level


Interkarma:
This is excellent! I was a bit unsure just from the description, but it's so nice in practice. A very nice QoL change indeed.

PR #1621, opened 2019-11-22
Add ability for mods to update the politic and climate for map pixels

ajrb:
This supports the ability to add new locations to the world data.

See https://forums.dfworkshop.net/viewtopic.php?f=22&t=3011 for more discussion.


Interkarma:
I'll sneak this into this weekend's builds so your mods can use it live.

PR #1622, opened 2019-11-23
fix negative rented room time

petchema:
Also, I rewrote CheckRent() because it took me a while to understand that **-1** was a magic value for remainingHoursRented. The logic should be unchanged.

Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&t=3043


Interkarma:
Thank you!

PR #1624, opened 2019-11-24
Use brighter green for selected fast travel options

petchema:
Many players seem to initially think the selected options are the "red" ones. Also, red vs green is an extremely poor choice for accessibility (color blindness).

Actually didn't use plain brighter green because it looks very ugly; Instead I added some relief using borders to make them appear more "concrete". Sadly I couldn't use thinner borders than 1 virtual pixel.

I also created another branch based on this one, that changes mouse click to select an option instead of toggling it, which also participates to new player's confusion; But that may be more controversial, so I kept that separate (interface modding material?): https://github.com/Interkarma/daggerfall-unity/compare/master...petchema:less-ambiguous-fasttravel-dialog
FastTravelGreen


Interkarma:
I like this a lot. The brighter value definitely helps differentiate it from the map underneath, and should be easier to see with red-green colour blindness.

Happy to accept other change from toggle to select as well. This has become the standard for radio buttons, and even confuses me sometimes despite a lifetime playing this game.

Interkarma:
Just one minor issue - the GreenCheckbox texture doesn't perfectly fill the checkbox area. That's not enough to stop merging though. I can help with another pass on this if needed so that it aligns perfectly. :)

petchema:
I tried adjusting it the best I could (tested with a 16:9 and 4:3 resolutions, with DREAM,...) but when you adjust one exactly, another one loses or gains a pixel... I selected some average position and size, but it's imperfect.
I feel like to get rid of this problem things should be done in a different way, but I don't know how.

Interkarma:
It's all good, nothing major at all. :) That UI was really early in the process and doesn't follow quite the same practices as the later UIs, so it's a mess to work with. I think I'll be able to help fix it.

PR #1626, opened 2019-11-28
wrap AudioSource.PlayClipAtPoint

petchema:
Add volume setting and asynchronous loading

Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&t=3078


petchema:
Fixing conflict with master in EnemyDeath.cs

Interkarma:
Looks good, thanks for this one. :)

PR #1632, opened 2019-12-01
make sure dungeons's SongManager updates song with each dungeon

petchema:
Use OnTransitionDungeonInterior to update song based on dungeon.

I don't expect we'll switch to full event-driven SongManager immediately, as song selection depends on many factors...

Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=3059


petchema:
OnTransitionDungeonInterior seems to be sufficient for game (re)loading updates

Interkarma:
This is a nice change all around. Not only fixes the initial problem, but makes the code a bit more readable too. Well done!

PR #1633, opened 2019-12-02
don't change footstep sounds during winter under snow-free climates

petchema:
Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&t=3108


Interkarma:
Thank you for this fix and all the others. I'm going to make this weekend's 0.10.14 builds all about your changes and merge almost all your pending PRs. I'll need to call it the "Pango Care Package" or something. :)

PR #1636, opened 2019-12-07
Sink nature flats slightly into ground on slopes

ajrb:
Sink nature flats slightly into ground as slope increases to hide floaty trees on sloped ground a bit. Sinks the flats proportional to angle of ground, with a max of 50/70 = 0.714 unity units on y axis.

Based on a suggestion from discord. Ratio seemed a good balance between no floating and buried sprites. Could be tuned more.


Interkarma:
This looks great!

PR #1637, opened 2019-12-07
Add reset of drained stats to free temple healing

ajrb:
The reset doesn't work correctly yet in this PR. Need Interkarma to take a look because I don't understand the effect system well enough to know why calling EndIncumbentEffect<DrainStrength>() for instance doesn't seem to work.

### DIFF
--- a/Assets/Scripts/Game/MagicAndEffects/Effects/Destruction/DrainEffect.cs
+++ b/Assets/Scripts/Game/MagicAndEffects/Effects/Destruction/DrainEffect.cs
@@ -99,6 +99,12 @@ namespace DaggerfallWorkshop.Game.MagicAndEffects.MagicEffects
                 forcedRoundsRemaining = 0;
         }
 
+        public virtual void CureAttributeDamage()
+        {
+            // Fully heal magnitude of attribute drained by this effect
+            HealAttributeDamage(drainStat, magnitude);
+        }
+
         void ShowPlayerDrained()
         {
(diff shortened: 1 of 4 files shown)


Interkarma:
Attaching a save of a character will all stats drained for testing. They're not a member of temple currently standing in (I just shorted out the guild.FreeHealing() check temporarily while testing).

[SAVE776.zip](https://github.com/Interkarma/daggerfall-unity/files/3935656/SAVE776.zip)

Interkarma:
Just realised I'll need rework my curing change a little bit before merging PR. It's possible for other effects to perform attribute damage, not just DrainEffect. I built the stat mod healing to be agnostic of which effects are causing the damage, but my change to this PR is specific to DrainEffect. This means the temple curing won't recover negative stat mods from other effects (e.g. custom effects not built on DrainEffect).

It's a simple enough change though. Will loop back to this before merging.

ajrb:
Wow, okay so I did the easy bit then. lol

Thanks for finishing it off, it's not completely intuitive what is going on now but that's fine. 

Still, I was wondering what the heal skill helper is for and whether the temple heal should also call that too?

Interkarma:
Skill damage and healing is unused, it's just framework for custom effects that might want to drain skills. Hard to say if temples should heal this, as no debuff spells to drain skills exist in classic. Anyone using this in future might have other plans on how skills should be restored. My feeling is to leave it out for now.

PR #1639, opened 2019-12-09
Climbing skill check formula exposed to mods.

ajrb:
Added skill check when initiating climbing and tuned the default success percentages.

See https://forums.dfworkshop.net/viewtopic.php?f=5&p=36243 for discussion. Happy to tweak the numbers more before merging this one.

### DIFF
--- a/Assets/Scripts/Game/Formulas/FormulaHelper.cs
+++ b/Assets/Scripts/Game/Formulas/FormulaHelper.cs
@@ -274,7 +274,7 @@ namespace DaggerfallWorkshop.Game.Formulas
         public static int CalculateClimbingChance(PlayerEntity player, DaggerfallEntity na, int basePercentSuccess)
         {
             Formula_2de_2i del;
-            if (formula_2de_2i.TryGetValue("CalculateClimbing", out del))
+            if (formula_2de_2i.TryGetValue("CalculateClimbingChance", out del))
                 return del(player, null, basePercentSuccess);
 
             int skill = player.Skills.GetLiveSkillValue(DFCareer.Skills.Climbing);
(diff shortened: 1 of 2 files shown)


TheLacus:
@ajrb Not strictly related to this PR, but did you consider using generics instead of named delegates? You could use a single dictionary where a value is a delegate with the same parameters as the method, instead of multiple dictionaries; this would also avoid unused parameters.

For example if you have a dictionary with an helper method for casting

```cs
public static Dictionary<string, object> FormulaOverrides = new Dictionary<string, object>();

private static bool TryGetDel<T>(string name, out T del)
{
    object data;
    if (FormulaOverrides.TryGetValue(name, out data))
    {
        del = (T)data;
        return true;
    }

    del = default(T);
    return false;
}
```

you can add/retieve a new override with the Func delegate

```cs
public static int CalcExampleValue(int a)
{
    Func<int, int> del;
    if (TryGetDel("CalcExampleValue", out del))
        return del(a);

    return 2 * a; // default implementation
} 
```

```cs
private static void ExampleRegistration()
{
    FormulaHelper.FormulaOverrides["CalcExampleValue"] = new Func<int, int>(a =>
    {
        return 3 * a;
    });
}
```

ajrb:
Oh, that's much nicer. No I didn't consider it because of my limited C# knowledge. :-)

Maybe this would be best discussed on forums. Main concern is if anyone else has mods that override formula yet. Curious why you've waited so long to suggest this, or is this knowledge new to you as well?

TheLacus:
I has never been involved with formulas but this got my attention when i checked open PRs. Custom formulas is a nice feature to have, i simply wanted to see if number of dictionaries could be reduced to a single one using casts and made a quick snippet.

Since you seems to be interested i'll work a bit more on this and reach you out when i have a more solid proposal :)

ajrb:
Sounds good to me. I'm happy to work on this, but didn't want us to replicate work. Look forward to seeing your proposed refinement!

ajrb:
@TheLacus How's the proposal for changing formula overriding coming?

TheLacus:
I have it in a local branch; i'll open a PR tomorrow.

ajrb:
Thanks! I'm adding more formula to override so don't want to use the old system especially since they need a different signature which should be no problem with your solution.

PR #1640, opened 2019-12-10
prevent 3d map from being rotated upside-down

petchema:
Can cause confusing with no benefits
upside-down map


Interkarma:
Simple and useful. Thank you!

PR #1641, opened 2019-12-12
Only iterate for location discovery within bounds of both arrays

ajrb:
Pretty sure this fixes the issue reported here: https://forums.dfworkshop.net/viewtopic.php?f=5&p=36523

Have not fully tested all discovery imports, but seemed to work okay with some classic saves I tried.


petchema:
Works for me too, importing my 6 classic slots worked again in fully modded 0.10.14

Issue #1422, opened 2019-08-13
Enhanced Mod Support for FormulaHelper


EPNW:
The FormulaHelper is a central place where all kind of things are calculated.
Most methods offer the possibility to be overridden by a mod using formula delegates.
I think this is a very good design pattern, but I'd like to take things two steps further:

1) As last parameter of each formula delegate add an `out bool apply` parameter.
If the value of this bool is `false` after the formula delegate returns, don't use the value the delegate provided, but go on with calculations.
This way I can for example override "DeadraSummoningCosts" only if the faction is "Glenmoril Witches" or stick to normal behaviour otherwise.

2) Change the Formula Dictionaries to take a SortedList as value instead of the formula delegate directly.
E.g. `Dictionary<string, Formula_1i>` to `Dictionary<string, SortedList<int,Formula_1i>>`.
Mods should register their formulas in the sorted list, with their mod load order as key.
Then, if we also implement 1) we can iterate through all registered formulas, and return the first where the `apply` parameter is true.

Changing this would break some of the current mods, but will provide a much more flexible solution for the future. I'd like to discuss this ideas before making a PR.

ajrb:
I will take a look at these suggestions when I'm back from holiday. They seem like a good idea to me, but I will want to check the code and think more about it. Originally I thought that only one mod could effectively override a formula, but this is a nice idea to allow more complex behaviour. Main concern is that this complex behaviour dependent on mod order may provide a lot of end use confusion and/or issues.

JorisVanEijden:
Here's an approach for allowing multiple mods not only to replace formulas but also to act on existing values: https://gist.github.com/JorisVanEijden/0e9c18bc35709e4eecd2b3b9719f3e80

Example usage by mod:
```
DaggerfallWorkshop.Game.Formulas.FormulaHelper.RegisterFormulaAlter("DamageModifier", AlterDamageModifier, 100);

public int AlterDamageModifier(ref currentValue, int strength) {
  currentValue -= strength / 2;
}
```

Just as food for thought.
You could even make it backwards compatible by adding
```
Formula_1i del;
if (formula_1i.TryGetValue(name, out del))
  result = del(input);
```
To the `ApplyAlterations1I()` method.

TheLacus:
Note that mod load order is supposed to be like Skyrim, where mods at lower positions override what comes before them. 

[...]

PR #1642, opened 2019-12-14
Improved support for formula overriding

TheLacus:
All formulas now are stored in a single dictionary and casted to appropriate type when retrieved. Also converted named delegates to generic `System.Func`, which allows to always use the exact signature without making compromises, and enforced load order.

```cs
FormulaHelper.RegisterOverride<Func<int, int, int>>(mod, "CalculatePlayerLevel", (startingLevelUpSkillsSum, currentLevelUpSkillsSum) =>
{
    return (int)Mathf.Floor((currentLevelUpSkillsSum - startingLevelUpSkillsSum + 28) / 15);
});
```

### DIFF
--- a/Assets/Scripts/External/Func.cs
+++ b/Assets/Scripts/External/Func.cs
@@ -1,0 +1,6 @@
+namespace System
+{
+    // Func delegates with more than four arguments are only available on .NET Framework 4.0 and above
+    // Here's a substitute while we are on .NET Framework 3.5
+    public delegate TResult Func<in T1, in T2, in T3, in T4, in T5, out TResult>(T1 arg1, T2 arg2, T3 arg3, T4 arg4, T5 arg5);
+}
--- a/Assets/Scripts/External/Func.cs.meta
+++ b/Assets/Scripts/External/Func.cs.meta
@@ -1,0 +1,11 @@
+fileFormatVersion: 2
+guid: 5ca3921a72f5fc44c8a0c9f5ab143a61
+MonoImporter:
+  externalObjects: {}
+  serializedVersion: 2
+  defaultReferences: []
+  executionOrder: 0
+  icon: {instanceID: 0}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
--- a/Assets/Scripts/Game/Formulas/FormulaHelper.cs
+++ b/Assets/Scripts/Game/Formulas/FormulaHelper.cs
@@ -4,9 +4,7 @@
 // License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
 // Source Code:     https://github.com/Interkarma/daggerfall-unity
 // Original Author: Gavin Clayton (interkarma@dfworkshop.net)
-// Contributors:    Hazelnut
-//                  ifkopifko
-//                  Numidium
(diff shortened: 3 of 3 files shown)


TheLacus:
I remembered there was also issue #1422, which I left to you as you made the formula overriding.

The suggestion was to keep a list of delegates and allow them to reject a call depending on arguments, and let the following one produce a value. I though about this and made two conclusions:
1. I think this can be nicely implemented with nullable types (`Func<int?>` instead of `Func<int>`). Delegates can return `null` if they don't want to generate a result. 
2. This proposal has value but I personally don't think it's worth the added performance overhead, which is potentially high...

ajrb:
Yeah, I've been leaving #1422 to see if there's actually any demand for this feature. There doesn't appear to be any so my view is the added complexity makes it not worth doing. If two mod authors want to change the same formula then they'll need to collaborate. I think this auto stuff would potentially lead to bugs based on execution (mod) order which is player controlled.

Will review this PR now. Thanks for doing this.

Assets/Scripts/External/Func.cs
```diff
@@ -0,0 +1,6 @@
+namespace System
+{
+    // Func delegates with more than four arguments are only available on .NET Framework 4.0 and above
+    // Here's a substitute while we are on .NET Framework 3.5
+    public delegate TResult Func<in T1, in T2, in T3, in T4, in T5, out TResult>(T1 arg1, T2 arg2, T3 arg3, T4 arg4, T5 arg5);
```
ajrb: So would additional delegates be required for 6 argument formula if they were added?

ajrb:
Is there a way to register an override using a named method rather than anonymous function?

I will try it but am not sure if it's possible or not. I would like to continue to keep the registration statement code separate from the formula code in my mods.

Assets/Scripts/External/Func.cs
TheLacus: Yes, you can define a new delegate like this one. System namespace from new .NET versions include [variations for up to 16 arguments](https://docs.microsoft.com/en-us/dotnet/api/system.func-17?view=netframework-4.8)...

TheLacus:
> 
> 
> Is there a way to register an override using a named method rather than anonymous function?
> 
> I will try it but am not sure if it's possible or not. I would like to continue to keep the registration statement code separate from the formula code in my mods.

Yes but the compiler needs to be able to resolve to correct type. Example above also works with  a method:

```cs
FormulaHelper.RegisterOverride<Func<int, int, int>>(mod, "CalculatePlayerLevel", CalculatePlayerLevel);
```

or

```cs
FormulaHelper.RegisterOverride(mod, "CalculatePlayerLevel", (Func<int, int, int>)CalculatePlayerLevel);
```

ajrb:
Oh I like  that second variant, much better. Thanks, am testing Archs guild override now.

ajrb:
Works beautifully! Thanks so much for this, I was not happy with what I'd done here but didn't know how to improve it. At some point I should look for any functions that had their args changed to fit into the signatures I created and refine them.

ajrb:
@TheLacus Are you happy for me to merge this now, or would you like to give other devs time to review first?

PR #1643, opened 2019-12-14
fix classic load by swapping free action and climbing effects

petchema:
Attachment: classic save with custom spells using all effects

Only other difference is missing effects (Shield, Dispell, Comprehend languages)

[SAVE4.zip](https://github.com/Interkarma/daggerfall-unity/files/3963960/SAVE4.zip)

### DIFF
--- a/Assets/Scripts/Game/MagicAndEffects/Effects/Alteration/Climbing.cs
+++ b/Assets/Scripts/Game/MagicAndEffects/Effects/Alteration/Climbing.cs
@@ -24,7 +24,7 @@ namespace DaggerfallWorkshop.Game.MagicAndEffects.MagicEffects
         public override void SetProperties()
         {
             properties.Key = EffectKey;
-            properties.ClassicKey = MakeClassicKey(26, 255);
+            properties.ClassicKey = MakeClassicKey(28, 255);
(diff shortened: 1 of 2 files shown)


Interkarma:
Did I really muck up these keys? How embarrassing.

Interkarma:
OK, so I didn't muck the keys up - I didn't verify that source information was correct for everything. The below (my source for keys) table has those effects swapped.

https://en.uesp.net/wiki/Daggerfall:SPELLS.STD_indices

>Only other difference is missing effects (Shield, Dispell, Comprehend languages)

Can you elaborate on this one? All of those effect have been implemented, and I can see they import fine using your test save. "shadow-shield-silence" has those effects, "detect-dispell-ele" has those effects, and "lang-item-cure" has those effects. I wasn't able to see anything missing or not importing other than the swapped effects you fixed here.

petchema:
> > Only other difference is missing effects (Shield, Dispell, Comprehend languages)
> 
> Can you elaborate on this one? All of those effect have been implemented, and I can see they import fine using your test save. "shadow-shield-silence" has those effects, "detect-dispell-ele" has those effects, and "lang-item-cure" has those effects. I wasn't able to see anything missing or not importing other than the swapped effects you fixed here.

I checked again, and it worked fine for me too, I'm not sure what happened

PR #1644, opened 2019-12-14
Climbing skill check, item damage, and weapon anim time - migrated to formula and exposed to mods.

ajrb:
These formula use the new override system implemented by @TheLacus - it's brilliant!! :-)

* Added skill check when initiating climbing and tuned the default success percentages.

See https://forums.dfworkshop.net/viewtopic.php?f=5&p=36243 for discussion. Happy to tweak the numbers more before merging this one since I can always mod them in R&R. @Interkarma going to take your lead on this once you've tried out in game. You may want default to be more forgiving for that 'fun' factor!

* Moved frame time to formula and optimised FPSWeapon so it doesn't calculate constantly.

* Moved item condition damage calculations to formula.

* Moved the is item stackable checks into a formula.

Apologies for combining, but it's much easier for my mod development to have all these changes in the same branch rather than separated.


Interkarma:
Excellent work as always mate. No problems found on my side. I'll merge this one for today's builds.

PR #1645, opened 2019-12-15
Weapon sounds when hitting walls

petchema:
Add clinging sound when you hit walls with a weapon
Limitations: 
- uses Parry6, I would have preferred a totally distinctive sound instead.
- works for all static objects, so hitting wooden furniture does a clinging sound too, even if one could expect the same sound as doors bashing.
- On the other hand non-static objects like elevators do no sound. I could have checked if the gameObject has a DaggerfallMesh component instead, but I don't know if that would have been okay

Forums: https://forums.dfworkshop.net/viewtopic.php?f=4&t=3172


Interkarma:
I like this a lot! Nice and simple change too.

PR #1646, opened 2019-12-15
bash-closing doors

petchema:
(Splitting #1645 in two separate PRs)

Sometimes enemies hide behind or "inside" opened doors, and player weapons miss them because they're bashing the door instead.

While trying to add door bashing sound in this case, I checked classic Daggerfall behavior and noticed that it does bash-closing of the door (usually; sometimes it gets confused and bash-opens already opened doors), and that attracts enemies attention just like bash-opening.


Interkarma:
Works well and indeed matches classic. Thank you. :)

PR #1647, opened 2019-12-16
For numeric input, accept numeric keys even without shifting

petchema:
In some locales, numeric keys normally require shifting to get digits.
Bypass this requirement for numeric inputs.

Added difficulty: on french keyboard, the 6 key is the shifted dash (-) key, so there's an ambiguity for relative numbers. So I restricted this to NATURAL numeric inputs.

Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&t=3192


Interkarma:
Thank you for this. I wouldn't catch many of these issue on this side by myself.

PR #1649, opened 2019-12-19
Corpse cleanup for player owned places

ajrb:
This was quite a mission doing this. See thread here for my journey: https://forums.dfworkshop.net/viewtopic.php?f=20&t=3213

The fix I made to StreamingWorld.InitPlayerTerrain() may break other stuff. I'm really not sure about it so would be good to have opinions from others.

### DIFF
--- a/Assets/Scripts/Game/Serialization/SerializableStateManager.cs
+++ b/Assets/Scripts/Game/Serialization/SerializableStateManager.cs
@@ -124,11 +124,22 @@ namespace DaggerfallWorkshop.Game.Serialization
             }
             else
             {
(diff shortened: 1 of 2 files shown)


ajrb:
Did some more testing and everything seems to work as expected. The only issue I found was recalling out of somewhere to outside gets rid of loot piles when exiting using the door doesn't. Generally players recall larger distances so I don't think it's a problem.

Interkarma:
No problems found on this end either. Good work sorting this. I'm sorry it was such a mission. Incredible how often the "easy" changes turn out to be much deeper than expected.

PR #1651, opened 2019-12-19
Classic quest fixes, Dec 2019

JayH2971:
Much of this is enabling working quests and fixing the Questlist, which I left in a brutal state.


Interkarma:
I'll roll this into today's builds so people will have access to the newly available quests and fixes. Thank you as always for all your work on this.

PR #1652, opened 2019-12-19
Add location index to ensure nameseed is unique

ajrb:
See https://forums.dfworkshop.net/viewtopic.php?f=5&t=3048


Interkarma:
Looks like a good solution without adding any complexity. I don't think this will break anything else, seems to be a very low impact fix. Will merge this one now.

PR #1654, opened 2019-12-22
<---> separator bug in S00000018

petchema:
Makes reading the letter a 50% chance of reading one line


petchema:
@JayH2971 ?

Interkarma:
Agreed this looks like a bug. Good catch.

PR #1655, opened 2019-12-22
Don't suppress world when inside a building

df-ferital:
Fix list of houses for sale not available when loading the game inside a bank: https://forums.dfworkshop.net/viewtopic.php?f=24&t=3034


Interkarma:
Perfect, thank you. No issues found in testing, and it should be a harmless change. I'll roll this into new builds now.

PR #1657, opened 2019-12-26
adjust door thickness

petchema:
Improvement over 5b1d37a4b992f61cad1d2400fbca2ca0f218f312

It needs more testing, are there examples of cases that required the fix above?

Forums: https://forums.dfworkshop.net/viewtopic.php?f=27&p=37823#p37823

### DIFF
--- a/Assets/Scripts/Internal/DaggerfallStaticDoors.cs
+++ b/Assets/Scripts/Internal/DaggerfallStaticDoors.cs
@@ -82,11 +82,6 @@ namespace DaggerfallWorkshop
                 go.transform.position += transform.position;
                 go.transform.rotation = transform.rotation;
 
-                // Sometimes doors injected from mods don't trigger because hit point is not exactly
-                // on the plane calculated from classic door data so we ensure the collider has some depth
-                if (c.size.x < 1)
-                    c.size = new Vector3(1, c.size.y, c.size.z);
-
                 // Check if hit was inside trigger
                 if (c.bounds.Contains(point))
                 {
--- a/Assets/Scripts/Utility/GameObjectHelper.cs
+++ b/Assets/Scripts/Utility/GameObjectHelper.cs
@@ -1376,7 +1376,10 @@ namespace DaggerfallWorkshop.Utility
                 Vector3 v2 = door.Vert2;
 
                 // Get door size
-                const float thickness = 0.05f;
+
+                // Sometimes doors injected from mods don't trigger because hit point is not exactly
+                // on the plane calculated from classic door data so we ensure the collider has some depth
+                const float thickness = 1f;
                 Vector3 size = new Vector3(v2.x - v0.x, v2.y - v0.y, v2.z - v0.z) + door.Normal * thickness;
 
                 // Add door to array
--- a/Assets/Scripts/Utility/GameObjectHelper.cs
+++ b/Assets/Scripts/Utility/GameObjectHelper.cs
@@ -1379,7 +1379,7 @@ namespace DaggerfallWorkshop.Utility
 
                 // Sometimes doors injected from mods don't trigger because hit point is not exactly
                 // on the plane calculated from classic door data so we ensure the collider has some depth
-                const float thickness = 1f;
+                const float thickness = 0.3f;
                 Vector3 size = new Vector3(v2.x - v0.x, v2.y - v0.y, v2.z - v0.z) + door.Normal * thickness;
 
                 // Add door to array


Interkarma:
The approach here is improved as it builds on how DaggerfallStaticDoors orients collider. But I'm not sure the collider needs to be 20x thicker, especially in the default game without mods. Why not 0.1 or 0.2?

Reason being there's a quirk of colliders where if ray originates from inside volume (e.g. player pressed right up against door) then raycast will not land. I reduced default thickness to 0.05 to avoid this, and I don't want to risk re-introducing. The smallest required thickness is best here.

@TheLacus Wasn't there a process where building mods could add custom static door colliders as needed? I recall that from a while back but may be misremembering.

petchema:
After more testing, for the case mentioned in the forums (Garenor Hall), door starts working with a thickness of exactly 0.3

TheLacus:
> @TheLacus Wasn't there a process where building mods could add custom static door colliders as needed? I recall that from a while back but may be misremembering.

Yes, mod authors on forums thought it was limitating to place doors at the same position of original building models, so i started working on [a component](https://github.com/Interkarma/daggerfall-unity/blob/master/Assets/Scripts/Utility/AssetInjection/Components/CustomDoor.cs) to define doors at any position. Nothing came out of that on the forums, but the component is functional.

The situation here is a bit different. A mod replaces the door model, to be used inside dungeons without altering other dungeon meshes. If increasing trigger size is inconvenient, the component above could be adapted to work with dungeons but i don't know when i'll be able to do it. @petchema is this something you could be interested with?

petchema:
That's way beyond my current knowledge of Unity, so I doubt it'll be finished sooner if I give it a try...

Interkarma:
I think increasing trigger size is a fine compromise here. Thank you @petchema for finding a more minimal fitting value for new collider thickness. :)

I have a few edge cases saved around doors from when I built this originally. I'll run it through those when I can and merge if no issues found.

HybOj:
Please - if possible - add this to next release. Since I introduced this module to the DREAM mod and released it to the public - and the bug reports started to rain - this makes it hard to play, because ppl cant exit the dungeons. Its quite a serious issue :X  Thank you guys! KoW

Interkarma:
Will regression test my old edge cases against this change today. I'll make sure this one gets priority in next builds. I'm hoping to have that release ready in another day or so.

Interkarma:
Zero issues found on my end. Merging this now.

petchema:
I have issue _entering_ some dungeons like Jubilon Cloister, Sentinel without mods
[SAVE498.zip](https://github.com/Interkarma/daggerfall-unity/files/4022751/SAVE498.zip)

petchema:
Those doors being inclined, and the box collider straight, they may very well have been the reason for @TheLacus workaround?
dungeonexternaldoor

Interkarma:
That's the kind of thing yes. Thanks for finding another case for us. I'll see what we can do here.

petchema:
Exiting Shedungent seems problematic with DREAM mod, you can still fall thru the door :(

petchema:
Yes, Shedungent exit is fine when not modded, but doesn't work when modded: you can't click it, and you can walk thru and fall in the Void...
[SAVE517.zip](https://github.com/Interkarma/daggerfall-unity/files/4024159/SAVE517.zip)

petchema:
Left exit door of Castle Sentinel broken, again you can't click it and you fall into the Void is you walk thru it. Right door works fine.
[SAVE523.zip](https://github.com/Interkarma/daggerfall-unity/files/4024206/SAVE523.zip)

Interkarma:
>Left exit door of Castle Sentinel broken, again you can't click it and you fall into the Void is you walk thru it. Right door works fine.

I don't experience this without mods, and seems unrelated to the issue of static door collider placement.

If you inspect modded left door in editor, does it have a BoxCollider or MeshCollider attached and active? If player can walk through exit quad replacement, I'd suggest it's missing a collider in DREAM or as part of replacement process.

The actual static door collider volume will remain the same, but player click needs to land on some geometry inside that volume. If the geometry isn't collidable then that click can't register and player will be able to pass through it.

petchema:
> > Left exit door of Castle Sentinel broken, again you can't click it and you fall into the Void is you walk thru it. Right door works fine.
> 
> I don't experience this without mods, and seems unrelated to the issue of static door collider placement.

Sorry for the confusion, yes like for Shedungent door it works fine when not modded. The problem mentioned happens with DREAM 2020R1 mod.

> If you inspect modded left door in editor, does it have a BoxCollider or MeshCollider attached and active? If player can walk through exit quad replacement, I'd suggest it's missing a collider in DREAM or as part of replacement process.

Will try, but I'm not sure how to run DFU in the editor with mods loaded.
Anyway, why would one door misbehave but not the other?

Interkarma:
>Will try, but I'm not sure how to run DFU in the editor with mods loaded.
>Anyway, why would one door misbehave but not the other?

That's OK will check in a bit and report back. This is most likely something specific to the mod or to the injection method.

petchema:
The left door hides a hole in the wall, the right one doesn't, that is why. Bummer.
castle sentinel doors

Could a collider be added if one is not found? Or do you prefer not to

Interkarma:
Thanks for checking so quickly. The replacement model injected from DREAM should have a collider attached. The exit quad always need to be collidable for clicks and so player can't walk through it.

I'm not sure though if this is something missed in the DREAM prefab for this model, or as part of the replacement process.

petchema:
@TheLacus  @HybOj how can you go about adding a collider to the door replacement model?

HybOj:
Hi I was given this module by Lacus, I have no idea on how he created it or
how it works, Im sorry. If there is a need to add a collider, than I will
add it to next release, if someone sends the updated package to me.

Here is the package as I have received it month or so back:

po 6. 1. 2020 v 1:24 odesílatel Pierre Etchemaïté <notifications@github.com>
napsal:

Interkarma:
I feel the problem of door trigger dimension is nicely solved at this point, but it's a separate system from adding colliders to replacement geometry. I'll await for more information from @TheLacus about the process of replacing these exit quads so they have a collider. It might just be a simple matter of KoW adding a component to the prefab in DREAM (assuming a prefab is the source).

I'll wait another day before releasing new builds in case we have a breakthrough. :)

petchema:
I imported unitypackage from https://forums.dfworkshop.net/viewtopic.php?f=23&t=2043&start=20#p33185
Then tried adding a MeshCollider to it, but can still fall thru modded exits :(
DungeonExitCollider

TheLacus:
I gave assets to KoW several weeks ago so i don't remeber the details, but it worked with the doors i tested; apparently i missed some issues so thank you for this fix! KoW should be able to add collider to prefab if missing (use Add Component button and add as in screenshot from pango).

I'm sorry if i'm not much of help but I have very little time these days. If you don't feel investing more time in it, i think a quick update for Dream can remove door and i'll investigate at a later time.

Interkarma:
>Then tried adding a MeshCollider to it, but can still fall thru modded exits :(

Is collider present on replacement model at runtime as well?

Interkarma:
>I'm sorry if i'm not much of help but I have very little time these days. If you don't feel investing more time on it, i think a quick update for DREAM can remove door and i'll investigate at a later time.

That's no problem at all. Thank you for your help so far. :)

Interkarma:
>Is collider present on replacement model at runtime as well?

Just answering my own question here. I can see the MeshCollider does indeed flow through to runtime. The problem appears to be the mesh is loaded after the collider, so it ends up with no collision shape.

For @HybOj, here's a solution for next release of DREAM. Setup your 70300 prefab like below using a tightly fitting BoxCollider instead. This will ensure the collision geometry is always present for this prefab no matter load order. This solves both clicking on the object and walking through it.

image

HybOj:
Ok I did that :)

út 7. 1. 2020 v 4:16 odesílatel Gavin Clayton <notifications@github.com>
napsal:

HybOj:
And thanks everyone for working on this issue, I really appreciate that!!

PR #1660, opened 2019-12-27
Year-end quest fixes

JayH2971:
All witch quests tested and working


bansheexyz:
Rather than replace "%g" with "the one", maybe there is simply a bug here in macros that needs to be resolved. I've tested classic and it always resolves the gender properly. "=man_" always results in a valid male description like "blond aristocrat" or "well-to-do merchant". It never shows up as something feminine like it does in DFU. Furthermore, I don't know how DFU is setting %g to "she" all the time, but then later %g2s are properly set. Is it because %g prints before =man_ in the paragraph?

JayH2971:
%g is probably reading from =questgiver, which would make the person female unnecessarily. I can only guess that classic omits that and does a pre-read of the quest NPCs first, and assigns %g to the last referenced NPC in that list.
That's all blind dart shooting, though, and 90% my imagination.

Interkarma:
Thank you Jay! :)

PR #1661, opened 2020-01-02
find revealed locations by a very common substring

petchema:
that's a consequence of search running in two passes: finding locations whose name best match the given string, then filter the list to only keep already revealed locations. And also the list produced by the first pass having a finite size.

They're screens upon screens of locations matching "Ruins..." (or "Ruins of..." for that matter) in the Wrothgarian Mountains, so only a subset of the names are kept by the first pass; And since the revealed dungeon name is not in the intermediate list (by pure bad luck), the second pass doesn't keep any names.

Simplest fix is just to increase the size of the intermediate list. The first pass creates the list using a priority queue, so it should stay very fast.
Other possibilities include swapping the two passes (search among the name of the locations that are revealed) but I don't know how fast we can create the list of all revealed locations; or add a predicate parameter to the function filtering names, so we can check if the location is revealed just before it's put in the intermediate list.

Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=3292


Interkarma:
I think a regression from this change - sometimes when searching for a location (e.g. "daggerf") in Daggerfall will drop back to an empty search box without any results. Doesn't always happen, no errors in logs. I'll update if I manage to pin it down any further.

PR #1662, opened 2020-01-04
complete all archives containing emissive textures, not only lights

petchema:
Had to replace test on isEmissive with resultEmissive l.285 because we can have one but not the other, I'm not sure it's to be expected.

Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=2917


Interkarma:
I can't see any issues with using resultEmissive here. No problems found in testing. Will merge this one now. Definitely a better solution than my quick, thank you Pango! :)

PR #1663, opened 2020-01-05
fix quicksave overwriting save0 if used first

petchema:
Reading the list of existing save slots only happens when opening save window interface; So if quicksave is used first, it will see an empty list of slots and overwrite save0.

Fixed by reading folders once at initialization, alternative would be to check if it needs to be done in several places (QuickSave, HasQuickSave, QuickLoad, at least)

Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&t=3303


petchema:
Seems insufficient; I created a new character, opened the Save interface, manually named my save "QuickSave" and that also overwrote save0

Interkarma:
That's a good fix thank you, I wasn't aware this was happening. Will merge this one now.

PR #1665, opened 2020-01-14
exclude player from weapon hit checks

petchema:
Seems to help hitting enemies at very close range

Forums: https://forums.dfworkshop.net/viewtopic.php?f=4&t=3172


Interkarma:
Thank you! :)

PR #1666, opened 2020-01-16
Stop headbobbing animation while player is not grounded

jefetienne:
One very small adjustment I wanted to make was the fact that the headbobbing animation, if enabled, would continue to play even if the player is in the air. I found it more noticeable if you are jumping while riding on a horse, and the camera still continues to shake. This simple added condition to check for if the player is grounded or not should fix this issue.


petchema:
Noticed that last night too

Interkarma:
Thank you! :)

Issue #1588, opened 2019-10-28
Forward momentum not retained on sprinting jumps


drakovyrn:
Jumping forward while walking retains the same speed while doing so. However, jumping forward while sprinting does not.

The first sprinting jump is at the same speed as sprinting, but every consecutive jump afterward is done at walking speed.

Easier to notice with a character with a high jumping skill. Very noticeable if set_runspeed = 30.

Current pattern of jumps:
Walking: Short, Short, Short, Short, etc.
Running: Long, Short, Short, Short, etc.

Expected pattern of Jumps:
Walking: Short, Short, Short, Short, etc.
Running: Long, Long, Long, Long, etc.

OS: Windows 7
DFUnity Version: 0.10.10

Not save-dependent, but here's a decently running skilled character in the Wayrest castle courtyard with plenty of space to run around and test this behavior:

[SAVE1.zip](https://github.com/Interkarma/daggerfall-unity/files/3777178/SAVE1.zip)

petchema:
How does it work in classic?
Does it help with "platforming" cases, like in Mantellan Crux?

drakovyrn:
It works as the "expected" pattern.

Specifically, the character will long jump whenever the sprint button is held.

It does help in platforming cases, but also getting away from enemies. Currently, the jump only long jumps the first time, and then goes back to short jumps.

bansheexyz:
The OP is talking about "Bunnyhopping". It's this ultra-unrealistic mechanic that makes spamjumping just as fast as sprinting. So everyone starts bunnyhopping exclusively to move because it makes you way harder to hit and there's no downside. It's also disorienting and unpleasant to watch. Imagine every let's play of an first-person game looking like this: https://www.youtube.com/watch?v=Lw5lKtvpv4g

I used to play a lot of HL1 deathmatch, so I know what it is. The mechanic died with HL2 and Counter-Strike, but people still have some kind of misguided reverence towards it.

drakovyrn:
You're comment is entirely misguided and unnecessary. It is already
possible to bunny hop in original Daggerfall and in Daggerfall Unity.

I am not asking to implement bunny-hopping (which is already implemented),
I'm asking to replicate vanilla behavior regarding forward speed while
holding sprint while jumping. So your point is moot and unnecessary here.

drakovyrn:
Needlessly angry point still irrelevant, as bhopping is already
implemented, and what I'm asking is for the behavior to work consistently,
[...]

PR #1667, opened 2020-01-17
Speedchanger refinements

petchema:
FIx for #1588, and then some (like lowering player's noise volume when sneaking).

Note: 
playerMotor.speed is determined by both world conditions and player's input. While it's only required during FixedUpdate() for modifying movement(*), player's input can change with each Update() so that input must be recorded separately.
*. Except in LevitateMotor, that can adjust movement with each frame. If we care, we could update speed with each Update() too.


Interkarma:
Works well and no problems found. Thank you. :)

PR #1669, opened 2020-01-19
emissive maps: support more lights

petchema:
Fix for forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=3289


petchema:
As a side note (this is not something modified by that patch), Fire god chamber looks wrong without emissive walls, but has strong stroboscopic effect with emissive walls. I don't know how to fix that.

PR #1670, opened 2020-01-19
Degrade equipped "when held" magic items over time

numidium:
I used @petchema 's findings here: https://forums.dfworkshop.net/viewtopic.php?f=24&t=2584&start=40

I tried using the "%" operator on the current game second but that was giving me race condition issues. I opted to use randomness since it wouldn't need to use the current game time. If introducing randomness to the mix doesn't feel right then I could rework the magic round loop.

Still need to find out how much to degrade on equip.


Interkarma:
This looks good, thank you! :)

petchema:
It would be great if periodic effect reroll could be implemented too, I think it's also an important aspect of "when held" magic...

PR #1671, opened 2020-01-20
Allow long unused textures to be unloaded

TheLacus:
Added experimental setting "AssetCacheThreshold". If different than zero, this is the time in minutes from last access for which a texture can be stored in cache. During fast travels textures that exceed this value are removed to allow Unity to unload them if unused.

### DIFF
--- a/Assets/Scripts/MaterialReader.cs
+++ b/Assets/Scripts/MaterialReader.cs
@@ -17,6 +17,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using DaggerfallConnect;
 using DaggerfallConnect.Utility;
 using DaggerfallConnect.Arena2;
@@ -99,6 +100,7 @@ namespace DaggerfallWorkshop
         public bool MipMaps = true;
         public bool ReadableTextures = false;
         public SupportedAlphaTextureFormats AlphaTextureFormat = SupportedAlphaTextureFormats.ARGB32;
+        public int AssetCacheThreshold;
 
         // Window settings
         public Color DayWindowColor = new Color32(89, 154, 178, 0xff);
@@ -347,7 +349,7 @@ namespace DaggerfallWorkshop
             int key = MakeTextureKey((short)archive, (byte)record, (byte)frame);
             if (materialDict.ContainsKey(key))
             {
-                CachedMaterial cm = materialDict[key];
+                CachedMaterial cm = GetMaterialFromCache(key);
                 rectOut = cm.singleRect;
                 return cm.material;
             }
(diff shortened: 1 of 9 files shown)


Assets/Scripts/MaterialReader.cs
```diff
@@ -864,6 +866,21 @@ public void ChangeWindowEmissionColor(Material material, WindowStyle windowStyle
             }
         }
 
+        /// <summary>
+        /// Removes from cache all materials that have not been accessed from the time in minutes defined
+        /// by <see cref="AssetCacheThreshold"/>. If equals to <c>0</c> assets are never removed from cache.
+        /// </summary>
+        internal void ValidateCache()
```
petchema: Rename as PruneCache()?

Assets/Scripts/MaterialReader.cs
```diff
@@ -864,6 +866,21 @@ public void ChangeWindowEmissionColor(Material material, WindowStyle windowStyle
...
+        {
+            if (AssetCacheThreshold == 0)
+                return;
+
+            float time = Time.realtimeSinceStartup;
+            float threshold = AssetCacheThreshold * 60;
+            foreach (var item in materialDict.Where(x => time - x.Value.timeStamp > threshold).ToList())
+                materialDict.Remove(item.Key);
+        }
```
petchema: Can you add some log, say "x materials expired"?

Assets/Scripts/MaterialReader.cs
```diff
@@ -864,6 +866,21 @@ public void ChangeWindowEmissionColor(Material material, WindowStyle windowStyle
...
+        /// <summary>
+        /// Removes from cache all materials that have not been accessed from the time in minutes defined
+        /// by <see cref="AssetCacheThreshold"/>. If equals to <c>0</c> assets are never removed from cache.
+        /// </summary>
+        internal void ValidateCache()
+        {
+            if (AssetCacheThreshold == 0)
+                return;
+
+            float time = Time.realtimeSinceStartup;
```
petchema: For a general purpose cache in a general purpose program, I would have used a virtual clock, that only progresses when the program itself makes progress; So that for machines that are, say, slow to load resources, the cache pruning is not also more aggressive (relatively speaking).
I'm not sure it would make any difference in the case of a game, if the program is slowed down by minutes there's probably other more important issues...

TheLacus: The purpose of removing assets from cache here is to free memory (especially vram); I don't think game performance should affect this behaviour. If you have low memory you want the cache not to take too much space, even if it means longer loading times. If you have more memory than what is normally used by the game, you can increse the threshold.

petchema: The problem I was thinking about is loaded assets that you benefit for less effective time, because they (or other assets) took time to load; That feels kinda silly

Assets/Scripts/MaterialReader.cs
TheLacus: I don't think this would be valuable information, the log is already a bit "spammy" at times...

Assets/Scripts/MaterialReader.cs
TheLacus: Now i get what you mean. I don't think this is really an issue here  but i appreciate that you pointed that out.

Interkarma:
I'm happy with this. Thank you. :)

PR #1672, opened 2020-01-24
Use correct indices for animations

numidium:
Addresses the issue discussed in https://forums.dfworkshop.net/viewtopic.php?f=24&t=3316


Interkarma:
Thanks for the fix! :)

PR #1673, opened 2020-01-24
fix static doors collider size

petchema:
In addition to using the max of width and depth as door thickness, swap Y and Z if Y size is less than Z size.

It seems to work fine (more testing welcome though), but I suspect the logic is not perfectly correct; I suspect Y and Z actually need to be swapped for grounded doors because you're actually entering by its top. As if door dimensions were considered _after_ rotations instead of before; They're really its X, Y and Z sizes, and not local width, height and depth. Or something like that. 
(and that could explain why width and depth can switch roles, probably depending on door orientation).

Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=566


Interkarma:
Thank you for this, a nice refinement. :)

In the case of the door laying along ground, virtual collider is slightly larger than target model. But I'm not concerned about getting it perfect. Player still needs to click on collidable geometry inside the virtual collider area. Provided the sizes remain reasonable around the door area, it won't present much problem in practice. These volumes aren't used for physics simulation or anything.

PR #1674, opened 2020-01-25
Display launcher options if a key is pressed during startup

petchema:
Should make it easier for players to get the options back than find and edit settings.ini.

This does not automatically turn alwayShowOptions back on, so it can also be used for one-time display of options.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallUnitySetupGameWizard.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallUnitySetupGameWizard.cs
@@ -149,7 +149,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             NativePanel.Components.Add(exitButton);
 
             // If actually validated and we just want to see settings then move direct to settings page
-            if (DaggerfallUnity.Instance.IsPathValidated && DaggerfallUnity.Settings.ShowOptionsAtStart)
+            if (DaggerfallUnity.Instance.IsPathValidated && (DaggerfallUnity.Settings.ShowOptionsAtStart || Input.anyKey))
             {
                 currentStage = SetupStages.Options - 1;
             }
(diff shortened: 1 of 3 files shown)


ajrb:
Great idea, nice.

ernierasta:
@petchema you are on fire! You are doing great job.

Interkarma:
I really love the idea, but I find it hard to use in practice. Always takes me a few attempts to make it work. It's not really clear if a key press or a key hold makes it work, or even when I need to press the key. Sometimes I have to spam keys rapidly just to catch it. Could be because I have a fast SSD and game is starting too quickly. But we don't want to encourage users to be spamming keys as it leads to a few other issues:
If user hits the Enter key a bit too soon on Windows while starting game, then they will launch a second instance of DFU, which is a bit confusing.
If user is spamming keys at the opening stage as I was, they can flip to a launcher menu like Advanced thanks to the new keyboard shortcuts. This gave me a brief feeling of "oh crap, what did I just change then?" And if they hit the P key again at the wrong time, it just starts the game anyway.
Rather than any key, I'd prefer if this was a nice stateful condition like "hold shift when starting game to show launcher again". Then check if either left or right right shift is held down. Any key is just inviting trouble. :)

petchema:
Yes, in practice I'm using some state key (Shift, Control, Alt, etc.) that I keep pressed during startup; and that's probably how it should be documented ("keep Shift key pressed [...]").
I'm not sure which ones can be safely kept pressed from even before launching DFU, if that helps with very fast setups (on mine it takes over 30s for the launcher to appear, so I can't really tell ;) ). Could depend on the platform too, 

However I'm not sure the code needs to be that precise; it adds complexity with not much benefit.

Interkarma:
I'm happy with guiding users to use a state key. Thanks :)

My system takes about 1 second before the "bethsoft presents" video starts, and by then it's too late. The time window to register key is probably much less. Even then just holding down a key can be iffy as I don't seem to time it just right. I need to hit that sweet spot between game actually starting up but before it reaches game scene. I only seem to have a few hundred ms and it's hard in practice to get right.

But I still think this is a good feature that will help more people than it confuses. And they can always just edit the .ini if they have trouble with the keypress. I'll merge it now and we'll see how it goes.

petchema:
For the record I found out why DFU has been slow to load on my rig for the last few months (thanks strace -e file)
I had created a symlink to access classic Daggerfall main directory from my documents directory.
But that, and symlinks created by Wine that pointed back from Wine's virtual environment to my document directory created some kind of loop that DFU directories scanning jumped into, to the limit of file paths, so it took 30s to complete!

PR #1675, opened 2020-01-25
Prevent backward timetravel issues.

JorisVanEijden:
Fix for https://forums.dfworkshop.net/viewtopic.php?f=24&t=3309
0.10.17 - Vampire fast travel DFU hangs CUREVAM if target dungeon not present in region

The dungeon had nothing to do with it.
It was the fast travel going back 40 minutes in time. That made the unsigned integer calculation for how many minutes had passed return a huge number. (136 years) which triggered creating the vampire cure quest 591 times, which quickly runs out of random dungeons to assign.


Interkarma:
This is a good catch. Thank you for the fix and extra hardening to time adjustments. :)

PR #1677, opened 2020-01-26
Disable soultrap after capturing soul

JorisVanEijden:
to prevent capturing multiple souls when killing an enemy by multiple damage sources.

Related form thread: https://forums.dfworkshop.net/viewtopic.php?f=24&t=3187


Interkarma:
Thank you for the fix. :)

PR #1678, opened 2020-01-26
Move racial paralysis immunity from saving throw to immunity check.

JorisVanEijden:
Forum thread: https://forums.dfworkshop.net/viewtopic.php?f=24&t=3157

This makes High Elves 100% resistant to paralysis, regardless of the source of the paralysis or magic weaknesses.


JorisVanEijden:
Ignore this for now. It allows taking paralysis weakness during character creation without taking the penalties.

JorisVanEijden:
Should be good now. Racial immunity doesn't work anymore if you've chosen to lower it.

Interkarma:
Superceded by https://github.com/Interkarma/daggerfall-unity/commit/9a4c8a2bff5b362e223b3531bc352712166342b7.

PR #1679, opened 2020-01-26
Less stroboscopic fire walls

petchema:
Lower the contrast of Fire god chamber fire walls, and slow down the animation, so it gets less aggressive on the eyes.


Interkarma:
Great work on this. It's the little things. :)

PR #1680, opened 2020-01-26
Changing the magicka in the get method causes bugs during load.

JorisVanEijden:
Logically the cap should only need to be applied when the MaxMagicka changes.
This is the same logic used for health and fatigue.

This fixes forum bug https://forums.dfworkshop.net/viewtopic.php?f=24&t=3075

This reverts commit e1df2c3c2bd72c08f0b8f9d5b7c5b413cbad9eb0

### DIFF
--- a/Assets/Scripts/Game/Entities/DaggerfallEntity.cs
+++ b/Assets/Scripts/Game/Entities/DaggerfallEntity.cs
@@ -388,6 +388,8 @@ namespace DaggerfallWorkshop.Game.Entity
         public void ChangeMaxMagickaModifier(int amount)
         {
             MaxMagickaModifier += amount;
+            if (currentMagicka > MaxMagicka)
+                currentMagicka = MaxMagicka;
         }
 
         public void SetIncreasedWeightAllowanceMultiplier(float amount)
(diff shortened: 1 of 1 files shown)


Interkarma:
>Logically the cap should only need to be applied when the MaxMagicka changes

No, something else could be changing player's current spellpoint value unrelated to changing MaxMagickaModifier. One is a quantity and the other is a cap modifier. They are only indirectly related to each other. The only universal place to enforce the ceiling is in the get method as implemented.

I'll think about the bug in question and use another solution. Thank you for your feedback here.

JorisVanEijden:
> No, something else could be changing player's current spellpoint value unrelated to changing MaxMagickaModifier.

This is already handled by `SetMagicka()` which the `CurrentMagicka` property uses.

> The only universal place to enforce the ceiling is in the get method as implemented.

I disagree. Health and Fatigue seem to do fine without changing values in a getter.
What is different about Magicka?

Interkarma:
>This is already handled by SetMagicka() which the CurrentMagicka property uses.

Thank you, I had missed that. My mental map is a bit out of date here. Happy to be wrong and look at changes at again. Will roll this into next round of builds. :)

PR #1681, opened 2020-01-26
Make constant effects go into effect immediately on load

JorisVanEijden:
Start constant effects when they are resumed on load. Otherwise things like increased magicka are capped before the new max goes into effect.

Related forum post: https://forums.dfworkshop.net/viewtopic.php?f=24&t=3075


Interkarma:
I don't want ConstantEffect() being triggered for every effect during restore. If an individual effect requires this, it can call this from Resume(). I'm happy for ExtraSpellPts to use this approach.

JorisVanEijden:
The problem with doing it in `Resume()` is that that is called before `RestoreSaveData()` so it might not have the correct data yet.
Looking at this I wonder if `RestoreSaveData()` shouldn't be moved from `DaggerfallWorkshop.Game.MagicAndEffects.EntityEffectManager.RestoreInstancedBundleSaveData()`  to `DaggerfallWorkshop.Game.MagicAndEffects.BaseEntityEffect.Resume()`

Or is there a specific need for resuming an effect before the effectSpecific data is loaded?

PR #1682, opened 2020-01-27
enemies fall damage

petchema:
While hard to reproduce, I've seen thieves bleed from a fall, so I assume fall damage exists for enemies too.

It's even harder to know what formula is used (beside decompilation maybe), for I've used the exact same formula as for player fall damage.

Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&t=3353#p39126


Interkarma:
This works well. You're right it's a bit hard to reproduce, so I resorted to spawning enemies against a ledge then smacking them off with a warhammer. :)

Issue #1676, opened 2020-01-25
Selecting "New Character" as the start game behavior gives nullpointer to biog file


Aelthien:
I assume that is because a default biog doesn't exist.

petchema:
Works for me

- How did you install classic Daggerfall?
- What version of Daggerfall Unity have you used?
- What choices have you selected while creation the character?

Aelthien:
> Works for me
> 
> * How did you install classic Daggerfall?
> * What version of Daggerfall Unity have you used?
> * What choices have you selected while creation the character?
> * Can you provide a game log?

This is running the project in Unity, past versions just let me choose the option and the game would work. But now do I need to make a biog before being able to run in Unity?

Readme says v0.10.x

`ullReferenceException: Object reference not set to an instance of an object
DaggerfallConnect.Arena2.BiogFile.ApplyEffects (System.Collections.Generic.List`1 effects, DaggerfallWorkshop.Game.Entity.PlayerEntity playerEntity) (at Assets/Scripts/API/BiogFile.cs:408)
DaggerfallWorkshop.Game.Utility.StartGameBehaviour.StartNewCharacter () (at Assets/Scripts/Game/Utility/StartGameBehaviour.cs:389)
DaggerfallWorkshop.Game.Utility.StartGameBehaviour.InvokeStartMethod () (at Assets/Scripts/Game/Utility/StartGameBehaviour.cs:144)
DaggerfallWorkshop.Game.Utility.StartGameBehaviour.Update () (at Assets/Scripts/Game/Utility/StartGameBehaviour.cs:117)
`

TheLacus:
So, this happens only when running inside the Unity Editor?

Aelthien:
> So, this happens only when running inside the Unity Editor?

Yeah.

PR #1683, opened 2020-01-29
Null check for list of effects when applying in the biog file

Aelthien:
If the list of effects is null the game will permanently lock up. So I simply added a check to see if the list is null and if so return the "ApplyEffects" function.


TheLacus:
When does this happen? There might be a deeper issue if the list is null.

Aelthien:
> When does this happen? There might be a deeper issue if the list is null.
> 
> EDIT. Issue #1676.

In the editor when you select NewCharacter as the start game behavior. It doesn't give the error if you create the character through character creation. Maybe there needs to be a way to set up a default character?

Interkarma:
I'm happy to have a null check here. Thank you. :)

PR #1684, opened 2020-01-29
Allow to get UIWindowType enum value from type

AsesinoBlade:
I was looking for a method to return the type of window that is currently up.   I created GetType in UIWindowFactory to achieve this functionality.

An example of use would be 

            var manager = (UserInterfaceManager)sender;
            if (UIWindowFactory.GetType(manager.TopWindow.GetType()) == UIWindowType.Trade) {
…

thanks 
a

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/UIWindowFactory.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/UIWindowFactory.cs
@@ -56,6 +56,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
         UseMagicItem,
         VidPlayer,
         WitchesCovenPopup,
+        NotFound,
     }
 
     public static class UIWindowFactory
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Game/UserInterfaceWindows/UIWindowFactory.cs
```diff
@@ -132,6 +134,16 @@ public static IUserInterfaceWindow GetInstanceWithArgs(UIWindowType windowType,
             return GetInstance(windowType, args);
         }
 
+        public static UIWindowType GetType(Type windowType)
+        {
+            foreach (var pair in uiWindowImplementations)
+            {
+                if (pair.Value == windowType)
+                    return pair.Key;
```
TheLacus: That's fine, but you can also use Linq for one line: 

```cs
var pair = uiWindowImplementations.FirstOrDefault(x => x. Value == windowType);
```

AsesinoBlade:
Thanks TheLacus, I made the changes you requested.  I hadn't used Linq before; it is very convenient :)

thanks
a

TheLacus:
Added check for default value (which is null for `Type` class). Sorry i wasn't completly clear, `FirstOrDefault` refers to `KeyValuePair` which is a struct so is never null. It's the value that is null if not found :)

AsesinoBlade:
> Added check for default value (which is null for `Type` class). Sorry i wasn't completely clear, `FirstOrDefault` refers to `KeyValuePair` which is a struct so is never null. The default value is `(default(UIWindowType), null)` so it's the value that is null if not found :)

Ah, thank you for explaining that.  I tested it by passing HUD into it and comparing against Trade.  It worked as expected so I assumed that it returned Null, but now I see why it passed.

I am really enjoying helping out; I am learning so much :)

Thanks again TheLacus!
A

PR #1685, opened 2020-02-02
Right-clicking clothing in inventory window changes its clothing variant

jefetienne:
This adjustment is both a quality of life improvement as well as a small bugfix.

Before this commit was made, if you click the "Use" button and then click an article of clothing, it will change its appearance. However, if you click an article of clothing that has a magical effect, it will not only change its variant but it will also incidentally cast the magic effect as well. This means it's only possible to change a magic clothing's appearance without casting only if you equip it and click it on the paper doll.

I made a nice little workaround and improvement to this by making it so if you right-click on the paper doll, or on the "local" and "remote" item panels, it will change its variant without casting a magic effect, and without having to click the "Use" button.

### DIFF
--- a/Assets/Scripts/Game/UserInterface/ItemListScroller.cs
+++ b/Assets/Scripts/Game/UserInterface/ItemListScroller.cs
@@ -126,6 +126,9 @@ namespace DaggerfallWorkshop.Game.UserInterface
         public delegate void OnItemClickHandler(DaggerfallUnityItem item);
         public event OnItemClickHandler OnItemClick;
 
+        public delegate void OnItemRightClickHandler(DaggerfallUnityItem item);
+        public event OnItemRightClickHandler OnItemRightClick;
+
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Game/UserInterface/ItemListScroller.cs
```diff
@@ -529,12 +533,25 @@ void ItemButton_OnMouseClick(BaseScreenComponent sender, Vector2 position)
 
             // Get item and raise item click event
             DaggerfallUnityItem item = items[index];
-            if (item != null && OnItemClick != null)
-                OnItemClick(item);
+
+            if(!rightClick && item != null && OnItemClick != null)
```
TheLacus: Please match the spacing with existing code (`if (`).

Interkarma:
This works very well, thanks for this change. :)

jefetienne:
You're welcome!

PR #1687, opened 2020-02-02
Add clearclick to S0000009

JorisVanEijden:
Clear _clickedElysana_ until the _giveLetter_ timer has expired. This prevents the quest prompt from popping up as soon as the _giveLetter_ timer expires when you clicked Elysana at any time after the quest started.
Forum post: https://forums.dfworkshop.net/viewtopic.php?f=5&t=3368


Interkarma:
Thank you. :)

PR #1690, opened 2020-02-02
Improved outdoor footstep tile noises with shallow water and stone paths

jefetienne:
If a player is walking on shallow water, the PlayerMotor won't recognize it as a water tile and will make the PlayerFootsteps play regular grass terrain footstep noises. In my commit, I added additional index tiles that are corner pieces between ground and water to make splash noises, without actually being submerged in the water and set to "swimming."

I also added stone paths (indices 46, 47, 55) to use the dungeon tile floor noise, which I think fits nicely.


Interkarma:
This is really good. I love the firmer footstep sound on paths, and clipping those partial water tiles and hearing a "splash" is very satisfying. Thank you. :)

PR #1691, opened 2020-02-02
Set player race before generating backstory

numidium:
Addresses the bug described here: https://forums.dfworkshop.net/viewtopic.php?f=24&t=2220
The %ra macro, which uses the PlayerEntity's BirthRaceTemplate, is used while generating player backstory. Since BirthRaceTemplate is not set until after the player exits the new game wizard, %ra expands into the default race name i.e. Breton.
Setting the race template before generating backstory gives the macro helper the info it needs.


Interkarma:
Thank you for the fix! :)

PR #1692, opened 2020-02-03
prevent mobile NPCs from entering stairs

petchema:
Use raycast as an additional check to localWeight == 0

Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=792


Interkarma:
This is exactly what I had in mind for the solution. Thanks so much for putting it together. :)

Issue #1688, opened 2020-02-02
Wrong questor data issues


JorisVanEijden:
When a quest adds a questor through the `add ... as questor` syntax, the wrong data is used.

In `Quest.AddQuestor()`
`person.IsQuestor = true;` calls `SetQuestorStatus(true)` which sets the questorData of the person to the data of the last clicked NPC. Which may very well not be the same person.
It then goes on the create a new questorData which is added to the quest, but not the Person.

This causes all kinds of subtle bugs with mixing up NPCs wherever the questorData of that Person is used.

Pull request that caused it: https://github.com/Interkarma/daggerfall-unity/pull/1583

Probably related reports:
Forum post: https://forums.dfworkshop.net/viewtopic.php?f=5&t=3368
Reddit post: https://www.reddit.com/r/daggerfallunity/comments/et6z8v/main_quest_glitched_please_help/

df-ferital:
Good catch! I will take a look at this to provide a better fix than my original PR.

df-ferital:
After further investigation, the issue looks quite hard to fix without breaking the original issue I wanted to fix (post success and failure greetings).

The problem is that DFU quest system relies on `StaticNPC.NPCData` to check if a clicked NPC was the questor of a particular quest. This works for all generic quest NPCs, but not for main quest ones, who are all added as questors using the `add _NPC_ as questor` command. In my previous PR, I incorrectly presumed that this command was always used after the questor NPC was clicked, but as you noticed, this is not always the case.

Now the problem is that there is absolutely no way to generate `StaticNPC.NPCData` on the fly for a given individual NPC, since this is related to a NPC location. I could use the NPC faction data, but since the Person `isQuestor` flag is set to false at the end of the quest by the  `drop _NPC_ as questor`, I can't know for sure that the individual the player is talking too was the questor.

One of the solution would be to keep a reference, in each quest instance, on the questor (since there is always one questor, as far as I know).

df-ferital:
I opted for the least intrusive solution which is to keep the `Person.isQuestor` flag to true even after the `drop _NPC_ as questor` command has been fired. This fixes the current issue while keeping PR #1583 intended fix and shouldn't break anything else.

df-ferital:
Fixed by PR #1693

PR #1693, opened 2020-02-03
Fix main quest questors issues

df-ferital:
This fixes issue #1688  which is related to my previous PR #1583.

Because main quest questors are added using the `add _NPC_ as questor` command, they are not always the last clicked NPCs, so we can't rely on this data to select quest related post success and post failure greetings. Instead, we can use faction IDs since all main quest questors are individual NPCs, having a unique faction ID. However to do that, we must also keep the `Person.isQuestor` flag as true, but this should not have any consequences since the `drop _NPC_ as questor` command effectively clears the NPC from the quest list of questors.

### DIFF
--- a/Assets/Scripts/Game/Questing/Person.cs
+++ b/Assets/Scripts/Game/Questing/Person.cs
@@ -84,7 +84,7 @@ namespace DaggerfallWorkshop.Game.Questing
         public bool IsQuestor
         {
             get { return isQuestor; }
-            set { SetQuestorStatus(value); }
+            set { isQuestor = value; }
         }
 
         public bool IsIndividualNPC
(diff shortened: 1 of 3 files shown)


Interkarma:
Thank you, I think I understand the issue and fix, and can't think of any problems with this right now. I'm sorry for not helping catch the problem introduced when I reviewed those changes earlier.

Interkarma:
OK, managed to find a problem. When starting a quest directly (either by console or from another quest) that sets an individual NPC questor, game will throw an exception in SetQuestorStatus() as LastNPCClicked is null (so cannot access QuestMachine.Instance.LastNPCClicked.Data). To reproduce, just load a game and use "startquest s0000009". Quest crashes out immediately due to exception in SetQuestorStatus().

Adding a check to SetQuestorStatus() so that it doesn't try to set NPCData for individual NPCs fixes the exception and shouldn't cause and further problems I can think of (as individual NPCs don't use that NPCData for conditional anyway, just faction id).

I'll add that change now ahead of builds and reference this comment.

PR #1694, opened 2020-02-04
adjust max loiter time message to setting

petchema:
Is there a way to embed a line return in a message instead of having to use several messages to match classic text formatting?


Interkarma:
Not at this time. It would be nice to have an approach that handles \n in message boxes and creates new lines automatically from a single string input.

PR #1695, opened 2020-02-05
fix spell casting female spellsword size

petchema:
Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=2961


petchema:
to test: tgm then cm 129 'til you get some female spellswords

Interkarma:
This is a huge improvement, thank you. :)

PR #1696, opened 2020-02-05
Do not update Magic Rounds when the game is loading

JorisVanEijden:
(or paused/inactive). This prevents issues where it would compare the worldtime of the previous game to the worldtime of the savegame. Which would run up to 2880 magic rounds when it should not.

See forum post https://forums.dfworkshop.net/viewtopic.php?f=5&t=3379

### DIFF
--- a/Assets/Scripts/Game/MagicAndEffects/EntityEffectBroker.cs
+++ b/Assets/Scripts/Game/MagicAndEffects/EntityEffectBroker.cs
@@ -191,8 +191,8 @@ namespace DaggerfallWorkshop.Game.MagicAndEffects
 
         void Update()
         {
-            // Don't tick if lastGameMinute not set (pre-init)
-            if (lastGameMinute == 0)
+            // Don't tick if lastGameMinute not set (pre-init) or the game is inactive.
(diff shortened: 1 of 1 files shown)


Interkarma:
This change will stop quest system from ticking as game time passes while player is in a UI - e.g. travel or resting. This breaks the intended design. We _want_ player to be able to rest/loiter to trigger timed events such as spawns in quests.

A different solution will be required. I'll take a look at the original issue when I can. Thank you for your feedback on this one.

Interkarma:
This change will stop effect system from ticking as game time passes while player is in a UI - e.g. travel or resting. This breaks the intended design. We still want effects to tick as player rests or travels. Some effects might even depend on this.

A different solution will be required. Just checking to see if game is loading should be adequate, rather than globally stopping effect system when a UI is open.

I'll also take a look at the original issue when I can. Thank you for your feedback on this one.

JorisVanEijden:
Ah, I did not realize the game was Paused during rest and travel.
So my change caused the magicRounds to be fired _after_ the game was unpaused rather than _during_ the pause.

PR #1697, opened 2020-02-06
Fixes region update in $CUREWER

JayH2971:
Region in letter would change when player moved. Now it'll be the region where letter is received only.


Interkarma:
Thank you Jay! :)

PR #1698, opened 2020-02-06
[RFC] DFUMessageBox.clickAnywhereToClose using any button but swing weapon one

petchema:
Forums: https://forums.dfworkshop.net/viewtopic.php?f=12&t=3332


Interkarma:
I like this solution, thank you. :)

PR #1699, opened 2020-02-09
use nearClipPlane in 2D dungeon automap

petchema:
Use nearClipPlane to "peel" dungeon layers as you get closer.
This should makes the 2D automap more useful, and closer to classic behavior


Nystul-the-Magician:
imho classic behavior is not a good measure since the "peeling"  was a horrible mechanism in my opinion which led to more confusion than it helped. I rather think it was a compromise because of technical limitations regarding the very low resolution in vanilla
I mean lets test your change and see how people like it ;)

petchema:
classic 3D map is barely usable, and veteran players know how to put that characteristic (if not that feature) of the 2D map to good use. Either that, or gave up using dungeon maps altogether ;)
On the other hand, as far as I've seen new players use the 3D map exclusively. So I guess we could try to make the two maps complementary.
As you said, let's test how it goes :)

Nystul-the-Magician:
the more I think about your change, the more I like it - I never thought about zooming in on the 2D map heavily but with your suggested change this could change ;)

Interkarma:
This definitely makes the 2D map easier to read. Thank you for this tweak. :)

PR #1700, opened 2020-02-09
Cemetery ambient sounds

petchema:
Around cemeteries, classic plays creepy bird calls infrequently (once per minute on average), and howlings even less often (about 1/4th of the time).

During the day it happens on top of normal bird calls, so it's another layer of ambient sounds.

### DIFF
--- a/Assets/Scripts/Game/AmbientEffectsPlayer.cs
+++ b/Assets/Scripts/Game/AmbientEffectsPlayer.cs
@@ -151,7 +151,7 @@ namespace DaggerfallWorkshop.Game
                 StartWaiting();
             }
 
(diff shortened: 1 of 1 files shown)


Interkarma:
Amazing! I'd forgotten how much I enjoyed those creepy ambient effects until now. Thank you for restoring them. :)

petchema:
Until recently I was confused about the context for those sounds. Did they happen during cold nights?
So when they happened again first hand during some experiments in classic I investigated, and concluded they only happen around cemeteries.
And yes, it's specific, it's infrequent, and yet memorable.

ajrb:
Sound plays such a big part in creating atmosphere, nice job.

PR #1701, opened 2020-02-10
Added screenshotting functionality

jefetienne:
Whether you are playing or paused on a DaggerfallWindow, pressing the _PrtScr/SysReq_ button will take a screenshot, display the PNG file's name through scrolling HUD text, and save it into a "Screenshots" folder, located in same path as where the "Saves" path is (e.g. "AppData\LocalLow\Daggerfall Workshop\Daggerfall Unity\Screenshots" for Windows).

The PrintScreenManager is added as a separate GameObject in the "DaggerfallUnityGame" scene.

**Notes**:

Because there is no "Screenshot" keybind shown on the classic keybind window texture, it means that  if a player used the _PrtScr/SysReq_ button for a different keybind, it would silently erase the "Screenshot" keybind. I figured then that it would be safe to restrict the _PrtScr_ button as rebindable in the DaggerfallControlsWindow (like the _Escape_ key), but I'm not sure if that is alright with you guys.

Also, if you screenshot while the game is paused, it will do so, but the scrolling HUD message won't display until it's unpaused. I tried looking into it and realized I couldn't get it to display without overhauling PopupText and preventing DaggerfallUI from disabling DaggerfallHUD upon pausing. It's a minor annoyance and I don't know if there are other ways I could display the screenshot message.


Interkarma:
I love the feature, looking forward to having this one in the game. Just one nitpick though, I'm sorry. :)

SysReq/PrintScreen is already used in Windows to screenshot display to the clipboard (it's how I take screenshots now). It's kind of an unwritten understanding in Windows that PrintScreen only copies stuff to the clipboard. I wouldn't want a game overriding this to drop files on my drive every time I'm trying to use the clipboard feature. I'd like a different keybind for this - perhaps F8? When the new keybind UI is written, this is something users can bind to their own preferred key.

The feature works really well though. I'm happy with the default Screenshot location as well. Excellent work. 👍

jefetienne:
Thanks! I can definitely change it to an F8 keybind.

Also, as a heads-up, I've been working on adding native gamepad functionality since this last week. Although I haven't directly asked if you wanted this feature (I assumed since the "Joystick" UI button in the controls window is unclickable, that it would be done later), and that I'm still working on it and it would need to be reviewed by you, I am actually fully committed to working on a secondary keybind UI interface, since they can kind of work hand-in-hand. If you're alright with this, we discuss more about the features of the window and what it would look like in the forums :)

Interkarma:
Sounds good to me. :) Hit me up with a PM if I don't reply to your forum post promptly. I'm finding it hard to stay on top of everything in the forums these days.

jefetienne:
Awesome! I'll do that then!

I cleaned up my PrintScreenManager a bit more, and also, there was a bug where the DaggerfallControlsWindow wasn't utilizing InputManager's ReservedKeys to check for new bindings, so I fixed that as well.

Interkarma:
Thank you for that. :) I didn't have time today to review this one for next builds. I'll definitely get into next round of builds though.

Interkarma:
One issue I've found is that going back to a previous version after upgrading will set all keybinds back to nothing. I'll see if I can help with this today, as I'd really like to get this one merged. :)

jefetienne:
That sounds good, thank you!

petchema:
One drawback is, that more people seem to post PNG screenshots in the forums when JPEGs would have worked just as well, often for a fraction of file size; And some pages then take ages to render...
But the culprit is more a lack of general awareness of posters around those considerations.

Interkarma:
I like that users can make take lossless screenshots, and also agree it would nice if they also converted them to JPEGs before uploading to forums. :)

PR #1702, opened 2020-02-10
force UnityEngine.Random reseeding when parsing quests

petchema:
No optimisation attempt, QuestMachine.ParseQuest can be called several times during the same frame and is not in a hot path so I'm using System.Random as randomness source.
Generating a list of quests feels just as fast as before.

Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&t=3399
https://forums.dfworkshop.net/viewtopic.php?f=27&t=924&p=39061#p39056

### DIFF
--- a/Assets/Scripts/Game/Questing/Actions/PickOneOf.cs
+++ b/Assets/Scripts/Game/Questing/Actions/PickOneOf.cs
@@ -82,7 +82,7 @@ namespace DaggerfallWorkshop.Game.Questing.Actions
             bool success = false;
             if(ParentQuest != null)
             {
-                UnityEngine.Random.InitState(System.Environment.TickCount);
+                UnityEngine.Random.InitState(GameManager.Instance.QuestMachine.InternalSeed);
                 selected = taskSymbols[UnityEngine.Random.Range(0, taskSymbols.Length)];
                 Task task = ParentQuest.GetTask(selected);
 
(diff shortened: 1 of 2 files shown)


ajrb:
This will not help with PickOneOf action as that is not selecting until runtime as far as I can see, although I may be mistaken. There is a commented out seed line in there commented out by Interkarma in 2017.

Interkarma:
If PickOneOf is compiled multiple times in a row, can more than one compile within 1ms of each other? That would result in them ending up with the same seed. I can't recall now, but I think that's why I commented out that InitState in 2017. I don't have a problem with seeding random here, just that TickCount might not have enough time resolution.

I do like the clever approach of using an internal seed and Next() in QuestMachine. If we expose that at a property then PickOneOf could just use QuestMachine.Instance.InternalSeed when seeding InitState. Would just need to make sure the property called internalSeed.Next() when returning result, and we'll have a constantly rolling seed value anyone can use.

petchema:
> I do like the clever approach of using an internal seed and Next() in QuestMachine. If we expose that at a property then PickOneOf could just use QuestMachine.Instance.InternalSeed when seeding InitState. Would just need to make sure the property called internalSeed.Next() when returning result, and we'll have a constantly rolling seed value anyone can use.

As the King of Worms would say, "Done."

Interkarma:
Nice one, cheers! :)

PR #1703, opened 2020-02-11
MeleeDamage raycast fallback

petchema:
One can't simply call WeaponDamage() again with a fallback raycast hit if no enemy was struck during a first call of WeaponDamage() with a spherecast, because environment damage may be done twice.
So the major refacto is to split WeaponDamage() into environment damage (WeaponEnvDamage) and enemy damage (WeaponDamage).
I also used that refacto to simplify missiles damage.

### DIFF
--- a/Assets/Scripts/Game/WeaponManager.cs
+++ b/Assets/Scripts/Game/WeaponManager.cs
@@ -838,12 +838,9 @@ namespace DaggerfallWorkshop.Game
             if (Physics.SphereCast(ray, SphereCastRadius, out hit, weapon.Reach, playerLayerMask))
             {
                 DaggerfallUnityItem strikingWeapon = usingRightHand ? currentRightHandWeapon : currentLeftHandWeapon;
-                if(!WeaponEnvDamage(strikingWeapon, hit))
-                {
-                    hitEnemy = WeaponDamage(strikingWeapon, false, hit.transform, hit.point, mainCamera.transform.forward);
-                }
-                // Fall back to simple ray for narrow cages https://forums.dfworkshop.net/viewtopic.php?f=5&t=2195#p39524
-                else if (Physics.Raycast(ray, out hit, weapon.Reach, playerLayerMask))
+                if(!WeaponEnvDamage(strikingWeapon, hit)
+                   // Fall back to simple ray for narrow cages https://forums.dfworkshop.net/viewtopic.php?f=5&t=2195#p39524
+                   || Physics.Raycast(ray, out hit, weapon.Reach, playerLayerMask))
                 {
                     hitEnemy = WeaponDamage(strikingWeapon, false, hit.transform, hit.point, mainCamera.transform.forward);
                 }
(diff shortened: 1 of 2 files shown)


petchema:
As seen before in my first attempt in the forums, you can't just use raycast hit as a fallback to the spherecast hit.
You cannot use a call to WeaponDamage() with the raycast hit as a fallback to a call to WeaponDamage() with the spherecast hit either, because WeaponDamage() does both objects hit checks and apply damage, so you may apply damage twice, specifically apply environment damage twice.

Cleanest solution would be to separate checks from damage computation, so you can check

- if spherecast hit struck an enemy;
- if raycast hit struck an enemy;
- if spherecast hit struck the environment

and apply one of the 3 possible damages, with that descending order of priority.

That would require a deep refactorization however, so I went for a simpler refacto: I split WeaponDamage() into environment damage (WeaponEnvDamage) and enemies damage (WeaponDamage), then used the logic

- check if spherecast hit struck the environment, and compute environment damage;
- if it did not hit, check if spherecast hit struck an enemy, and compute damage;
- if it hit, check if a raycast hit can struck an enemy, and compute damage.

The drawback is that if you hit an enemy with the fallback raycast, environment damage will have been computed also (say, if you hit the enemy in the cage, you'll also hear the clinging of the weapon on the cage). It seems like a reasonable compromise.

Interkarma:
I see the issue here - definitely deeper than I realised. I'll loop back to this one soon. :)

petchema:
I think that in classic you can hit several enemies at once with a melee weapon (to be confirmed).
If it's confirmed and we want to reproduce that, we may have to look into SphereCastAll() or other different approaches...

Interkarma:
I'll loop back to this one when I can. Thanks for all your work on this.

Interkarma:
No issues found with this, I think you've done excellent work. I've also noticed it's much easier to fight enemies in doorways and against corners now, which was always problematic with the spherecast approach only. Thank you again. :)

Assets/Scripts/Game/WeaponManager.cs
```diff
@@ -413,49 +414,43 @@ public void Reset()
...
+            if (GameObjectHelper.IsStaticGeometry(hit.transform.gameObject))
+            {
+                DaggerfallUI.Instance.PlayOneShot(strikingWeapon == null ? SoundClips.Hit2 : SoundClips.Parry6);
+                return true;
             }
 
-            // Set up for use below
-            Transform hitTransform = arrowHit ? arrowHitCollider.gameObject.transform : hit.transform;
-            Vector3 impactPosition = arrowHit ? hitTransform.position : hit.point;
+            return false;
```
petchema, line 448: There's a small issue with non-static environment, say portcullis or blue bars, that prevent you from hitting enemies thru: WeaponEnvDamage returns false so RayHit fallback does not happen.
Not sure how to improve test here to take into account non-static environment (or mark portcullis and blue bars as static env, if that makes sense)

PR #1705, opened 2020-02-12
Added scale support for 3D objects

UncannyValley21:
Unmodified objects in unmodified locations will always spawn with a scale of  (1,1,1).
Works with mesh replacement.
Works with mesh combiner.
Replaced many calls for GameObjectHelper.QuaternionFromMatrix(matrix) to matrix.rotation, as the former would result in strange rotation in combination with below 1 and non-uniformed scaling.


Assets/Scripts/API/DFBlock.cs
```diff
@@ -411,6 +411,15 @@ public struct RmbBlock3dObjectRecord
             /// <summary>Z position in 3D space.</summary>
             public Int32 ZPos;
 
+            /// <summary>X position in 3D space.</summary>
+            public float XScale;
```
TheLacus: Change _position_ to _scale_ in description.

Assets/Scripts/Utility/AssetInjection/MeshReplacement.cs
```diff
@@ -104,7 +104,13 @@ public static GameObject ImportCustomGameobject (uint modelID, Transform parent,
             go.name = GameObjectHelper.GetGoModelName(modelID) + " [Replacement]";
             go.transform.parent = parent;
             go.transform.position = matrix.GetColumn(3);
-            go.transform.rotation = GameObjectHelper.QuaternionFromMatrix(matrix);
+            go.transform.rotation = matrix.rotation;
+
+            //We can't simply apply scale, since we can't guarantee that custom models will have a defualt (1,1,1) scale
+            go.transform.localScale = new Vector3(go.transform.localScale.x * matrix.lossyScale.x,
+                                                    go.transform.localScale.y * matrix.lossyScale.y,
+                                                    go.transform.localScale.z * matrix.lossyScale.z);
```
TheLacus: I think you can use `go.transform.localScale = Vector3.Scale(go.transform.localScale, matrix.lossyScale);`

ajrb:
Do we really need individual scales for all 3 dimensions? Just wondering if a single scale would be sufficient here?

UncannyValley21:
> Do we really need individual scales for all 3 dimensions? Just wondering if a single scale would be sufficient here?

Having the option to scale in all 3 dimensions is a powerful tool for level designers, allowing for more creative use of existing assets.  It is also a very standard thing to have, and to only have single scale variable would frankly be very weird. There is nothing to gain from only allowing uniform scaling.

HybOj:
3 dimensional scaling is useful, tall slim pillars, longer tables, better fitting of the carpets...

Interkarma:
This looks good, no issues found here.

Yeah, Matrix4x4.rotation should be preferable to QuaternionFromMatrix() in modern Unity. I don't believe Matrix4x4.rotation existed back in the Unity 4.x days when I originally wrote this. That's why I ended up using the helper method. I made sure I checked all the problem spots for dungeon object rotations that I could remember, and no problems found with this PR. I'll merge this one for next builds. Cheers!

ajrb:
@Interkarma This has broken the world data override system since these scales default to zero. Well I think that's what's broken things anyway. Am investigating... :(

UncannyValley21:
@ajrb Good catch. I had completely overlooked existing world replacement mods that didn't use the scale feature.

Issue #1707, opened 2020-02-12
Is there a way to pause keyboard shortcuts when focus is on input text field.


AsesinoBlade:
**Describe the bug**
With the introduction of keyboard shortcuts (which are fantastic by the way 👍 ), a bug was introduced in my inventory filter mod.   If I try to enter a filter that uses a keyboard shortcut character, the shortcut runs, which causes an annoying issue for the data entry.

For example, if I want to filter to only weapons from inventory screen, as soon as I type W, the wagon opens if I have one, or provides the message that there is no wagon. 

What I was wondering if there was some code I could place in my mod to pause the keyboard shortcuts when someone is typing in the inventory filter input filed.

**To Reproduce**
1.  Install and enable Inventory Filter mod
2.  Open inventory window 
3.  Click on Filter
4. Type weapon to filter to only weapons
5. As soon as you type W, the wagon shortcut runs.

**Additional context**
Note that there is nothing wrong with the shortcut logic, it works as intended.  I am only looking for some logic to pause it during data entry.

Thanks all
a

petchema:
Ah, damn. Hotkey sequences processing currently happens very early in DaggerfallUI.OnGUI(), so before controls can do anything about it.
I wrote a patch that allows BaseScreenComponents to tell they don't want hotkey sequences to be processed while they have to focus, but do you think it's okay @Interkarma ?
https://github.com/Interkarma/daggerfall-unity/compare/master...petchema:per-control-hotkeys-override

With that patch in place, localFilterTextBox could:

- declare it wants to override hotkeys
```
             localFilterTextBox = DaggerfallUI.AddTextBoxWithFocus(new Rect(new Vector2(1, 24), new Vector2(47, 8)), "filter pattern", localTargetIconPanel);
+            localFilterTextBox.OverridesHotkeySequences = true;
```

- grab focus when the filter textbox is revealed (UpdateFilterButton):
```
                 localFilterButton.Enabled = false;
                 localFilterTextBox.Enabled = true;
                 localCloseFilterButton.Enabled = true;
+                localFilterTextBox.SetFocus();
```

- relinquish focus when enter key is pressed (Update):
```
+            if (Input.GetKeyDown(KeyCode.Return))
+            {
+                if (localFilterTextBox.HasFocus())
+                    SetFocus(null);
+            }
```

AsesinoBlade:
Wow, this would be exactly what I need.  Man, you are quick :)
[...]

PR #1708, opened 2020-02-12
Allow controls to override hotkey sequences

petchema:
Disable hotkey sequences when controls with overridesHotkeySequences flag have the focus

Part of fix for issue #1707


petchema:
That could be made even a bit more flexible if controls could provide some keyboard events handler, returning true if they handled the event on their own; That way one could implement partial overrides.
I don't know if we'll require the extra flexibility, but it doesn't seem too difficult to implement.

PR #1710, opened 2020-02-15
Classic quest fixes, Mid-Feb 2020

JayH2971:
C0B00Y02 was complex enough to merit a whole variable/structure rewrite. Only a few quests remaining now: A0C00Y16, K0C00Y06, M0B11Y18, N0B11Y18, and S0000007.


Interkarma:
Legendary work as always, thank you Jay! :)

PR #1711, opened 2020-02-15
Allow mods to override allocating player starting equipment and spells, enemy equipment, and modify found items

ajrb:
- Introduced delegates for player starting equipment and spell assignment. (Moved them to be called after items from biog questions are added to player, so mods can adjust)
- Introduced delegate for assigning enemy starting equipment.
- Added a new formula that is called when loot items are generated for loot piles. This is for mods and LootRealism will use it to have found armor, weapons, and books be in a random condition.

You can see an example of usage here: https://github.com/ajrb/dfunity-mods/blob/startingStuff/LootRealism/Scripts/LootRealismMod.cs

### DIFF
--- a/Assets/Scripts/Game/Entities/EnemyEntity.cs
+++ b/Assets/Scripts/Game/Entities/EnemyEntity.cs
@@ -375,7 +375,6 @@ namespace DaggerfallWorkshop.Game.Entity
                 // they are better than the value in the definition.
                 for (int i = 0; i < ArmorValues.Length; i++)
                 {
-                     // BUG? this is using mobs armor val if it's less than equipment, contrary to comment above
                     if (ArmorValues[i] > (sbyte)(mobileEnemy.ArmorValue * 5))
                     {
                         ArmorValues[i] = (sbyte)(mobileEnemy.ArmorValue * 5);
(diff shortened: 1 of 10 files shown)


ajrb:
Also found a possible bug where conditional doesn't match the preceding comment in EnemyEntity line 378.

Assets/Scripts/Game/Entities/EnemyEntity.cs
```diff
@@ -462,31 +375,13 @@ public void SetEnemyEquipment(int variant)
                 // they are better than the value in the definition.
                 for (int i = 0; i < ArmorValues.Length; i++)
                 {
+                     // BUG? this is using mobs armor val if it's less than equipment, contrary to comment above
                     if (ArmorValues[i] > (sbyte)(mobileEnemy.ArmorValue * 5))
```
ajrb: Possible bug here.

Interkarma: I think this is because "no armor" = 100. Equipping armor subtracts from armor values, and unequipping adds it again (see DaggerfallEntity.UpdateEquippedArmorValues()).

A higher ArmorValues[i] means a lower actual armor rating, so in this case the code is working as commented by using the mobs armor rating if lower than equipment rating.

ajrb: Wow that's confusing but you're right. Backwards numbers confuse me :-)

ajrb:
Hopefully this is good to be merged now?

Interkarma:
No issues found, thank you. :)

PR #1712, opened 2020-02-20
More texture fixes

df-ferital:
I resumed my work on the texture runtime patch. I fixed all models referenced by @bansheexyz on the forums, most of @petchema ones as well as some other I discovered by myself.

@Interkarma If you want some saves to check all the patches, I can provide them.


Interkarma:
I'll take this one on trust, as you've done such great work on this in the past. Thanks so much. :)

PR #1713, opened 2020-02-20
Animation fix for female spellswords apply to the texture archive rather than the class

ajrb:


Assets/Scripts/Utility/GameObjectHelper.cs
```diff
@@ -1167,7 +1167,7 @@ public static GameObject[] CreateFoeGameObjects(Vector3 position, MobileTypes fo
                 {
                     // Assign gender randomly
                     MobileGender gender;
-                    if (UnityEngine.Random.Range(0f, 1f) < 0.5f)
+                    if (UnityEngine.Random.Range(0f, 1f) < 0.55f)
```
petchema: Change reverse-engineered from classic?

ajrb: No, something I agreed with Interkarma a while ago. The existing < 0.5 actually gives slightly more female enemies than male which has always seemed a bit odd in DFU compared to classic. This is just me skewing it 5% towards male since in a psudo-medieval setting adventuring and stuff like that would generally be more male dominated. I thought this change went in a while ago, seems I never committed it

petchema: okay

PR #1714, opened 2020-02-21
Fix covens offering different Daedra summons in the same day

ajrb:
See https://forums.dfworkshop.net/viewtopic.php?f=24&t=2649


Interkarma:
Thanks for fixes. Looks OK on a read, will merge this one in today.

PR #1715, opened 2020-02-21
Add ability to create specific books and random or specific potions in quest scripts

ajrb:


Interkarma:
Awesome thank you! :)

PR #1716, opened 2020-02-21
Remove escorting faces more reliably

petchema:
In some cases faces creation involve some randomness (say, for children), so FaceDetails structural equality could mismatch.

Also disable face panels until placed, so they don't show up in screen top left corner for a frame.

Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&t=3430


Interkarma:
Cheers :)

PR #1717, opened 2020-02-22
Allow apostrophies in classic spell names for quest actions

ajrb:


Interkarma:
Awesome, thank you. :)

PR #1718, opened 2020-02-23
Additional fixes and improvements for footstep noises

jefetienne:
- Static geometry (such as buildings) now also make the hard dungeon footstep noise instead of the grassy terrain noise
- Improved detection for what tile the player is on by extending the raycasting distance
- Fixed an issue where sometimes jumping would make a grass noise, and also improved jumping by playing the tile's footstep noise upon landing

When I was designing the functionality for the landing noise, it also happened to get triggered when the player would load a save, or move into/out of an interior location, because the player would be floating in the air before being repositioned on the ground. To remedy this, I created a _StreamingWorld.IsRepositioning_ bool property that checks if the player is currently being repositioned so it can ignore that footstep noise.

### DIFF
--- a/Assets/Scripts/Game/PlayerMotor.cs
+++ b/Assets/Scripts/Game/PlayerMotor.cs
@@ -517,7 +517,7 @@ namespace DaggerfallWorkshop.Game
 
             // Must be outside and actually be standing on a terrain object not some other object (e.g. player ship)
             RaycastHit hit;
-            if (GameManager.Instance.PlayerEnterExit.IsPlayerInside || !Physics.Raycast(transform.position, Vector3.down, out hit, rayDistance))
+            if (GameManager.Instance.PlayerEnterExit.IsPlayerInside || !Physics.Raycast(transform.position, Vector3.down, out hit, rayDistance * 2))
             {
                 return false;
             }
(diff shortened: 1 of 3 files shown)


Interkarma:
I'll get this one into today's builds, thank you. :)

Assets/Scripts/Game/PlayerMotor.cs
```diff
@@ -491,7 +491,7 @@ public bool GetOnExteriorGroundMethod()
 
             // Must be outside and actually be standing on a terrain object not some other object (e.g. player ship)
             RaycastHit hit;
-            if (GameManager.Instance.PlayerEnterExit.IsPlayerInside || !Physics.Raycast(transform.position, Vector3.down, out hit, rayDistance))
+            if (GameManager.Instance.PlayerEnterExit.IsPlayerInside || !Physics.Raycast(transform.position, Vector3.down, out hit, rayDistance * 2))
```
petchema: That * 2 triggers splash sounds when horsejumping over water tiles

jefetienne: Thanks, I'll get that fixed.

jefetienne: I added a fix #1738. I kept the multiplication in place because of a problem when walking on static slopes such as the stairs of Sentinel Castle, which resorted to a grass noise because the raycast didn't hit far enough. I instead made a change to the block that checks groundedness, where instead of checking if the player is currently not using the submerged noise, it will instead check if the player is swimming.

PR #1719, opened 2020-02-23
Fix equipped monster enemies always using hand to hand

ajrb:
Noticed this bug while looking though the code. Affects orcs and centaurs. No wonder they always felt so weak.


Interkarma:
Ha, good catch!

PR #1720, opened 2020-02-23
Ensure that the scaling vectors are not initialised with zeros. URGENT FIX!

ajrb:
Tried to use a fsSerializer post-processor but couldn't as the DFBlock is a struct and that means there's no way to adjust the values because the obj is passed by value.
This enables world data override files to continue to work without every single model requiring 3 new float values to be added.

### DIFF
--- a/Assets/Scripts/API/BlocksFile.cs
+++ b/Assets/Scripts/API/BlocksFile.cs
@@ -867,9 +867,6 @@ namespace DaggerfallConnect.Arena2
                 recordsOut[i].XPos = reader.ReadInt32();
                 recordsOut[i].YPos = reader.ReadInt32();
                 recordsOut[i].ZPos = reader.ReadInt32();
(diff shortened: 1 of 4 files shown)


ajrb:
This is a quick fix, I've made them internal so it stops data dumps from adding these for every single model when 99.9% are scale of 1. Authors can add scale values, but if not required they can be ignored providing backward compatability.

Interkarma:
Good thing I didn't get time to publish those builds from yesterday. I'll merge this and make new builds today.

ajrb:
Yeah, was quite fortunate. Especially since my mods would have broken. :-) 

Was just playtesting a new character for the Loot mod and went to find the Archs guild... it was there, just really really tiny I guess. Loading a save inside one just threw a bunch of exceptions.

Interkarma:
Thank you for the fix. :)

PR #1721, opened 2020-02-24
Texture fix strikes back

df-ferital:
More texture fixes, including the small ship (@HybOj will be happy), which was a real pain to fix, a lot of its planes UV coordinates were wrong, alongside a bunch of incorrect texture records. I also fixed a few other models reported on the forums, as well as a ruined tombstone which was in a very bad state (this one can be found in the cemetery of Aldingwall, Daggerfall).


Interkarma:
Thank you so much! I'll merge this one on faith again, you've done consistently impeccable work on this.

PR #1722, opened 2020-02-24
Add enhancement setting to use bows in left hand

ajrb:
This allows weapon switching between a melee and a bow without having to access the inventory. Think... sling the bow over shoulder and draw a blade. Introduces a switching delay for all weapon switches which is half the average weapon equip delay of the two weapons.

Playing a character who uses a mix of ranged and melee is a pain, and it seems reasonable to me for this to be a balanced tweak of the Daggerfall weapon system. If you would rather it goes in a mod, then I'll do that, but I feel it's a nice enough change for a core enhancement.

### DIFF
--- a/Assets/Resources/defaults.ini.txt
+++ b/Assets/Resources/defaults.ini.txt
@@ -120,3 +120,4 @@ CombatVoices=True
 EnemyInfighting=True
 EnhancedCombatAI=True
 GuildQuestListBox=False
+BowLeftHandWithSwitching=False
--- a/Assets/Scripts/Game/Items/ItemEquipTable.cs
+++ b/Assets/Scripts/Game/Items/ItemEquipTable.cs
@@ -123,11 +123,7 @@ namespace DaggerfallWorkshop.Game.Items
             }
 
             // Equipping a shield will always unequip 2H weapon
-            if (item.ItemGroup == ItemGroups.Armor &&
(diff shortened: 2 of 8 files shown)


jefetienne:
As a left-handed person, I appreciate this.

ajrb:
Great, but this doesn't make the bow appear on the left - there's already a setting for that. This just allows bow in left hand (internally) and a 1H weapon in right so you can switch between using 'H' key.

jefetienne:
Oh yeah I got that, I was just trying to be funny, haha.

petchema:
I can't test that PR at the moment... Can you explain how it's supposed to work? When bow is removed from hands by equipping a dual handed weapon, it stays equipped on second hand but not displayed?
So if you want to switch to hand-to-hand, you must just unequip both weapons?

ajrb:
No it works just as you would expect - basically like shields do already. The only confusing part is when you just equip a bow it's to left hand so when you bring up weapon if it's set to right hand you just get a fist and have to press H to switch to bow. This will default to off, and I think people who switch it on will know about 'H' and it will be obvious pretty quickly.

Interkarma:
This works great! And you found a clever way to handle this without really changing much in the equip slot tracking. You also made it an option for user. I very much appreciate both of those things, thank you. :)

ajrb:
Glad you like it. :-)

I do love elegant code tweaks that get a big bang for buck. It's my bag for sure. lol

PR #1724, opened 2020-02-26
[RFC] More discreet create foe place freely 

petchema:
Previous algorithm to spawn foes freely would first find a direction slightly outside of player's field of view, check if there's any obstacle in that direction, and if so find a spawning point along the normal of collision point.

Using that normal helps finding a clear location (it should be at least clear from the obstacle just hit, but further tests are still needed), however it changes the direction of the spawning point, moving it away from field of view borders; Either inward (making the foe suddenly appear in front of the player) or outward (not as bad).
Also the code checks the distance of the hit point, not spawning point, making it possible for spawns to happen at a much lower distance than minDistance constraint.

Stream: https://www.twitch.tv/videos/556821863?t=27m40s
lycanthrope spawn
(Lycanthrope appearing in a tavern right in the center of player's field of view, probably taking the normal off the fireplace on the right)

This patch:
- selects a spawn point along the direction selected at first, to be on the border of player's field of view, so foes should never "appear" in front of the player (even if sometimes in places that do not make much logical sense, exactly like before);
- assuming the obstacle is a plane, the new algorithm may choose a spawn point much closer than before as clear from the obstacle (can't beat using the normal here);
- however it will always check that spawn happens further than minDistance, and only add some randomness to the distance if there's enough room to do so in the first place.

### DIFF
--- a/Assets/Scripts/Game/Questing/Actions/CreateFoe.cs
+++ b/Assets/Scripts/Game/Questing/Actions/CreateFoe.cs
@@ -290,13 +290,19 @@ namespace DaggerfallWorkshop.Game.Questing
             RaycastHit initialHit;
             if (Physics.Raycast(ray, out initialHit, maxDistance))
             {
-                // Separate out from hit point
-                float extraDistance = UnityEngine.Random.Range(0f, 2f);
-                currentPoint = initialHit.point + initialHit.normal.normalized * (separationDistance + extraDistance);
+                float cos_normal = Vector3.Dot(- spawnDirection, initialHit.normal.normalized);
+                if (cos_normal < 1e-6)
+                    return;
+                float separationForward = separationDistance / cos_normal;
 
                 // Must be greater than minDistance
-                if (initialHit.distance < minDistance)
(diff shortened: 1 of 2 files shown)


petchema:
There's another bug in the original code, camera.fieldOfView returns vertical field of view in Unity, not horizontal field of view.

Interkarma:
>There's another bug in the original code, camera.fieldOfView returns vertical field of view in Unity, not horizontal field of view.

When I wrote that code early on, I unfortunately assumed it was horizontal FOV. I'm not sure if I missed this or it just wasn't well documented back in Unity 5.x days. I don't even remember seeing a horizontal FOV setting until fairly recently. Definitely a mistake on my part which seems obvious now.

Interkarma:
Thanks for improvements. Will merge this and we'll see how it rolls. :)

petchema:
To be perfectly clear I didn't fix the horizontal vs vertical FoV bug in this PR, because I didn't like the enemies becoming extremely sneaky.
As you said, that's something that will need experimenting. I think we should also think about other ways to spawn enemies, like right when you switch scene or wake up from resting/loitering, like classic does.

PR #1726, opened 2020-02-27
Need to allow replacement of potion recipes for mods

ajrb:
Just propagating the already passed flag to potion indexing method, pretty sure that it was just an oversight but it should be verified before merging.


Interkarma:
This was an oversight on my part. Thank you for fixing. :)

PR #1727, opened 2020-02-28
Don't assume player is standing in PlayerActivate raycast

petchema:
WeaponManager uses camera position and direction, just do the same.

Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&t=3454

There may be more to the problem discussed in the forums than this bug, but it's a start.


Interkarma:
I think this is a sensible fix in any case, even if there's something else going on. Thank you. :)

PR #1728, opened 2020-02-29
Make inventory window more accessible to mods

numidium:
This is for myself and any future modders who want more customization for the inventory window.


ajrb:
The changes to inventory window look fine, but I'm not sure about the changes to the UI component ItemListScroller. It's fully configurable via constructor and events - overrides of inventory window should be changing how the component is configured and not overriding & extending the component.

Please can the protections remain unchanged for the ItemListScroller component? If there's something missing from the component setup let me know and it can be added.

numidium:
@ajrb I concur with your comment. The ItemListScroller code was different when I started this branch but at this point it doesn't need to be changed as far as I can tell.

TheLacus:
Please squash all commits into one to avoid reverted changes in the commit history if you can; if not, this can also be done by the person that accepts the PR so this is not an issue :)

PR #1729, opened 2020-02-29
Same as #1484 with EnableTextureArrays = False

petchema:
Same as #1484 but in DaggerfallTilemap.shader

No regression found on my hardware with EnableTextureArrays = False, but ideally it should be tested with hardware that require that setting...

Forums https://forums.dfworkshop.net/viewtopic.php?f=5&t=3456


Interkarma:
I think this should be fine based on previous success of the approach. I'm sure we'll hear about it if others experience problems on their specific hardware. :)

PR #1730, opened 2020-02-29
fix rare mobile orientation blind spot

petchema:
Sometimes, when encountering an idle mobile, it look like it's facing you until you slightly move around.

The problem is that the choice of mobile orientation is based on cases using only strict inequalities, so boundary values did not match any case and orientation was left at its initial value of 0.

I replaced everything with formula, that could be extended more easily to any number of orientations.

### DIFF
--- a/Assets/Scripts/Internal/DaggerfallMobileUnit.cs
+++ b/Assets/Scripts/Internal/DaggerfallMobileUnit.cs
@@ -40,6 +40,9 @@ namespace DaggerfallWorkshop
     [RequireComponent(typeof(MeshRenderer))]
     public class DaggerfallMobileUnit : MonoBehaviour
     {
+        const int numberOrientations = 8;
+        const float anglePerOrientation = 360f / numberOrientations;
+
         [SerializeField]
         MobileUnitSummary summary = new MobileUnitSummary();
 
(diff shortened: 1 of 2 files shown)


Interkarma:
Just a bit of background on this one. I used the divisor approach in my first implementation, and I disliked the feeling of each orientation having an equal number of radii. Monsters will too readily angle away from player during combat if player stands slightly to one side or the other (e.g. strafing). It felt to me that the front facing should be weighted more than other facings so the facing sprite orientation feels more "locked on" and less "flickery" during combat. That's why I ended up with the hand-tuned approach to increase facing angle.

To see what I mean, spawn a skeleton with cm 15 then strafe left and right, you'll see the orientation rapidly flip around more with the equal division of orientations than with the weighted orientations I ended up with. It's very obvious with the skele because the shield changes hands when the sprite is mirrored during orientation changes. Both approaches have the issue, just the wider facing angle experiences it a bit less.

With that said, I don't think the issue is that dramatic or even problematic now that I look at it again. I implemented that facing code way before these objects were mobile. They had no pursuit AI or steering at the time, I was just working with standing anims to get the basic systems in. The problem I observed back then is nowhere near as dramatic as I remember now the enemies can naturally chase and turn by themselves, and I'm more engaged trying to actually fight them I don't really notice anymore.

The divisor approach is definitely cleaner, and if it solves another rare issue then I'm happy to merge it. Thanks for the input. :)

petchema:
I see...
Except this PR does not change the output orientation for any angle (but for the exact boundary values where the result was wrong):
enemyFacingAngle
I assume your formula didn't have the correct rounding

Interkarma:
You're right, I'm using equal 45 degrees per orientation as well now that I look at it. I could have sworn I used a wider frontal angle and apportioned the rest out amongst the remaining orientations. Not sure when/if that changed since I first wrote this back in 2014. Could just be faulty memory and confirmation bias. Sorry for not looking closer.

In any case, your change is definitely superior and I was happy to merge the edit. Cheers.

PR #1731, opened 2020-02-29
Copy building array in block before modifying it

ajrb:
See https://forums.dfworkshop.net/viewtopic.php?f=24&t=3417

Only seems to be an issue when two identical blocks are next to each other in west-east direction.


Interkarma:
Great work on this obscure little bug. Thank you. :)

PR #1732, opened 2020-02-29
prevent collisions between all missiles

petchema:
Both arrows and spell missiles gameObjects didn't inherit SpellMissiles layer from their parent, so they would collide with one another.

Forums: https://forums.dfworkshop.net/viewtopic.php?f=23&t=3459


Interkarma:
Nice catch, cheers.

PR #1733, opened 2020-03-01
AI improve doorways finding

petchema:
This PR does not change the current doorway finding algorithm, that looks good, but tweaks 2 parameters so far:
- limit the height of the capsulecast used to find doorways to 1.75, to match the hack used to allow tall enemies to get thru doors. Without this, tall enemies could technically get thru doors, yet are very reluctant to do so;
- lower capsulecast distance from 1 to 0.8. Long casts was making it hard for enemies to find door openings coming from the sides. Now it's a bit better.


Interkarma:
Looks good to me, cheers! :)

PR #1735, opened 2020-03-02
Added version label to setup wizard

jefetienne:
https://imgur.com/a/tdHhoRw

I felt like the setup wizard could use a small version label for users to notice which .exe they're opening. It's especially useful in case if a user uses multiple versions for mod compatibility, or if they're transitioning from one version to another and accidentally left the older version pinned to their taskbar, start menu, or on their desktop as a shortcut.


Interkarma:
Looks good, thank you. :)

PR #1738, opened 2020-03-05
Fix water footstep playing while jumping over water

jefetienne:
Fixed from petchema's review #1718.


Interkarma:
Thank you! And thanks @petchema for the help reviewing. :)

petchema:
Probably a regression with this patch or the previous one:
https://forums.dfworkshop.net/viewtopic.php?f=5&t=3619

jefetienne:
Thanks for sharing that!

I did take a look at his safe, and it actually is a bug independent of these changes - it's a different water sound (`FootstepSoundShallow` instead of `FootstepSoundSubmerged`) and it has to do with wit the player's position and the `blockWaterLevel` in the dungeon (`PlayerFootsteps.cs` line 189).

The problem is that I'm not really sure what the best fix is because I do not know how or when it is supposed to be played in classic behavior. My recommendation is that it should play if the player is colliding with `DungeonWater` but isn't swimming, like they're walking in a kiddie pool.

It is weird behavior and I'll go ahead and fix it if everyone feels alright with defining `FootstepSoundShallow` as that.

PR #1739, opened 2020-03-07
Fixes + improvements for opening/closing UI windows and console

jefetienne:
Besides the map windows, if the player had previously opened a window and then changed the keybind for it, the window would not refresh the toggle button and would stick to the old keybinding. This was also true for the console window as well. I fixed it by refreshing the bindings in `OnPush()` and `ToggleConsole()` for the windows and console, respectively.

Also, in preparation for the secondary keybind window, I've made the pause window utilize a rebindable toggle like all of the other windows. Previously, it relied on a hardcoded `KeyCode.Escape` and `DaggerfallPopupWindow`'s default 'exitKey' to leave the window without actually using `InputManager.Actions.Escape`. Also, when making this change, I noticed that if the toggled key was held down for too long after exiting the window, that it would pop back up again. I saw that the problem was `DaggerfallPopupWindow` using `GetKeyDown(exitKey)`, so I changed it to `GetKeyUp(exitKey)` to resolve the issue.

EDIT: Added rebindability for PrintScreenManager (I can't believe I forgot about my own submitted feature for a second)

### DIFF
--- a/Assets/Game/Addons/UnityConsole/Console/Scripts/ConsoleUI.cs
+++ b/Assets/Game/Addons/UnityConsole/Console/Scripts/ConsoleUI.cs
@@ -41,6 +41,9 @@ namespace Wenzil.Console
         /// </summary>
         public void ToggleConsole(bool force = false)
         {
+            //refresh keybinds if player rebinded it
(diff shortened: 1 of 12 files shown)


Interkarma:
Thank you! :)

PR #1742, opened 2020-03-07
Fix automap partial blocks wrong shaders

petchema:
When some gameObject without MeshRenderer is found in a dungeon block, shader of remaining gameObjects is left to Standard shader in the automap.

Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&t=3476


Interkarma:
Thank you!

PR #1744, opened 2020-03-09
Fix left handed bow damage and condition lowering 

ajrb:
The way bows are accessed for the formula differs from melee so it was using the weapon in right hand for damage calc.


Interkarma:
Thank you! :)

PR #1748, opened 2020-03-11
[Secondary Keybinds] Reworked controls window

jefetienne:
(I already have the secondary keybind window finished, but I'm breaking into smaller PRs)

- Rewrote CheckDuplicates() to use a shared static helper method called GetDuplicates(), which will be used by the secondary keybind window. 
- All keybinds are shared in a static dictionary of unsaved keybinds, and updated upon save or reset to default.
- WaitForKeyPress is also static for it to be shared with the secondary keybind window.
- Fixed a bug where players could press the bottom buttons while the window is waiting for input


Interkarma:
This looks good. I'll merge this one now. Sorry for the delay. :)

jefetienne:
No problem at all! There's a lot going on right now in the world, haha.

I'll make a PR soon with the actual secondary keybind window, which can also get merged first before the controller support (why add support if the player can't even rebind the pause button? lol)

PR #1749, opened 2020-03-11
[Secondary Keybinds] Created Singleton out of StartGameBehaviour

jefetienne:
Used for the secondary keybind window.

Singleton code was copied and reworked from InputManager. `ApplyStartSettings()` needed to be public for saving additional settings in that window, such as mouse look sensitivity, weapon sensitivity, movement acceleration, inverting mouse Y axis, and for weapon attack threshold.

### DIFF
--- a/Assets/Scripts/Game/Utility/StartGameBehaviour.cs
+++ b/Assets/Scripts/Game/Utility/StartGameBehaviour.cs
@@ -99,6 +99,48 @@ namespace DaggerfallWorkshop.Game.Utility
 
         #endregion
 
+        #region Singleton
+
+        static StartGameBehaviour instance = null;
+        public static StartGameBehaviour Instance
+        {
+            get
+            {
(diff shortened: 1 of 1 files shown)


ajrb:
I thought that GameManager handled the singleton-ness of StartGameBehaviour, and I don't see why it was needed to make apply settings public. Could you explain please?

jefetienne:
Ooh, thanks, I didn't catch that StartGameBehaviour property in GameManager. I do need the ApplySettings to be public so I can call it from the secondary keybind window, which is currently named as DaggerfallMouseControlsWindow. I consulted with @Interkarma about adding some of the same settings that are found in the setup wizard to the secondary keybind window for ease of access when adjusting input sensitivities. I could copy and paste the same serialization for those certain settings but I think it would be best to stick to DRY principle and reuse the method. I will close this issue though since a Singleton is unnecessary.

PR #1750, opened 2020-03-11
Fixes discrepancies in item stats from classic

ajrb:
I've just realised that the fix here for the enchantmentPoints is not getting used in the game. That's probably why it was not noticed up until now.

@Interkarma when you implemented item enchanting you created formula methods (GetItemEnchantmentPower etc) to calculate this from the base points value in the template instead of using the value calculated by Allofich code in ItemBuilder. I'm not sure if this was because of the bug that this PR fixes, or because you were not aware of its existance.

Having it in the item means it is easier for a custom item implementation to alter the points in a different way, which is useful in at least one case but not necessarily a reason to switch if there are good reasons not to.

Either way I think one or the other of these 2 implementation of enchantment point material scaling should be removed. I am happy to do that, but you should decide which we keep I think.

### DIFF
--- a/Assets/Scripts/Game/Items/ItemBuilder.cs
+++ b/Assets/Scripts/Game/Items/ItemBuilder.cs
@@ -47,7 +47,7 @@ namespace DaggerfallWorkshop.Game.Items
         static readonly short[] valueMultipliersByMaterial = { 1, 2, 4, 8, 16, 32, 64, 128, 256, 512 };
 
         // Condition multipliers by material type. Iron through Daedric. MaxCondition is baseMaxCondition * value / 4.
-        static readonly short[] conditionMultipliersByMaterial = { 4, 4, 6, 8, 12, 16, 20, 24, 28, 32 };
+        static readonly short[] conditionMultipliersByMaterial = { 4, 6, 6, 8, 12, 16, 20, 24, 28, 32 };
 
         // Enchantment point multipliers by material type. Iron through Daedric. Enchantment points is baseEnchanmentPoints * value / 4.
         static readonly short[] enchantmentPointMultipliersByMaterial = { 3, 4, 7, 5, 6, 5, 7, 8, 10, 12 };
@@ -680,9 +680,9 @@ namespace DaggerfallWorkshop.Game.Items
         {
             item.value *= 3 * valueMultipliersByMaterial[(int)material];
             item.weightInKg = CalculateWeightForMaterial(item, material);
-            item.maxCondition *= conditionMultipliersByMaterial[(int)material] / 4;
+            item.maxCondition = item.maxCondition * conditionMultipliersByMaterial[(int)material] / 4;
             item.currentCondition = item.maxCondition;
-            item.enchantmentPoints *= enchantmentPointMultipliersByMaterial[(int)material] / 4;
+            item.enchantmentPoints = item.enchantmentPoints * enchantmentPointMultipliersByMaterial[(int)material] / 4;
 
             return item;
         }


Interkarma:
>I'm not sure if this was because of the bug that this PR fixes, or because you were not aware of its existance.

I think GetItemEnchantmentPower() path in FormulaHelper is clearer to read than a lookup table, especially with the comments provided, and it supports formula overrides if we want to add them. The calculation in ItemBuilder doesn't appear to be rounding down, something that I recall was important to the value being correct. I prefer the GetItemEnchantmentPower() path overall.

ajrb:
> 
> 
> > I'm not sure if this was because of the bug that this PR fixes, or because you were not aware of its existance.
> 
> I think GetItemEnchantmentPower() path in FormulaHelper is clearer to read than a lookup table, especially with the comments provided, and it supports formula overrides if we want to add them. The calculation in ItemBuilder doesn't appear to be rounding down, something that I recall was important to the value being correct. I prefer the GetItemEnchantmentPower() path overall.
> 
> Edit: BTW, we still need to ensure item.enchantmentPoints is populated as GetItemEnchantmentPower() uses this value for items not scaled by materials.
> 
> Can you give me some background on what this PR is looking at fixing? Cheers. :)

ajrb:
Whoops I mis-clicked.

ajrb:
> I think GetItemEnchantmentPower() path in FormulaHelper is clearer to read than a lookup table, especially with the comments provided, and it supports formula overrides if we want to add them. 
Yes it's more readable & verbose, but downside is it differs from the way the other material scaling is done using lookups in ItemBuilder.

If you decide you want to keep the formula, which do not currently support overrides BTW but that can be added easily, then that is fine. However I will want to route the calls via the item class so that custom items can do their own thing, and remove the redundant lookup table. Also I would consider maybe using the same switch and subtracting 0x0200 from armor materials before calling instead of having two large identical switches in the codebase.

> The calculation in ItemBuilder doesn't appear to be rounding down, something that I recall was important to the value being correct. I prefer the GetItemEnchantmentPower() path overall.
> 
This PR fixes that rounding bug. I found it was happening for steel once I entered the correct fractional power of 4 for the condition multiplier and fixed it at the same time. 

> Edit: BTW, we still need to ensure item.enchantmentPoints is populated as GetItemEnchantmentPower() uses this value for items not scaled by materials.
Not as far as I can see - looks like it's always the item.template.enchantmentPoints that's used.

> Can you give me some background on what this PR is looking at fixing? Cheers. :)
This PR just fixes the rounding bugs from Allofiches lookup reverse engineered from classic. basically by using the *= operator the division was done before the multiplication and with ints this gave errors unless the scaling factor was a multiple of 4, which most are hence it going undetected for so long.

Interkarma:
It sounds like you're happier with the lookup tables and this also gives you more flexibility with custom items. In that case, I'm completely happy for you to ditch the switch statements in FormulaHelper and just fix Allofich's method.

ajrb:
Nope, not at all. Only thing I care about is that we remove one or the other so it doesn't confuse anyone like it did me in the future. As for whether we remove Allofiches code or yours, that's what I'd like you to decide... I focussed on advantages for keeping the lookup because I assumed you had the ones for your code in mind already.

If you also don't care either way (or are just too busy right now to make the call) then I will keep your formula, and route their calls through item object which is an established pattern used for other things like min/max damage etc to allow mods to change the existing items by overriding the formula, but also have complete freedom for custom item implementations. 

Feel free to take a couple of days to decide, or just hand the decision to me if you want. As i said I just don't want to leave redundant code in that can mislead people.

Interkarma:
I'm happy for you to decide here. The main thing for me is the right values come out the other end. If Allofich's fixed code is better for custom items and more consistent, I'm totally happy to preserve it and zap my formulas.

On my side, I've been bogged down lately with business continuity design and mobilising staff to work from home. I don't know when this is going to end, but I think my workload is probably going to level out within a month or two. There's only so much that can be done and my client base is finite. Have reserved some time this weekend to catch up on sleep. :)

ajrb:
Good hope you can get some rest. I'll handle this and then drop the PR in next few days.

take care!

PR #1751, opened 2020-03-12
Fix bug with breaking armor and enchanting unworn armor

ajrb:
Should fix the following issues:
https://forums.dfworkshop.net/viewtopic.php?f=5&t=3497
https://forums.dfworkshop.net/viewtopic.php?f=24&t=3491


Interkarma:
Legend, thank you mate. :)

PR #1752, opened 2020-03-14
Improved robustness of ModManager.TryGetAsset()

TheLacus:
`ModManager.TryGetAsset()` now doesn't return null if the first asset found can't be loaded but it keep seeking for a valid result instead.

This fixes an issue where a texture could be retrieved properly if a mod contains both texture and xml configuration file (with the same name), while it failed if it contains only the xml file because our internal code only considers the filename (unlike Unity API that also checks type).

`Mod.LoadAsset()` returns null without logging errors if asset is not found, which is consistent with Unity API (i.e. `Resources.Load()`).


Interkarma:
Thank you for the fix. I'll merge this one for next builds. :)

PR #1755, opened 2020-03-19
make frost daedra emissive, like in classic

petchema:
Noticed by KoW / @HybOj

Took me a while to understand that emission was wrong for some records because of IsExteriorWindow(). That predicate sure is not pretty.


HybOj:
Thank you, so the corpse will be emissive as well? Same as fire daedra is? That would be great, sorry if I was not easy to understand :)

petchema:
The corpse will be emissive, just like Fire Daedra's one

Interkarma:
Nice to see more work on this. Cheers. :)

PR #1756, opened 2020-03-20
Fix windows not toggling closed after rebinding key

numidium:
If the user changes the key binding for the inventory, character sheet, journal, or travel map windows then the toggle bindings do not update.


jefetienne:
I actually have this fix waiting to be approved for #1739, haha.

numidium:
> 
> 
> I actually have this fix waiting to be approved for #1739, haha.

Ah, my mistake. I see yours is more comprehensive too. Well done!

PR #1759, opened 2020-03-21
Safer unstick handling using Physics.CapsuleCast()s

petchema:
Previous unstick handling used a SphereCast to check whether the player could be teleported forward across some geometry issue.
However the fact that the player controller shape is a capsule instead of a sphere could lead to an erroneous teleportation inside obstacles (shop banners, graveyard gate,...)

Just replacing one shape with another is not sufficient, because hte fact that a controller shaped capsule couldn't move forward is exactly why the player was stuck in the first place. To be safe, what should be checked is whether a capsule could freely follow a path to the place we want to teleport the player to.
Simplest path after the straight line is a step (go up, forward, then down), that's what this PR does.

Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=3291

### DIFF
--- a/Assets/Scripts/Game/Player/FrictionMotor.cs
+++ b/Assets/Scripts/Game/Player/FrictionMotor.cs
@@ -130,7 +130,6 @@ namespace DaggerfallWorkshop.Game
             //  -Works best when player is standing as spherecast test has more clearance
             //  -Enemies will still become stuck as their motor does not have this handling
             const float stuckMovementThreshold = 0.07f;
-            const float stuckSampleDistance = 0.5f;
             const int stuckFrameThreshold = 3;
             bool tryingToMoveForwards = InputManager.Instance.HasAction(InputManager.Actions.MoveForwards);
             bool tryingToMoveBackwards = InputManager.Instance.HasAction(InputManager.Actions.MoveBackwards);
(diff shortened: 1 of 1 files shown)


petchema:
PlayerHeightChanger is not 100% safe either

jefetienne:
I actually had an issue with this method while developing the controller support. If the joystick was moved very slightly vertically for slower movement, this function would think the player was stuck and make a weird jittery movement caused by the teleportation.

petchema:
> I actually had an issue with this method while developing the controller support. If the joystick was moved very slightly vertically for slower movement, this function would think the player was stuck and make a weird jittery movement caused by the teleportation.

What you can try to tweak is the tests done before the code this PR changes; Either change stuckMovementThreshold and stuckFrameThreshold values, or find a better stuck detection algorithm...

jefetienne:
My solution for the issue was to have an arbitrary minimum force applied if Unity detects input from the joystick, which came out to be 0.68. I could try testing this PR and merging it with a duplicate of my controller branch to see how it feels.

petchema:
> I could try testing this PR and merging it with a duplicate of my controller branch to see how it feels.

The two problems are unrelated (detect player stuck on geometry vs finding a workaround to unstick the player), so  it should be work as expected, unless you totally rewrote UnstickHandling()

Interkarma:
This still work well in all my test saves against the broken geometry I know about. Thanks for the refinements. :)

PR #1760, opened 2020-03-21
Prevent the "end quest" action from firing twice

JorisVanEijden:
… when the _oneday_ timer expires without having stolen the guarded item.

Forum post: https://forums.dfworkshop.net/posting.php?mode=reply&f=24&t=3422#pr40022


Interkarma:
Thank you. :)

PR #1766, opened 2020-03-29
Fix dungeon exit and other static door overlap

petchema:
I case they're several static door hits, give higher priority to dungeon exit
Alternate implementation would be to add static doors deduplication in RemoveOverlappingDoors(), but I'm not sure it's worth the extra complexity.

Forums: https://forums.dfworkshop.net/viewtopic.php?f=24&t=3390


Interkarma:
This is a nice simple resolution, thank you. I was looking at deduplication as you mentioned, but this is much more straightforward. Considering what a rare case the double exit door is, I think this is a perfectly fine solution.

PR #1769, opened 2020-03-30
Made RdbResource.TriggerFlag_StartingLock public so it gets exported …

BadLuckBurt:
…/ imported properly for WorldData overrides. Without it, it's not possible to assign the trigger type for classic activation behaviours of models / flats, making custom dungeons with classic activators impossible.


ajrb:
Looks good to me, I didn't really delve into actions when I did the world data for RDBs.

PR #1770, opened 2020-03-31
[RFC] Don't stack equipped items during player deserialization

petchema:
Equipped items are not stackable, but this check doesn't work during deserialization, because inventory items are deserialized first, then they're equipped while deserializing the equipTable.

Simplest fix is to disable stacking while deserializing the inventory: use AddItem(..., noStack: true) in ItemCollection.DeserializeItems().

However, if this was done on purpose (so that stacking "self heals" during game loading), this PR instead only prevents the stacking of equipped items.
It's not very satisfying, because equipTable serialized data is used twice, so I submit it as a RFC.

Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&t=3560

### DIFF
--- a/Assets/Scripts/Game/Items/ItemCollection.cs
+++ b/Assets/Scripts/Game/Items/ItemCollection.cs
@@ -531,7 +531,7 @@ namespace DaggerfallWorkshop.Game.Items
         /// Existing items will be destroyed.
         /// </summary>
         /// <param name="itemArray">ItemData_v1 array.</param>
-        public void DeserializeItems(ItemData_v1[] itemArray)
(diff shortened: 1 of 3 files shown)


ajrb:
I think I'd prefer the simpler fix of just disabling auto-stacking when de-serialising the item data instead of doing it only conditionally. This is already done for custom item classes since until the item is instantiated, the IsStackable() cannot be queried and the auto-stack on deserialisation doesn't work properly - IIRC.

Obviously that needs agreement from other core devs, mainly @Interkarma I think, but my suspicion is that this emerged from evolution of the class rather than an intentional behaviour that's required. I could be wrong of course.

petchema:
Yup, I think #1771 is sufficient, and possibly more robust too

PR #1771, opened 2020-04-01
Don't (re)stack items during player deserialization

petchema:
Simpler alternative to #1770


Interkarma:
This is great, thank you. :)

PR #1772, opened 2020-04-02
Add a blinking cursor to daedra quest Y/N question

petchema:
Should convey that the player is supposed to type something, while having a minimal impact on visuals.

I put the cursor near the bottom right corner of the screen because I didn't manage to fully grasp the UI coordinates system, but it looks okay (and could be improved later).


ajrb:
Noooo, it's a test from the Daedra!! :-) And is like classic... but yeah, a few people have been a bit stumped by it.

Interkarma:
Thank you! Yeah, I've had a couple of emails about this one where folks assume the game has crashed because they didn't know to press Y or N keys. :)

Issue #1420, opened 2019-08-12
Weather effect animations reset when walking to a new zone in the overworld


BlueBolt70:
**Describe the bug**
When moving on foot/on a horse/cart to a new zone in the overworld (ex. moving from one pixel on the world map to the next), weather effects such as rain and snow reset and play independently from the previous zone. Moving back and forth between the previous and next zone will show that the snow falling effects doesn't match with eachother.

**To Reproduce**
Just run into a direction in the overworld while it's raining/snowing (it's easier to notice when it's snowing.)

**Expected behavior**
The snow/rain will briefly disappear and start falling down all over again.

**Desktop (please complete the following information):**
 - OS: Windows 10 Home 64-bit
 - Version Alpha 0.10.4

petchema:
Yes, I noticed the issue a while back:
https://forums.dfworkshop.net/viewtopic.php?f=5&t=1916&start=10#p23383

petchema:
This should be fixed as of 0.10.22

PR #1773, opened 2020-04-02
Relocate weather particles when world offset changes

petchema:
Fix for #1420

I implemented a fix, then noticed there was 5 years old code doing the same thing, just a bit better!


Interkarma:
Assuming you've added PositionUpdate component to PlayerAdvanced.prefab so that it recursively updates particles, but not the transform? Yeah, that should do it. :)

I'd forgotten about that component as well, as the later iterations became more automatic. I'll test this one and merge. Cheers!

PR #1775, opened 2020-04-03
Support for runtime materials

TheLacus:
# Summary
Added a component that holds a list of archives and records and uses them to assign materials when the prefab is instantiated at runtime. Materials are automatically applied on Awake but can also be manually reapplied.

It allows to "reference" a classic Daggerfall material from a prefab provided by a mod.

runtime_mat_0

runtime_mat_2

# More Informations

Mod authors can use this component to provide a list of archive-record tuples associated to a material index. This is a better alternative than including classic textures with the mod because it doesn't require texture duplicates for each mod and has full support for texture replacement with load order. 

The reason to specify material index (instead of using array order) is to allow the use of classic materials along with custom ones. As a convenience i added an entry in the context menu named "Init from MeshRenderer".

Materials can be manuallly reapplied with `void ApplyMaterials(RuntimeMaterial[] materials)` or from the inspector with the context menu entry "Reapply Materials".

runtime_mat_1

# How To Test

- Create a prefab and add component **Runtime Materials**.
- Set target archive-record (for example 23-0).
- Launch play mode, pause and drag the prefab inside the scene. Materials are assigned immediately from Awake().

runtime_mat

### DIFF
--- a/Assets/Scripts/Utility/AssetInjection/Components/RuntimeMaterials.cs
+++ b/Assets/Scripts/Utility/AssetInjection/Components/RuntimeMaterials.cs
@@ -1,0 +1,160 @@
+// Project:         Daggerfall Tools For Unity
+// Copyright:       Copyright (C) 2009-2020 Daggerfall Workshop
+// Web Site:        http://www.dfworkshop.net
+// License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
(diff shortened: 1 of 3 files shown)


XJDHDR:
For context, this was created in response to this post on the forum: https://forums.dfworkshop.net/viewtopic.php?p=41861#p41861

Interkarma:
Looks good to me. Will merge this one now. Thank you! :)

PR #1777, opened 2020-04-08
Improved output of 'startquest' command

jefetienne:
The console command 'startquest' will now say "Added quest {quest display name}" if it is a valid ID and "Quest ID '{id}' could not be found" otherwise.

It used to only return "Finished", which seemed ambiguous if it meant I finished that quest or if it got completed it for me. It also didn't say if the player used a valid quest id or not.


Interkarma:
Makes sense to me. Thanks for the change!

PR #1780, opened 2020-04-12
Make NPCs present if you have HallAccessAnytime privilege

petchema:
Seems to work that way in classic, and makes that privilege actually useful.

Forums: https://forums.dfworkshop.net/viewtopic.php?t=3502

### DIFF
--- a/Assets/Scripts/Game/PlayerActivate.cs
+++ b/Assets/Scripts/Game/PlayerActivate.cs
@@ -85,6 +85,7 @@ namespace DaggerfallWorkshop.Game
             get { return currentMode; }
         }
 
+        // Public opening hours; Guilds' HallAccessAnytime can override that
         public static bool IsBuildingOpen(DFLocation.BuildingTypes buildingType)
         {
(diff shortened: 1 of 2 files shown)


petchema:
Test save
[SAVE340.zip](https://github.com/Interkarma/daggerfall-unity/files/4466362/SAVE340.zip)

ajrb:
Hmm, if this is matching classic then I can't really argue I guess, however I thought that it seemed sensible that guild staff go home at night when the guild is closed. I cannot recall now what my testing of classic was.

petchema:
Yes, I was wondering if maybe only part of the crew would be there, or maybe only part of the week...
But from the test I did with a high rank in Mages Guild everybody was there. Could certainly use more tests, say with other guilds, but it would make sense to be the way guilds work in general, as having access to an empty guild is not very useful, and does not look like a high rank privilege...

ajrb:
Well it means you can access the library or rest for free in some guilds. Or just to escape from vengeence. :-)

petchema:
You're right... but that's still a bit underwhelming ;)

Interkarma:
Thank you. :)

PR #1781, opened 2020-04-12
Prevent the use of weapon while climbing

petchema:
Seems to be what classic does according to my observations.

This is the simplest implementation, No weapon drawn sound is played when you're finished climbing.


Interkarma:
That's fine with me, thank you for fix. :)

Issue #1539, opened 2019-09-29
Closing UI windows makes sound skip


petchema:
**Describe the bug**
Using Resources.UnloadUnusedAssets() in UserInterfaceWindow.CloseWindow() makes sound skip or stutter for many people, even with no mods (hence minimal memory usage).

Forums: https://forums.dfworkshop.net/viewtopic.php?f=27&t=10&p=25780#p25780

**To Reproduce**
Listen to sound while closing any UI window.

**Expected behavior**
Music and sounds to play seamlessly while using UI

**Desktop (please complete the following information):**
Tested with 0.10.7 on Linux, but it can be noticed on many gameplay videos

**Additional context**
Resoures.UnloadUnusedAssets() is supposed to be called automatically by Unity when new scene is loaded, I guess explicit call was added because Daggerfall Unity doesn't load many scenes (beside very early during execution). But could it be called in places that would be equivalent to scene loading: loading game, fast travel, teleport, entering/leaving buildings?
Alternatively, I personally used a patch that leaves the call in UserInterfaceWindow.CloseWIndow() but throttles garbage collection so it doesn't happen more than once every 5 minutes, which is sufficient to prevent eventual long term memory usage runaways.
Distant Terrain mod also seems to call Resoures.UnloadUnusedAssets()

petchema:
WIP:
https://github.com/Interkarma/daggerfall-unity/compare/master...petchema:load-scenes-gc

Nystul-the-Magician:
As far as I remember we put this in on purpose, we had a lot of ppl running out of memory and experiences crashes without because gc would not happen often enough

petchema:
I'm not advocating removing those calls altogether, because they're probably required (mainly because DFU doesn't use LoadScene() much I suppose).
However calling it from UserInterfaceWindow.CloseWindow() doesn't seem like a good idea, because it produces a noticeable skip for many people. Once you noticed that issue, it becomes unbearable.
I think we should investigate other options, like the patch I linked above. (investigating again could also be a good idea because we changed the version of Unity we use, so memory management may have improved).

Distant Terrain also contains a call to Resources.UnloadUnusedAssets() when origin is moved, as long as it does it's not concerned by how frequently we call UnloadUnusedAssets from DFU itself.

Nystul-the-Magician:
alright, makes sense ;)

petchema:
[...]

PR #1782, opened 2020-04-12
Reduce GC alloc of SongManager

jefetienne:
I was looking at #1539 and while I can't seem to reproduce the issue, I minimized GC allocation of SongManager in hopes that it might help with the sound issue. It reduced the GCAlloc from 284 B to 0 B.

Calling struct.Equals(object o) by default will "call Object.Equals(Object) on each field of the current instance and obj and returns true if all fields are equal."

https://docs.microsoft.com/en-us/dotnet/api/system.valuetype.equals?view=netframework-4.8#System_ValueType_Equals_System_Object_

Let me know if this fix was useful.

### DIFF
--- a/Assets/Scripts/Game/SongManager.cs
+++ b/Assets/Scripts/Game/SongManager.cs
@@ -62,6 +62,14 @@ namespace DaggerfallWorkshop.Game
             public PlayerMusicWeather weather;
             public PlayerMusicTime time;
             public bool arrested;
+
+            //minimize GC alloc of struct.Equals(object o) with this method instead
+            public bool Equals(PlayerMusicContext pmc) {
+                return  environment == pmc.environment
+                        && weather == pmc.weather
+                        && time == pmc.time
+                        && arrested == pmc.arrested;
+            }
         }
 
         PlayerMusicContext currentContext;


petchema:
That kind of work should always start with some profiling, because otherwise you'll:

- not know if you're heading in the right direction, doing lot of changes (and possibly introducing bugs or making code less readable) for basically no gain;
- have no idea how much you improved speed or reduced garbage; Or even if you improved the performance at all, because managed environments can behave in unexpected ways, and what you thought was an optimization is actually worse than the code it replaces.

jefetienne:
Definitely. I was using the Unity profiler to find that SongManager was using 284 B in the Update() function, then commented out code to isolate where the GCAlloc was happening and found that it was with that Equals function.

petchema:
You mean it's a significant source of garbage?
Well, creating an object in Update() is not good, but I didn't think this one would be significant...
But I take your word for it, interesting!

jefetienne:
Yeah, I'm not sure how significant this will be, especially since I haven't noticed any sound skipping previously, but it definitely is a positive optimization according to the Unity profiler.

So I'm thinking that by eliminating the GC overhead in SongManager, closing a UI window and calling `Resoures.UnloadUnusedAssets()` won't make an impact on the sound, but again, I can't reproduce the sound issue so I need someone else to see if they still have the problem by testing with this PR.

Interkarma:
While it seems likely skip is from a large GC overhead across multiple sources, a positive reduction in garbage creation anywhere is always a good thing. Thank you for the PR @jefetienne. :)

I don't experience this one either. My feeling is it mostly effects people with lower-end PCs who mod their games too heavily for those specs.

jefetienne:
Yeah, it could be that as well, too.

petchema:
I've seen it in all kind of streams, including using unmodded Daggerfall Unity on capable hardware, so that's probably not the only factors. I suspect low-level details like driver and soundcard buffering, but who knows.
I think it also gets worse with game run time (probably because of some memory fragmentation?), so the problem may never show up in, say, 1 hour sessions.

Interkarma:
>I think it also gets worse with game run time (probably because of some memory fragmentation?), so the problem may never show up in, say, 1 hour sessions.

That's interesting. Because garbage is collected on UI close, there should only be small semi-frequent GC flushes. The GC itself will also perform collections on its own schedule while the game is running. The problem shouldn't be cumulative over time unless something is leaking and fragmenting memory as you say.

Interkarma:
As this improves GC alloc and has been profiled, I can't see any reason not to merge. Every little bit helps, even if it doesn't solve the other issue of sound skipping on window close.

PR #1783, opened 2020-04-12
Require confirmation before spellbook sorting

petchema:
A drawback of adding keyboard shortcuts to form buttons is that sometimes people trigger actions by inadvertedly pressing game keys.

Most common I've seen is W (forward) key pressed in inventory. Either selects the wagon or displays a message if the player has no wagon, it's relatively harmless.

Another a more annoying I'm starting to see is people pressing S (backward) key in the spellbook, reordering their spellbook.

This patch adds a confirmation before spellbook sorting.
Alternatively, the player could be asked what sort order to use.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallSpellBookWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallSpellBookWindow.cs
@@ -870,17 +870,29 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
         // Not implemented in Daggerfall, could be useful. Possibly move through different sorts (lexigraphic, date added, cost etc.)
         public void SortButton_OnMouseClick(BaseScreenComponent sender, Vector2 position)
         {
-            var spellsBefore = GameManager.Instance.PlayerEntity.GetSpells();
-            GameManager.Instance.PlayerEntity.SortSpellsAlpha();
-            var spellsAfter = GameManager.Instance.PlayerEntity.GetSpells();
-            if (spellsAfter.SequenceEqual(spellsBefore))
+            DaggerfallMessageBox mb = new DaggerfallMessageBox(uiManager, DaggerfallMessageBox.CommonMessageBoxButtons.YesNo, TextManager.Instance.GetText(textDatabase, "sortSpells"), this);
+            mb.OnButtonClick += SortSpellsConfirm_OnButtonClick;
+            mb.Show();
+        }
+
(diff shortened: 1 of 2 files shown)


jefetienne:
Thank you, both can be pretty annoying as well.

What do you think about maybe changing the shortcut key to something like Shift-S? That way it will be harder for the player to accidentally sort it.

petchema:
I considered it, but I don't think I like it better than an extra prompt. But that's debatable.

Interkarma:
This works well, thank you! The prompt isn't as bothersome as I thought it might be. I think this is a pretty good solution for the hotkey problem.

petchema:
I see some streamers hitting this _all th time_ because they open their spellbook while retreating...
That's a pain, I'm not sure that prompt will be the final word on this issue :(

Interkarma:
Man goes to the doctor and says "Doc, it hurts every time I do this" (waves around elbow violently). Doctor says "well just don't do that then".

I think the best answer here is to simply not to have a hotkey on spell sort button. It adds very little value by itself, causes more problems for players and thus problems for us. Users can just click sort when they want it. Solved. :)

petchema:
Well, that will solve that one case, but it's a general problem; With different game key bindings or UI shortcuts localized differently, it can happen again with another UI shortcut...

petchema:
Switching back HotkeySequences to KeyDown triggering is easy enough (just two places to adjust), that's with deferring CloseWindow() to the KeyUp event of the same KeyCode that I have trouble with;
Because the code activating HotkeySequences has no idea that the event handler will call CloseWindow(), and the event handler or even better UserInterfaceWindow.CloseWindow(), has no idea that it's being called because of a HotkeySequence, I'm not sure how to get everything to cooperate nicely...

jefetienne:
Yeah, that's kind of why I was thinking if we make most our HotKeySequences as "Ctrl/Shift/Alt" + "key", we won't need to worry about those kind of problems when users decide to change keybinds for certain actions. It will avoid any conflicts since it's near impossible to accidentally be holding one of those 'control' keys while pressing one of their keybind actions.

Interkarma:
Fair enough, thanks for the feedback @petchema. Sounds like you're already chasing down a better solution. :)

petchema:
My current idea is that HotkeySequences should emit KeyDown and KeyUp events, so that handlers that want to close a window (for example) during KeyDown can defer it to the KeyUp event they should receive next.
That requires to upgrade the events handler model, that needs to keep handling mouse clicks too, but that's the most flexible and prevents adding dependencies all over the place...

PR #1785, opened 2020-04-17
Remove invalid assets from cache

TheLacus:
If a mod loads an asset, unloads it and then requests it again, the invalid asset is removed from cache to allow it to be loaded again. Previously, the load operation would have failed.

Mods should not unload cached assets (of course additons to the API can be proposed  if actually needed) but this should handle the issue more graciously for the player.


Interkarma:
Thank you. :)

PR #1786, opened 2020-04-18
Changed protection level of Affiliation method to allow for Mod overr…

AsesinoBlade:
I would like to override the ShowAffiliationDialog in a mod.  My mod simply shows the number of days until the next Guild Promotion evaluation.

To do so, i need to change the method from private to protected virtual.

That is the only change I made to the code.

Thanks 
a


Interkarma:
That's OK with me. :)

AsesinoBlade:
Thank you sir!
A

PR #1787, opened 2020-04-18
Do not call Resources.UnloadUnusedAssets() more than once every 3 mins

petchema:
Daggerfall Unity needs to explicitly call Resources.UnloadUnusedAssets() because it barely uses SceneManager.LoadScene() which is the way games usually indirectly call it.

This is currently done in two places, including while closing UI windows, a path taken often enough during normal gameplay to prevent unused assets to pile up too much.

However this can be a lengthy operation that blocks the whole environment as it calls GC.Collect() internally, and can make sound skip which can be pretty annoying once you noticed it (issue #1539).

This patch makes sure the call to UnloadUnusedAssets() does not happen more than once every 3 minutes, a tradeoff sufficient to make that issue much less noticeable, yet should prevent unused assets to stay around forever.

It should be noted that, while not required, for optimal efficiency all code using Resources.UnloadUnusedAssets() (say mods) should call either ThrottledUnloadUnusedAssets() or ForcedUnloadUnusedAssets() instead to prevent unnecessary collections.

### DIFF
--- a/Assets/Scripts/Internal/DaggerfallGC.cs
+++ b/Assets/Scripts/Internal/DaggerfallGC.cs
@@ -14,16 +14,12 @@ using UnityEngine;
 
 namespace DaggerfallWorkshop
 {
-    public class DaggerfallGC
+    public static class DaggerfallGC
     {
         // Min time between two unused assets collections
         private const float uuaThrottleDelay = 180f;
 
         private static float uuaTimer = Time.realtimeSinceStartup;
-
-        private DaggerfallGC()
-        {
-        }
 
         public static void ThrottledUnloadUnusedAssets()
         {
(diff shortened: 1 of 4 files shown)


Assets/Scripts/Internal/DaggerfallGC.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        // Min time between two unused assets collections
+        private const float uuaThrottleDelay = 180f;
+
+        private static float uuaTimer = Time.realtimeSinceStartup;
+
+        private DaggerfallGC()
+        {
+        }
+
```
TheLacus, line 27: Is there any reason you used a private constructor that never get called instead of a static class or you just preferred this way?

petchema: Oops, just Java habits!

Interkarma:
Thank you, happy for this workaround. Even just minimising the problem goes a long way to making the experience better. Cheers! :)

PR #1788, opened 2020-04-18
Remove unused assets from mod cache

TheLacus:
#1671 introduced removal from cache of materials that have not been accessed for a fixed time. This is now extended to mod cache in order to improve memory management for unique assets used by mods.

Timestamp of last access is updated when an asset is requested with `GetAsset()`. As a compromise, assets loaded with `LoadAllAssetsFromBundle()` or `LoadAllAssetsFromBundleAsync()` are not removed before they are actually retrieved.


Interkarma:
Thank you. :) I'll just merge this one, as I trust your knowledge of the mod system more than my own.

PR #1789, opened 2020-04-19
[Experimental] Add posterization postprocessing to retro mode

petchema:
Uses 4bits per color component instead of using a palette.
I added some non-linearity (gamma) to get more colors in dark places.


petchema:
Capture du 2020-04-19 05-20-14
Capture du 2020-04-19 05-21-18

jefetienne:
I thought those were classic screenshots for a second. This looks great!

petchema:
Yes! Another using DREAM, Handpainted models and few other usual suspects, just for funzies :D
Capture du 2020-04-19 05-41-07

Interkarma:
This is wildly good! I love it and happy for it to be an option for retro rendering mode.

Interkarma:
I'm keen to merge this one for next builds. Can you see any issues @petchema?

BadLuckBurt:
When this is in, I'll be switching to retro mode. I could've sworn I was looking at classic DF, fantastic!

petchema:
> When this is in, I'll be switching to retro mode. I could've sworn I was looking at classic DF, fantastic!

Thanks! 
Yes, I played a bit with that shader on yesterday, and my brain was nicely deceived too :D

> I'm keen to merge this one for next builds. Can you see any issues @petchema?

The feature itself seems okay. The drawback of allocating more precision to dark colors is that snow can be a bit too tinted under light, but it's not so bad.
Do you mind me using Shift-F11 to toggle it at runtime?
I'd also appreciate a review of the shader code; I used a template provided by Unity, so maybe some code could still be removed, or some pragma added so the shader is only compiled for the relevant targets, stuff like that.

Interkarma:
I don't mind using Shift+F11 to toggle. I used to use this for runtime font/SDF toggle but disabled that a while back.

Interkarma:
This one has some conflicts with recently merged PRs. Might need a rebase. Thank you Pango. :)

petchema:
done!

petchema:
For the record, I just implemented true palettisation with nearest-color to classic palette.
It uses naive implementation, which still runs at 120 FPS on my hardware:
https://github.com/Interkarma/daggerfall-unity/compare/master...petchema:retro-mode-nearest-color

Interkarma:
Shut the front door! I'm dying to check this out.

petchema:
I created some code generator to implement k-d tree based nearest neighbour search
https://en.wikipedia.org/wiki/K-d_tree#Nearest_neighbour_search
https://github.com/petchema/ocaml-palettisation-shader-generator
https://github.com/Interkarma/daggerfall-unity/compare/master...petchema:retro-mode-nearest-color-kdtree
However if I generate code for the whole tree, the shader becomes too large for the ShaderCompiler, so I had to limit the tree when I reach 32 color clusters :(
It's still an improvement (> 200 FPS), but I guess it's a deadend... larger trees would have to be materialized as a datastructure, not functions...

petchema:
landscape-posterization
landscape-palettization

jefetienne:
@petchema 
This is really nice! Since I'm not really educated on post-processing, which one would you say is closer to the DOS version? Or would different DOS computers back in the day vary and some would use posterization while others use palettization? Sorry if this sounds like a dumb question, haha.

petchema:
TL;DR: palettization, but even that is not exactly the same as DOS

Posterization just reduces the number of levels of red, green, and blue (separately), the code above uses 17 levels for each color component (that was a bug initially, I intended to use 16, but I prefer the results). This give some color banding artifacts that "fake" classic look, but it's not very accurate because even if there's still 17 * 17 * 17 = 4913 colors available for each pixel, those colors do not match the colors of the classic palette, so details can be lost in "translation" (see the mountain backdrop above for example).
For palettization I took a palette from classic, and for each pixel I pick the nearest match in the palette. This is more expensive, but artwork assets usually pop up nicely because they use available colors.
But classic Daggerfall uses neither of that, for efficiency its whole graphic stack must use indexed color and lookup tables to adjust lighting, add fog, etc. And sometimes use different colors than what palettization can find by looking at the whole palette...

PR #1793, opened 2020-04-21
Added backup escape key method (Shift-Escape)

jefetienne:
Working with the controller PR with the added advanced controls window, if someone were to rebind the pause button to the controller, then replayed the game without the controller, then they would be stuck and unable to change their keybinds without modifying Keybinds.txt.

This PR adds a hardcoded Shift-Escape keybind allowing the player to access the DaggerfallPauseOptionsWindow in case if they fall into this situation.


Interkarma:
Good idea. I didn't think of that outcome. :)

PR #1795, opened 2020-04-22
Added ability to filter quests to Active Only in Quest Debugger

AsesinoBlade:
Hi, 
I had thought that the quest debugger only showed active quests, but it was always showing all quests for me.  I added some functionality to the quest debugger to allow you to show all quests or just active quests.  
You can flip between all quests and active quests by pressing "A".   
I default to all quests upon start so that it aligns with current functionality.'
Thanks 
a

Added Dialog Shortcut to select between all quests and active quests only (shortcut = letter A)
Defaults to all quests upon starting the quest debugger.

### DIFF
--- a/Assets/Scripts/Game/UserInterface/DaggerfallShortcut.cs
+++ b/Assets/Scripts/Game/UserInterface/DaggerfallShortcut.cs
@@ -63,6 +63,7 @@ namespace DaggerfallWorkshop.Game.UserInterface
             DebuggerToggle,
             DebuggerPrevQuest,
             DebuggerNextQuest,
+            DebuggerActiveQuestToggle,
 
             // Rest menu
             RestForAWhile,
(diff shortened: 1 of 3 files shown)


Interkarma:
I'm not a fan of this one in its current state. The A key is shared with strafe left by default, so just moving normally with debugger open will trigger this. It's also not setting quest count correctly on toggle and will still display total quest count then only cycle between active quests. Then count and index no longer align.

Showing all active quests loaded on the QuestMachine is an intentional choice. Even when a quest is tombstoned, the ending state might be important to quest authors, and it may still be delivering ongoing rumours. A quest isn't really finished until it leaves the QuestMachine completely, at which point it disappears from debugger anyway. There's usually only 1-2 finished quests at a time, and they drop off after 7 game days. I'm just not convinced this change is needed.

AsesinoBlade:
Fair enough, thank you for considering it.

we can remove the PR.
thanks
a

PR #1796, opened 2020-04-22
Removed accidentally merged changes from custom-dialog-buttons.

BadLuckBurt:
My most sincere apologies, I accidentally merged changes from custom-dialog-buttons into custom-flat-model-activation. While the changes don't interfere with DFU's normal operation, after consulting @ajrb who thankfully spotted this, I decided to remove them for now until they are properly implemented.


ajrb:
Thanks for fixing this so quickly, mistakes happen.

PR #1799, opened 2020-04-26
Apr 2020 Quest Fixes

JayH2971:
Related discussions:
https://forums.dfworkshop.net/viewtopic.php?f=25&t=3473#p40923
https://forums.dfworkshop.net/viewtopic.php?f=5&t=3651&p=42749#p42745
Spawning slowed in pest quests due to some forum discussion, can't find it but every 5 seconds is too quick for novice players


Interkarma:
Great work as always Jay.

PR #1800, opened 2020-04-26
Prevent saves while rappelling

petchema:
Discussion in https://github.com/Interkarma/daggerfall-unity/commit/6a7ca24c52ec7c0cff684feab98a6c274b55b94e


Interkarma:
This works well. Quite a clever way to go about it as well.

PR #1801, opened 2020-04-26
Disable head bobbing while in cursor mode

petchema:
While in cursor mode the player can't adjust his/her view direction with the mouse, but if head bobbing is on it will change view direction, making the view roll over as the player walks around.

This leads to embarrassing moments, because whether the mouse cursor is displayed or not becomes totally secondary to the player when (s)he starts looking at his/her feet, then the landscape upside-down, then the sky, and repeat. People get utterly confused, and the obvious conclusion is usually "the game bugged out", and fixed by restarting it.

I'm not sure whether cursor mode should be totally disabled until fully implemented, but at bare minimum the interaction between cursor mode and head bobbing has to be handled somehow; That's what this PR does.

Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&t=2776

### DIFF
--- a/Assets/Scripts/Game/Player/HeadBobber.cs
+++ b/Assets/Scripts/Game/Player/HeadBobber.cs
@@ -80,6 +80,7 @@ namespace DaggerfallWorkshop.Game
         {
             if (!DaggerfallUnity.Settings.HeadBobbing ||
                 GameManager.Instance.PlayerEntity.CurrentHealth < 1 ||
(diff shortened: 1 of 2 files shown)


Interkarma:
Thank you for PR. That's a good patch for now.

BTW, I have no plans to implement Cursor mode (move character by pointing and clicking at edges of screen). But the ability to capture/release cursor is also a requirement of large HUD to click buttons. This will function just like classic in View mode, which also does not allow cursor-mode movement.

PR #1804, opened 2020-05-02
Update support for mod manifest file

TheLacus:
- Updated json schema that is used by some text editors to validate file.
- Do not serialize null values.


Interkarma:
Thank you Lacus! :)

PR #1806, opened 2020-05-02
Improvements to joystick attack swinging

jefetienne:
(This is pertaining to having `ClickToAttack` disabled)

As of right now, the `WeaponAttackThreshold` is shared with both the mouse and the joystick. For gameplay experience, this threshold isn't really needed for a joystick since it does not move like a mouse, and makes it harder to attack. If the `WeaponAttackThreshold` is set very low (<0.001), the player can hold their joystick in one direction and the game will automatically constantly swing the weapon, giving an unfair advantage over the mouse.

This change removes the WeaponAttackThreshold from being calculated if the player is using a controller (via `InputManager.Instance.UsingController`), but also prevents automatic swinging by having the player let go of the joystick (or moving it close to the middle) in order to swing again.

### DIFF
--- a/Assets/Scripts/Game/InputManager.cs
+++ b/Assets/Scripts/Game/InputManager.cs
@@ -177,6 +177,11 @@ namespace DaggerfallWorkshop.Game
             get { return (vertical < -deadZone || vertical > deadZone) ? vertical : 0; }
         }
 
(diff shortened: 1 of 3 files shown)


Interkarma:
Thank for the refinements. :)

PR #1807, opened 2020-05-02
Retro mode add palettization

petchema:
Add nearest-color palettization shader using ART_PAL classic palette as an alternative to posterization shader.

The shader uses k-d tree based nearest neighbour search, with a cutoff below 32 color clusters.
Most of the source has been generated for extra efficiency.

Example of result already posted in a comment:
https://github.com/Interkarma/daggerfall-unity/pull/1789#issuecomment-623002019


petchema:
With that shader, the night sky is grey instead of deep blue.
I wonder if I should use NIGHTSKY instead of ART_PAL, or if I'll have to switch (yuck!) or combine the differen t palettes...

Interkarma:
I think combining the different palettes is the best option. This is pretty much what classic DF did anyway using colour translations from HAZE and SHADE files.

https://en.uesp.net/wiki/Daggerfall:Image_formats/Colour_Translation

Adding a few more colours into the palette will likely a yield more classic-like result than just sticking with the standard un-translated palette.

petchema:
Well, from what I understand that's only a translation inside the same palette, so it doesn't add any displayable color.

Anyway, ART_PAL and NIGHTSKY don't share as many colors as I would have guessed:
ART_PAL 255 unique colors
NIGHT_SKY 251 unique colors
both together 461 unique colors
So for the moment I only added the 32 first entries of NIGHTSKY.
I modified my tools to externalize palettes and easily handle such cases :)

Interkarma:
>Well, from what I understand that's only a translation inside the same palette, so it doesn't add any displayable color.

This is true, but what I'm saying is that DF technically does have a larger colour space than 256. It's just clever about how the larger virtual palette space is manipulated through the physical palette space based on VGA limitations. On our end, we have no such physical limitations so can simply represent the full virtual palette at once. What you've done is perfect. :)

Loving this BTW, it's one of my favourite cosmetic features.

petchema:
That PR can be squashed, no need to keep the 3 different generated code versions around...

Interkarma:
Looks lovely, I'll merge this one now. I'm aiming to get out 0.10.23 builds this weekend if there are any last-minute fixes/improvements needed before then. Well done! :)

petchema:
Well, right now I don't see any way to get a faster implementation that gives the exact same result. Maybe I overlooked some tricks to get better performance from shaders, but I tested a few things already and that was always slower.
I guess the main performance path would be computing just approximations that would not have much impact on visual result. With just algorithmic tweaks I think there will be visual differences, say it will be difficult not to systematically favor some palette colors against some others. But maybe some machine learning could be used to learn the output of the palettizer shader, then used to replace it. I should give an approximation with (hopefully) not too blatant bias, and replace conditionals with matrices operations that shaders can compute very efficiently...

petchema:
Interesting discussion on Reddit, at any given time (outdoors at least) 32 colors of the palette are reserved for sky colors...
https://www.reddit.com/r/daggerfallunity/comments/ghvkzr/the_new_retro_shader_brings_the_dark_nights_of/

Interkarma:
This is correct, DFU reads that palette data when reading sky images. And it's no problem with full 32-bit colour, as DFU just converts the SKY images to a 32-bit texture internally. I can see how this might cause issues with only a subset of that larger virtual colour space available though.

It's a hard one to improve further without just adding more colours to the virtual palette available to retro shader. Right now, I think the results look really good almost all of the time. And the posterization shader is a great alternative if this bothers anyone.

It's very easy to get caught up in the minutia of something like this. But the rendering tech is so wildly different between a bespoke 1996 software renderer and today's GPUs, that it's quite amazing what you've been able to accomplish. It looks and feels just like classic to me.

petchema:
I had trouble sleeping on it lol
I was wondering if we could use the z-buffer from the first rendering to tell apart the sky and not palettize it... After all classic skies should already palletized, and even if some seem to have up to 64 colors instead of 32 that should look fine.
That would incur no additional overhead for other non-retro mode rendering, and retro mode is already incompatible with Enhanced Sky... 
hybrid

petchema:
Mmmh I give it a try, but so far I couldn't get access to depth texture. 
https://github.com/Interkarma/daggerfall-unity/compare/master...petchema:exclude-sky-from-palettization

Also, that blurb from https://docs.unity3d.com/2018.2/Documentation/ScriptReference/Graphics.Blit.html didn't look engaging:
"Note that if you want to use depth or stencil buffer that is part of the source (Render)texture, you'll have to do equivalent of Blit functionality manually - i.e. Graphics.SetRenderTarget with destination color buffer and source depth buffer, setup orthographic projection (GL.LoadOrtho), setup material pass (Material.SetPass) and draw a quad (GL.Begin)."
I'm not sure it's the problem I'm experiencing though

Interkarma:
Thank you for all the information. You've got me intrigued, I'll see what I can do when I can later today. :)

Interkarma:
I've had a play around with this. I can't say I've made significant progress, but the following method gave me good results as a starting point. However, I encountered a couple of issues that I haven't been able to resolve yet. Need more time. Here's the gist:

1. When Retro Mode is enabled, redirect DaggerfallSky rendering to its own RenderTexture. This can be a fixed size like 640x400. Don't use postprocessing.
2. Clear and render to 320x200 or 640x400 target as normal, with postprocessing if enabled.
3. Combine RenderTexture output from 1 (sky) & 2 (world) and present to display.

Step 3 is where I've encountered some problems. It works OK, but blending and clearing everything in the right order was giving me some grief. Just using clip() to drop the fully alpha sky pixels works OK, but also cuts off some things in the foreground for some reason. I'll take another pass at it when I can. I had some other things to work on so wasn't able to get my head fully into it yet.

petchema:
Okay, looks like a good alternate way of reaching the same goal. I assume blitting is relatively cheap?
Just for science, I continued on my idea, and reading more docs the replacement for Graphics.Blit that should do the trick looks like
```c#
// yes, depthBuffer passed as target depthBuffer
Graphics.SetRenderTarget(destination.colorBuffer, retroTexture.depthBuffer);
postprocessMaterial.SetTexture("_MainTex", retroTexture);
GL.PushMatrix();
GL.LoadOrtho();
postprocessMaterial.SetPass(0);
GL.Begin(GL.QUADS);
GL.Vertex3(0, 0, 0);
GL.Vertex3(1, 0, 0);
GL.Vertex3(1, 1, 0);
GL.Vertex3(0, 1, 0);
GL.End();
GL.PopMatrix();
```

Except I had no luck with that either... You'll probably have something working before me

petchema:
I also played with your solution, I didn't get a totally working solution either, but things I've found along the way:
- the sky is no longer drawn on the main camera texture, so you have to change the main camera's clear flags;
- main camera's clear flags are modified by CameraClearManager, so you have to get it to cooperate
Things still get the wrong color as they're further away than in previous frames, like the depth buffer is not cleared correctly...
Also, I get different results depending on whether I start the game inside or outside, with leads me to believe I'm doing something wrong. Maybe we'll get somewhere by sharing experiments ;)
https://github.com/petchema/daggerfall-unity/compare/petchema:lut-palettization...petchema:non-palettized-sky

petchema:
DaggerfallSky also like to mess with main camera clearFlags...
So I'm getting there using hacks, the remaining problem is that trees are cut at the horizon, weird
cut trees

petchema:
That's a problem with "Queue" = "Transparent" tag in billboards batch shader...
with the obvious problem of self shadowing it was there to prevent if you remove it :(
self shadowing

Interkarma:
Thank you, the transparent queue is what I was missing earlier. I'll recreate and share what I did earlier in a PR. Will need to write again, as I just tossed the experiment. Shouldn't take long though, it was only a small number of changes.

PR #1808, opened 2020-05-03
Fix horse gallop sound clipping when repeatedly jumping

jefetienne:
Holding space while jumping makes an unpleasing clipped horse noise. This fix delays the stopping of the audio by 0.2 seconds, just enough to get at least one gallop noise in when colliding with the ground. It's not perfect, and sometimes it might fail to make a sound, but I think it's a step up from what it currently is at the moment.

(Also, I think my VSCode editor is automatically normalizing the line endings and I'm having trouble finding another editor that won't do this, sorry about that)


Interkarma:
This works well, thank you! :)

PR #1809, opened 2020-05-03
Added option to toggle sneak

jefetienne:
You can now enable the sneak action to be toggled rather than held. Running will break you out of the sneak mode.


Interkarma:
Thanks for adding this so quickly, amazing work! :)

Interkarma:
Works very well, thanks again! :)

PR #1810, opened 2020-05-05
Add genuine keyboard event handlers to UI controls

### LINKED ISSUE #1805: Key press event leak from video player to Start menu
**Describe the bug**
I am on first dungeon i died and when i click load saved game the game just crashed

i have no mods installed

**To Reproduce**
1. Dead
2. Click load saved game
3. Crash

**Expected behavior**
The game not crashed and i can load my saved game without exiting 

 - OS: Windows 10 1903 OS Build 18362.418
- CPU : AMD A9-9425
- RAM : 4 GB
- GPU : Radeon R7 M440
- Version Alpha v0.10.22

petchema:
This is a fix for the issue discussed in #1783:
    
Keyboard events can leak from game to UI or from UI to game:
- to prevent leaks from UI to game, UI controls can be triggered on KeyUp, so that the key is released before the game takes control back. However if the player releases game keys after opening the UI, that can activate UI buttons ("W" in inventory, "S" in spellbook,...);
- to prevent leaks from game to UI, UI controls can be triggered on KeyDown, so that keys have to be pressed again inside the UI to have an effect. However this means keys can still be pressed after the UI closed, leaking key events into the game.
    
It should be clear now that those two situations are symetric, and that none fixes all problems. UI controls need to be able to handle KeyDown and KeyUp explicitly.
    
With this PR, buttons can:
- (unchanged) have a OnMouseClick handler; They will only react to mouse events;
- (unchanged) also have a Hotkey hotkey sequence; When the sequence is pressed, the OnMouseClick handler receives the event as a faked click;
 - but now also have a OnKeyboardEvent handler; In that case no click will be faked, and the keyboard handler will receive both KeyDown and KeyUp events.
    
With that in place, controls should react to KeyDown events in general, but the buttons that close the UI, or maybe even yield control to another UI window, should use a keyboard handler. For safety, they should be armed by KeyDown events (possibly giving user feedback by emitting a sound), and activated by KeyUp events if previously armed.
    
I'm very pleased with this, feels like sound design.I even used it to fix #1805 and simplify fast travel popup code :)
It needs more testing, but I'm releasing it for public scrutiny.

### DIFF
--- a/Assets/Scripts/Game/DaggerfallUI.cs
+++ b/Assets/Scripts/Game/DaggerfallUI.cs
@@ -377,6 +377,7 @@ namespace DaggerfallWorkshop.Game
             // Possible to get multiple keydown events per frame, one with character, one with keycode
             // Only accept character or keycode if valid
             lastKeyModifiers = HotkeySequence.GetKeyboardKeyModifiers();
+            hotkeySequenceProcessed = false;
 
             if (Event.current.type == EventType.KeyDown)
             {
@@ -388,11 +389,7 @@ namespace DaggerfallWorkshop.Game
 
                 if (lastCharacterTyped > 255)
                     lastCharacterTyped = (char)0;
-            }
 
-            hotkeySequenceProcessed = false;
-            if (Event.current.type == EventType.KeyUp)
-            {
                 hotkeySequenceProcessed = ProcessHotKeySequences();
             }
 
--- a/Assets/Scripts/Game/DaggerfallUI.cs
+++ b/Assets/Scripts/Game/DaggerfallUI.cs
@@ -393,6 +393,11 @@ namespace DaggerfallWorkshop.Game
                 hotkeySequenceProcessed = ProcessHotKeySequences();
             }
 
+            if (Event.current.type == EventType.KeyUp)
+            {
+                hotkeySequenceProcessed = ProcessHotKeySequences();
+            }
+
             if (Event.current.type == EventType.Repaint)
             {
                 RenderTexture oldRT = null;
--- a/Assets/Scripts/Game/UserInterface/BaseScreenComponent.cs
+++ b/Assets/Scripts/Game/UserInterface/BaseScreenComponent.cs
@@ -88,6 +88,9 @@ namespace DaggerfallWorkshop.Game.UserInterface
         float minAutoScale = 0;
         float maxAutoScale = 0;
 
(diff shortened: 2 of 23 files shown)


petchema:
Switching in-game shortcuts to also work on KeyDown would require the same "arming" mechanism to work correctly, maybe I'll work on that in a separate PR

jefetienne:
I think I misread your comment from earlier - after testing this PR I now what see what you were saying and trying to do, sorry about that! This looks great. It was annoying to be holding 'W' while accessing the inventory screen which would prompt you to select 'Wagon' after the KeyUp event.

petchema:
No problem, you know first hand that input handling can be tricky and I didn't take the time before to explain exactly what the problem was and what I had in mind, I'm glad my short explanation and actual code made everything clearer (and even more glad you like it ;) )
I don't think it conflicts with anything you added, but if you detect some problem let me know.

Interkarma:
I'd love to spend some more time with this one. Are you happy if this is scheduled for 0.10.24? I'm aiming to get 0.10.23 out this weekend with all currently merged PRs. Have a lot of final testing before then, and some writing for the next article. .23 will a big release already. :)

petchema:
Sure, I don't want to delay 0.10.23, so much good stuff needs to be released to the world :)
I'll work on switching the in-game shortcuts to KeyDown triggered so that this PR feels complete and we'll have more time to review and test it after 0.10.23

Interkarma:
This works really well! I'm happy to merge this into master. Thank you for all your work on this, it's a nice improvement.

I spent a little while thinking about the below repeating pattern and how it could be made more generic or moved into base class support. I couldn't conceive of anything simpler and more readable than just doing it the way you did, although it feels like there should be a nice robust solution I just didn't see. But it's not important at all, and I don't need you change anything, it's just something I was idling around in my head. :)

    if (keyDown)
    {
        PlayOneShot(sound)
        isDeferred=true;
    }
    else if (keyUp && isDeferred)
    {
        isDeferred=false;
        PostMessage(message);
    }

petchema:
It's possible to bundle the hotkey and the boolean in some object, and simplify those calls (and reduce the scope of those booleans)
I'm sure we could build better abstractions around keyboard events (or user events in general now), but I wasn't sure that this abstraction was good enough, so I left the code as-is.
But yeah.

PR #1811, opened 2020-05-06
In game shortcuts on keydown

petchema:
Open UI windows on KeyDown event, to make the UI feel snappier.
This requires UI to handle spurious KeyUp events, so is a PR on top of #1810

### DIFF
--- a/Assets/Scripts/Game/DaggerfallUI.cs
+++ b/Assets/Scripts/Game/DaggerfallUI.cs
@@ -377,6 +377,7 @@ namespace DaggerfallWorkshop.Game
             // Possible to get multiple keydown events per frame, one with character, one with keycode
             // Only accept character or keycode if valid
             lastKeyModifiers = HotkeySequence.GetKeyboardKeyModifiers();
+            hotkeySequenceProcessed = false;
 
             if (Event.current.type == EventType.KeyDown)
             {
@@ -388,11 +389,7 @@ namespace DaggerfallWorkshop.Game
 
                 if (lastCharacterTyped > 255)
                     lastCharacterTyped = (char)0;
-            }
 
-            hotkeySequenceProcessed = false;
-            if (Event.current.type == EventType.KeyUp)
-            {
                 hotkeySequenceProcessed = ProcessHotKeySequences();
             }
 
--- a/Assets/Scripts/Game/DaggerfallUI.cs
+++ b/Assets/Scripts/Game/DaggerfallUI.cs
@@ -393,6 +393,11 @@ namespace DaggerfallWorkshop.Game
                 hotkeySequenceProcessed = ProcessHotKeySequences();
             }
 
+            if (Event.current.type == EventType.KeyUp)
+            {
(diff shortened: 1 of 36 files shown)


petchema:
One problem with such change is the compatibility with some mods (say, Tedious Travel that overrides the travel map)

petchema:
Is InputManager state still updated while UIs are displayed?
If so maybe I should depend more on ActionStarted/ActionCompleted etc.
All I may need is some ActionAcknowledged() or similar that would filter KeyDown/KeyUp/ActionStarted/ActionCompleted from a key until it's no longer pressed...

jefetienne:
> Is InputManager state still updated while UIs are displayed?

Unfortunately the `InputManager.Update` method checks for an `IsPaused` state and `return`s if so, preventing `FindKeyboardActions` from running.

There should be just a `GetKey` function that you can use to check if something is currently being held?

petchema:
Well, as you can see in the code of this PR, this is highly insufficient.
I want action keys to close UIs on KeyUp so that no event for that key leaks into the game, but since I also want those keys to open the UI on KeyDown, I have first to wait for the key to be released. I have to implement that "rearm" logic in each UI window, but doing so I still break modded UIs that won't immediately implement the same logic: right now the action keys being used as toggle, when releasing the action key the modded UIs immediately close.
That's why I'm looking for solutions that move that "rearm" logic responsability in InputManager.Instance.GetKeyUp() so that modded UIs benefit from it immediately...

petchema:
Actually what Update() still does may be sufficient, I must investigate that further...

petchema:
Coded something, but Tedious Travel uses Input.GetKeyUp(), so my trick in InputManager.GetKeyUp() doesn't work...
https://github.com/TheNewBob/TediousTravel/blob/master/Scripts/TediousTravelMap.cs#L378

jefetienne:
Well I can say for one thing that's it recommended that any future keybindings, mod or feature, should switch to using InputManager.GetKeyX instead of Unity's Input.GetKeyX so that they can be recognized by the controller (unless if it's checking for a hardcoded keyboard key like LeftShift or Escape).

Also, I found out that I forgot to change it to that in the DaggerfallPauseOptionsWindow and it generates an Exception if I bind Pause to a joystick axis, ouch.

jefetienne:
I tested your PR and it seems to work well!

Interkarma:
I'll need to take a look at this one again later. It works well as it stands, but we're introducing a lot of complexity to support opening UIs on keydown. And potentially breaking mods as you mention. I don't generally notice the difference between keydown and keyup opens unless I specifically look for it, I just tap the key and the UI opens. But when I do look for it, I do notice the delay.

Anyway, will loop back to this one when I can. :)

jefetienne:
This PR might just need to be rewritten with the changes from the combo codes and secondary keybinds. It actually might even be easier to do since I now introduced a polling system that keeps track of held keys and that this `AcknowledgeKey` functionality in this PR is probably redundant.

PR #1812, opened 2020-05-06
Improve handling of mod dependency issues

TheLacus:
- All dependency issues are shown on message box with a scrollbar if needed.
- Prompt to auto sort mods is shown only if at least one issue can be solved by sorting.

checkDependencies


Interkarma:
Thank you, I'll fast-track this one into 0.10.23.

PR #1814, opened 2020-05-07
Early May 2020 quest fixes

JayH2971:
Baltham Greyman quests still not ready, but much progress so far. Also fixing some typos in quests.

### DIFF
--- a/Assets/StreamingAssets/Quests/10C00Y00.txt
+++ b/Assets/StreamingAssets/Quests/10C00Y00.txt
@@ -127,7 +127,6 @@ Person _qgfriend_ group Librarian anyInfo 1011 rumors 1012
 Place _mondung_ remote dungeon
 
 Clock _1stparton_ 00:00 0 flag 1 range 2 5
-Clock _delay_ 01:00 0 flag 1 range 0 1
 
 Foe _monster_ is Sorceror
 Foe _daedra1_ is Fire_daedra
@@ -148,7 +147,8 @@ variable _2ndparton_
 _pcgetsgold_ task:
 	when _npcclicked_ and _mondead_ 
 	give pc _artifact_ 
-	start timer _delay_ 
+	end quest 
+--removed unnecessary delay
 
 _mondead_ task:
 	killed 1 _monster_ 
@@ -162,9 +162,6 @@ _clearclick_ task:
 	when _npcclicked_ and not _mondead_ 
 	say 1003
 	clear _npcclicked_ _clearclick_ 
-
-_delay_ task:
-	end quest 
 
 _S.07_ task:
 	injured _monster_ 
--- a/Assets/StreamingAssets/Quests/20C00Y00.txt
+++ b/Assets/StreamingAssets/Quests/20C00Y00.txt
@@ -138,7 +138,6 @@ Place _tavern_ remote tavern
 Place _hideout_ remote dungeon11
 
 Clock _1stparton_ 00:00 0 flag 1 range 0 5
-Clock _delay_ 01:00 0 flag 1 range 0 1
 Clock _S.08_ 00:01 00:04
 
 Foe _monster_ is Mage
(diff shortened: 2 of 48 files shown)


Interkarma:
I'll hold this one for 0.10.24 and provide some help on the Baltham quest when I can. :)

Thank you Jay for all your help.

JayH2971:
Yes sir! It's an odd glitch and I don't understand it, so any help is welcome.

Interkarma:
@JayH2971 I can make time to help with the Baltham Greyman issue now. Can you please bring me up to speed and I'll see what I can do. Is it just that sprite remains visible on S.14 route, as in quest comment?

JayH2971:
Yep, that's the only issue remaining for that quest. The rest of it
should work fine. The other Baltham Greyman quest can't start due to
invalid placemarker, presumably in Direnni Tower.

Interkarma:
I've started the S.14 route by saying "no" to going it alone and met with Baltham back at the inn the requisite number of days later (player really needs to keep track of that, yikes). When clicking on Baltham to pick him up again, his escort portrait gets added almost every frame by S.19 "add _rebel_ face". Is this the issue you mean (click on Baltham to trigger)?

[SAVE895.zip](https://github.com/Interkarma/daggerfall-unity/files/4665111/SAVE895.zip)

JayH2971:
Oh heck, that didn't even happen to me. You're seeing the persistent sprite bug already, but I never saw the portrait spam you also have.

Interkarma:
Gotcha. Will look at both today. Thanks Jay. :)

Interkarma:
I understand the portrait spam. "When" is an "always on" trigger condition and there are two different "when" actions on that block.

After clicking Baltham, the following "when" condition is evaluated first. As the primary condition, this can switch task ON or OFF again. This one is always false so switches task OFF each tick of quest machine.

    when _S.03_ and _house1_ and _spouse_

Then the next line evaluates to true. As a secondary condition, this can only switch task ON. But because these are true, task is enabled.

    when _S.14_ and _S.03_

So what's happening is the task gets switched off then back on every tick, rearming everything under that task. And because "add face" doesn't protect itself against rearming, it just spams the face into the HUD. I'll tighten this up first and make "add face" protect itself against rearm spam. Swapping the order of evaluation around like below should also fix that problem. But if these conditions are both contingent on different pathways, probably best not to evaluate them together as one or the other will always be false first.

    when _S.14_ and _S.03_
    when _S.03_ and _house1_ and _spouse_

Haven't looked into the sprite removal yet. It might be related or not. Will add another post once I've investigated.

Interkarma:
For the NPC not disappearing, this is because Baltham is being assigned to tavern _twice_ somewhere along the line.

image

So one gets disabled by `hide npc _rebel` but the other is still standing there. Baltham The Many strikes again.

I'll chase this one down and see if I can remediate in backend first.

Interkarma:
Both issues should be resolved now in 03f8e29 and 87d709d with no changes required to quest script.

Please note that Baltham Two-Times can still happen if player has triggered issue in their save prior to patch. Probably won't happen often, but good to keep in mind in case someone reports it again sometime in future. Would definitely need a save off them to confirm and review resource placements to confirm duplicate in save.

JayH2971:
Cool! Thanks for taking a look at that. I hope to require that as little as possible.

Interkarma:
It's my pleasure Jay. I'll also get to those Classic Quest bugs you've added as soon as possible. I'm getting pulled in a lot of different directions at the moment, so just bouncing around doing what I can when I have a time window.

Interkarma:
Hi @JayH2971 thank you for all the fixes! I'm happy to merge this one into 0.10.24, just confirming you're happy too and there were no remaining tweaks to PR?

JayH2971:
I'm done with all current fixes! :)

PR #1815, opened 2020-05-08
Added ability to abort spell via Ready Weapon action

jefetienne:
Added functionality where pressing "Ready Weapon" when having a spell ready to cast will cancel it as an alternate way to abort a spell. The weapon will be unsheathed regardless if it was previously or not.

### DIFF
--- a/Assets/Scripts/Game/WeaponManager.cs
+++ b/Assets/Scripts/Game/WeaponManager.cs
@@ -60,6 +60,7 @@ namespace DaggerfallWorkshop.Game
         private int _longestDim;                    // Longest screen dimension, used to compare gestures for attack
 
         PlayerEntity playerEntity;
+        EntityEffectManager entityEffectManager;
         GameObject player;
         GameObject mainCamera;
         bool joystickSwungOnce = false;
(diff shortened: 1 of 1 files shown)


petchema:
I wonder if we could display the hands of the caster while a spell is readied, that would be another way to avoid confusion...
The gfx exists, it's record 0 of each casting animation:
casting

jefetienne:
I agree, I think it would be great visual input for the user. I remember my first time playing and being so confused why I couldn't press anything.

Assets/Scripts/Game/WeaponManager.cs
```diff
@@ -228,25 +230,40 @@ void Update()
...
             // Hide weapons and do nothing if spell is ready or cast animation in progress
             if (GameManager.Instance.PlayerEffectManager)
             {
                 if (GameManager.Instance.PlayerEffectManager.HasReadySpell || GameManager.Instance.PlayerSpellCasting.IsPlayingAnim)
                 {
-                    ShowWeapons(false);
-                    return;
-                }
-            }
+                    if (!isAttacking && InputManager.Instance.ActionStarted(InputManager.Actions.ReadyWeapon) && entityEffectManager)
```
Interkarma, line 247: You can use GameManager.Instance.PlayerEffectManager here rather than grab another reference in Start.

jefetienne: Updated!

jefetienne:
> I wonder if we could display the hands of the caster while a spell is readied, that would be another way to avoid confusion...
> The gfx exists, it's record 0 of each casting animation:
> casting

After discussing about it in the forums, I've now created a [mod](https://forums.dfworkshop.net/viewtopic.php?f=14&p=44449) that implements this feature. :)

Interkarma:
This works great, thank you! :)

PR #1816, opened 2020-05-09
[Joystick Controls] Added and applied joystick INI settings

jefetienne:
- Decoupled the mouse sensitivity from the joystick camera movement with `JoystickLookSensitivity`
- Added `JoystickCursorSensitivity` (TODO: figure out how to apply ini setting to the setup wizard)
- Added `JoystickMovementThreshold`, which allows the player to set a threshold to reach maximum without moving the cursor all the way. Sometimes, worn controller joysticks might not reach the '1.0' maximum axis movement, which would hinder their movement speed.
- Force `MouseSmoothing` in `PlayerMouseLook` when using a controller

For now, adjusting these settings will have to be done in the settings.ini file. I will release the UI for the joystick controls window after this and another PR get merged.


Interkarma:
Looks good! Thank you. :)

PR #1818, opened 2020-05-10
Strengthens/Weakens Armor bug fix

dmirizzi:
Fixed strengthen armor enchantment - armor modifier is now correctly applied to player entity and correctly reflected in paperdoll


Interkarma:
Hey, thanks for taking a look! :)

When I credit you, do you prefer to be called dmirizzi or MaxRebo like on forums?

dmirizzi:
> Hey, thanks for taking a look! :)
> 
> When I credit you, do you prefer to be called dmirizzi or MaxRebo like on forums?

MaxRebo is fine :)

PR #1819, opened 2020-05-10
Added Autorun action and keybind

jefetienne:
Pressing the middle mouse button will toggle autorun mode, which defaults to sprinting forward without holding any keys. While in autorun mode, the run key also becomes togglable, so you can resort to simply pressing the run key once to toggle it between autowalking and autorunning.

Pressing the Autorun key again, or just moving backwards, will cancel it.

I rearranged the order of the advanced control keybinds so similar actions are grouped on top of each other.

### DIFF
--- a/Assets/Scripts/Game/InputManager.cs
+++ b/Assets/Scripts/Game/InputManager.cs
@@ -267,7 +267,6 @@ namespace DaggerfallWorkshop.Game
             Crouch,
             Slide,
             Run,
-            AutoRun,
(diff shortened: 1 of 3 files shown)


Interkarma:
Nice work! I like the choice of middle mouse button as default. :)

Interkarma:
I particularly like how you can toggle between autowalk/autorun by tapping the toggle run key. It's a very nice implementation.

PR #1823, opened 2020-05-15
Lookup table palettization

petchema:
Very fast palettization (up to 280~300 fps on my hardware) using a Texture3D as lookup table.
I get results pretty close to my previous shader by dropping 1 bit from each color component (128x128x128 lookup table). Texture takes 8MB and requires ~~2.4s~~ 850ms (on my hardware, again) to initialize.
Other tradeoffs can be made by modifying lutShift in RetroRenderer.cs (that I haven't exposed as a setting... yet?)
I could also lower memory requirements to 1/4th by only storing palette indexes in the lookup table, at the price of an additional lookup in color palette...


petchema:
Benefits:
- no external tools required, the lookup table is generated on the fly from the provided palette;
- could potentially be used with higher resolutions

petchema:
initialization time down to 850ms using Color32

petchema:
0 10 23 no postprocessing
0 10 23 posterization
0 10 23 kdtree palettization
lut palettization lutshift=0
lut palettization lutshift=1
lut palettization lutshift=2
lut palettization lutshift=3
lut palettization lutshift=4
lut palettization lutshift=5
lut palettization lutshift=6

petchema:
I'm reverting FilterMode.Point, because it's rounding access and I want "rounding down" (floor) instead.
Why is it so hard to get that kind of information...

petchema:
Fix the issue differently; Since FilterMode = Point does rounding in truecolor space, remove the rounding inside the lookup table. The result is not strictly the same, but very similar and the implementation simpler.

Interkarma:
Superseded by #1828. Thank you @petchema! :)

PR #1825, opened 2020-05-17
[RFC] Travel map dot outlines

petchema:
Make region maps more accessible to color blinds by adding an outline to location pixels.
It actually makes it easier to point at locations, so it could be seen as a general improvement.

Forums: https://forums.dfworkshop.net/viewtopic.php?f=4&p=43718#p43717

The implementation uses Panels for the outlines, works well but may be too hackish?
Also, ir probably needs to be made an option or have a keyboard toggle.
outlines 2 - 128

### DIFF
--- a/Assets/Resources/defaults.ini.txt
+++ b/Assets/Resources/defaults.ini.txt
@@ -61,6 +61,8 @@ AccelerateUICopyTexture=False
 SDFFontRendering=True
 EnableGeographicBackgrounds=False
 DungeonExitWagonPrompt=True
+TravelMapLocationsOutline=True
+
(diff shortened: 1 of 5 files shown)


petchema:
Made it a new setting (default on, it really helps with visibility)

ajrb:
Nice to have an accessibility option, but not convinced it should be on by default as it changes the look from classic quite a bit.

petchema:
Yup, seems fair

Interkarma:
I like this one a lot. Thank you for your efforts on this @petchema and communicating with forum members to get a good result.

PR #1827, opened 2020-05-20
Cast when held item condition lowering

df-ferital:
The behavior was reverse engineered on DF forums by Numidium and Elenwel and confirmed by my own reverse engineering, see here: https://forums.dfworkshop.net/viewtopic.php?f=23&t=3621

Since @Allofich did most of the job some time ago, there was very little to do there.

The only difference from classic is that DFU takes all item "Cast when held" effect costs to lower the item condition, where classic takes only the last one.

### DIFF
--- a/Assets/Scripts/Game/Formulas/FormulaHelper.cs
+++ b/Assets/Scripts/Game/Formulas/FormulaHelper.cs
@@ -2223,11 +2223,13 @@ namespace DaggerfallWorkshop.Game.Formulas
         }
 
         /// <summary>
-        /// Reversed from classic. Calculates cost of casting a spell.
+        /// Reversed from classic. Calculates cost of casting a spell. This cost is also used
+        /// to lower item condition when equipping an item whith a "Cast when held" effect.
         /// For now this is only being used for enchanted items, because there is other code for entity-cast spells.
         /// </summary>
         /// <param name="spell">Spell record read from SPELLS.STD.</param>
-        public static int CalculateCastingCost(SpellRecord.SpellRecordData spell)
+        /// <param name="enchantingItem">True if the method is used from the magic item maker.</param>
+        public static int CalculateCastingCost(SpellRecord.SpellRecordData spell, bool enchantingItem= true)
         {
             // Indices into effect settings array for each effect and its subtypes
             byte[] effectIndices = {    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, // Paralysis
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Game/Formulas/FormulaHelper.cs
```diff
@@ -2340,11 +2357,8 @@ public static int CalculateCastingCost(SpellRecord.SpellRecordData spell)
             byte[] rangeTypeModifiers = { 2, 2, 3, 4, 5 };
 
             int cost = 0;
-            int skill = 50;  // 50 is used for item enchantments, which is all this spell is used for now.
-                             // This function could be used as in classic to get casting costs for spells
-                             // cast by entities by replacing skillUsedForEnchantedItemCost with the skill
-                             // of the caster in the effect's spell school.
-
+            int skill = 50; // 50 is used for item enchantments
```
numidium, line 2360: When I was working on reversing this I found that it actually uses the player's skill in classic. Elenwel confirmed this here: https://forums.dfworkshop.net/viewtopic.php?f=23&t=3621&sid=a1aa5bf0451e688b46ee586f57093325#p43434

It makes more sense to me to have it use a constant skill value for enchanted items but if we're going for parity with classic it should use the player's skill.

df-ferital: It uses the payer skill, see the code I added below. The 50 constant is only kept for the Magic item maker. :)

numidium: That'll teach me to code review while distracted. Great work!

Assets/Scripts/Game/Formulas/FormulaHelper.cs
```diff
@@ -2223,11 +2223,12 @@ public static int GetSpellEnchantPtCost(int spellIndex)
         }
 
         /// <summary>
-        /// Reversed from classic. Calculates cost of casting a spell.
-        /// For now this is only being used for enchanted items, because there is other code for entity-cast spells.
```
Interkarma, line 2227: Please leave this comment in place. Spell point costs really do have a different formula for entity-cast spells that emulates the classic formula while allowing modded effects to set their costs. In DFU it's possible to have any number of custom spells effects, and they need to provide costings too.

At some point enchantment costings need to be moved over the new formula anyway, as it's currently limited to the classic effect only. This is another TODO item I need to loop back to as part of spell mod support.

Interkarma:
Amazing work on this everyone involved @numidium, Elenwel, @df-ferital. :)

Assets/Scripts/Game/Formulas/FormulaHelper.cs
df-ferital: No problem, I restored the comment. While doing so, I also removed the last element of magic effect `settingsTypes` array in `CalculateCastingCost` since it is supposed to be an array of 51 bytes, not 52. This matches classic and was probably a bad copy/paste from the original exe data.

Interkarma: Thank you. :)

PR #1828, opened 2020-05-20
New retro postprocessing setup with depthbuffer filtering for sky

Interkarma:
@petchema I'm pretty happy with this one now. Promoting to a PR for final review.


Interkarma:
Have made those final changes and tested in builds. Everything looks good so far.

petchema:
Yes, standalone player works fine again

petchema:
By the way, since I merged #1823 in this PR, it supersedes it.

Interkarma:
I've run through the main graphical mods (excluding enhanced sky of course) and they work great under new setup. I'm happy we spent the extra time to fix particles. It took three different approaches between us with lots of iteration on each one to get here, but the end result was worth it. :)

petchema:
Haha, yes I saw the twitts of yours. Had to check the well myself, it's pretty sweet indeed.
I played/tested with retro mode improvements on too, no problem to report. The only case where the last trick is slightly visible is under foggy conditions:
retro fog
But it's okay I guess, we'll see if anybody notices without knowing beforehand where to look at.

My two personal favorite effects of palettization are

- landscapes, with mipmaps on. This turns many landscapes into abstract paintings, it may not be the greatest setup for finding dungeons but in so many places I really enjoy the results;

- dungeons. I remember a player (Grimith I think) saying that the long view distance is great, but at the same time removes a lot of the claustrophobic sensation you get from only seeing 20 feet away. Well, with palettization and lowered ambient light, enemies have again many corners to hide in, no matter how you setup your gamma. That lead to interesting situations, from getting shot by arrows while starring at a dark corridor, or finding unexpected enemies while retracing your steps back, etc. So it has an impact on gameplay too...

petchema:
By the way I used Texture3D, it seemed well supported but I could have been optimistic thee.
At least retro mode is only an option, but maybe some checks or fallback will be necessary...

Nystul-the-Magician:
does one of you know what causes the problems with the enhanced sky. I would like to take a look and see what causes the issue so any good starting point for hunting down the issue would be welcomed from my side ;)

petchema:
Hi Nystul!
Most likely, the extra camera(s) have to switch to the right RenderTexture target.
If that's already supposed to happen, then I don't know...

Nystul-the-Magician:
ok, should be possible to make this work ;)

Interkarma:
Hey Nystul! As Pango says, it should be enough just to set the correct render target of enhanced sky camera(s). See the Update() method in RetroRenderer.cs to see how this is done for the builtin sky.

You can get the current retro texture target from RetroRenderer.RetroTexture property. And RetroRenderer component is found on the MainCamera object as of #1828 so should be easy enough to GetComponent<RetroRenderer>() off that object.

Enchanced sky should already be using a camera depth < 0 (default sky is -3). Provided you render into correct target before camera depth 0, everything should come out OK. I'm assuming distant terrain cameras are also < 0 depth.

Nystul-the-Magician:
ok, thanks, I know where to look ;)

Nystul-the-Magician:
there is a new version 2.1.2 in first post of forum thread: https://forums.dfworkshop.net/viewtopic.php?f=27&t=1542

petchema:
The timing is strange though, because in 0.10.24 the main change in retro mode postprocessing was that the sky would not be postprocessed...
The issue we had in 0.10.23 is that, because in classic the sky uses colors from a reserved subset of the palette that the game would adjust for each sky texture, palettizing the sky using the world colors would give wrong results. Cheap fix was to not palettize the sky, since sky textures use some limited amount of colors already.
I suspect non-palettized enhanced sky is not what most players expect, so I may have to extend the shader to palettize or not the sky...

petchema:
Done!
Retro mode + Enhanced Sky
no postprocessing:
no pp
posterization (full):
posterization full
posterization, not including the sky:
posterization no sky
palettization (full):
palettization full
palettization, not including the sky:
palettization no sky

Nystul-the-Magician:
looks good

petchema:
For classic sky, I thnk it works better without palettization.
But with the last commit I added, the choice will be up to the player

Nystul-the-Magician:
excellent ;)

Interkarma:
This looks amazing Nystul!

Interkarma:
I'm loving all the work done on this. Merging to master now.

PR #1829, opened 2020-05-22
Fix daedra summoning to match classic behavior

df-ferital:
I reverse-engineered classic function regarding daedra summoning. And as often, I noticed it somewhat contradicts the information found on the UESP. In particular, there is not such thing in classic code as a 95% maximum chance of summoning a dadrea (it's always a roll between 1 and 100), where the "5 % chance" left would be the chance for Sheogorath to replace the daedra.

In classic, it is as follows:
Sheogorath has a 5% chance of replacing the daedra (15% under thunderstorms), so he gets a dedicated roll for that, even with The Glenmoril Witches who usually summon Hircine (this is intentional).
Then, and only after Sheogorath has replaced or not the daedra to summon, the summoning chance is computed and the summoning success or failure roll is done.

I modified DFU's code accordingly.


ajrb:
Looks good at a quick read.

Interkarma:
Thank you @df-ferital :)

PR #1830, opened 2020-05-22
fix encumbrance rounding errors

petchema:
Convert weight computations to integers using mulitples of gold piece weight.
This fixes possibles minor errors, the most visible being you can sometimes pick an additional item by trying again. Round weights may also be displayed in the interface with .00 decimals when they shouldn't.(and removes the warnings in the editor that this problem could happen).
I kept the getters for weights in Kg for mods compatibility.


petchema:
By the way this PR makes encumbrance computation a little less obvious, so this can be discussed.
Classic using 025Kg unit for storing weights makes me believe it uses sum(classic weights) + ceiling(gold/100) kind of formula, which requires special handling for gold in each container and is not obvious either.
This can also be fixed by using "decimal" type. If somebody is a specialist in numeric stability (I'm not), it may also be possible to fix rounding issues with plain floats.
We may also decide such non game-breaking bugs are not worth fixing...

Interkarma:
>This can also be fixed by using "decimal" C# type.

I can follow what you've done, but it does seem like a lot to fix a minor precision issue when a higher precision type is available. Would you be happy to try that approach instead?

petchema:
Fact is, I already tried:
https://github.com/Interkarma/daggerfall-unity/compare/master...petchema:decimal-weights
I'm not sure what are the consequences of this solution though. That type works fine, but seems a bit obscure and bulky, outside of financial applications I guess; I assume it''s not be compatible with other numerical types at the signatures level, so it probably breaks mods compatibility, if things are not done carefully (keep around a method that returns a float type for compatibility, something like that).

Interkarma:
Fair enough, thank you for the detail. In that light, your approach in PR is probably best overall for now.

I'll leave this one open for now and we'll see where it goes. :)

PR #1831, opened 2020-05-23
Refactor CreateCharClassQuestions for greater readability

numidium:
My current iteration of the GetHeaderIndex method is messy and hard to understand. This is mainly because I manually translated it from assembly code that may have had compiler optimizations applied to it. 
I initially left it as is for fear of breaking it. However, now that I've run it through a driver program to compare results with every possible input on my old and new versions of the method, I'm confident that this won't change functionality.


Interkarma:
I trust your work here. That's a substantial improvement to get same results. Very clever! :)

PR #1833, opened 2020-06-01
Targeted Changes To Access Modifiers For Some UI Windows

magicono43:
Changed access modifiers for UIWindows to allow for more simple modification and overriding of textures and already existing buttons, etc.

Please feel free to make changes if they don't seem necessary.

### DIFF
--- a/Assets/Scripts/Game/TransportManager.cs
+++ b/Assets/Scripts/Game/TransportManager.cs
@@ -135,32 +135,32 @@ namespace DaggerfallWorkshop.Game
         private PlayerPositionData_v1 boardShipPosition;    // Holds the player position from before boarding a ship.
 
 
-        const string horseTextureName = "MRED00I0.CFA";
-        const string cartTextureName = "MRED01I0.CFA";
-        const float animFrameTime = 0.125f;  // Time between animation frames in seconds.
+        protected const string horseTextureName = "MRED00I0.CFA";
+        protected const string cartTextureName = "MRED01I0.CFA";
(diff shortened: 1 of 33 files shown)


Interkarma:
Hey Magicono43! :)

This is a very large number of changes to review, and doesn't support any specific mod. I generally don't put through changes just in case someone needs them later. I prefer to have a clear and stated goal and only the minimum number of changes to support that.

We have put through similar UI protection changes in the past, but they were to support specific mods and of smaller scope. As we're heading towards the beta goalpost, I'm getting shy about introducing regressions to game and prefer to see the smallest number of changes required to support an outcome.

The original design for classic UI system was to be baked in and mods would need to add a replacement UI completely rather than override the existing windows. But as replacement window mods aren't yet supported, I understand the desire to open everything up as a short term measure. It's just not in line with the final destination for UI mods, so I'm being selective. Perhaps we can reduce this to a smaller number of changes that are posted only as required for a specific mod? That makes it easier to review and to regression test.

There are also some issues with formatting - tabs instead of spaces. Please set your code editor to use spaces to ensure the formatting isn't broken. Thank you!

Gav.

magicono43:
Yeah, I had a feeling it would be a bit too broad of a net to be casting just for the possibility of someone wanting to change some UI elements possibly. I do have some more focused changes for mods I am working on, and already have finished, so i'll alter this PR, or make another one with those specifics, thanks Gavin.

Interkarma:
This is all good. Thanks so much. :)

PR #1834, opened 2020-06-02
Set Montalion spell to Recall

df-ferital:
This fixes a classic issue where Montalion vampires are given the Free Action spell, which is completely useless for vampires since they are all immune to paralysis.

Instead, it gives them Recall, as was originally intended, since we can read from from "Vampires of the Iliac Bay":

_Montalion alone have the gift for teleportation, but the other eight have powers of their own._

This classic bug was originally mentionted by DelphiSnake: http://slushpool.dfworkshop.net/BUGS.html


Interkarma:
I'm happy with this, it's a good change. Thank you.

PR #1835, opened 2020-06-02
fix ComputeCanHoldAmount() rounding errors locally

petchema:
Round all inputs into multiples of a gold piece weight then use an integer division.
This works fine with classic weights, because beside gold pieces, all weights are multiple of "classic unit" (250g), which is also a multiple of gold piece weight.

Simplest alternative to #1830


Interkarma:
This works well for me. Thank you for continuing to look at this one. :)

PR #1836, opened 2020-06-04
fix for missing buildings bug (when doing round-trips between locations)

Nystul-the-Magician:
see this forum-thread where the bug was reported:https://forums.dfworkshop.net/viewtopic.php?f=4&t=3391&start=10)

the reason was that StreamingWorld class can reuse terrain array elements when needed - it marks no longer needed terrains as "Pooled" and keeps them in memory until memory is needed for new terrains. If pc turns around and travels back the "Pooled" terrains can get reused. If the "Pooled" terrain  was a location map pixel buildings get destroyed as soon as terrain becomes "Pooled" - so this change enforces that buildings of reused "Pooled" terrains which contain a location are loaded again.

### DIFF
--- a/Assets/Scripts/Terrain/StreamingWorld.cs
+++ b/Assets/Scripts/Terrain/StreamingWorld.cs
@@ -861,6 +861,9 @@ namespace DaggerfallWorkshop
                     terrainArray[index].active = true;
                     terrainArray[index].terrainObject.SetActive(true);
                     terrainArray[index].billboardBatchObject.SetActive(true);
+                    // also set flag for updating locations if this terrain contains a location - so building data gets created again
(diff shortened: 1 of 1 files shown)


Interkarma:
Thank you for looking into this! :)

Other than the case this fixes, what's the outcome here when a pooled terrain that had a location is recycled for world tile that does not have a location? That will need to be regression tested. Will do that as part of review if you haven't had a chance already.

Nystul-the-Magician:
there is a if branch for each case, I only touched the branch that handles reactivating already loaded terrains, if a pooled terrain is recycled for a new map pixel this branch is not reached

Interkarma:
Simple and works well, doesn't introduce any problems I can see. Thank you for the fix @Nystul-the-Magician!

PR #1837, opened 2020-06-04
[RFC] Support all cursor sizes

petchema:
Hardware cursor on Windows only supports up to 32x32 pixels cursors; That has been detected because DREAM defines a 64x64 pixels cursor, but it could only be seen on other platforms (at least on Linux) whereas on Windows you only got a 32x32 downscaled version.

Another consequence is that tooltip text was not correctly positioned relative to the mouse pointer on Linux when using large custom pointers.
(Forums:
https://forums.dfworkshop.net/viewtopic.php?f=5&t=3805&sid=565b12db579c16e22326106c391d1ada )

This can be fixed in several ways:
- require custom cursors to be 32x32 pixels. No code change needed;
- forcefully downscale cursor textures to 32x32 pixels to get uniform results on all platforms. Getting good results is actually a difficult matter, I'm not sure what algorithm does Windows use, many classic algorithms don't handle transparency very well (edit: wrote one: https://github.com/Interkarma/daggerfall-unity/compare/master...petchema:downscale-cursors-larger-than-32x32 );
- try to stick with original cursor size, which may require to turn off hardware cursor depending on platform and cursor size. This is what this PR does. Tooltip text positioning is also adjusted to match.

The main drawback I've experienced with software cursor is that it feels a little less responsive than the hardware accelerated counterpart.

(if anyone has an idea how to make Tooltip.cs change more efficient, I'm interested)

DREAM with hardware cursor:
hardware cursor

DREAM with software cursor:
software cursor

@HybOj any opinion on this, was larger cursors what you expected?


HybOj:
Hi mate, thanks for letting me know!

I did not expect the cursor to be this big, I got used to the size it had
since I was able to import it to the game last year.

I have no idea how this bigger version actually feels in the game, so I
cant say if its good or not for certain. But my feeling is that its TOO big
as it is almost the same size as the item tiles.
Also, compared to standard size of windows cursor, this appears too large.

In my opinion, the gameplay/user experience is the top priority over the
looks, so if the HW accelerated cursor movement feels better, Id vote for
that, even if it requires me to downscale the cursor for next release.

I think I will leave the final decission up to you guys, just please let me
know if I should adjust the cursor to match the new environment,

Thanks :)





čt 4. 6. 2020 v 21:39 odesílatel Pierre Etchemaïté <notifications@github.com>
napsal:

jefetienne:
I haven't actually played with D.R.E.A.M. yet (sorry @HybOj!) but now after installing it, I see that the joystick cursor does not match the overridden texture. I think once this PR is completed and my other PRs are added, I'll see about integrating the cursor code with this so the tooltips work nicely with it.

petchema:
Well, at that point we have to decide what we do next.
Not knowing that cursor size on Linux was not intentional, it didn't feel too large. I was just wondering, looking at Windows  streamers, if they were playing in 4K or something (which could be another concern, are 32x32 cursors sufficient in 4K? Maybe because they used on larger monitors? I'm sure Windows will lift that limitation at some point, if that's not already the case and is just a Unity limitation)
But the lag associated with software cursors in sensible. It's not huge, but studies have shown that every millisecond matters when it comes to cursors...
If anyone wants to experiment the two points above, here's the windows 64bit build I created for the screenshots above: https://ks369561.kimsufi.com/~petchema/gfx/Daggerfall/dfu-win.zip

As I said before, the two other alternatives are to only allow for 32x32 custom cursors (which means a change in next DREAM releases), or to write a good cursor downscaling algorithm...

Other possible improvements concerning custom cursors would be the possibility to define the hotspot position (with an optional xml file? Does it look feasible, @TheLacus?). Right now the hotspot is hardcoded in the top left corner, the cursor provided by DREAM has a slight shadow on the left, so the arrow cannot be perfectly aligned with the left margin and the hotspot is slightly to the left of the arrow tip. If the hotspot could be positioned precisely this could be fixed, or allow for more varied cursor designs...

petchema:
Alternate way to see a larger cursor in action, Zaric Zhakaron just did a Daggerfall Unity stream under KUbuntu: https://www.twitch.tv/videos/640655283
Yet another route would be to embrace different cursor sizes by operating system, and just fix tooltip texts position; The main difficulty is that I haven't seen a way to read cursor size, so it would have to be guessed, and guesswork could "break" in the future...

jefetienne:
> Yet another route would be to embrace different cursor sizes by operating system, and just fix tooltip texts position; The main difficulty is that I haven't seen a way to read cursor size, so it would have to be guessed, and guesswork could "break" in the future...



This probably isn't the best solution, but do you think we could try something like:

1. Get OS information
2. If it's Linux, determine the cursor size by the size of the currently texture
3. Else, assume 32x32

Now I'm not sure if macOS behaves like Linux or Windows when it comes to cursors and downscaling, but it's either going to be one of those two. Another thing I'm unsure about is if a Windows user changes their mouse size in accessibility settings, is it still within a 32x32 box?

petchema:
Well, that's basically what I was suggesting

jefetienne:
Sorry if that sounded redundant, I just wasn't sure if what you meant by "guessing" was getting the cursor texture or not.

Interkarma:
Thank you for your work on this one Pango. I will say that I'm not a fan of software cursors. Well, "aggressively opposed" might be a more fitting description. :) I do like that you made the software cursor a fallback option, but for anyone currently affected and generally happy with a less-than-ideal framerate are going to find their cursor suddenly choppy as well. This might be a deal-breaker for playing the game (it probably would be for me).

I'll play with this one some more and compare it your downsample variant when I can.

petchema:
Agreed, software cursors are not desirable.
One of the reasons I'm also exploring alternatives is because I kind of like the 64x64 hardware cursor on Linux; Even if not intended by KoW, and the resulting cursor bigger than the "system" cursor I use, the VMblast artwork is great and got used to it.
But I see possible problems arising from cursor size depending on operating system, so I'd adjust to downsampled cursor...

Interkarma:
Lots of good stuff in this PR too. Sorry for being down on the hardware cursor. Fixing the position for larger cursors is a huge plus just by itself.

Will loop back to this one once I've had some more time to unpack it. :)

Interkarma:
Will resolve minor conflict and merge this one now, my apologies for delay @petchema.

jefetienne:
I wanted to mention that as of right now using the fake cursor with the controller defaults to the 32x32 cursor texture regardless of platform. I haven't done anything about it because of the conversation above - since it is a software cursor, do we feel alright with modded controller cursor textures being its actual size, or force every texture to be 32x32? I'm leaning towards the former if on Linux a cursor would be 64x64 as a hardware, and keep the controller cursor as 64x64 as well.

Interkarma:
As we're now forcing a software cursor on Windows when greater than 32x32, I think it's OK for fake controller cursor to use modded cursor texture at actual size.

My personal preference on Windows would be to downscale to 32x32 and never use software cursors, but I understand the conflicting requirements across platforms, input methods, and mods here that make it challenging. The default cursor will be fine if nothing else, and mods can adjust their cursor sizes based on feedback if too many people experience cursor lag in software based on their system performance and mod loadout. I'll leave that choice up to mod authors now.

petchema:
@HybOj Well, we can wait for players feedback, but since you already mentioned it, I guess the next release of DREAM will ship with a 32x32 cursor;
A short term workaround would be a mod that provides only a 32x32 cursor to replace to 64x64 one...

petchema:
It seems some systems have issues with software cursor:
https://www.reddit.com/r/daggerfallunity/comments/iyqofh/i_cant_see_my_cursor_in_the_main_menu_or_any_menu/

PR #1838, opened 2020-06-09
Typo fixes in quests part 1

JayH2971:
Mainly using UESP's suggestions as a guide. Order of fixing is random, not yet finished.


JayH2971:
Also accidentally included a notice in M0B00Y17 if you kill the tiger, to go talk to the client to end the quest. Didn't mean to include it in this PR but ¯\_(ツ)_/¯

JayH2971:
Thinking about it, rather submit all typo fixes in one PR. Less searching that way.

PR #1839, opened 2020-06-15
Prompt warnings for modlist mismatches in saves

jefetienne:
[[Forum post](https://forums.dfworkshop.net/viewtopic.php?f=22&p=44732&sid=3c2b933a14b59a211a0b54c797f6cb4d#p44732)]

[[Images](https://imgur.com/a/sDN4tUb)]

SaveData.txt will now serialize information about mods used at the time of saving - `fileName`, `title`, `guid`, `version` and `loadPriority`. While this serialized information is mainly used for displaying mod incompatibilities when a player loads or quickloads a save, they can also be useful for mod creators to warn players about particular versions or incompatibilities with mods in their own way as well.

Loading a game will compare the saved `ModInfo_v1[]` in SaveData.txt by the currently enabled and loaded-in mods against its GUID. If it cannot find the a mod with the same GUID, then it will say "Mod is either not loaded or has been altered." If it has been found, then it will compare against their versions and warn if the loaded mod is older than the recorded one, or if the versions are incompatible (e.g. '1.0' vs 'banana').

There is an issue with the scrollbar getting in the way of the first line of text, but I think that should be resolved as a separate PR just to extend the message box width a bit more.

### DIFF
--- a/Assets/Scripts/Game/Serialization/SaveLoadManager.cs
+++ b/Assets/Scripts/Game/Serialization/SaveLoadManager.cs
@@ -1190,9 +1190,20 @@ namespace DaggerfallWorkshop.Game.Serialization
 
                         if (mod.ModInfo.ModVersion != record.version)
                         {
-                            message.Add("- " + record.title + " (v. " + record.version + ")");
-                            message.Add("Incoming version: '" + mod.Title + " (v. " + mod.ModInfo.ModVersion + ")'");
-                            message.Add(String.Empty);
+                            bool? comp = ModManager.IsVersionLowerOrEqual(record.version, mod.ModInfo.ModVersion);
+
+                            if (comp == false)
+                            {
+                                message.Add("- " + record.title + " (v. " + record.version + ")");
+                                message.Add("Incoming version is older: '" + mod.Title + " (v. " + mod.ModInfo.ModVersion + ")'");
+                                message.Add(String.Empty);
+                            }
+                            else if (comp == null)
+                            {
+                                message.Add("- " + record.title + " (v. " + record.version + ")");
+                                message.Add("Incoming version is different: '" + mod.Title + " (v. " + mod.ModInfo.ModVersion + ")'");
+                                message.Add(String.Empty);
+                            }
                         }
                     }
                     else
--- a/Assets/Scripts/Game/Serialization/SaveLoadManager.cs
+++ b/Assets/Scripts/Game/Serialization/SaveLoadManager.cs
@@ -811,7 +811,7 @@ namespace DaggerfallWorkshop.Game.Serialization
         ModInfo_v1[] GetModInfoData()
         {
             List<ModInfo_v1> records = new List<ModInfo_v1>();
-            foreach (var mod in ModManager.Instance.GetAllMods())
+            foreach (var mod in ModManager.Instance.Mods)
             {
                 if (mod.Enabled)
                 {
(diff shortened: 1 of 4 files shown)


Assets/Scripts/Game/Serialization/SaveLoadManager.cs
```diff
@@ -754,10 +803,31 @@ SaveData_v1 BuildSaveData()
...
             saveData.advancedClimbingState = GameManager.Instance.ClimbingMotor.GetSaveData();
+            saveData.modInfoData = GetModInfoData();
 
             return saveData;
         }
 
+        ModInfo_v1[] GetModInfoData()
+        {
+            List<ModInfo_v1> records = new List<ModInfo_v1>();
+            foreach (var mod in ModManager.Instance.GetAllMods())
```
TheLacus, line 814: You can use the property _Mods_ instead of _GetAllMods()_ to avoid creating a new collection.

Assets/Scripts/Game/Serialization/SaveLoadManager.cs
```diff
@@ -1093,6 +1163,85 @@ IEnumerator SaveGame(string characterName, string saveName, bool instantReload =
...
+                return null;
+            else
+                path = GetSaveFolder(key);
+
+            // Read save game data
+            string saveDataJson = ReadSaveFile(Path.Combine(path, saveDataFilename));
+            SaveData_v1 saveData = Deserialize(typeof(SaveData_v1), saveDataJson) as SaveData_v1;
+
+            // Use dictionary for faster indexing
+            Dictionary<string, Mod> dict = ModManager.Instance.GetAllMods().ToDictionary(m => m.GUID);
```
TheLacus, line 1190: Same as above.

Assets/Scripts/Game/Serialization/SaveLoadManager.cs
```diff
@@ -1093,6 +1163,85 @@ IEnumerator SaveGame(string characterName, string saveName, bool instantReload =
...
+
+            if (saveData.modInfoData != null && saveData.modInfoData.Length > 0)
+            {
+                // Verify that every mod recorded in the save is included in the current mod list, or has a newer version loaded
+                // Otherwise add a warning for each missing mod or version conflict
+                foreach (ModInfo_v1 record in saveData.modInfoData)
+                {
+                    if (dict.ContainsKey(record.guid))
+                    {
+                        Mod mod = dict[record.guid];
```
TheLacus, line 1202: Can you use a _TryGetValue_ instead?

Assets/Scripts/Game/Serialization/SaveLoadManager.cs
jefetienne: Resolved!

Interkarma:
Thank you with your help on review @TheLacus. Just confirming you're happy with this one now?

TheLacus:
> 
> 
> Thank you with your help on review @TheLacus. Just confirming you're happy with this one now?

Yes. I'd suggest is not optimal to parse the main save file twice (1: check mods, 2: actual load) but i'll leave to you and @jefetienne to decide if you want to improve this now. I also think it might benefit of some caching (mods can't be enabled or disabled in game at this time). Unfortunately i don't have much time these days and i don't want to cause this feature to be delayed unnecessarily :)

jefetienne:
> > Thank you with your help on review @TheLacus. Just confirming you're happy with this one now?
> 
> Yes. I'd suggest is not optimal to parse the main save file twice (1: check mods, 2: actual load) but i'll leave to you and @jefetienne to decide if you want to improve this now. I also think it might benefit of some caching (mods can't be enabled or disabled in game at this time). Unfortunately i don't have much time these days and i don't want to cause this feature to be delayed unnecessarily :)

I'll look into trying a form of caching. I think I already have an idea for it :)

jefetienne:
@TheLacus my two commits now use cache loading. I've tested the contents of the cache when switching between saves and I think it looks fine.

I also came across an odd bug a couple of times between hopping saves and prompts where the QuickLoad button stops working, but I can't figure out how to reproduce it. I don't believe it's an issue from this PR though, since it wouldn't even print a debug statement inside PromptQuickLoad() when it happens.

Interkarma:
I really like the feature. I've tested as thoroughly as I can and can't find any issues. I'll merge this into master now.

@TheLacus if you have time to give it one last look after the caching changes, it would be appreciated. There's still about a week before 0.10.24 is released. If you don't find time, that's OK too. :)

TheLacus:
Looks fine! @jefetienne thank you for working on this :)

jefetienne:
No problem, happy to help :)

PR #1841, opened 2020-06-20
Controller fixes, improvements and additional settings related to cursor issue

jefetienne:
- Added "joystick deadzone" setting, a value between 0 and 1.0 (0-100%) that is the minimum percentage of joystick movement required to be recognized by the game. This should help with joysticks that have issues such as drifting or are too sensitive and generate small amounts of movement that get captured in other ways.

- Added "Enable Controller" setting to the setup wizard that can disable the controller in realtime, and will have to be re-enabled by the use of a mouse.

- Improvements to the cursor flickering and clicking bug. _(**Note**: If you do not have the flickering issue and want to test this, just keep a joystick moved in one direction with one hand and move your mouse with the other.)_ The flickering is caused by a constant tug-of-war between the cursor going into "mouse mode" and "controller mode" if the controller is generating movement on its own. Even if the deadzone setting still doesn't help, now the mouse takes priority over the controller whenever the mouse is moving, minimizing the flickering. I also changed the logic of the `GetMouseButton()` so it can recognize both the mouse or controller button clicking better even while flickering.

- Joystick deadzone and cursor speed is now applied upon loading the setup wizard. If in the future these options get room to have their own sliders in the advanced settings menu, I think any adjustments also should be applied in realtime.

### DIFF
--- a/Assets/Resources/defaults.ini.txt
+++ b/Assets/Resources/defaults.ini.txt
@@ -79,6 +79,7 @@ MouseLookSensitivity=1.0
 JoystickLookSensitivity=1.0
 JoystickCursorSensitivity=1.0
 JoystickMovementThreshold=0.9
+JoystickDeadzone=0.0
 HeadBobbing=True
 Handedness=0
 WeaponAttackThreshold=0.03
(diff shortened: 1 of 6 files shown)


Interkarma:
Thank you, this works well! :)

PR #1842, opened 2020-06-24
Properly end resting when stopped prematurely

jefetienne:
Previously, if the player is currently resting, and they stop it prematurely via the 'Stop' GUI button, the toggle window key, or the escape key, the hours they slept for would not actually be accounted to raise skills, and neither would there be a popup saying "you wake up" or "you are healed" when done so.

This PR resolves this issue by isolating all exits to `EndRest()` when the player stopped while resting, or `CloseWindow()` if the player was on the selection menu. I also made a small change where if you choose to rest "until fully healed", but end it prematurely, the message will say "You wake up" instead of "You are healed", since the player wouldn't be fully healed.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallRestWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallRestWindow.cs
@@ -447,7 +447,8 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
                 }
                 else if (currentRestMode == RestModes.FullRest)
                 {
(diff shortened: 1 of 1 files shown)


jefetienne:
I realized that by routing premature exiting to `EndRest()`, that displaying the MessageBox  every time is now an added feature. While I think it feels good, I'm not sure if this is classic behavior, or if most people prefer just an immediate exit. If either of these is the case, I can rewrite it so it will skip the message box but take the skill-raising into account.

Interkarma:
I also find this more aesthetically pleasing and appreciate the fixes. Thank you. :)

jefetienne:
I'm glad you like it :) I caught this bug while working on a mod, too!

PR #1843, opened 2020-06-25
WIP: ghosts and wraiths shader

petchema:
Use semi-transparent shader for ghosts and wraiths. This works quite well with modded textures, but SetSoectral() is lacking for
classic textures (not extracting emissive map for the eyes, nor extracting all the details of the original textures)
ghost and wraith
modded (DREAM):
modded ghosts

### DIFF
--- a/Assets/Scripts/Game/SetupDemoEnemy.cs
+++ b/Assets/Scripts/Game/SetupDemoEnemy.cs
@@ -125,6 +125,11 @@ namespace DaggerfallWorkshop.Game
                 MeshRenderer meshRenderer = dfMobile.GetComponent<MeshRenderer>();
                 if (meshRenderer)
                 {
+                    if (dfMobile.Summary.Enemy.Behaviour == MobileBehaviour.Spectral)
+                    {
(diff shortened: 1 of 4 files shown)


petchema:
It would make sense that those frames are not color-coded (paletted), but input indices for a colour transform, mapping each color to a darker variant, except for the indice used to eyes that map all colors to red. But I'm not sure such transformation is materialized anywhere...

Interkarma:
This is a lovely bit of work. Thank you for making a start on this! :)

Would you like any help with SetSpectral() to support eye pixels and emissive setup?

petchema:
> Would you like any help with SetSpectral() to support eye pixels and emissive setup?

Definitely, I'm a bit stuck on that aspect. Like, how to extract two textures (albedo and emission) in the current image processing architecture?

Interkarma:
Happy to help! I'll dig in a bit tomorrow to scope out work.

Interkarma:
Merging this to master now, it's good as-is for a starting point. Then we can build from there.

PR #1844, opened 2020-06-27
Added Joystick Controls Window

jefetienne:
[[Image](https://imgur.com/a/XouZv5R)]

Added the Joystick Controls Window, which allows players to:

- Enable/disable the controller in realtime, like the startup wizard
- Rebind their joystick axes, in case of controller or OS differences. It can also be used for preference, like using the D-pad for movement instead
- Invert axes, both horizontally and vertically. The "Invert Mouse" setting in the Advanced Controls Window can also coincide with the vertical camera axis, so it has been renamed to "Invert Look-Y".
- Rebind left- and right-click keybinds
- Adjust look sensitivity, cursor sensitivity, maximum movement threshold, and deadzone. Deadzone can only be set up to 0.9, as 1.0 is near impossible to reach consistently with most controllers

Except for the Enable Controller setting, all settings get applied and saved after clicking 'continue' in the main controls window. Resetting to default in the main controls window will also reset joystick bindings.


Interkarma:
You're an absolute machine! :)

I'll get this one reviewed and merged ahead of 0.10.24.

PR #1845, opened 2020-06-27
Implement vampire guild membership switching

ajrb:
Switches the guild memberships when player is a vampire, switching back when cured. Nothing else is changed, no reputations or TG/DB initiation processes. I implemented it differently from classic, instead of switching round two data structures I simply access the appropriate one. Trying to minimise the dependencies here and not rely on vampirism triggering the swap.

Classic imports should work too, but I don't have any classic vampire saves so unable to test this. Could someone else test that it works correctly please?

Also fixed a bug with finding region factions.

### DIFF
--- a/Assets/Scripts/Game/Entities/PlayerEntity.cs
+++ b/Assets/Scripts/Game/Entities/PlayerEntity.cs
@@ -979,7 +979,7 @@ namespace DaggerfallWorkshop.Game.Entity
         /// <summary>
         /// Assigns guild memberships to player from classic save tree.
         /// </summary>
-        public void AssignGuildMemberships(SaveTree saveTree)
+        public void AssignGuildMemberships(SaveTree saveTree, bool vampire = false)
         {
             // Find character record, should always be a singleton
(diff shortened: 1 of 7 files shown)


Interkarma:
This looks good and thorough on a read. Will test and merge as soon as I can. Thank you for adding this. :)

ajrb:
No problem. Needed to shake the rust off with something. :D

Interkarma:
I'm happy with this. I really like the way you've handled it.

The main problem is one of support - anyone that's been rocking a vamp for a while will suddenly find themselves out of guilds until they rejoin. I'll be sure to highlight this ahead of new builds and in blog update.

Interkarma:
Hi mate, this change hit an exception with some classic save data. See issue #1849 for test save. Don't worry about the CastWhenHeld comments in original report, that's unrelated.

I've wrapped ImportMembershipData() foreach loop in a try/catch block to solve this one for now. The problem is most likely just bad data in the game save, as we weren't importing old memberships before. I thought you might like to have a closer look before I close issue.

df-ferital:
I'm also looking into it, since importing the classic character I imported had OldGuild data (he was a former vampire).

PR #1846, opened 2020-06-28
[RFC] make ghosts and wraiths highlights slightly emissive

petchema:
I do like the version of ghosts and wraiths in master, it's really close to classic in term of look and stealthiness, with just more details up close with some lighting. This PR is just an alternative that I find interesting.

It make ghosts and wraiths highlights slightly emissive (more focused on highlights than just a constant emissive value). Shouldn't make much difference under bright light, including classic high ambient light conditions.
In dark places however, it gives a hint of the baddies, as if they were slightly phosphorescent
spectres highlights 3

### DIFF
--- a/Assets/Scripts/API/BaseImageFile.cs
+++ b/Assets/Scripts/API/BaseImageFile.cs
@@ -12,8 +12,8 @@
 #region Using Statements
 using System;
 using System.IO;
-using UnityEngine;
 using DaggerfallConnect.Utility;
+using UnityEngine;
 #endregion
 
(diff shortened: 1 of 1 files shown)


Interkarma:
I love the look. Thanks for making another pass over it. :)

petchema:
That's funny, I found that x^1.9 (or is it x^2.9, given it's an amplification of an existing color? anyway) purely based on experiment, but looking at the result, how the bones look, that's just reinventing specular highlight formula...

PR #1850, opened 2020-07-02
Fix import of classic old guild records

### LINKED ISSUE #1849: Unable to load classic save whith character equipped with a CastWhenHeld magic item
**Describe the bug**
Since I implemented CastWhenHeld item condition lowering in PR #1827, I'm unable to load a classic save with a character equipped with such kind of magic item. 

**To Reproduce**
1. Load the provided classic save.
2. The item will broke immediately in an infinite loop.

**Expected behavior**
The game should load and the item isn't supposed to break since such items condition lowering should be done on equip only.

As I'm not quite familiar with how the current system is implemented, it might be better if you check what's happening there, @Interkarma.

[SAVE2.zip](https://github.com/Interkarma/daggerfall-unity/files/4863670/SAVE2.zip)

df-ferital:
This fixes #1849 properly, by importing old guild membership records. @Interkarma I removed your exception handling code since it shouldn't be required anymore. However if you prefer that it remains, I will put it back.

@ajrb, could you review the changes as well? I provided a classic save with a character cured from vampirism in #1849, having a single old guild record (Kynareth at novice rank). I had to modify GuildMembership.RestoreMembershipData to avoid an importation bug where this old guild membership was overwriting the actual player membership in Kynareth at patriarch rank.

### DIFF
--- a/Assets/Scripts/API/Save/GuildMembershipRecord.cs
+++ b/Assets/Scripts/API/Save/GuildMembershipRecord.cs
@@ -87,7 +87,8 @@ namespace DaggerfallConnect.Save
         void ReadNativeGuildMembershipData()
         {
             // Must be a guild membership type
-            if (recordType != RecordTypes.GuildMembership)
+            if (recordType != RecordTypes.GuildMembership &&
+                recordType != RecordTypes.OldGuild)
                 return;
 
             // Prepare stream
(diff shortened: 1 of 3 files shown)


Interkarma:
No worries removing exception, that was just me working around exception quickly before I crashed for the night. :)

I'm happy with this. Will merge ahead of this weekend's builds unless @ajrb wants to change something.

ajrb:
Ha, didn't give me a chance to review... but looks good to me. Great work Ferital finishing this off! Thanks.

Interkarma:
Sorry mate, I wanted to get the builds ready so I could start testing. I figured I could rebuild if you found anything wrong before weekend was out. Thanks for checking. :)

PR #1853, opened 2020-07-04
prevent initating fast travel during day with damage-from-sunlight disadvantage

petchema:
(I did you try factorizing code with plain vampirism. Could probably be done using something similar to  #1847 but for fast travel?)

Forums: https://forums.dfworkshop.net/viewtopic.php?f=31&t=3710


Interkarma:
Thank you for fix! I'm happy with this for now, and to look at refinements later. :)

petchema:
There's a loophole in this fix, I thought about damage from sunlight given by class disadvantage, but not given by an enchanted item
https://www.twitch.tv/talonhatesnames/video/893637162?t=1h49m14s

PR #1856, opened 2020-07-06
Correct %lp macro implementation

df-ferital:
This macro is used to get the current province name (reverse-engineered from classic). It is only used in TEXT.RSC record 8375, which corresponds to Mid Year 16th celebration text.


Interkarma:
Nice work! :)

jotoho:
@Interkarma 
Are you sure that you wanted to merge master into this and not the other way around?

Just want to make sure, since it seems a bit strange...

Assets/Scripts/Internal/PlayerGPS.cs
```diff
@@ -385,27 +385,10 @@ public NameHelper.BankTypes GetNameBankOfCurrentRegion()
...
-            {
-                case FactionFile.FactionRaces.Redguard:
-                    return Races.Redguard;
-                case FactionFile.FactionRaces.Nord:
-                    return Races.Nord;
-                case FactionFile.FactionRaces.Breton:
-                default:
-                    return Races.Breton;
-            }
+            return (Races) MapsFile.RegionRaces[GameManager.Instance.PlayerGPS.CurrentRegionIndex] + 1;
```
Interkarma, line 391: I don't agree with this change. Classic definitely generates Nord NPCs around those subclimates, and RegionRaces array only defines Breton or Redguard (0 or 1). This change would remove that diversity and will not match classic. For example, NPC selection in Wrothgarion Mountains/Whitetry in classic and DFU with existing code will match:

**Classic**
fall_102


**DFU**
image


And then after this code change the NPCs no longer match.

**After Edit**
image


Happy to take the %lp macro, just not with this edit included. Perhaps we need to rethink the accuracy of MapsFile.RegionRaces instead.

Interkarma:
> @Interkarma
> Are you sure that you wanted to merge master into this and not the other way around?
> 
> Just want to make sure, since it seems a bit strange...

The PR was out of date and referenced code that no longer exists in master (HardStrings). It also had a conflict that had to be resolved. What you're seeing is the result of making those edits and resolving conflict.

PR #1860, opened 2020-07-12
Improve doorways pathfinding (again)

petchema:
This is basically an improvement over #1733.

In #1733 is conservatively lowered pathfinding probe distance from 1 to 0.8f. After more testing, lowering this distance further improves AI performance.
The first thing to notice, is that because probing is done using a CapsureCast with half the player's capsule radius, everything in probing geometry is proportional to player's capsule radius.
Second, if you consider the cast of following a flat surface (typically a wall), there's some relation between the ratio of probe length to capsule radius and the angle of incidence of probing: sin a = r / 2l
spherecast pathfinding
Very low incidence will make it hard to find doorways, very high would probably slow down progression along the wall, and a = 45° <=> l = r / sqrt(2) seems to give good results. The only other values I tried were a = 30° <=> l = r is worse (and so should be even higher values of l) and a = 60° <=> l = r / sqrt(3) that felt okay but somewhat slower.
The only bad side effect I could think of, and the reason I was conservative the first time around, is that enemies start avoiding obstacles from a shorter distance, however I have no evidence this is a real problem (the original probe length of 1 seems to have been chosen arbitrarily in retrospect).

I also slightly modified probe height from 1.75f to 1.65f (and factorized that height), since it has been modified since c282fc667a43f486be91e09fe77cafc351ad5e3f
By the way, it seems Unity 2019.4 seems to also help tall enemies not getting stuck into doors

### DIFF
--- a/Assets/Scripts/Game/EnemyMotor.cs
+++ b/Assets/Scripts/Game/EnemyMotor.cs
@@ -33,6 +33,7 @@ namespace DaggerfallWorkshop.Game
         public float OpenDoorDistance = 2f;         // Maximum distance to open door
         const float attackSpeedDivisor = 2f;        // How much to slow down during attack animations
         float stopDistance = 1.7f;                  // Used to prevent orbiting
+        const float doorCrouchingHeight = 1.65f;    // How low enemies dive to pass thru doors
         bool flies;                                 // The enemy can fly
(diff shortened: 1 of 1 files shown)


Interkarma:
A definite improvement, well done. 👍

PR #1861, opened 2020-07-12
Prevent point blank spellcasting into low obstacles

petchema:
When enemy casters can see their target but are right behind some obstacle (say low obstacles like tables, fountains,...), the SphereCast used to check whether they're free to cast can miss the obstacle and allow them to cast, potentially damaging themselves (in classic lich style). An easily reproducible example of such behavior can be triggered in Orsinium entrance hall, if you agro the orc shaman but keep tables in his way.

Starting SphereCast from some position behind the caster could be a workaround, but since its radius is usually larger than caster's capsule radius, it's not very satisfactory (you don't start the SphereCast from inside a "known empty space"; The right offset would also depend on enemy capsule radius).
So instead I prepend the SphereCast test with a SphereCheck at its origin, temporarily excluding caster's collider.

I do not check whether that SphereCheck hits the target (point blank shooting), so HasClearPathToShootProjectile() will return false in this case, favoring other attack types (melee, touch spells,...)

### DIFF
--- a/Assets/Scripts/Game/EnemyMotor.cs
+++ b/Assets/Scripts/Game/EnemyMotor.cs
@@ -77,6 +77,7 @@ namespace DaggerfallWorkshop.Game
         Vector3 detourDestination;
         CharacterController controller;
         DaggerfallMobileUnit mobile;
+        Collider myCollider;
         DaggerfallEntityBehaviour entityBehaviour;
         EnemyBlood entityBlood;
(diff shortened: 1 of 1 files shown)


petchema:
One drawback of this PR is that it prevents daedra lords from attempting to shoot at you in the "crossbow room" at the end of the Mantellan Crux. (Without it they sometimes hit their own "balcony", but enough fireballs make it thru to make them dangerous)

PR #1862, opened 2020-07-13
Refactor TalkManager for More Readability

dyanikoglu:
I'm just playing around with the project, and decided to refactor a random class as a warming up while trying to understand how it works at the same time :)

I'm not sure if you have a coding standard for C# scripts, but I assume my changes adds some readability.

### DIFF
--- a/Assets/Scripts/Game/TalkManager.cs
+++ b/Assets/Scripts/Game/TalkManager.cs
@@ -1122,7 +1122,7 @@ namespace DaggerfallWorkshop.Game
                     {
                         int index = currentDFRegion.MapNameLookup[location];
                         DFRegion.RegionMapTable locationInfo = currentDFRegion.MapTable[index];
-                        position = MapsFile.LongitudeLatitudeToMapPixel((int)locationInfo.Longitude, (int)locationInfo.Latitude);
+                        position = MapsFile.LongitudeLatitudeToMapPixel(locationInfo.Longitude, locationInfo.Latitude);
                         positionLocation = new Vector2(position.X, position.Y);
                     }
                 }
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Game/TalkManager.cs
```diff
@@ -1056,22 +1031,21 @@ public string DirectionVector2DirectionHintString(Vector2 vecDirectionToTarget)
 
             if ((angle >= 0.0f && angle < 22.5f) || (angle >= 337.5f && angle <= 360.0f))
                 return TextManager.Instance.GetText(textDatabase, "east");
-            else if (angle >= 22.5f && angle < 67.5f)
+            if (angle >= 22.5f && angle < 67.5f)
```
petchema, line 1034: Maybe that could be simplified using some maths like in
https://github.com/Interkarma/daggerfall-unity/pull/1730/files#diff-c4ce0a36388a746d93049250a142647fR350
? Putting direction strings in an array...

dyanikoglu: I also noticed that code block while reviewing it, that sounds like a good idea, instead of hardcoding them into if conditions.

Interkarma:
Hello dyanikoglu! :)

Thank you for taking the time to contribute. I don't wish to discourage anyone, but I generally don't take PRs that just change the style from one person to another. I'm happy with the overall readability and function of Nystul's work here and don't want to risk introducing regressions for the sake of style changes alone.

At this point, I'm actively trying to limit changes for my own time management and sanity. I hope you will understand.

https://forums.dfworkshop.net/viewtopic.php?f=4&t=2503

Nystul-the-Magician:
I would like to add some of my thoughts here:
- I like many of the suggested changes, but some makes stuff harded for me to read/follow
- I learned new c# stuff like "??=" operator and I like using the IsNullOrEmpty function
- unsure about the for each loops - I agree that it looks a bit shorter/clearer but I am still not used to it making it harder to read for me personally
- I dislike the if statement replacement with condition?ifcase:elsecase - I always have to look twice to see what is going on
- I like the one switch statement making the one if-elseif block that grew over time much more concise
- If all of you agree I will try to incorporate the suggestions where it does not obfuscate stuff for me (considering that a future bug report will make it necessary to work on the code again)

PR #1863, opened 2020-07-14
Quest typo fixes

JayH2971:
Just used UESP as a guide to find typos in quests, to fix them. Also updates the quest list to show that everything enabled has passed testing.

### DIFF
--- a/Assets/StreamingAssets/Quests/$CUREVAM.txt
+++ b/Assets/StreamingAssets/Quests/$CUREVAM.txt
@@ -153,7 +153,7 @@ Message:  1025
 <ce>                    will not need to hunt you down.
 
 Message:  1030
-<ce>               Stripling, do not disturb my mediations!
+<ce>               Stripling, do not disturb my meditations!
 <ce>                    Begone, or my hunger will force
 <ce>                            me to slay you!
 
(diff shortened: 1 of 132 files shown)


JayH2971:
My first commit said "Order of fixing is random," but I went over and re-did the entire thing to make sure all quests are fixed. All typos in quests should be fixed now.

df-ferital:
Hi @JayH2971. I noticed that in quests A0C00Y14 and A0C00Y15, if the package to deliver is a painting, quest giver says (message 1015) that the commissioner wants a `_painting_ painting`. But since macro _painting_ expands to the item name, which is `Painting`, the text becomes `Painting painting`. So one of the occurrence has to be removed, probably the second one if we want to remain consistent with the fact that all quest items are referred to using a capital letter.

JayH2971:
I hadn't noticed that, and agree on the change.

ajrb:
That's some work right there, nice stuff Jay.

df-ferital:
Before this get merged, I noticed several quests are using %dat instead of %qdt for some log messages, resulting in those entries always displaying the current date instead of the date it happened:
- A0C01Y03: messages 1015 and 1016
- N0B21Y14: 1021
- P0B00L04: 1014
- P0B00L06: 1019 and 1021
- P0B10L07: 1013
- P0B10L10: 1014
- P0B20L09: 1013

JayH2971:
Excellent, thanks for pointing them out. I've added the change.

Interkarma:
This is quite the effort Jay. Really incredible work. Thank you for all those fixes. 👍

PR #1865, opened 2020-07-17
Added support for secondary keybindings

jefetienne:
[[Images](https://imgur.com/a/7wrMblw)]

Adds support for secondary keybinds, which can be accessed via a small toggle button at the top right of the controls window. By default, they are unpopulated, although later in the future I will create ControllerConfigs that will have loadout configurations specific to the controller type and OS.

Any cross-unary conflicts will be highlighted as blue (to help colorblind people distinguish it from yellow or red/"dark yellow"), and the player cannot "Continue" until resolving them, but can switch between primary and secondary. Any internal (red) conflicts made, and they cannot switch between primary and secondary until that gets resolved first.

To not bloat the controls window and avoid static code, I made a ControlsConfigManager that sets aside much of the cross-window logic from the main controls window. Also fixed a bug with unknown actions - if the player had an unknown action that was bound to a certain key, using that same key to rebind to an existent action would cause an exception and fail to update Keybinds.txt. Now, if it is rebinded, it will erase the unknown action (as it should be repopulated anyway if they switch to a newer version.)

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallControlsWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallControlsWindow.cs
@@ -204,7 +204,6 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             foreach (InputManager.Actions a in Enum.GetValues(typeof(InputManager.Actions)))
             {
                 UnsavedKeybindDict[a] = InputManager.Instance.GetKeyString(InputManager.Instance.GetBinding(a));
-                Debug.Log(a.ToString()+": "+UnsavedKeybindDict[a]+","+((int)InputManager.Instance.GetBinding(a)));
             }
         }
 
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallControlsWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallControlsWindow.cs
@@ -58,28 +58,6 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
 
         #endregion
 
-        #region Static Fields
-
-        private static Color crossDupeColor = new Color(0, 0.58f, 1);
-        private static Color internalDupeColor = new Color(1, 0, 0);
-
-        private static Dictionary<InputManager.Actions, string> PrimaryUnsavedKeybindDict = new Dictionary<InputManager.Actions, string>();
-        private static Dictionary<InputManager.Actions, string> SecondaryUnsavedKeybindDict = new Dictionary<InputManager.Actions, string>();
-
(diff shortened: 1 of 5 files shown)


jefetienne:
I also think it would be good to remove bindings to "None" by right-clicking a keybind button and prompting "are you sure you want to clear the binding for ActionName ('ButtonName')?", but I will add that as a separate PR after this.

jotoho:
> I also think it would be good to remove bindings to "None" by right-clicking a keybind button and prompting "are you sure you want to clear the binding for ActionName ('ButtonName')?", but I will add that as a separate PR after this.

I wanted to suggest allowing players to unbind keys they do not want by "assigning" it to the Escape key, like in many other games, to make it more likely that players will find it, but then noticed that the pause menu can be reassigned through the advanced controls menu, so that's not really an option, is it?

Being able to unbind keys at all would be a useful improvement but I feel like right-clicking is not necessarily obvious enough, so I don't know how many players would know it is possible. On the other hand, I do not know how it could be made any more easy to find, besides outright saying it in the menu.

Curiosity: Was key unbinding a thing in the original engine? I never played it for more than 5 minutes.

jefetienne:
> I wanted to suggest allowing players to unbind keys they do not want by "assigning" it to the Escape key, like in many other games, to make it more likely that players will find it, but then noticed that the pause menu can be reassigned through the advanced controls menu, so that's not really an option, is it?

Typically I think the Escape key is more for canceling than removing a keybind, but yes that is a reason. Maybe in the future I'll set Escape back to a reserved keybind but treat the Pause keybind as an exception. I have seen games though where right-clicking does ask to remove keybinds.

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallControlsWindow.cs
```diff
@@ -294,6 +247,8 @@ private void UpdateKeybindButtons()
             SetupKeybindButtons(lookKeys, 32, 36, 269, 102, false);
             SetupKeybindButtons(uiKeys, 36, 40, 269, 147, false);
 
+            currentBindingsButton.Label.Text = ControlsConfigManager.Instance.UsingPrimary ? "Primary" : "Secondary";
```
Interkarma, line 250: Don't let me forget to add these strings to the Internal_Strings collection for localization later. :)

Interkarma:
As always, this just works perfectly @jefetienne. No problems found on this end, I think it's well worth getting into 0.10.26.

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallControlsWindow.cs
jefetienne: Got it :)

PR #1866, opened 2020-07-17
Fix parchment tearing

petchema:
Round all coordinates where textures need to rejoin
Simple fix, not a general solution for changed GUI.DrawTexture() behavior change


petchema:
Need more testing

PR #1867, opened 2020-07-17
fix parchment tearing

petchema:
Simplest fix: fix parchment tearing by rounding coordinates where textures must meet

### DIFF
--- a/Assets/Scripts/Game/UserInterface/Panel.cs
+++ b/Assets/Scripts/Game/UserInterface/Panel.cs
@@ -268,56 +268,56 @@ namespace DaggerfallWorkshop.Game.UserInterface
             // Top-left
             topLeftBorderRect.x = drawRect.x;
             topLeftBorderRect.y = drawRect.y;
-            topLeftBorderRect.width = virtualSizes.TopLeft.x * LocalScale.x;
-            topLeftBorderRect.height = virtualSizes.TopLeft.y * LocalScale.y;
+            topLeftBorderRect.xMax = Mathf.Round(drawRect.x + virtualSizes.TopLeft.x * LocalScale.x);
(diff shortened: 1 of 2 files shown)


petchema:
For skybox, I assume a patch like

```patch
--- a/Assets/Scripts/Internal/DaggerfallSky.cs
+++ b/Assets/Scripts/Internal/DaggerfallSky.cs
@@ -240,8 +240,8 @@ namespace DaggerfallWorkshop
             float yShear = (Mathf.Tan(angleXRadians) * zoom) * 0.50f;
             float scrollY = Mathf.Clamp(baseScrollY - (yShear * Screen.height), -height, 0f);
 
-            westRect = new Rect(westOffset + scrollX, scrollY, width, height);
-            eastRect = new Rect(eastOffset + scrollX, scrollY, width, height);
+            westRect = new Rect(Mathf.Round(westOffset + scrollX), Mathf.Round(scrollY), width, height);
+            eastRect = new Rect(Mathf.Round(eastOffset + scrollX), Mathf.Round(scrollY), width, height);
         }
 
         private void DrawSky()
```

could help, however I don't like adding code for a problem I've been unable to reproduce...

Interkarma:
This is very good and certainly helps fix the problem almost completely. I was able to still create a small tear using this, but it's very uncommon now and I had to work really hard to find just the right spot. :)

image

Happy for a PR for sky if you have time. Thank you again!

petchema:
Yes, I can still create the issue in 1440x960
1440x960 seam
Not sure why it still happens in that case...

Interkarma:
Would be interesting to step back through Unity versions to see exactly where this started. The change notes for that release might offer some clue.

I suppose it's possible this will be fixed by a Unity point release later. Other than a few minor tweaks like this PR, I'm not keen to overhaul UI layout just yet to work around problem. I have a sense this one will be a fix that finds us rather than the other way around.

PR #1868, opened 2020-07-20
Fix %god macro to match classic behavior

df-ferital:
This fixes %god macro expansion to match classic, see the issue reported by @JayH2971 here: https://forums.dfworkshop.net/viewtopic.php?f=31&p=45914#p45914. I fixed only `QuestMCP.God()` here, but the method could have been used globally.

While doing so, I moved an array extracted from FALL.EXE to `MapsFile` since it is also used in classic for the %god macro.

@ajrb

### DIFF
--- a/Assets/Scripts/API/MapsFile.cs
+++ b/Assets/Scripts/API/MapsFile.cs
@@ -72,6 +72,19 @@ namespace DaggerfallConnect.Arena2
         };
 
         /// <summary>
+        /// Region temple faction IDs, extracted from FALL.EXE.
+        /// </summary>
+        private static readonly int[] regionTemples = {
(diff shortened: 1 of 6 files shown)


ajrb:
When you say it could be used globally, what do you mean exactly? The guild (temple) MCP gets the deity from the temple instance and works just fine as far as I know, and I don't think this would be an appropriate replacement, or did you mean something else?

df-ferital:
I meant that I basically translated the original classic function and put it in `QuestMCP.God()`, but as it is, it can be used anytime %god is required. However, from what I saw, %god is only used in quests or when talking to NPCs inside temples. So as you say, it works just fine as it is now :).

Interkarma:
Great work as always. :)

PR #1869, opened 2020-07-20
Upgrade support for providers of mobile person graphics

TheLacus:
Fixes some regressions:

- Unity 2019 doesn't allow to destroy or reparent gameobjects that are part of prefab instances. When a mod provides an override, the default gameobject is only deactivated instead.
- Custom providers of MobilePersonAsset that use colliders need to define a layermask to support functionality that checks for geometry in the way of npc path (regression introduced some time ago with `MobilePersonMotor.IsDirectionClear()`).


TheLacus:
This covers all the evident issues i found.

I wrote a simple test mod, which is available at the bottom of [this tutorial](https://forums.dfworkshop.net/viewtopic.php?p=3635#p3635). It should be helpful to help mod authors to familiarize with this feature and to easily check against new regressions.

PR #1876, opened 2020-07-28
Fix warning messages

TheLacus:


jefetienne:
You can undo the Input manager variable. The secondary keybinds PR deletes it. :)

PR #1877, opened 2020-07-28
Use faction race ID for oath

df-ferital:
This actually fixes a bug from classic to make NPCs use the correct oath from their respective races.

Oaths are declared by NPC race according to the race index used in FACTION.TXT. Unfortunately, classic never uses this faction race ID but, instead, uses the hardcoded index race of each region, which is not the same. In classic, this results in all NPCs from High Rock saying Nord oaths, while all NPCs in Hammerfell will say Khajiit oaths.

Instead, DFU uses the faction race ID to return the correct oath.

So now, when talking to Morgiah when she doesn't like you, be aware:

Morgiah oath

@deepfighter, finally, FACTION.TXT race ID has found its intended use :)

### DIFF
--- a/Assets/Scripts/API/FactionFile.cs
+++ b/Assets/Scripts/API/FactionFile.cs
@@ -596,27 +596,25 @@ namespace DaggerfallConnect.Arena2
         }
 
         /// <summary>
-        /// Faction race value does not map to usual race ID.
-        /// Instead it selects from a smaller pool as below.
-        /// For example:
-        ///  * "Daggerfall", and most others, have a race of 3 (Breton)
-        ///  * "Sentinel" has a race of 2 (Redguard)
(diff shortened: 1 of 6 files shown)


JayH2971:
What happens if the %oth isn't directly attached to a specific NPC? Kamer and I have used it in some occasions like the start of an ambient quest, where no NPC is defined.

df-ferital:
For quests, oath is always tied to questor race. If this NPC does not have a particular race, race from the current region is used.

Classic does almost the same, except it takes an invalid race index, resulting in the bug I mentioned (Breton using Nord oaths and Redguard using Khajiit oaths).

Interkarma:
Thank you @df-ferital. I appreciate the constant refinement and matching to classic over my approximate first-pass implementations on some of these items.

PR #1879, opened 2020-07-28
Add water to dungeon automap

petchema:
Looks better than expected, given how simple is the shader modification.

water3

The MR can be squashed, I attempted another approach first on this branch that doesn't need to stay around...

### DIFF
--- a/Assets/Scripts/Game/Automap.cs
+++ b/Assets/Scripts/Game/Automap.cs
@@ -1293,6 +1293,7 @@ namespace DaggerfallWorkshop.Game
                 slicingPositionY = gameObjectPlayerAdvanced.transform.position.y + Camera.main.transform.localPosition.y + slicingBiasY;
             else
                 slicingPositionY = float.MaxValue;
+            // Debug.Log("slicingPositionY = " + slicingPositionY);
             Shader.SetGlobalFloat("_SclicingPositionY", slicingPositionY);
         }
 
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Game/Automap.cs
```diff
@@ -2002,10 +2002,11 @@ private void CreateDungeonGeometryForAutomap()
             {
                 if (elem.name.Contains("DaggerfallDungeon"))
                 {
+                    DFLocation.DungeonBlock[] blocks = GameManager.Instance.PlayerEnterExit.Dungeon.Summary.LocationData.Dungeon.Blocks;
```
petchema, line 2005: I had to base the automap on Summary.LocationData.Dungeon.Blocks instead of location.Dungeon.Blocks, because in the latter all block.WaterLevel are always 0.
Maybe I'm missing something.

petchema:
Another screenshot of the MR in action...
water4
"now I remember why I didn't take that shortcut"

Nystul-the-Magician:
wow this looks amazing! great work!

petchema:
Glad you like it :)

Assets/Scripts/Game/Automap.cs
Interkarma: I think this is fine, even with SmallerDungeons enabled. The new dungeon layout is generated at a lower level and should be correct in this data at runtime.

PR #1880, opened 2020-07-29
Add motion limits for mods, and expose applying force publicly

ajrb:
@Interkarma - would like you to check you're okay with these changes. I can manage without them, they allow autowalking without using reflection and limiting horse strafe and backup speed.


Interkarma:
This looks OK to me. @jefetienne will this be any issue on your side with upcoming input PRs?

jefetienne:
Content-wise it should be fine - secondary keybinds doesn't touch anything with horizontal or vertical forces, but I'm not sure if there might be any conflicts with whitespace or code on the same line numbers.

Interkarma:
Thanks for checking! :)

PR #1882, opened 2020-07-31
remove unstick handling

petchema:
Problematic cases seem now handled correctly by Unity 2019.4 capsule


Interkarma:
Thank you Pango! :)

PR #1885, opened 2020-08-07
Increase location extra clearance for cities

ajrb:
I think this is good change giving a wider area outside city walls before natures can get placed, and it incidentally also feeds into a mod I am making very nicely.

Please say if anyone objects.


Interkarma:
No objections here with change limited to large cities. The extra clear area is hardly noticeable.

PR #1893, opened 2020-08-14
Add run quest action

TheLacus:
Added an action that allows to start a quest and wait for it to end, setting task for success or failure. This is the syntax:

```
run quest QUESTNAME then _success_ or _failure_
```

and this is a simple test example:

```
Quest: __DEMORUN

QRC:

QBN:

Clock _clock_ 5

    start timer _clock_

_clock_ task:
    give pc nothing
    end quest
```

```
Quest: __DEMORUNMAIN

QRC:

QBN:

Clock _clock_ 5

    start timer _clock_

_clock_ task:
    run quest __DEMORUN then _success_ or _failure_

_success_ task:
    give pc nothing
    end quest

_failure_ task:
    end quest
```

The purpose of this action is to track progress in a chain of quests (and/or reusable subquests) with a more modern approach compared to global variables, at least when a single success flag is enough. Success state is scoped to the instance of the quest that started the other, so it forces a design that is easier to manage compared to a global flag that might be used in multiple locations and cause consequences that are harder to debug.

### DIFF
--- a/Assets/Scripts/Game/Questing/Actions/RunQuest.cs
+++ b/Assets/Scripts/Game/Questing/Actions/RunQuest.cs
@@ -1,0 +1,147 @@
+// Project:         Daggerfall Tools For Unity
+// Copyright:       Copyright (C) 2009-2020 Daggerfall Workshop
+// Web Site:        http://www.dfworkshop.net
+// License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
+// Source Code:     https://github.com/Interkarma/daggerfall-unity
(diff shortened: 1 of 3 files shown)


JayH2971:
Wow, wow, WOW. This is AMAZING!! So glad I haven't done a MQ2.0 before seeing stuff like this :laughing:

ajrb:
Nice idea, I like it. :-)

Interkarma:
Very smart, and nice implementation. Thank you @TheLacus! :)

PR #1897, opened 2020-08-18
More Disease info, can be charged with Breaking and Entering

DFIronman:
Currently, only "Attempted Breaking and Entering" is supported. This allows for the potential upgrade of the crime to full Breaking and Entering when you successfully break into the location. The check for Attempted Breaking and Entering is a Dice100.SuccessRoll(10). I did the same for Breaking and Entering (10% chance "someone saw you" breaking into the location). Guards are currently not being called for Breaking and Entering

Should guards ever be called for Breaking & Entering? I figure if this is implemented, being charged with Breaking and Entering in classic DFU is going to be a fairly edge case (probably: guards got summoned for Attempted B&E, you broke in, went back outside, and the guards arrested you).

Also, made IsDiseasePermanent() public instead of protected and added IsDiseaseCompleted() for modders.


Interkarma:
Happy to add this. Thank you @DFIronman :)

DFIronman:
> Happy to add this. Thank you @DFIronman :)

Sure thing - thanks!

PR #1898, opened 2020-08-21
Keybind right-click removals

jefetienne:
Added ability to right-click a keybind button which prompts if the player if they would like to set it to 'None'. This is useful for secondary keybinds where someone might accidentally bind the wrong button or no longer would like to use it. As of right now, they  would have to find an unused key to bind to it or set it to None in the Keybinds.txt file, but actually having an in-place removal would help with this issue.

One problem I encountered was the fact that DFU always assumed the primary keybinds (`InputManager.actionKeyDict`) to be non-`None` since there wasn't any need for them to be in the first place. This made finding secondary KeyCode equivalents based off of the primary KeyCode a problem if it was set to `None`, since there can always be duplicate `None`s. So I created an `existingKeyDict` that attempts to fill missing KeyCodes in primary binds with the secondary for the `FindKeyboardActions` and `GetKey` methods.

### DIFF
--- a/Assets/Scripts/Game/ControlsConfigManager.cs
+++ b/Assets/Scripts/Game/ControlsConfigManager.cs
@@ -16,6 +16,7 @@ using System.Collections.Generic;
 using System.Linq;
 using DaggerfallConnect.Arena2;
 using DaggerfallWorkshop.Utility;
+using DaggerfallWorkshop.Game.UserInterfaceWindows;
 using DaggerfallWorkshop.Game.UserInterface;
 using DaggerfallWorkshop.Game.Utility;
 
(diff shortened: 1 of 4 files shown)


jefetienne:
I'm writing a feature for custom hotkey sequences and the ability to bind them to actions - coincidentally part of my work on this PR will actually fix a cornercase bug that currently exists, but there aren't any new changes or features yet to allow the bug to reveal itself.

Interkarma:
>It also is probably good if the message box utilizes the localization feature, but I'm not sure how to use it properly :)

I can help with that. :) I'll move to localization post-merge now and @ you on the changes so you can see how it was done.

Interkarma:
Hey @jefetienne! :) A user in below forum topic is experiencing some issues when clearing Primary bindings to "None". Their main issue is that Primary bindings set as "None" are reset to defaults on next launch, which could conflict with other bindings user has configured.

https://forums.dfworkshop.net/viewtopic.php?f=5&t=4943

Please let me know if anything I can do to help.

KABoissonneault:
The save format for the keybinds seems to use the input button as the key and the mapped action as the value. This means the format has no way of distinguishing an action mapped to "None" versus a binding that's just not set.

Maybe we could have another two fields in the save file that's an array of actions set to None in the Primary and Secondary sets? This way, when loading the save, we'd know which actions should be set to None rather than go to the default binding
Users would have to re-save their config for it to work, of course

Interkarma:
I'm open to differentiating between unset bindings and "None" as a valid permanent binding. This is probably the better outcome for users.

Just to fix the default conflict issue, it would also be adequate to not set default binding if this would conflict with a defined binding.

I'll leave for jefetienne to decide first, as he's the one with most time invested to the bindings and controls. I'm also happy to work on this once I start grinding bug reports again. I'll move that topic into bug reports so we don't lose track of it.

PR #1899, opened 2020-08-21
Change screenshot capture to JPEG format

jefetienne:


petchema:
there's another issue I've seen on stream:
dungeon automap uses F8 to change background color, so if you use that, when you exit the automap you notice that some screenshots have been taken...

jefetienne:
Oh interesting, I didn't know that. I think it's worth keeping the ability to screenshot in menus, but maybe we change the background shortcut to a hotkey sequence? I know we could simply change the default "F8" to to something like "F7", but that wouldn't prevent players from rebinding it to F8 again.

petchema:
automap uses F5 to F8 to select different backgrounds:
https://github.com/Interkarma/daggerfall-unity/blob/master/Assets/StreamingAssets/Text/DialogShortcuts.txt

Interkarma:
Thank you @jefetienne looks good. And well spotted @petchema. I'll merge JPG change now and we can work out that issue as a refinement.

PR #1901, opened 2020-08-22
Dungeon shallow water splash sound

petchema:
Make sure water splash sound plays when walking in dungeon shallow water
[SAVE646.zip](https://github.com/Interkarma/daggerfall-unity/files/5112087/SAVE646.zip)


petchema:
I don't seem to experience regression with
https://forums.dfworkshop.net/viewtopic.php?f=30&t=3619

Interkarma:
>I don't seem to experience regression with https://forums.dfworkshop.net/viewtopic.php?f=30&t=3619

Your change is fine. I was just a little more conservative with the threshold. :)

PR #1904, opened 2020-08-24
Update joystick minimum force and acceleration

jefetienne:
After having unstick handling removed, there no longer needs to be an arbitrary float value in place that prevented it from going off. Additionally, the `JoystickMovementThreshold` (the ceiling value for maximum movement via joystick) is now used to scale with the distance of the joystick. So if the `JoystickMovementThreshold` is 0.5 and the user moved their joystick at 20% distance from the center, their force would be 0.2 / 0.5 = 40%. Compare this method to having 49% distance as 49% force, but 50% distance = 100% force.


jefetienne:
I was also thinking that maybe with having softer movement, we could actually dynamically change between "walking" and "sneaking" if the player movement falls within the range of sneaking speed, but I'm not sure which values would I need to look at in order for this to happen.

PR #1905, opened 2020-08-25
Preprocessor definiton for separate development persistent path

jefetienne:
Added a commented-out preprocessor definition under `SettingsManager` called `SEPARATE_DEV_PERSISTENT_PATH`, which allows development under the Unity editor to use a separate data directory called "DaggerfallUnity.devenv", which is alongside the default path. This is useful for separating recreational saves and settings from the development side for organization, and to prevent wiping of settings not yet implemented in the stable release.

I grepped and replaced all instances of `Application.persistentDataPath` with `DaggerfallUnity.Settings.PersistentDataPath`.


Interkarma:
Looks good. Thank you for implementing a way that end users aren't likely to tangle themselves up. :)

Issue #1902, opened 2020-08-22
Got no response from Aubk-I when completing Mynisera's Letters main quest


petchema:
**Describe the bug**
After recovering the burnt letters from Castle Necromoghan, clicking on Aubk-I does apparently nothing. The quest is marked completed though.

**To Reproduce**
1. Load gamesave below
2. Click on Aubk-I
3. Get no reply, but quest is finished

**Expected behavior**
Aubk-I comment on letters' content
https://en.uesp.net/wiki/Daggerfall:Mynisera%27s_Letters#Reward

**Desktop (please complete the following information):**
 - OS: Linux
 - Version master (61c8c05)

**Additional context**
Error in log:
[S0000013] Error in quest follows. Terminating quest runtime. 
(Filename: ./Runtime/Export/Debug/Debug.bindings.h Line: 35)

Specified cast is not valid. 
(Filename: ./Runtime/Export/Debug/Debug.bindings.h Line: 35)

[SAVE729.zip](https://github.com/Interkarma/daggerfall-unity/files/5112199/SAVE729.zip)

Interkarma:
> P.S.: Fixed by rebuilding addressables

Glad you sorted it. :)

If using Play Mode Script > Use Existing Build, then Addressables will need to be rebuilt on developer side anytime text is updated.

If using Play Mode Script > Use Asset Database, then it should always pull from assets directly rather than from built addressables. I recall you had some trouble with that play mode script before though. Not sure if that's a condition unique to the Linux editor or not.

petchema:
Probably because I barely ever use the editor UI for running the game.
Instead, I used it as a command-line compiler, scripting a build (based on the Makefile solution that was once published in the forums) and a deployment.
But I don't know if addressables can be rebuilt that way...

PR #1906, opened 2020-08-25
Set the race of static NPCs

df-ferital:
Static NPC race is now stored in its own dedicated field. If not available from faction data, fallback to race of current region.

This also fixes an issue reported by @petchema in #1902 where Aubki-i was using %oth in a quest before being assigned as questor. The previous workaround was a fallback to race of current region. Now, Aubk-i will properly use the Redguard pool of oathes.

### DIFF
--- a/Assets/Scripts/Game/Questing/QuestMCP.cs
+++ b/Assets/Scripts/Game/Questing/QuestMCP.cs
@@ -156,8 +156,14 @@ namespace DaggerfallWorkshop.Game.Questing
                 Symbol[] questors = parent.GetQuestors();
                 if (questors.Length > 0)
                     race = parent.GetPerson(questors[0]).Race;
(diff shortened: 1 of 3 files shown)


jefetienne:
Thank you so much for adding this. I have been working with someone for mod development for having static NPC replacement to replace certain textures based on the region/climate they are currently in, e.g. redguard taverns have redguard variants of static NPCs.

I had a small bit of progress where the textures got replaced at runtime but grinded to a halt trying to figure out how to deal with animated replacements. I think this PR might help a bit.

df-ferital:
Glad to know this will be helpful :) And I'm really looking forward to the mod you're working on!

Interkarma:
Thanks for taking a look at this and improving on the earlier quick patch. :)

PR #1907, opened 2020-08-25
Don't show holiday text message unless HUD is top window

ajrb:
It can mess with non-modal UI's like those used by Tedious Travel etc.

I think this works well, but I would be interested in other devs opinions before this one gets merged. Motivated by holiday message boxes messing up the travel UI of my upcoming Travel Options mod when they pop up at the wrong time.


Interkarma:
No problems with this change. :)

PR #1908, opened 2020-08-25
Implemented cn2 macro for jokes

ajrb:
Just grabs first city in region list since it's only used in one joke but appears twice so either needs context or deterministic behaviour... I chose the easy latter.

### DIFF
--- a/Assets/Scripts/Utility/MacroHelper.cs
+++ b/Assets/Scripts/Utility/MacroHelper.cs
@@ -541,8 +541,15 @@ namespace DaggerfallWorkshop.Utility
         }
 
         private static string CityName2(IMacroContextProvider mcp)
-        {   // %cn2 (only used in msg #200 where cn and cn2 are random? places)
-            throw new NotImplementedException();
+        {   // %cn2 (only used in msg #200 where cn2 is a random? city)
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Utility/MacroHelper.cs
```diff
@@ -541,8 +541,15 @@ public static string CityName(IMacroContextProvider mcp)
...
-            throw new NotImplementedException();
+        {   // %cn2 (only used in msg #200 where cn2 is a random? city)
+            // Return first city of the region, unless it's current location, not worth doing more as the macro is hardly ever used.
+            DFRegion dfRegion = GameManager.Instance.PlayerGPS.CurrentRegion;
+            for (int i = 0; i < dfRegion.LocationCount; i++)
+            {
+                if (GameManager.Instance.PlayerGPS.CurrentLocationIndex != i && dfRegion.MapTable[i].LocationType == DFRegion.LocationTypes.TownCity)
+                    return dfRegion.MapNames[i];
+            }
+            return "Daggerfall";
```
Interkarma, line 552: Thanks for fix. :) Ideal to use GetLocalizedText here. Regions are in database with region name as key. I'll patch this one after merge.

PR #1909, opened 2020-08-27
Allow mods to set extra blend space around locations for placing content

ajrb:
Basically gives control over flattening a wider area based on location type. Requested by Kamer for his stables modding work.


Interkarma:
No problems with this one. :)

PR #1910, opened 2020-08-28
Replace mispelled "sorceror" with "sorcerer" in quests

df-ferital:
Fixes done to quest texts only, to avoid any potentially breaking change.


Interkarma:
Thank you!

PR #1916, opened 2020-09-06
Change default bindings of Run and Sneak to Left Shift and Left Alt, respectively

jefetienne:
It seems that it was intended to have both Shifts and Alts be able to work Run and Sneak but when written at the time, there wasn't any support for secondary bindings, thus the "right" versions of the keys accidentally overrode the "left" versions in the default bindings.

I think it's better to just default them to the "left" versions since they are closer to WASD, but leave it up to the player if they want to include the "right" versions in their secondary bindings.

I also fixed some space formatting issues made by earlier PRs.


Interkarma:
A minor conflict on this one now after merging an earlier PR.

jefetienne:
@Interkarma fixed. Also forgot to do the same thing for Slide, which has been done now.

Interkarma:
Thanks for fixing so quickly, you rock! :)

PR #1917, opened 2020-09-07
Ignore secondary binding equivalents for GetMouse methods and HotkeySequences

jefetienne:
Small oversight when designing secondary keybinds allowed `GetMouse` methods and `HotkeySequences` to look for the secondary equivalents of a specific key when it wasn't supposed to. Created a default parameter for GetKey methods to look for the secondary by default, but set them to false instead for the `HotkeySequence` methods, and used the `GetSingleKey` methods for the `GetMouse` methods.


Interkarma:
Thank you :)

PR #1921, opened 2020-09-08
Improved keybind label text and tooltips

jefetienne:
https://imgur.com/a/rBGxQzS

- Changed the display of KeyCodes without altering the strings that are serialized in Keybinds.txt. Non-alphabetic keys are displayed as their true character, except for numbers. Joystick buttons are displayed as "Joy BX" and Joystick axis buttons are displayed as "JoyX BY".
- When in non-SDF rendering, it will display certain KeyCode strings in a Daggerfall DOS-inspired format (barring some such as "SPCEBAR", mouse buttons like "M 1" and "M 2", etc.)
- If the KeyCode string is too long (in preparation for custom combo codes), the label text will display `...` and you can hover over it to see its full name. The tooltip gets updated/disappears if a smaller-length KeyCode string is binded. If you would like to test this, change the `ControlsConfigManager.maxButtonTextLength` to a small number.

### DIFF
--- a/Assets/Scripts/Game/ControlsConfigManager.cs
+++ b/Assets/Scripts/Game/ControlsConfigManager.cs
@@ -14,6 +14,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.RegularExpressions;
 using DaggerfallConnect.Arena2;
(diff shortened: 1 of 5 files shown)


Interkarma:
This is nice! One of those little things that make a big difference to how this UI looks with classic text.

PR #1928, opened 2020-09-12
Bindable KeyCode combinations

jefetienne:
Introduces a way for players to bind a combination to a single action in the controls window, e.g. "LeftShift + K", "RightControl + K", "K + V", "[ + /", "Joy3B0 + JoyB2".

Since KeyCodes are 32-bit integers, the max length of the KeyCode enum is around 300, and the axis KeyCodes are less than 16-bits long, I managed to store two KeyCodes into one as *15* bits each. (I chose 15 instead of 16 so the most significant bit isn't 1 and becomes a negative number.) You can create a combo by using `GetComboCode(key1, key2)`.

When binding in the controls window, it will make sure that:
 - There is no duplicate combo (e.g. two "LeftShift + K"s)
 - The modifier is not used independently in any other binding (e.g "LeftShift + K" and "LeftShift")
 - The modifier is not used as a key (e.g. "LeftShift + K" and "F + LeftShift")

When pressing a combo, it will make sure that:
 - The modifier is "held first" - meaning it is being held solely without any other keys that are combo'd to that modifier (e.g. LeftShift+K jumps, LeftShift+L opens a menu. It checks to make sure that either 'K' or 'L' are not being held. It will ignore keys that are not combo'd to it, like 'W' for forward.)
- It will also check to make sure no other modifiers are being held.

If a modifier is being pressed, the system will ignore the combo'd key as being pressed independently (e.g. Shift+Space sheathes weapon, Space jumps. When holding Shift, `GetKey(KeyCode.Space)` will return false so it prevents jumping.)

In order to detect other non-independently binded keys as being held, the way input got handled had to be rewritten to a polling system that examines every `KeyCode`. It sounds expensive, but I managed to optimize it as best as possible and for it to be only 0.05-0.08 ms more time spent in the `Update()` function according to the Unity Profiler, which is unnoticeable (at least on my end). Also, because of the new system, I was able to remove much of the old joystick axis code.

`HotkeySequences` used the Unity `Event.current` system in `OnGUI()` for key detection, but couldn't sync to the polling system and detect any GetKey methods. I had to rewrite that part so `DaggerfallUI` has an event-driven `OnKeyPress` method and uses a custom `Event` that gets modified in order to get the `HotkeySequences` working again.

### DIFF
--- a/Assets/Scripts/Game/ControlsConfigManager.cs
+++ b/Assets/Scripts/Game/ControlsConfigManager.cs
@@ -147,14 +147,75 @@ namespace DaggerfallWorkshop.Game
         {
             HashSet<String> recorded = new HashSet<String>();
             HashSet<String> dupes = new HashSet<String>();
+            Dictionary<String, HashSet<String>> modifiers = new Dictionary<string, HashSet<string>>();
             String none = KeyCode.None.ToString();
 
-            foreach (String str in texts)
+            // Sort by combos first to record their modifiers
+            var list = texts.OrderByDescending(str =>
+            {
+                if (str.Contains('+'))
+                {
+                    var c = InputManager.Instance.GetCombo(InputManager.Instance.GetComboCode(str));
+                    // Only record the modifier (i.e. the button that gets held down first)
+                    var kc1 = InputManager.Instance.GetKeyString(c.Item1);
+
+                    // Add modifier to 'recorded' so it cannot be used as an independent keybind
+                    if (!recorded.Contains(kc1))
+                        recorded.Add(kc1);
+
+                    HashSet<String> mods;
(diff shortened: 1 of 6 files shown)


jefetienne:
#1929 is made independently to test and make sure that the automap window bindings work fine.

Basically, this PR alone won't fix an issue where a player creates a combo involving Shift/Ctrl/Alt and binds it to the Automap window. If they do, they cannot close the window with that binding because the HotkeySequence toggle wants *no* modifiers to be held, so it will refuse to close the window. I talked to petchema about this and he said that PR should be fine.

Interkarma:
No issues found and works as described for me. I'm happy to get this one into upcoming 0.10.26 for live testing. The next release will also be a preview, so we can catch any issues ahead of next stable release.

PR #1931, opened 2020-09-13
Remove GC in drawing Panels

jefetienne:
Using the Unity profiler, I found that using a foreach loop of a BaseScreenComponentCollection class allocates memory when getting the Enumerator, and Panel.Draw() is called every frame using that loop. The setup wizard alone took 0.7 KB before under DaggerfallUI.OnGUI(), now it is 0.


Interkarma:
This should be fine, thank you! :)

PR #1933, opened 2020-09-13
Apply bonuses and phobias to all attacks

numidium:
Addresses the bug discussed here: https://forums.dfworkshop.net/viewtopic.php?f=31&t=3923

Bonuses and phobias are now applied to player AND enemy armed/unarmed attacks.

Note: This will break any mods that override GetBonusOrPenaltyByEnemyType

### DIFF
--- a/Assets/Scripts/Game/Formulas/FormulaHelper.cs
+++ b/Assets/Scripts/Game/Formulas/FormulaHelper.cs
@@ -644,6 +644,8 @@ namespace DaggerfallWorkshop.Game.Formulas
 
                             damage += hitDamage;
                         }
+
+                        damage += GetBonusOrPenaltyByEnemyType(attacker, target);
                         ++attackNumber;
                     }
                 }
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Game/Formulas/FormulaHelper.cs
```diff
@@ -958,60 +960,66 @@ public static int CalculateBackstabDamage(int damage, int backstabbingLevel)
...
                 {
-                    damage -= attacker.Level;
+                    if (((int)attacker.Career.AnimalsAttackModifier & (int)DFCareer.AttackModifier.Bonus) != 0)
+                        damage += attacker.Level;
+                    if (((int)attacker.Career.AnimalsAttackModifier & (int)DFCareer.AttackModifier.Phobia) != 0)
+                        damage -= attacker.Level;
                 }
             }
-            else if (AITarget.GetEnemyGroup() == DFCareer.EnemyGroups.Animals)
+            else // Assume target is player if not an enemy
```
Interkarma, line 1007: This seems reasonable, but I'd prefer an `else if (target is PlayerEntity)` just for safety though.

Assets/Scripts/Game/Formulas/FormulaHelper.cs
```diff
@@ -958,60 +960,66 @@ public static int CalculateBackstabDamage(int damage, int backstabbingLevel)
...
             {
-                if (((int)attacker.Career.AnimalsAttackModifier & (int)DFCareer.AttackModifier.Bonus) != 0)
+                if ((target as PlayerEntity).RaceTemplate.Name == TextManager.Instance.GetLocalizedText("vampire")) // Vampires are undead, therefore add undead modifier
```
Interkarma, line 1009: Better to use `GameManager.Instance.PlayerEffectManager.HasVampirism()` here than match by string.

numidium: Ah, I didn't notice that method. Much better.

PR #1934, opened 2020-09-14
More accurate pursuit formula

petchema:
Current pursuit algorithm in EnemySenses compute how much time will be needed to reach the current position of the target, then computes by how much the target will have moved by then. This is only an approximation, because it does both the hypothesis that the target won't move and that it will move. The fastest the target, the less accurate it gets.

One can exploit that inaccuracy by rushing at enemies, that will turn their back, pursuing a position behind them. Hence current algorithm has an additional test to detect that condition, but it doesn't seem to work perfectly, the exploit still exists.

Accurate approach is to compute at what moment the target will be on a sphere centered at pursuer position, growing at its interception speed each second; Or, in 4 dimensions (x, y, z, t), compute the intersection of a cone and a line. This is slightly more expensive, solving a quadratic equation then taking its smallest positive solution, but doesn't seem to suffer from the previous exploit.

### DIFF
--- a/Assets/Scripts/Game/EnemySenses.cs
+++ b/Assets/Scripts/Game/EnemySenses.cs
@@ -542,34 +542,38 @@ namespace DaggerfallWorkshop.Game
             float b = 2 * Vector3.Dot(d, v);
             float c = d.sqrMagnitude;
 
-            Vector3 prediction;
+            Vector3 prediction = assumedCurrentPosition;
 
-            float disc = b * b - 4 * a * c;
-            if (disc >= 0)
+            float t = -1;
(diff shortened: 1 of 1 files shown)


Interkarma:
This is a lovely mathy solution. No issues found and the additional overhead is not noticeable. Cheers!

petchema:
Coefficients a, b, c turned out to be surprisingly simple and meaningful (difference in speed between the opponents, dot product that tells whether the target is closing by or moving away, distance to the target). I'm very pleased with the result.
It's only incremental improvements over the nice existing enemies AI logic, but I've found that enemies moving more swiftly and making less mistakes about how to block you path makes the game feel more natural but also more challenging...

PR #1935, opened 2020-09-14
Allow TerrainSamplers to disable the default job and do their own location blending / flattening

ajrb:


Interkarma:
Nice idea, thanks for the refinement. 👍

PR #1936, opened 2020-09-14
Move material selection into overridable formulas

ajrb:
Feels like the calculation of random material for a given level should be exposed for mods


Interkarma:
Good change, thank you mate.

PR #1937, opened 2020-09-15
fix Clocks that would return 0 days

petchema:
Clocks with
- both timerValue0 = timerValue1 = 0 and
- neither flag & 1 or flag & 16

Will have a final clockTimeInSeconds of 0, not overridden by a travel time.
Capture d’écran du 2020-09-15 20-56-58
(source https://www.twitch.tv/videos/741689738?t=4h16m16s )

Until Clock implementation is changed, a workaround for those cases is to modify flag to have either flag & 1 or flag & 16.
Fortunately, there's only 2 such instances.


petchema:
What a coincidence, ProJared has just been doing the other quest, that is also broken (couldn't deliver the sapphire)
https://www.youtube.com/watch?v=mTn5wW_cc3A&t=8m12s

Issue #1911, opened 2020-08-30
0.10.25: world borders have become sticky


petchema:
**Describe the bug**
When trying to go past world borders, you get stuck in place and can only free yourself by mean of, say, fast travel
Tested with north, east and south borders.
tdbg shows that players goes slightly further than the border coordinate (say going south, coordinate becomes 9995, when the border is at 10000).

**To Reproduce**
1. Load game save below
2. Travel north until you hit world northern limit
3. Try to get away from the border

**Expected behavior**
Be blocked at the world border, but be able to move away from it

**Desktop (please complete the following information):**
 - OS: Linux
 - Version 0.10.25

Tested to work with 0.10.24 compiled with Unity 2018.2.x.

**Additional context**
Game save near northern world border
[SAVE752.zip](https://github.com/Interkarma/daggerfall-unity/files/5147330/SAVE752.zip)

Interkarma:
This is a good find. Cheers :)

SoullessHumanoid:
Tested this in Unity editor. I never got permanently stuck to the border, but did experience visual jitter and was temporarily suspended above ground when jumping against the border. Found some possibly unsafe code:
`gameObject.transform.position = lastFramePosition;`
If somehow `lastFramePosition` is set outside the world border, it would be impossible to move. I think it would be better to clamp only the coordinate that is outside the limits (Z for northern border), making it behave similar to a wall.
I could try to implement this, after researching how world coordinates (`worldX, worldY`) correspond to player transform position.

petchema:
Fixed in master branch

PR #1939, opened 2020-09-18
Made world borders non-sticky

SoullessHumanoid:
Fixed issue #1911 by changing world borders only affect the coordinate that goes out of bounds, and pushing the player slightly away from the border, to prevent getting stuck if `lastFramePosition` is somehow set outside allowed area.

### DIFF
--- a/Assets/Scripts/Internal/PlayerGPS.cs
+++ b/Assets/Scripts/Internal/PlayerGPS.cs
@@ -318,12 +318,23 @@ namespace DaggerfallWorkshop
 
             // Snap back to physical world boundary to prevent player running off edge of world
             // Setting to approx. 10000 inches (254 metres) in from edge so end of world not so visible
-            if (WorldX < 10000 ||       // West
-                WorldZ > 16370000 ||    // North
-                WorldZ < 10000 ||       // South
-                WorldX > 32750000)      // East
-            {    
-                gameObject.transform.position = lastFramePosition;
+            if (WorldX < 10000)
+            {
+                // West
+                gameObject.transform.position = new Vector3(lastFramePosition.x + 0.01f, gameObject.transform.position.y, gameObject.transform.position.z);
+            } else if (WorldX > 32750000)
+            {
+                // East
+                gameObject.transform.position = new Vector3(lastFramePosition.x - 0.01f, gameObject.transform.position.y, gameObject.transform.position.z);
+            }
+            if (WorldZ < 10000)
+            {
+                // South
+                gameObject.transform.position = new Vector3(gameObject.transform.position.x, gameObject.transform.position.y, lastFramePosition.z + 0.01f);
+            } else if (WorldZ > 16370000)
+            {
+                // North
+                gameObject.transform.position = new Vector3(gameObject.transform.position.x, gameObject.transform.position.y, lastFramePosition.z - 0.01f);
             }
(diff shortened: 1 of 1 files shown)


petchema:
That probably works, any idea why this regression exists in the first place though? The logic looked solid...
Some change in the way Unity handles Update() and FixedUpdate()?

SoullessHumanoid:
I can't be sure about  what caused this, as I never got permanently stuck, all I got was a few seconds suspended above ground, but that is not too good either, even if encountered rarely (jumping just before hitting world border). Someone who got permanently stuck should confirm that the bug is fixed.

petchema:
I don't have access to my rig this weekend, but if nobody beats me to it, I'll check on monday

petchema:
Tested a bit ahead of schedule, and it works fine for me!

Interkarma:
This one does work. I hit a strange problem when hitting the borders at speed then trying to back up - player can be flung backwards through the air.

I'll take another look at this one later. Possibly a new method needed to clamp movement at edges of world.

petchema:
Since Unity recommends moving "physics" to FixedUpdate(), I tried moving the bounding code to FixedUpdate() and that made the borders non-sticky:

```diff
diff --git a/Assets/Scripts/Internal/PlayerGPS.cs b/Assets/Scripts/Internal/PlayerGPS.cs
index 52d855af9..39ed59dd1 100644
--- a/Assets/Scripts/Internal/PlayerGPS.cs
+++ b/Assets/Scripts/Internal/PlayerGPS.cs
@@ -315,6 +315,9 @@ namespace DaggerfallWorkshop
                 UpdateNearbyObjects();
                 nearbyObjectsUpdateTimer = 0;
             }
+        }
+
+        void FixedUpdate() {
 
             // Snap back to physical world boundary to prevent player running off edge of world
             // Setting to approx. 10000 inches (254 metres) in from edge so end of world not so visible
```

Interkarma:
Strange, I still experience stick borders with this logic in FixedUpdate(). I'd still like to find a different solution to this one.

petchema:
Mmmh, I must not have started from a clean state, because indeed it does not seem to fix the issue.
Sorry for the noise!

Interkarma:
It's no problems at all, your feedback is always welcome! You easily hit the nail on the head 99% of the time.

petchema:
I wonder if the issue is the order in which FixedUpdate()s are executed; If, withing a frame, coordinates are first clamped then the player moves, that could explain the issue.
And I'm not sure anything currently ensures that the FixedUpdate()s are evaluated in a specific order, so that would explain why the whole thing is brittle, and could have broken with something seemingly unrelated like an engine update...

petchema:
Using LateUpdate (https://docs.unity3d.com/Manual/ExecutionOrder.html) instead of FixedUpdate() in my previous fix proposal does seem to work. I'm not sure it's the best possible fix, but it seems to confirm my new hypothesis.

Interkarma:
Sounds good, happy to try this! I'd really like to find a better (less brittle) solution for this overall, but just fixing the sticky behaviour is a great start. Thank you for all your experimentation on this one.

Interkarma:
Thank you again for PR. I've gone with Pango's fix of moving the detection code to LateUpdate. It's still something that might be revisited later.

PR #1943, opened 2020-09-24
Ensure shrines are handled correctly

ajrb:
See bug report https://forums.dfworkshop.net/viewtopic.php?f=31&t=4117

I changed the localization table for internal strings to add:

temple = temple

but it doesn't seem to have got picked up in the commit and I don't know what to do.


Interkarma:
Cheers. I'll need to add "temple" to Internal_Strings and rebuild addressables. Will take care of that when I merge commit.

PR #1947, opened 2020-09-30
Enemy + Mobile NPC Spawn Events

joshcamas:
Hooking into enemy / npc spawning isn't super doable currently, making modding a bit tough. This makes modifying enemies itself a lot harder than it should be! Adding spawn events is a major way to make this easier.

All cases I could find correctly trigger these events - dungeon enemies, quest spawns, sleeping encounters, etc.


Interkarma:
This looks fine to me, thank for the contribution! 👍

PR #1948, opened 2020-10-01
Change crouch action to "key-down" (ActionStarted)

jefetienne:
The sneak action either uses ActionStarted when in toggled mode, or HasAction when not. To keep a consistency for all physical actions, I believe the crouch action should be changed to "ActionStarted" as well.


Interkarma:
No problems here. Cheers!

PR #1949, opened 2020-10-01
Remove changes from PR#1808

ajrb:
The changes to improve horse jumping sounds broke the stopping of sounds when horse/cart stops. This commit removes those changes, as I think this is worse than clipped no sound if a player holds space to bunny hop the horse.

See: https://forums.dfworkshop.net/viewtopic.php?f=5&t=4191

I'd be happy to put in a change to prevent bunny hopping horses which would be a better approach to the issue than allowing the sound to play longer. TBH if players want to hold space, I care very little what it sounds like as it's a silly thing to do when playing, which is why preventing that action would be my preference here if something has to be changed.

### DIFF
--- a/Assets/Scripts/Game/TransportManager.cs
+++ b/Assets/Scripts/Game/TransportManager.cs
@@ -151,8 +151,6 @@ namespace DaggerfallWorkshop.Game
         DaggerfallAudioSource dfAudioSource;
         AudioSource ridingAudioSource;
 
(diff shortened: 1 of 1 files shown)


ajrb:
This can be closed if you accept #1950 from Pango. Depends which one you prefer. I'm really not bothered, I just don't want to have to do any more work on it right now - hence the removal.

PR #1950, opened 2020-10-01
fix horse/cart sound stopping (regression in 0.10.23)

petchema:
- reset flag so that new coroutines can be started
- And rename stoppedRidingAudio as pendingStopRidingAudio to better convey
its purpose.

Fixes a regression introduced in 0.10.23
Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&t=4191

### DIFF
--- a/Assets/Scripts/Game/TransportManager.cs
+++ b/Assets/Scripts/Game/TransportManager.cs
@@ -151,7 +151,7 @@ namespace DaggerfallWorkshop.Game
         DaggerfallAudioSource dfAudioSource;
         AudioSource ridingAudioSource;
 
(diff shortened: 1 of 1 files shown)


petchema:
This PR is an alternative to #1949

Interkarma:
Thank you @petchema. I thought about this for a while and decided with this one rather than remove changes in @ajrb's other PR.

I do agree with Hazelnut the bunny hopping is a bit silly. I didn't implement bunny-hopping early on and got a lot of requests for this, and eventually buckled and put it in. If we're going to let horses jump, might as well keep the behaviour consistent with on foot.

PR #1951, opened 2020-10-05
fix lunar computations

petchema:
Make sure full moons happen on 21st and 25th of Morning Star, then
23rd and 27th of Sun Dawn, etc.

Forums: https://forums.dfworkshop.net/viewtopic.php?f=22&t=4203

### DIFF
--- a/Assets/Scripts/Utility/DaggerfallDateTime.cs
+++ b/Assets/Scripts/Utility/DaggerfallDateTime.cs
@@ -678,7 +678,7 @@ namespace DaggerfallWorkshop.Utility
             int offset = (isMasser) ? 3 : -1;
 
             // Find the lunar phase for current day
-            int moonRatio = (Day + offset) % 32;
+            int moonRatio = (DayOfYear + Year * 360 + offset) % 32;
             LunarPhases phase = LunarPhases.None;
             if (moonRatio == 0)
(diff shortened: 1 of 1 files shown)


petchema:
Without that PR, full moons seem to happen every 2nd and 30th of each month (I would have guessed 1st and 29th, maybe Day(of month) is 0 based? either way, it's broken)

Interkarma:
Thank you for fix! I suspected it might be off with this but didn't investigate. Appreciate your attention to detail as always.

As an aside, I wonder if the Enhanced Sky mod will need lunar calculations fixed so the visible full moons line up with game calendar now? The old lunar formula is a remnant from the very early days before mod system when some things like Enhanced Sky were actually part of the code base. If the mod now tracks this differently to core then moon graphics might not line up. Tagging @Nystul-the-Magician for comment. :)

Nystul-the-Magician:
tbh, I don't know, I never dived deep into Enhanced Sky mod's code, I just tried to get Lypyl's mod to run whenever a change breaked something.

petchema:
At a glance, it looks like MoonController.GetDay() already used the right formula:
https://github.com/Nystul-the-Magician/dfunity-mods/blob/master/EnhancedSky/Scripts/MoonController.cs#L210

PR #1953, opened 2020-10-11
Create a setting to choose the FM soundtrack

numidium:
Adds a new checkbox to advanced settings under audio. If checked, the game will use the FM version of the soundtrack if an OPL3 soundfont is in use or custom music files that replace FM songs are loaded.

It works well with @petchema's mod here: https://www.nexusmods.com/daggerfallunity/mods/79/
The OPL3 soundfont found here: https://musical-artifacts.com/artifacts/15 doesn't make the songs sound quite the same as they do in classic, unfortunately. It's theoretically possible to mimic Classic's OPL3 soundtrack with a proper soundfont in conjunction with this PR, however.

### DIFF
--- a/Assets/Resources/defaults.ini.txt
+++ b/Assets/Resources/defaults.ini.txt
@@ -30,6 +30,7 @@ RandomDungeonTextures=0
 
 [Audio]
 SoundFont=
+AlternateMusic=False
 
 [ChildGuard]
 PlayerNudity=False
--- a/Assets/Scripts/Game/SongManager.cs
+++ b/Assets/Scripts/Game/SongManager.cs
@@ -4,7 +4,7 @@
 // License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
 // Source Code:     https://github.com/Interkarma/daggerfall-unity
 // Original Author: Gavin Clayton (interkarma@dfworkshop.net)
-// Contributors:    
+// Contributors:    Numidium
 // 
 // Notes:
 //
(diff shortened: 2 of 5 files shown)


petchema:
Well, my mod consist of sampled tracks, so it disables MIDI synth rendering.
Using this PR to switch to OPL3-optimized tracks and using an OPL-3 soundfont is an alternate way to approximate classic SoundBlaster sound; But the two approaches cannot be combined.

numidium:
@petchema, the PR is able to work with the .ogg music files in your mod because the FM versions of the tracks are included with those. As an example: player enters a house. If alternate soundtrack is enabled then song_23fm.ogg will play instead of song_23.ogg. 

The FM soundtrack uses a different set of tracks so this PR would simply switch out certain track lists such as dungeon music, fm_sqr2 in place of gshop, etc.

ajrb:
Forgive my ignorance but what is an "FM soundtrack"?

numidium:
@ajrb, Daggerfall has a separate set of midi tracks that are optimized for FMSynth/OPL3 midi devices like the one built into Soundblaster 16. If you choose Soundblaster for your music device in classic you'll hear that version of the soundtrack. You'll notice that it has a few tracks missing such as the shopping theme which is replaced by one of the tavern songs or the mages guild songs which are replaced with the night theme from Arena.

Interkarma:
This works well. Thank you for the option! :)

petchema:
Bug report that seems related to that new setting: https://forums.dfworkshop.net/viewtopic.php?f=5&t=4401

petchema:
The number of dungeon songs is hardcoded in some formula. Either there must also be 15 FM songs, or the formula must be fixed:
```diff
diff --git a/Assets/Scripts/Game/SongManager.cs b/Assets/Scripts/Game/SongManager.cs
index 78ee9fb87..6f526d31c 100644
--- a/Assets/Scripts/Game/SongManager.cs
+++ b/Assets/Scripts/Game/SongManager.cs
@@ -332,7 +332,7 @@ namespace DaggerfallWorkshop.Game
                     }
                     DFRandom.srand(unknown2 ^ ((byte)region << 8));
                     random = DFRandom.rand();
-                    index = (int)(random % 15);
+                    index = (int)(random % DungeonInteriorSongs.Length);
                 }
                 else if (currentPlaylist == SneakingSongs)
                 {
```

petchema:
I submitted a fix: #1985

PR #1955, opened 2020-10-16
Use HUD text for light source depletion message

numidium:
Currently uses a message box which stops the player in their tracks.


Interkarma:
This makes sense to me and happy to make change. Are you happy with this one @ajrb?

ajrb:
Okay, happy for it to change if it's bothering people.

PR #1956, opened 2020-10-17
Publicize parts of PlayerActivate and NativePanel for mod support

jefetienne:
I'm writing a mod for displaying tooltip information in the world and some of these methods need to be public in order to retrieve all of the information about it. Also publicized `ChangeInteractionMode` and `Pickpocket` as they would probably be useful for future mods.

`NativePanel` would be useful as a public property to get information about certain information about windows and adding components to them without overriding the class itself.


Interkarma:
Thank you :)

Issue #1942, opened 2020-09-24
System performance extremely bad with Daggerfall in background


keenerb:
Any way to decrease framerate or CPU usage, especially with Daggerfall as a background process?  I can hardly switch to a different application, rendering seems to run at 100% even with app minimized.

petchema:
Just saw that Unity sends events when losing/receiving focus, so it should be at least possible to disable parts of the main loop when the game doesn't have the focus: 
https://docs.unity3d.com/ScriptReference/MonoBehaviour.OnApplicationFocus.html

Interkarma:
It's possible to disable "run in background" in player settings so that runtime will be suspended when game is minimised or otherwise loses focus.

ajrb:
Isn't it just easier to press escape to bring up the main pause menu and the alt-tab? That should remove most of the work Unity is doing shouldn't it?

keenerb:
That doesn't seem to help, actually.

Interkarma:
Are you sure this is an issue with DFU and not your mods? DFU only uses around 10% CPU and <1GB RAM when in a large city.

Please use the Help & Support forums for support with individual setups.

https://forums.dfworkshop.net/

keenerb:
The mods exacerbate the issue, but DFU with or without mods still seems to consume the same CPU and GPU resources in the foreground as it does minimized.  Whether you want to spend the time pausing the render thread (or whatever you programmers do) is up to you of course!

I managed to bring cpu/GPU usage down to acceptable levels by setting targetframerate to 30 in settings.ini

Interkarma:
I can change this to suspend the runtime when game is minimized or loses focus. This should resolve that problem.

This was known to cause issues in older builds, which is why the setting was disabled. Now that engine has been updated to a later version, this shouldn't be an issue any longer. I'll roll this change out with 0.10.27.

PR #1957, opened 2020-10-18
make runInBackground a setting

petchema:
Some people seem to be annoyed by runInBackground being false now.
It's hard to tell what value satisfies the most people, however I suspect it's when runInBackground is enabled, and only a minority really prefers it to be disabled.
So, making it a hidden setting (.ini only), defaulting to true.

Original issue:
https://github.com/Interkarma/daggerfall-unity/issues/1942
Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&t=4236

### DIFF
--- a/Assets/Resources/defaults.ini.txt
+++ b/Assets/Resources/defaults.ini.txt
@@ -13,6 +13,7 @@ VSync=True
 TargetFrameRate=0
 Fullscreen=False
 ExclusiveFullscreen=False
+RunInBackground=True
 FieldOfView=65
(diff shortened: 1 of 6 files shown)


petchema:
I put the setting in [GUI] section, but thinking about it it's not exactly GUI-related...

ajrb:
Video settings may be a better fit for this, maybe?

petchema:
I changed that in the last commit already

Interkarma:
Excellent work Pango. Thank you so much. :)

PR #1958, opened 2020-10-24
Increase micromap size

numidium:
The micromap dimensions for Ruins of Moorford Manor in Daggerfall exceed the current boundaries. Hopefully this will be large enough for all dungeons.

### DIFF
--- a/Assets/Scripts/Game/Automap.cs
+++ b/Assets/Scripts/Game/Automap.cs
@@ -1844,8 +1844,8 @@ namespace DaggerfallWorkshop.Game
             }
 
             int microMapBlockSizeInPixels = 2;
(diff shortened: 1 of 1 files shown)


petchema:
Actually that changes the micromap resolution, not its size, so it will make micromap pixels smaller in most dungeons.
What about autoadjusting resolution?
https://github.com/Interkarma/daggerfall-unity/compare/master...petchema:autoadjust-micromap-size

numidium:
It alters the size of the texture, though I'm not sure what all that entails. My tests showed it working as intended - the pixels may have been smaller but I didn't notice. That said I like your approach a lot better. I'm fine with closing this PR and using that instead.

PR #1959, opened 2020-10-26
Autoadjust micromap size

petchema:
Adjust dungeon micro map resolution to dungeon size.
Fixes issues with some dungeons like The Ruins of Moorford Manor, Daggerfall.

This is an alternative to #1958


petchema:
This implementation justifies the micromap to the left of screen, which classic doesn't

Interkarma:
Thank you! :)

PR #1961, opened 2020-11-03
Allow custom Renderer for ObjectPositioner

TheLacus:
By default  a mesh renderer is seeked on gameobject to get bounds. Renderer field can be assigned from the inspector to override this behaviour and use a different renderer. For example this is useful when a lod group is used and root gameobject doesn't have a mesh renderer attached.


Interkarma:
A simple enough change and adds some flexibility. Thank you. :)

PR #1963, opened 2020-11-07
Correct Bhoriane spelling everywhere

df-ferital:
Fix Bhoriane county mispelled as Bhoraine in FACTION.TXT, which resulted in similar mistakes in DFU source files.
Discussion here: https://forums.dfworkshop.net/posting.php?mode=reply&f=5&t=4294.

### DIFF
--- a/Assets/Scripts/API/FactionFile.cs
+++ b/Assets/Scripts/API/FactionFile.cs
@@ -165,7 +165,7 @@ namespace DaggerfallConnect.Arena2
             Menevia = 209,
             Alcaire = 210,
             Koegria = 211,
-            Bhoraine = 212,
+            Bhoriane = 212,
             Kambria = 213,
             Phrygia = 214,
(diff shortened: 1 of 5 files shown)


Interkarma:
Thank you for the fixes Ferital. 👍 I don't _think_ this will cause any issues. The faction data is handled by ID, region loading is done by index. The display name is generally just that. I haven't been able to provoke any gameplay problems testing this PR so far. I even tried setting up quests before PR then played through them successfully after PR applied.

I was surprised to see that DFTFU import by name in editor and Atlas editor window still work OK after name change, I thought that would throw it for sure. Happy to see this all working as expected. An excellent result!

df-ferital:
Great, I feared that the fix could break quests especially, glad to know this is not the case!

Interkarma:
Still no issues found here. If there are any problems, I've missed them. I'll merge this now. Thank you!

PR #1964, opened 2020-11-14
Handle interior guard spawns for taverns and houses

numidium:
Guards now spawn at the ground-level door in taverns and residences. The door farthest from the building's origin is chosen to prevent guards entering from the building's courtyard.
Note that guards can still spawn in an unreachable place in disconnected interiors. This PR does not cover that issue.


Interkarma:
All seems to work perfectly here. Nice way to cover that issue of spawning near centre of building for those with courtyards. Thank you! :)

Issue #1962, opened 2020-11-05
Can we make `CanRest()` in `DaggerfallRestWindow` virtual?


lonegamedev:
Would help a lot, cause now I have to copy-paste the entire `DaggerfallRestWindow` class to extend the default `CanRest()` method.

jefetienne:
Hey there, so there is a method for modding that can allow for a mod to prevent a player from resting without having to override anything, called RegisterPreventRestCondition. It is found in GameManager.

```summary: Registers a boolean method that prevents the player from starting or continuing to rest
'handler' - The method returning a boolean whether the player can rest
'message' - The message to be displayed
public void RegisterPreventRestCondition(Func<bool> handler, string message)
```

I believe the handler method should return true to prevent resting, false to continue resting.

lonegamedev:
Hi, thank you for replying. 

In that case, can we have:
public getter for `currentRestMode` 
and 
'enum RestModes' public?

Cause in my case, I want to allow loitering, but prevent them from resting.

I can make a pull request if that helps.

jefetienne:
Hey lonegamedev, sorry for the wait. I made a PR (#1965) that should help with this.

lonegamedev:
Hey, no worries. Thank you for that. Looking forward to the merge and the release 😃

Interkarma:
I've merge this PR now. :)

PR #1965, opened 2020-11-16
Publicize CurrentRestMode

jefetienne:
This is for issue #1962.

This PR essentially would replace the utility of `PlayerEntity.IsLoitering` and `PlayerEntity.IsResting`, but may be kept for mod compatibility reasons. I also have no problem with removing them if requested. :)

I publicized the `RestModes` enum and kept it in `DaggerfallRestWindow`, but I can move it elsewhere to `DaggerfallUnityEnums` if mod compatibility is also not an issue.

### DIFF
--- a/Assets/Scripts/Game/Entities/PlayerEntity.cs
+++ b/Assets/Scripts/Game/Entities/PlayerEntity.cs
@@ -144,6 +144,7 @@ namespace DaggerfallWorkshop.Game.Entity
         public bool PreventNormalizingReputations { get { return preventNormalizingReputations; } set { preventNormalizingReputations = value; } }
         public bool IsResting { get { return isResting; } set { isResting = value; } }
         public bool IsLoitering { get; set; }
+        public DaggerfallRestWindow.RestModes CurrentRestMode { get; set; }
         public Races Race { get { return (Races)RaceTemplate.ID; } }
         public RaceTemplate RaceTemplate { get { return GetLiveRaceTemplate(); } }
         public RaceTemplate BirthRaceTemplate { get { return raceTemplate; } set { raceTemplate = value; } }
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallRestWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallRestWindow.cs
@@ -68,7 +68,6 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
         protected const float loiterWaitTimePerHour = 1.25f;
         protected const int cityCampingIllegal = 17;
 
-        protected RestModes currentRestMode = RestModes.Selection;
         protected int minutesOfHour = 0;
         protected int hoursRemaining = 0;
         protected int totalHours = 0;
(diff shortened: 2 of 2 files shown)


Interkarma:
I'm happy to leave in IsLoitering and IsResting for overall compatibility. Thank you for fix! :)

petchema:
I'm currently tracking some regression:
When trying to sleep (say at night, to increase the likeliness of an encounter) and get interrupted, getting out of enemy reach and trying to rest again, the 3 rest options show up but your rest is immediately interrupted by a "You wake up." message
you wake up
And git bisect points at commit 29e261610b54c74568113a24c47ead8d6506b3f7

petchema:
Test game save, if you need one
[SAVE871.zip](https://github.com/Interkarma/daggerfall-unity/files/5692261/SAVE871.zip)

Interkarma:
Maybe the removal of `protected RestModes currentRestMode = RestModes.Selection;`? This default no longer appears to be set anywhere.

If not, should we create a quick issue for this one? I'm starting to unplug a little for Christmas and don't want to lose track.

jefetienne:
I think the issue is because when the window is pushed, it may still be using the last rest mode stored in GameManager because of an interruption. I'll take a look at it.

petchema:
That was my guess too, but I'm not sure how to emulate that back with a property.
I did
```diff
diff --git a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallRestWindow.cs b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallRestWindow.cs
index c359ac45c..273777315 100644
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallRestWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallRestWindow.cs
@@ -116,6 +116,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             : base(uiManager)
         {
             this.ignoreAllocatedBed = ignoreAllocatedBed;
+            currentRestMode = RestModes.Selection;
         }
 
         #endregion
```
and it seems to work...

jefetienne:
> That was my guess too, but I'm not sure how to emulate that back with a property.
> I did
> 
> ```diff
> diff --git a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallRestWindow.cs b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallRestWindow.cs
> index c359ac45c..273777315 100644
> --- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallRestWindow.cs
> +++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallRestWindow.cs
> @@ -116,6 +116,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
>              : base(uiManager)
>          {
>              this.ignoreAllocatedBed = ignoreAllocatedBed;
> +            currentRestMode = RestModes.Selection;
>          }
>  
>          #endregion
> ```
> 
> and it seems to work...

I think maybe putting it in OnPush() might be better? I forget how often the methods Start(), Setup() and the Constructor() are called when opening the window multiple times.

petchema:
Well, I think removed line
```
protected RestModes currentRestMode = RestModes.Selection;
```
was only executed once too, when the window was built...
I'm not sure if setting it explicitly in the constructor is correct, but I think it should be equivalent to what was there before

petchema:
> If not, should we create a quick issue for this one? I'm starting to unplug a little for Christmas and don't want to lose track.

I created a PR

Interkarma:
Great work, thank you @petchema and @jefetienne.

PR #1966, opened 2020-11-20
Ensure models with registered actions are not combined for RMB and interiors

ajrb:
As discussed here: https://forums.dfworkshop.net/viewtopic.php?f=22&t=4309


Interkarma:
Simple and functional, just how I like them. Thank you mate! :)

PR #1967, opened 2020-11-22
Fix exploit for Kightly Order armor service

ajrb:
See: https://forums.dfworkshop.net/viewtopic.php?f=5&t=4306


Interkarma:
Thanks for fix! :)

PR #1968, opened 2020-11-24
Immediately apply volume changes in launcher

petchema:
Immediately apply volume changes in launcher.
Play door effect to check sound volume.

I had to create a temporary GameObject to get a separate
DaggerfallAudioSource. Not sure how I feel about that.


Interkarma:
There's a DaggerfallAudioSource attached to DaggerfallUI that might be handy for this. Try `DaggerfallUI.Instance.PlayOneShot(clip);`

Love the choice of creaky door to test volume btw :)

petchema:
It shares volume with the music, so sadly that's not an option

Interkarma:
I see what you mean. Then it's as fine a solution as any for the advanced settings window, and only persists while that window is open. Thank you for the refinement. :)

PR #1969, opened 2020-11-25
Added Custom Sky Loading

joshcamas:
This adds support for custom day and night image loading from loose files and mods.

To load a custom sky texture, it would load this filename from the textures folder:
"SKY03_0-0.DAT"

This would replace sky 03 at frame 0 and east.

Nighttime is a bit more complex, since it has several options for it to scan:

1) NITE03I0.IMG --> This will act identically to the vanilla daggerfall, with a total of 4 night types being supported. No stars!

2) NITE03I0-0.IMG / NITE03I0-1.IMG --> This will act identically to the vanilla daggerfall, with a total of 4 night types being supported. HOWEVER: East and West will be supported by the final index. No stars!

3) NITEFULL28I0.IMG --> This will act more like the daytime skies, with a total of 32 night sky types being supported. No stars!

4) NITEFULL28I0-0.IMG / NITEFULL28I0-1.IMG --> This will act more like the daytime skies, with a total of 32 night sky types being supported, and east and west being supported by last index. No stars!

The system will go through 1-4, and if it successfully finds a file matching that name, it will use it. And of course if it fails all of them, it will load the vanilla files.

**A few other things to note:**

- As mentioned, stars are not supported in custom images. This is because there isn't much reason for it, especially since you can have east+west images separate, so you can manually paint stars and the moons or whatever. However, if stars are something that are wanted, I could make it so stars are default, then a filename with a "_NOSTARS" postfix would disable the stars.

- For custom images no night seam is fixed - the images are supposed to be mirrored, just like daytime. In my eyes, the seam fixer exists due to the malformed night sky images in vanilla.

- Any sky image size is supported, as long as the west+east variants are the same size.


Assets/Scripts/Internal/DaggerfallSky.cs
```diff
@@ -400,31 +388,189 @@ private void LoadCurrentSky(int targetFrame)
...
+        }
+
+        /// <summary>
+        /// Loads day sky from arena files
+        /// </summary>
+        /// <param name="skyIndex"></param>
+        /// <returns>SkyColors loaded</returns>
+        private SkyColors LoadVanillaDaySky(int skyIndex, int frame)
+        {
+            SkyFile skyFile = new SkyFile(Path.Combine(dfUnity.Arena2Path, SkyFile.IndexToFileName(skyIndex)), FileUsage.UseMemory, true);
```
Interkarma, line 545: I would like for SkyFile to be loaded and cached only when the sky file itself changes, rather than loading and discarding away the entire SkyFile every frame. This is why the SkyFile data is currently handled the way it is in this class. This is particularly noticeable with increased timescale (e.g. set_timescale 1000), it generates large GC spikes and more disk accesses that could be avoided.

Interkarma:
Thank you for the work on this. :) It looks like it should work well, I just need to test some more before merging. It would be nice to see some handpainted sky backgrounds someday. There's a particular charm to this the procedural skies don't quite match.

Assets/Scripts/Internal/DaggerfallSky.cs
joshcamas: From a quick glance, I'm surprised it runs EVERY frame, since there is a check on line 163 that checks if the skyindex is different. regardless, I see what you mean about there lacking any caching of the skyfile, I agree that this should be fixed in the future.

Interkarma: Yep, I just mean each time the sky frame changes (not each frame rendered).

My mental map was that it cached the SkyFile and only loaded a new SkyFile when the sky file changed - but doesn't look like I implemented that way for some reason. Not sure why my memory doesn't mesh up here. Maybe it was something I meant to do back in 2014/15 when I first wrote this and just never got back to it.

joshcamas: One way to do it would be instead of caching the skyfile itself, it could simply cache the textures in a dictionary, with the key being the filename.

Interkarma: I'd rather not cache each Texture2D. These are substantially larger than the compact SkyFile format, and modded textures could be larger still. I mainly just want to avoid hitting the disk for the entire SkyFile every time the sky frame changes, as most of the time it will be the same SkyFile as previous frame. I also bet disk caching makes that a pretty minimal problem anyway, so the whole concern is probably misplaced on my end. :)

I'll loop back to this one later. I just made an incorrect assessment based on faulty memory when looking at your changes without properly comparing to original. So much time has passed since this class was created that I should have spent more time re-familiarising myself with my own work first. I apologise, and will try to avoid this in future.

PR #1970, opened 2020-11-26
Combine SetupDefaults and ResetDefaults as one method

jefetienne:
To remove code duplication and prevent binding inconsistencies in the future, this PR combined both functions by using a boolean parameter called `autofill` that will determine whether to forcefully reset the bindings or simply fill in what is missing.


Interkarma:
Thank you, I'll merge this one for 0.10.28.

PR #1975, opened 2020-11-29
Fix guild greetings for unique NPCs

### LINKED ISSUE #1973: Very weird dialog from Karethys
**Describe the bug**
When I talk to Karethys, her greeting is a generic "faction member" greeting for the faction "Karethys", eg: "Fellow members of Karethys have to stick together". It doesn't seem like a faction greeting should be used here. UESP says "Karethys also exists as a political faction and is associated with the Mages Guild." and I am a mages guild member, so I guess that is the cause. I don't know if this would count as a bug in the original game data, or the engine. (Also, what is the policy on fixing bugs in the original data? I'm doing my first playthrough of daggerfall now that df unity is ready to go, and I'm recording any bugs I encounter along the way, so it would be good to know not to report these if the policy is to leave them alone.)

**To Reproduce**
1. Load the attached save
2. Talk to Karethys
3. ???

**Expected behavior**
Some dialog that approximates sense.

**Screenshots**

 - OS: Linux (Ubuntu 20.10)
 - Version: 0.10.27

**Additional context**
Save file: [dfu_dialog_bug.zip](https://github.com/Interkarma/daggerfall-unity/files/5611280/dfu_dialog_bug.zip)

df-ferital:
Fix greetings when talking to unique NPCs affiliated to guilds, like Karethys or Popudax.

Fix #1973

### DIFF
--- a/Assets/Scripts/Game/TalkManager.cs
+++ b/Assets/Scripts/Game/TalkManager.cs
@@ -180,6 +180,7 @@ namespace DaggerfallWorkshop.Game
             public FactionFile.SocialGroups socialGroup;
             public FactionFile.GuildGroups guildGroup;
             public FactionFile.FactionData factionData;
+            public string npcFactionName; // kept for guild related greetings
             public string pcFactionName; // kept for guild related greetings
             public string allyFactionName; // kept for guild related greetings
             public string enemyFactionName; // kept for guild related greetings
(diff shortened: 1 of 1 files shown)


Interkarma:
This is excellent and fast work, thank you Ferital! :)

PR #1976, opened 2020-12-04
Prevent backstabbing daedra seducers during transformations

petchema:
Note: Changes the signatures of CalculateAttackDamage and CalculateBackstabChance delegates.
It seemed weird to pass around a frame record number that's tied to current enemies animation.


Interkarma:
Thank you Pango. I wonder if we should be using anim state at all here, rather than checking actual facing of enemy controller. I can see a day where custom mobile enemies won't be compatible with this setup.

petchema:
Agreed. I removed the condition from damage computation, but I'm not sure where it should be put exactly.

Interkarma:
An interesting idea. I'll merge this one for now and we can take a deeper look later. Cheers!

PR #1977, opened 2020-12-05
only apply reputation drop once in case of bank default

petchema:
Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&t=4346

Save less than 24h before default (but with enough on account to repay the loan). Neutral current reputation.
[SAVE867.zip](https://github.com/Interkarma/daggerfall-unity/files/5646217/SAVE867.zip)


Interkarma:
A good fix, thank you. :)

PR #1978, opened 2020-12-13
Fix spurious "can't carry any more" when picking maps

petchema:
If you encumbrance is maxed out, picking maps will work but be followed by a "can't carry any more" message that was stacked below by the call to CanCarryAmount().

This doesn't happen with arrows because their effective weight is 0.
Extend this to maps, which is the simplest fix.


petchema:
[SAVE839.zip](https://github.com/Interkarma/daggerfall-unity/files/5684483/SAVE839.zip)

PR #1982, opened 2020-12-22
Default binding fixes

jefetienne:
There were a couple of odd overlapping bugs I found. Here is one of them that I could reproduce:

1. Reset bindings to default
2. Remove `TurnLeft` and `TurnRight` in the primary bindings
3. Set `TurnLeft` and `TurnRight` as  "Left Arrow" and "Right Arrow", respectively, in the secondary bindings
4. Close windows, observe that TurnLeft has stopped working

Changes:
- Prevent binding `KeyCode` "None" to an action dictionary, so it won't get serialized in KeyBinds.txt
- Fix repopulating `existingKeyDict` when `actionKeyDict` did not contain every action by looping through the `Actions` enum list instead
- In `TestSetBinding`, do not set a default binding if the default `KeyCode` for it is already being used in the secondary bindings

### DIFF
--- a/Assets/Scripts/Game/InputManager.cs
+++ b/Assets/Scripts/Game/InputManager.cs
@@ -744,14 +744,17 @@ namespace DaggerfallWorkshop.Game
 
             ClearBinding(action, primary);
 
-            if (!dict.ContainsKey(code))
+            if (code != KeyCode.None)
             {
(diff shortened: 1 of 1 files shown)


Interkarma:
Thanks for additional fixes @jefetienne! :)

PR #1984, opened 2020-12-26
Fix and improve a few xml docs

TheLacus:
Fixed some issues with xml syntax and improved docs.


Interkarma:
Thank you for the update. This kind of thing is really important.

PR #1987, opened 2021-01-04
Return to using baseSpeed if riding

ajrb:
This makes no change to vanilla DFU, but allows mods to allow sprint riding. This will fix the Roleplay & Realism mod exhanced riding sprint function to be faster rather than slower as it currently is.

I can take a different approach if this change is a problem.


Interkarma:
This looks good to me. Thanks @ajrb!

PR #1991, opened 2021-01-12
Made DoorModelIndex public

### LINKED ISSUE #1989: Dumped Map Block json causes all doors to use model 9000
**Describe the bug**
I've noticed that when you use a Map Block Record override, all of the doors placed by that block will use model 9000 regardless of what the door models being used was before.

**To Reproduce**
1. Load the attached save. Note that all the ground floor doors use models 9003 or 9004.
2. Open the console and type "DumpBlock MAGEBA00.RMB". "DumpBuilding" while inside will also work. Quit the game.
3. Drop the json file created above into the WorldData folder.
4. Start the game and load the save again.
5. You will now see that all of the ground floor doors are using model 9000.

**Screenshots**
Here are two screenshots showing two of the doors without the dumped MBR loaded:
Good door 1
Good door 2

Here are those same doors after dropping that json into the WorldData folder:
Bad door 1
Bad door 2

**Desktop:**
 - OS: Windows 10 Pro x64
 - Version: 0.11.0

[Save.zip](https://github.com/Interkarma/daggerfall-unity/files/5799254/Save.zip)

XJDHDR:
This solves an issue I discovered and reported which causes all doors inside the overridden block to use model 9000.

Fixes #1989

### DIFF
--- a/Assets/Scripts/API/DFBlock.cs
+++ b/Assets/Scripts/API/DFBlock.cs
@@ -566,7 +566,7 @@ namespace DaggerfallConnect
             public Int16 OpenRotation;
 
             /// <summary>Model index to use for door as offset from base door ID.</summary>
-            internal Byte DoorModelIndex;
+            public Byte DoorModelIndex;
 
             /// <summary>Unknown.</summary>
             internal Byte Unknown;


ajrb:
Looks good to me. Seems to fix dumpbuilding as well (as I'd expect) contrary to your assertion in the bug. I never looked at doors, so thanks for catching this one.

Interkarma:
@ajrb I can't see any issue with this one. Happy for you to merge whenever, otherwise I'll get to another round of PR merging soon.

PR #1993, opened 2021-01-15
Fix some naming macros

df-ferital:
Fix an issue reported on the forums regarding the %fn macro: http://forums.dfworkshop.net/viewtopic.php?f=31&t=4447&sid=26748510f65c64d9c1d3237b74cb59cb

This issue can be seen on character biography when creating a Warrior. The same bug also affects the %mn macro when creating a Knight.

There is also one affected noble quest using both macros, R0C11Y28, where names on the autopsy report are always changing without this fix.

I also took the time to refactor the `MacroHelper.Name` method as its current state was not satisfying.


Interkarma:
Thank you for the fixes @df-ferital! :)

PR #1997, opened 2021-01-17
Refactor nature layout to it's own class and interface

ajrb:
This allows new nature layout algorithms to be used by mods in the same way terrain height samples and textures can be replaced.

Example usage:
` DaggerfallUnity.Instance.TerrainNature = new BasicRoadsNature();
`


Interkarma:
Looks good to me! Nice change btw, looking forward to seeing where this takes some of the wilderness mods starting to emerge.

PR #1999, opened 2021-01-19
Refresh cursor on game start

TheLacus:
Custom assets are tipically only seeked in the Game scene. Cursor is a special case because is also needed in the Startup scene, so there was an issue where a mod providing the cursor, initially enabled, was disabled resulting in a missing cursor, as well as an issue where a custom cursor was provided by a disabled mod.

Cursor is now also seeked on game launch to ensure changes to mod configuration during setup (i.e mod enabled or disabled) are applied.


TheLacus:
[Issue on forum](https://forums.dfworkshop.net/viewtopic.php?f=31&t=4436).

Interkarma:
Thanks for fix!

PR #2002, opened 2021-01-25
Call UpdateMeshRendererDungeonState() on all children of Transform Models

### LINKED ISSUE #1995: Replacement OBJ models invisible on dungeon map after loading save
**Describe the bug**
While modding DFU, I noticed that every model I created was invisible on the explored dungeon map after loading a save and only became visible again after I walked to those invisible models and re-explored the area. However, these re-explored models would disappear again after loading any saves after that as well. 

Recently, I discovered that this only happens if the replacement models are OBJs. As a test, I replaced a staircase model that I was using an OBJ model for with an FBX model and loaded it into my mod. After loading the save, the staircases using that model were now visible on the dungeon's map. 

So there is something about the way DFU is deciding what dungeon pieces to reveal on the map while loading a save that breaks if the piece uses an OBJ model. I'm wondering if this is a similar issue to one Petchema fixed in his pull request #1988. That this part of DFU was only designed to look at the root of the prefab and not at it's children? I've attached a save and the two staircase models to help reproduce this:
[Save.zip](https://github.com/Interkarma/daggerfall-unity/files/5824089/Save.zip)
[Models.zip](https://github.com/Interkarma/daggerfall-unity/files/5824090/Models.zip)

**To Reproduce**
1. Load the OBJ model into your game however you prefer. Make sure you rename the OBJ prefab to "61018.prefab".
2. Load the save game. Notice how the staircases in the corridor in my screenshot is not visible on the dungeon map even though that corridor was explored.
3. Load the FBX model into your game however you prefer.
4. Load the save game. Notice that the staircase pieces are now visible on the map. Please note that if you notice that the model doesn't become transparent, Petchema has already submitted a pull request that fixes this so don't worry.

**Screenshots**
Here are some screenshots of what I am talking about. Please don't worry about these models being completely white. It is a separate issue that I've already reported elsewhere.
This first one is where I'm only using OBJ replacement models. The corridor across the middle of the picture is supposed to be fully explored but pieces of it are missing.
OBJ hidden corridor pieces without FBX test

This second one shows what happened after I replaced the staircase model in that corridor with an FBX model. You can see that now the staircase pieces of that corridor are visible after loading my save.
OBJ hidden corridor pieces with FBX test

 - OS: Windows 10 Pro x64
 - Version: 0.11.0

XJDHDR:
This change makes the UpdateMeshRendererDungeonState() function be called on all MeshRenderer components of all children of the current transform model instead of only the first one in it's root. 

This fixes an issue where already discovered models on the dungeon map are not revealed after loading a save game if an OBJ file has been used to replace the model. This change was adapted from pull request #1988 . Without my changes, the dungeon map looks like the first screenshot attached to the issue. With my changes, the same corridor after loading the attached save looks like this, as it should. Don't worry about the wrong textures on my models, that's a separate issue I've reported elsewhere:
Undiscovered models fixed

If this PR is merged, it will fix #1995 

I'm leaving this as a draft because I would like some comments from the devs. 
On the one hand, Nystul said that this is the wrong way to fix this issue because he believes that if the model hierarchy is different to when the game was saved, it will break stuff.
On the other hand, other changes have been made to DFU in response to other OBJ related issues I've reported. One example is [this pull request](https://github.com/Interkarma/daggerfall-unity/pull/1988) by Pango which fixes it's issue in a similar way to this PR. 
That said, in [this forum post](https://forums.dfworkshop.net/viewtopic.php?p=47664#p47664) - even though TheLacus did change his script to work with OBJ's model hierarchy - he specifically mentioned avoiding GetComponentInChildren() for performance reasons.

I also had to create a new variable to make this work (Unity complained that I can't use a ForEach iteration variable as a ref). Does anyone feel I could have given it a better name? 

Otherwise, it is ready to go.

### DIFF
--- a/Assets/Scripts/Game/Automap.cs
+++ b/Assets/Scripts/Game/Automap.cs
@@ -2264,16 +2264,20 @@ namespace DaggerfallWorkshop.Game
                     foreach (Transform currentTransformMesh in currentTransformModel.transform)
                     {
                         AutomapGeometryModelState model = new AutomapGeometryModelState();
-                        MeshRenderer meshRenderer = currentTransformMesh.GetComponent<MeshRenderer>();
-                        if ((meshRenderer) && (meshRenderer.enabled))
+                        MeshRenderer[] meshRenderers = currentTransformMesh.GetComponentsInChildren<MeshRenderer>();
+                        for (int i = 0; i < meshRenderers.Length; i++)
                         {
-                            model.discovered = true;
-                            model.visitedInThisRun = !meshRenderer.materials[0].IsKeywordEnabled("RENDER_IN_GRAYSCALE"); // all materials of model have the same setting - so just material[0] is tested
-                        }
-                        else
-                        {
-                            model.discovered = false;
-                            model.visitedInThisRun = false;
+                            if ((meshRenderers[i]) && (meshRenderers[i].enabled))
+                            {
+                                model.discovered = true;
+                                model.visitedInThisRun = !meshRenderers[i].materials[0].IsKeywordEnabled("RENDER_IN_GRAYSCALE"); // all materials of model have the same setting - so just material[0] is tested
+                                break;
+                            }
+                            else
+                            {
+                                model.discovered = false;
+                                model.visitedInThisRun = false;
+                            }
                         }
                         models.Add(model);
                     }
--- a/Assets/Scripts/Game/Automap.cs
+++ b/Assets/Scripts/Game/Automap.cs
@@ -2397,7 +2397,7 @@ namespace DaggerfallWorkshop.Game
             DestroyTeleporterMarkers();
         }
 
-        void UpdateMeshRendererDungeonState(ref MeshRenderer meshRenderer, AutomapDungeonState automapDungeonState, int indexBlock, int indexElement, int indexModel, bool forceNotVisitedInThisRun)
+        void UpdateMeshRendererDungeonState(MeshRenderer meshRenderer, AutomapDungeonState automapDungeonState, int indexBlock, int indexElement, int indexModel, bool forceNotVisitedInThisRun)
         {
             if (automapDungeonState.blocks[indexBlock].blockElements[indexElement].models[indexModel].discovered == true)
             {
@@ -2481,10 +2481,13 @@ namespace DaggerfallWorkshop.Game
                     {
                         Transform currentTransformModel = currentTransformElement.GetChild(indexModel);
 
-                        MeshRenderer meshRenderer = currentTransformModel.GetComponent<MeshRenderer>();
-                        if (meshRenderer)
+                        MeshRenderer[] meshRenderers = currentTransformModel.GetComponentsInChildren<MeshRenderer>();
+                        for (int i = 0; i < meshRenderers.Length; i++)
                         {
-                            UpdateMeshRendererDungeonState(ref meshRenderer, automapDungeonState, indexBlock, indexElement, indexModel, forceNotVisitedInThisRun);
+                            if (meshRenderers[i])
+                            {
+                                UpdateMeshRendererDungeonState(meshRenderers[i], automapDungeonState, indexBlock, indexElement, indexModel, forceNotVisitedInThisRun);
+                            }
                         }
                     }
                 }


Nystul-the-Magician:
Thanks for the pr!
We just need to test it and see if it works ;) lets just use this pragmatic approach ;)
The things we should test are:
- no undiscovered model parts are shown with this fix
- if vanilla map segments can have several sub-gameobjects - so that some of these could have be discovered while others are not (I just don't know but would like to check) - so making all subobjects visible automatically for a parent go could make undiscovered stuff appear

Nystul-the-Magician:
tested it and did not find any problems so far - so my fears were unjustified
good work!

Assets/Scripts/Game/Automap.cs
```diff
@@ -2481,11 +2481,15 @@ private void RestoreStateAutomapDungeon(bool forceNotVisitedInThisRun = false)
                     {
                         Transform currentTransformModel = currentTransformElement.GetChild(indexModel);
 
-                        MeshRenderer meshRenderer = currentTransformModel.GetComponent<MeshRenderer>();
-                        if (meshRenderer)
-                        {
-                            UpdateMeshRendererDungeonState(ref meshRenderer, automapDungeonState, indexBlock, indexElement, indexModel, forceNotVisitedInThisRun);
-                        }
+						MeshRenderer[] meshRenderers = currentTransformModel.GetComponentsInChildren<MeshRenderer>();
```
TheLacus, line 2484: I'd suggest to preserve original indenting.

Assets/Scripts/Game/Automap.cs
```diff
@@ -2481,11 +2481,15 @@ private void RestoreStateAutomapDungeon(bool forceNotVisitedInThisRun = false)
...
+						foreach (MeshRenderer meshRendererIteration in meshRenderers)
+						{
+							if (meshRendererIteration)
+							{
+								MeshRenderer meshRenderer = meshRendererIteration;
+								UpdateMeshRendererDungeonState(ref meshRenderer, automapDungeonState, indexBlock, indexElement, indexModel, forceNotVisitedInThisRun);
```
TheLacus, line 2490: Passing by reference here is unneeded, you can change method signature if @Nystul-the-Magician agrees.

Nystul-the-Magician: no objections from my side ;)

XJDHDR: Done.

XJDHDR:
I've discovered that my PR is not complete as there is another problem that needs to be solved before this will be fixed. The issue is that even though the OBJ models are now visible with my change, there is some part of the code that thinks that all of my replacement models have not been discovered (I suspect the part that saves their discovery state). As a result, when you save the game, the models will be invisible on the map once again after you load and this time, they will only be revealed if you re-explore them.

You can test this by loading my test save attached to the issue. Then, without moving, save the game with a new name then load that new save. You will see that the models I replaced are invisible. Then compare the contents of the AutomapData.txt for the two saves. You will see that the later save now has a number of entries that have flipped from True to False.

XJDHDR:
Never mind, I just figured out how to solve this problem.

Nystul-the-Magician:
tested with your recent changes and it still works without breaking stuff, great work!

XJDHDR:
Thanks Nystul. However, I was thinking about that change and thought of a potential issue. Let's say hypothetically, we have a dungeon model that is discovered but a MeshRenderer in it's prefab was added in a disabled state after the last time UpdateMeshRendererDungeonState was called and this disabled MeshRenderer was last in the list. My change as it stands means that the entire model will be marked as not discovered even if all the other MeshRenderer components are enabled. Is this a realistic concern or not?

If so, I propose changing my additions on lines 2267 - 2280 to this:
```c#
                    List<AutomapGeometryModelState> models = new List<AutomapGeometryModelState>();
                    foreach (Transform currentTransformMesh in currentTransformModel.transform)
                    {
                        AutomapGeometryModelState model = new AutomapGeometryModelState();
                        MeshRenderer[] meshRenderers = currentTransformMesh.GetComponentsInChildren<MeshRenderer>();
                        model.discovered = false;
                        model.visitedInThisRun = false;
                        foreach (MeshRenderer meshRenderer in meshRenderers)
                        {
                            if ((meshRenderer) && (meshRenderer.enabled))
                            {
                                model.discovered = true;
                                model.visitedInThisRun = !meshRenderer.materials[0].IsKeywordEnabled("RENDER_IN_GRAYSCALE"); // all materials of model have the same setting - so just material[0] is tested
                                break
                            }
                        }
                        models.Add(model);
                    }
```

Nystul-the-Magician:
not sure what you mean with "a meshrenderer in its prefab was added in a disabled state"
isn't a prefab just the blueprint for a hierarchy of a gameobject with all scripts and settings.
so how would a prefab change at runtime?
or do you mean the gameobject holding the model might have several sub-gameobjects?

XJDHDR:
Essentially, any children attached to ModelA will be marked as discovered when ModelA is discovered. I can't think of any reason why you would only want to selectively discover only some of the multiple MeshRenderers attached to a model's prefab. Besides, the original code meant that there was never going to be a situation where those child MeshRenderers were going to be marked. The code's original granularity seems to be at the level of models placed by the Block and this is still the case with my changes.

In the situation you described, the tables and chairs are separate models from the room itself; placed by the block. Thus, the discovery state of the room won't affect any models in the room. 

With that said, my concern is the prefab changing at runtime. I'm not concerned about a prefab that has multiple MeshRenderers at the time of replacement because my changes on lines 2487 - 2493 iterate through all of the MeshRenderers attached to the prefab and set all of them to the state defined for that model.

I was thinking if maybe a mod is designed in such a way that it adds MeshRenderers to the model while the game is running. One idea could be a switch that when activated, makes a section of wall disappear and it accomplishes this by disabling the MeshRenderer for that section. Hence my question of whether this is a valid concern.
But either way, I don't think my foreach loop needs to check all of the MeshRenderers to see if the model has been discovered. I think only the first one that passes is necessary.

TheLacus:
> With that said, my concern is the prefab changing at runtime. I'm not concerned about a prefab that has multiple MeshRenderers at the time of replacement because my changes on lines 2487 - 2493 iterate through all of the MeshRenderers attached to the prefab and set all of them to the state defined for that model.
> 
> I was thinking if maybe a mod is designed in such a way that it adds MeshRenderers to the model while the game is running. One idea could be a switch that when activated, makes a section of wall disappear and it accomplishes this by disabling the MeshRenderer for that section. Hence my question of whether this is a valid concern.
> But either way, I don't think my foreach loop needs to check all of the MeshRenderers to see if the model has been discovered. I think only the first one that passes is necessary.

An instantiated prefab is a gameobject. If you disable the meshrenderer of an active gameobject, it doesn't affect meshrenderers of its children. If you disable a gameobject (`GameObject.SetActive()`), it makes each component to appear inactive, including meshrenderer, as well as all of its children.

`UpdateMeshRendererDungeonState()` doesn't only toggle meshrenderer, it also set material keyword (`"RENDER_IN_GRAYSCALE"`) so it must loop on all chidren for which this is needed.

XJDHDR:
> If you disable a gameobject (GameObject.SetActive()), it makes each component to appear inactive, including meshrenderer, as well as all of its children.

Indeed but my concern is someone disabling a MeshRenderer while leaving it's GameObject active. 

> In fact, i suppose that a mod that want to implement your example should make sure that changes in scenes are also syncronized in the automap.

It's the mod that does this which I'm concerned about.

As my script is currently set up, the portion that saves the AutoMap models' discovered and VisitedInThisRun states iterates through every MeshRenderer attached to the model, setting the value of a model's discovered and visitedInThisRun variables each time. The value of those two variables will be equal to what they were for the last MeshRenderer the loop encountered. This means that if the last MeshRenderer just happens to be disabled, that one will make the entire model be considered undiscovered even if all the others are enabled.

Another thing is that my proposed change means the loop stops running the first time it encounters a MeshRenderer that has been discovered, meaning that the loop has a chance of only iterating through some of them rather than always iterating through all of them as it does now. My thought is that if at least one of the AutoMap model's MeshRenderers are enabled, it means that it has been explored and I can't think of any reason why the rest need to be tested as well.

Nystul-the-Magician:
> Another thing is that my proposed change means the loop stops running the first time it encounters a MeshRenderer that has been discovered, meaning that the loop has a chance of only iterating through some of them rather than always iterating through all of them as it does now. My thought is that if at least one of the AutoMap model's MeshRenderers are enabled, it means that it has been explored and I can't think of any reason why the rest need to be tested as well.

what do you mean with "at least one of the Automap model's MeshRenderes" - I don't quite understand
 - do you mean your replacement model does have several meshrenderers on the same gameobject (this should not be)
 - or do you mean your replacement model does have one meshrenderer on every sub-gameobject?

XJDHDR:
> or do you mean your replacement model does have one meshrenderer on the root gameobject and one on every child-gameobject (on possibly several child-gameobjects)?

Yes, this is what I meant.

> in the hierarchy every mesh (dungeon segment) is listed there. e.g. mesh 67017 does exactly have one mesh renderer

I'm aware of this. I use this info to figure out which models I need to modify for my mods.

> in case of a replacement object there might be child-gameobjects inside the main gameobject (root gameobject) if I understand right, but every (sub-)gameobject should only have one meshrenderer at most

That's correct.

> if it should be possible to save different discovery state for different parts of the replacement model independently or not (it would make sense since discovery will always only discover the model parts whos meshcollider was hit)

I agree that it's probably a good idea to do this. That said ...

> I would rather like to have their discovery state independent and also saved this way - so stopping the iteration early would cause false results

I am submitting this PR for what the discovery system currently does. At the moment, it only saves and loads this info on the model level and making it work on the MeshRenderer level would require some changes to a number of places, including the parts I've changed. This is the case regardless of whether the code stays the way it is or changes to either of the two proposed changes. If we go with my first version, someone would still need to change it to make it save each MeshRenderer's state rather than the Model's. Same thing with my second version and they can easily remove the `break;` statement while doing so. 

My concern is that until if and when this MeshRenderer granularity gets implemented, the game will be in a state where either it doesn't work with OBJ models or has a risk of undiscovering discovered models, depending on whether my PR as it stands is implemented or not. That said, I would like to change my proposal to this instead. It should be a bit easier than my first to convert into a state that saves at the MeshRenderer level.

```c#
                        MeshRenderer[] meshRenderers = currentTransformMesh.GetComponentsInChildren<MeshRenderer>();
                        foreach (MeshRenderer meshRenderer in meshRenderers)
                        {
                            if ((meshRenderer) && (meshRenderer.enabled))
                            {
                                model.discovered = true;
                                model.visitedInThisRun = !meshRenderer.materials[0].IsKeywordEnabled("RENDER_IN_GRAYSCALE"); // all materials of model have the same setting - so just material[0] is tested
                                break;
                            }
                            else
                            {
                                model.discovered = false;
                                model.visitedInThisRun = false;
                            }
                        }
                        models.Add(model);
```

Nystul-the-Magician:
I am fine with your changes ;)
I think for most of cases this should be enough (subobjects being discovered with the parent object) - the special case I mentioned with different subobjects was more a artificially constructed one ;)

Interkarma:
Hey @XJDHDR! I'll loop back to your changes for 0.11.2. This should be ready in March sometime. I want to get 0.11.1 and localization tutorials completed first.

Thank you @TheLacus and @Nystul-the-Magician for your help with review. This means a great deal to me and takes away some pressures on my side.

XJDHDR:
When I opened this PR, I didn't realise that I was going to make more contributions or that this one would take so long to get merged. As a result, I added my changes to the master branch of my fork. Unfortunately, this oversight is hampering my ability to make further contributions. So I'm going to close this PR and move my changes into a separate branch. Once I've done that, I'll open a new one.

XJDHDR:
I've just opened a new PR: #2024. Any further discussion can happen there.

PR #2003, opened 2021-01-26
Some fixes for Action Doors with child objects

XJDHDR:
This change fixes a bug I found where the Action Doors in a Block don't have their Starting Lock Level (and presumably other variables) set to the block's defined values if the Action Door script is attached to a child of the model's prefab. An example of this is if the door has been replaced by an OBJ model.

I reported this bug on the forum [here](https://forums.dfworkshop.net/viewtopic.php?p=51629#p51629).

### DIFF
--- a/Assets/Scripts/Game/Serialization/SerializableActionDoor.cs
+++ b/Assets/Scripts/Game/Serialization/SerializableActionDoor.cs
@@ -24,7 +24,7 @@ namespace DaggerfallWorkshop.Game.Serialization
     {
         #region Fields
 
-        DaggerfallActionDoor actionDoor;
+        DaggerfallActionDoor[] allActionDoors;
 
         #endregion
 
@@ -32,8 +32,8 @@ namespace DaggerfallWorkshop.Game.Serialization
 
         void Awake()
         {
-            actionDoor = GetComponent<DaggerfallActionDoor>();
-            if (!actionDoor)
+            allActionDoors = GetComponentsInChildren<DaggerfallActionDoor>();
+            if (!allActionDoors[0])
                 throw new Exception("DaggerfallActionDoor not found.");
         }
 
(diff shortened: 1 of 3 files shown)


XJDHDR:
I just pushed another commit that fixes some errors I encountered where the game couldn't find the DaggerfallActionDoor script attached to my door model because the script was on the child object created on all OBJs. 
Action Door OBJ error

If needed, you can test this commit with the following files: 
[Files.zip](https://github.com/Interkarma/daggerfall-unity/files/5935670/Files.zip)

Interkarma:
Replacing the model should not break composition. SerializableActionDoor and DaggerfallActionDoor are intended to be peered, not split apart into child objects. The operating parts of the component must remain the same, only the visual parts should be different. I'd like to look into that first before working around it by hunting for misplaced components that shouldn't be on children to start with.

@TheLacus Could I please ask for your expertise on this one?

XJDHDR:
The thing is that I haven't split those two scripts apart; they are both attached to the same child:
Door 9000

For some reason, `SerializableActionDoor` couldn't find `DaggerfallActionDoor`, even though they were both attached to the same place, until I made my change. 

I should test what happens if I attach these scripts to the root of the prefab and see if things still work properly. If they do, it just becomes a question of whether the scripts being at the root is the only supported configuration.

That reminds me though. I need to fix RDBLayout to bring it in line with my changes to the other two scripts. Namely, make it only process the first instance of DaggerfallActionDoor it finds rather than iterate through all of them.

Interkarma:
GetComponent<T>() always looks on its own gameobject. Are there really multiple DaggerfallActionDoors in the hierarchy here - i.e. one in the root and one in the child object?

I've downloaded your example prefab files on forum post linked above. Will dig into this and try to understand what's happening.

XJDHDR:
> Are there really multiple DaggerfallActionDoors in the hierarchy here - i.e. one in the root and one in the child object?
No, there is only one of each script attached to the entire model.

TheLacus:
Components that are required by core game for door functionalities must be placed on expected position, which is the root of gameobject. I altered your prefab and moved all components (including collider) and everything works fine on a first look.

You can organize visual parts of model as you prefer, but i don't see why you would want `DaggerfallActionDoor` and others on children.

XJDHDR:
Fair enough. I only did it this way because I didn't think it would make a difference and this is similar to other fixes that have been done for OBJ models.

Interkarma:
I see it now, thank you TheLacus. Yeah, replacement prefabs should follow same initial composition and structure as the one in core. You can child stuff to it and add new components, but the object should otherwise allow DFU to find the components its looking for in the same positions.

Thanks for your help everyone. :)

PR #2005, opened 2021-01-27
Ensure that the last location is set to current loc when entering a building.

ajrb:
Turns out that live builds can sometimes update locations in a different order leaving the last location set to someplace other than current which breaks building variants for first entry.

Also added some docs for the terrain nature interface.


ajrb:
I'd like to merge this one soon, please say if you have any objection

Interkarma:
No objections, here. I'll merge this one for 0.11.1 now.

PR #2006, opened 2021-01-28
Extra features for the CustomDoor script

XJDHDR:
My changes here add two additional functions to the CustomDoor script:
- You can specify what DoorType the doors created by this script will be in the Editor's Inspector, giving a modder the ability to override the DoorType set by analysing the textures.
- ~~You can add multiple BoxColliders to the model and allow more than one of them to be treated as a door by this script.~~ Edit: Original code already did this.

I've marked this as a draft to get the opinions of the DFU devs and see if this is something you are willing to add to the base game. Especially @TheLacus since this was originally his work.

### DIFF
--- a/Assets/Scripts/Game/PlayerActivate.cs
+++ b/Assets/Scripts/Game/PlayerActivate.cs
@@ -470,7 +470,7 @@ namespace DaggerfallWorkshop.Game
             Transform doorOwner)
         {
             StaticDoor door;
-            if (CustomDoor.HasHit(hit, out door) || (doors && doors.HasHit(hit.point, out door))
+            if (CustomDoor.HasHit(hit, out door) || (doors && doors.HasHit(hit.point, out door)))
             {
                 // Check if close enough to activate
                 if (hit.distance > DoorActivationDistance)
--- a/Assets/Scripts/Game/PlayerActivate.cs
+++ b/Assets/Scripts/Game/PlayerActivate.cs
@@ -469,9 +469,8 @@ namespace DaggerfallWorkshop.Game
             int buildingLockValue,
             Transform doorOwner)
         {
-            bool noCustomDoorFound;
             StaticDoor door;
-            if (CustomDoor.HasHit(hit, out door, out noCustomDoorFound) || (doors && noCustomDoorFound && doors.HasHit(hit.point, out door)))
+            if (CustomDoor.HasHit(hit, out door) || doors.HasHit(hit.point, out door))
             {
                 // Check if close enough to activate
                 if (hit.distance > DoorActivationDistance)
@@ -1004,8 +1003,7 @@ namespace DaggerfallWorkshop.Game
             Transform doorOwner;
             DaggerfallStaticDoors doors = GetDoors(hit.transform, out doorOwner);
             StaticDoor door;
-            bool noCustomDoorFound;
-            if (CustomDoor.HasHit(hit, out door, out noCustomDoorFound) || (doors && noCustomDoorFound && doors.HasHit(hit.point, out door)))
+            if (CustomDoor.HasHit(hit, out door) || (doors && doors.HasHit(hit.point, out door)))
             {
                 // Discover building - this is needed to check lock level and transition to interior
                 GameManager.Instance.PlayerGPS.DiscoverBuilding(door.buildingKey);
--- a/Assets/Scripts/Game/PlayerActivate.cs
+++ b/Assets/Scripts/Game/PlayerActivate.cs
@@ -470,7 +470,7 @@ namespace DaggerfallWorkshop.Game
             Transform doorOwner)
         {
             StaticDoor door;
-            if (CustomDoor.HasHit(hit, out door) || doors.HasHit(hit.point, out door))
+            if (CustomDoor.HasHit(hit, out door) || (doors && doors.HasHit(hit.point, out door))
             {
                 // Check if close enough to activate
                 if (hit.distance > DoorActivationDistance)
--- a/Assets/Scripts/Game/PlayerActivate.cs
+++ b/Assets/Scripts/Game/PlayerActivate.cs
@@ -470,7 +470,7 @@ namespace DaggerfallWorkshop.Game
             Transform doorOwner)
         {
             StaticDoor door;
-            if (doors.HasHit(hit.point, out door) || CustomDoor.HasHit(hit, out door))
+            if (CustomDoor.HasHit(hit, out door) || doors.HasHit(hit.point, out door))
             {
                 // Check if close enough to activate
                 if (hit.distance > DoorActivationDistance)
--- a/Assets/Scripts/Game/PlayerActivate.cs
+++ b/Assets/Scripts/Game/PlayerActivate.cs
@@ -469,8 +469,9 @@ namespace DaggerfallWorkshop.Game
             int buildingLockValue,
             Transform doorOwner)
         {
+            bool noCustomDoorFound;
             StaticDoor door;
-            if (CustomDoor.HasHit(hit, out door) || doors.HasHit(hit.point, out door))
+            if (CustomDoor.HasHit(hit, out door, out noCustomDoorFound) || (doors && noCustomDoorFound && doors.HasHit(hit.point, out door)))
             {
                 // Check if close enough to activate
                 if (hit.distance > DoorActivationDistance)
@@ -1003,7 +1004,8 @@ namespace DaggerfallWorkshop.Game
             Transform doorOwner;
             DaggerfallStaticDoors doors = GetDoors(hit.transform, out doorOwner);
             StaticDoor door;
-            if (doors && doors.HasHit(hit.point, out door) || CustomDoor.HasHit(hit, out door))
+            bool noCustomDoorFound;
+            if (CustomDoor.HasHit(hit, out door, out noCustomDoorFound) || (doors && noCustomDoorFound && doors.HasHit(hit.point, out door)))
             {
                 // Discover building - this is needed to check lock level and transition to interior
                 GameManager.Instance.PlayerGPS.DiscoverBuilding(door.buildingKey);
--- a/Assets/Scripts/Utility/AssetInjection/Components/CustomDoor.cs
+++ b/Assets/Scripts/Utility/AssetInjection/Components/CustomDoor.cs
@@ -25,55 +25,24 @@ namespace DaggerfallWorkshop.Utility.AssetInjection
     public class CustomDoor : MonoBehaviour
     {
         private StaticDoor? staticDoor;
+        
+        [Tooltip("If ticked, tells the game not to check for a static door hit after checking for custom doors. NOTE: This is enabled per model, not per script.")]
+        public bool NoStaticDoorCheck = false;
+        
+        [Tooltip("Choose which of the doors in this model's Static Door array will be copied for this custom door.")]
+        public int StaticDoorCopied = 0;
 
-        [Tooltip("Number of attached Box Colliders that will be used as doors.")]
-        public int NumberOfTriggers = 1;
-        
-        [Tooltip("Type of door that is being defined. None means building textures are used to determine this.")]
-        public DoorTypes DoorType = DoorTypes.None;
-        
         /// <summary>
-        /// The triggers for the doors.
+        /// The trigger for this door.
         /// </summary>
-        [HideInInspector]
-        public BoxCollider[] DoorTriggers;
+        [Tooltip("The trigger for this door. Drag and drop your Box Collider here.")]
+        public BoxCollider DoorTrigger;
 
         private void Awake()
         {
-            if (DoorTriggers.Length < 1)
-                DoorTriggers = GetComponents<BoxCollider>();
-
-            string objectName;
-            if (DoorTriggers.Length < 1)
-            {
-                if (name == "default")
-                    objectName = transform.parent.name;
-                else
-                    objectName = name;
-                Debug.LogErrorFormat("No Box Colliders are attached to {0}. Did you forget to add them?", objectName);
-                return;
-            }
-            if (NumberOfTriggers < 1)
-            {
-                if (name == "default")
-                    objectName = transform.parent.name;
-                else
-                    objectName = name;
-                Debug.LogErrorFormat("Number of Triggers for {0} was less than 1. Did you put in the wrong number?", objectName);
-                return;
-            }
-            if (NumberOfTriggers > DoorTriggers.Length)
-            {
-                if (name == "default")
-                    objectName = transform.parent.name;
-                else
-                    objectName = name;
-                Debug.LogErrorFormat("Number of Triggers for {0} was greater than the number of attached Box Colliders. Are you missing Box Colliders? Did you put in the wrong number?", objectName);
-                return;
-            }
-
-            for (int i = 0; i < NumberOfTriggers; i++)
-                DoorTriggers[i].isTrigger = true;
+            if (!DoorTrigger)
+                DoorTrigger = GetComponent<BoxCollider>();
+            DoorTrigger.isTrigger = true;
         }
 
         /// <summary>
@@ -82,18 +51,30 @@ namespace DaggerfallWorkshop.Utility.AssetInjection
         /// <param name="building">The imported gameobject which provides building and doors.</param>
         /// <param name="staticDoors">The list of static doors for the vanilla building.</param>
         /// <param name="buildingKey">The key of the building that owns the doors.</param>
-        public static void InitDoors(GameObject building, StaticDoor[] staticDoors, int buildingKey)
+        public static void InitDoors(GameObject building, StaticDoor[] staticDoors, int buildingKey, out bool dontCreateStaticDoors)
         {
-            // Make data for the static door; only the common data for the building is needed.
-            StaticDoor staticDoor = staticDoors[0];
-            staticDoor.buildingKey = buildingKey;
-
+            dontCreateStaticDoors = false;
             // Set data to all doors in the building
             CustomDoor[] allCustomDoors = building.GetComponentsInChildren<CustomDoor>();
             for (int i = 0; i < allCustomDoors.Length; i++)
             {
-                if (allCustomDoors[i].DoorType != DoorTypes.None)
-                    staticDoor.doorType = allCustomDoors[i].DoorType;
+                if (allCustomDoors[i].NoStaticDoorCheck == true)
+                    dontCreateStaticDoors = true;
+
+                if ((allCustomDoors[i].StaticDoorCopied < 0) || (allCustomDoors[i].StaticDoorCopied > staticDoors.Length))
+                {
+                    string objectName;
+                    if (allCustomDoors[i].name == "default")
+                        objectName = allCustomDoors[i].transform.parent.name + "\\default";
+                    else
+                        objectName = allCustomDoors[i].name;
+                    Debug.LogErrorFormat("StaticDoorCopied for model {0} is outside the valid range for it's Static Door array. Defaulting to 0", objectName);
+                    allCustomDoors[i].StaticDoorCopied = 0;
+                }
+                // Make data for the static door; only the common data for the building is needed.
+                StaticDoor staticDoor = staticDoors[allCustomDoors[i].StaticDoorCopied];
+                staticDoor.buildingKey = buildingKey;
+
                 allCustomDoors[i].staticDoor = staticDoor;
             }
         }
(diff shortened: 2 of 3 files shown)


TheLacus:
Hi, I'll review more in deep as soon as possible. On a first look, i'm not sure about a single `CustomDoor` component actually referring to multiple doors. You should already be able to have as many doors you want as long as you assign box colliders in the inspector.

Assets/Scripts/Utility/AssetInjection/Components/CustomDoor.cs
```diff
@@ -24,20 +24,29 @@ namespace DaggerfallWorkshop.Utility.AssetInjection
     [RequireComponent(typeof(BoxCollider))]
     public class CustomDoor : MonoBehaviour
     {
+        [Tooltip("Type of door that is being defined. None means building textures are used to determine this.")]
+        public DoorTypes doorType = DoorTypes.None;
+        
+        [Tooltip("Number of BoxColliders alongside script that will be used as door activators")]
+        public int numColliders = 1;
+
```
TheLacus, line 32: A minor syntax note: please use pascal case for public members and place them after `DoorTrigger`.

XJDHDR:
> You should already be able to have as many doors you want as long as you assign box colliders in the inspector.

Okay, I don't know what happened but I could have sworn that multiple colliders didn't work with the original code but now it does. Maybe it was something else. I've reverted that commit and so, it's just the first two now.

Assets/Scripts/Utility/AssetInjection/Components/CustomDoor.cs
XJDHDR: Done.

Assets/Scripts/Utility/AssetInjection/Components/CustomDoor.cs
```diff
@@ -26,10 +26,13 @@ public class CustomDoor : MonoBehaviour
     {
         private StaticDoor? staticDoor;
 
+        [Tooltip("Type of door that is being defined. None means building textures are used to determine this.")]
+        public DoorTypes DoorType = DoorTypes.None;
```
TheLacus, line 30: There are only three types of door: `Building`, `DungeonEntrance` and `DungeonExit`. Choosing anything other than the correct one results in a broken door, in one case even with an exception being thrown. Can you explain what's the intended use case here?

Assets/Scripts/Game/PlayerActivate.cs
```diff
@@ -470,7 +470,7 @@ void Update()
             Transform doorOwner)
         {
             StaticDoor door;
-            if (doors.HasHit(hit.point, out door) || CustomDoor.HasHit(hit, out door))
+            if (CustomDoor.HasHit(hit, out door) || doors.HasHit(hit.point, out door))
```
TheLacus, line 473: Is this for when you have a classic door and a custom door with a partial overlap? I'm not necessarily against it, but i'm not sure this should happen in the first place. If this actually solves an issue you found i'm happy to merge :)

TheLacus (changes requested):

XJDHDR:
@TheLacus 
I feel that I may not have properly explained my motivation for these changes and there may be a difference in our visions for what we want this script to do.

First, as some background, I was trying to fix model 729 for my mod which Pango reported a problem with [here](https://forums.dfworkshop.net/viewtopic.php?f=30&t=2059). The left side of the castle model's main entrance was using a texture that was invalid for dungeon entrances; it's for settlement buildings only. As a result, this model had one door defined as the `Building` type and three defined as the `DungeonEntrance` type.

I created a model that replaces the bad texture with the correct one. This didn't fix the problem; the game still acted like the left side of that door was a settlement building's entrance. After a lot of investigation I realised, partly thanks to BadLuckBurt, that the part of DFU's code that detects where a building model has doors is hardcoded to get this info from the models inside ARCH3D.BSA and it probably isn't possible to change this (would require knowing how to decode the mesh format for FBX, OBJ and any other file format a model can use). This mean that the DaggerfallStaticDoors script will check for a door hit against the vanilla model data even when the model has been replaced and it's door situation doesn't necessarily reflect the vanilla model.

After that, I looked at your CustomDoor script to see if it could help. There were just two issues I noticed when I tried it. 

1. It originally set the DoorType for all of it's door triggers to that of the first door in the object's staticDoors array. This info is ultimately created by GetModelData() in MeshReader, which gets this info from the vanilla models. For model 729, the first door in that array just happens to be the door created from texture archive 474, the texture causing the bug I'm trying to fix. This meant that every door trigger created by CustomDoors acted as if it was a door defined as the `Building` DoorType. 
2. Due to the order that PlayerActivate did it's checks, it meant that when I placed a BoxCollider over the bugged door, CustomDoors never had a chance to tell the game how to handle that activation because DaggerfallStaticDoors was checked first and it told the game to do the bugged door activation. 

Since those first two commits, I've noticed three other things that I've changed as well.

- The part that checks for door bashing doesn't take CustomDoors into account at all so I added a check to that section as well.
- It seems like my initial attempt to add support for multiple colliders wasn't a mistaken hunch. Without my changes, the script only checks for a hit in the first Box Collider on the object. I did some debugging and found that the HasHit function returns false if the hit was on any other collider. When I initially thought that the original code was working for multiple BoxColliders, what was actually happening was that HasHit() for CustomDoors returned false and allowed the DaggerfallStaticDoors script to have a go and it's function did the work that I thought CustomDoors was doing. So in fact, CustomDoors needs my changes to test multiple Box Colliders. That is unless supporting only one door trigger is intended. In which case, model 729 has 3 separate doors attached and I would like to see CustomDoors support this scenario.
- Probably the most controversial of my new changes but I've also made it so that attaching CustomDoors to a model blocks PlayerActivate from checking DaggerfallStaticDoors for a door hit. This is because I realised that there could be a situation where the modder has changed their model such that the parts of the model that originally had door textures are now just ordinary walls and they use CustomDoors to define where the new doors are. Without my changes, the player can activate a wall where there used to be a door and the game will think that there is still a door there. My change means that the game relies entirely on CustomDoors to tell it if a door was hit. If you don't like this change, I just thought of an alternative. I could change the noCustomDoorFound variable to instead be a tickbox that the modder can tick to opt their model out of a DaggerfallStaticDoors check. That way, the DaggerfallStaticDoors GetHit() check will still be done after the CustomDoors one if that box is not ticked. Something like this:
```c#
[Tooltip("Tells the game to not check for a Static Door activation on this object if ticked.")]
public bool OptOutStaticDoorsCheck = false;

...

public static bool HasHit(RaycastHit raycastHit, out StaticDoor door, out bool dontDoStaticDoorsCheck)
{
    dontDoStaticDoorsCheck = OptOutStaticDoorsCheck;
...
```

Assets/Scripts/Utility/AssetInjection/Components/CustomDoor.cs
```diff
@@ -53,9 +56,13 @@ public static void InitDoors(GameObject building, StaticDoor[] staticDoors, int
             staticDoor.buildingKey = buildingKey;
 
             // Set data to all doors in the building
-            var doors = building.GetComponentsInChildren<CustomDoor>();
-            for (int i = 0; i < doors.Length; i++)
-                doors[i].staticDoor = staticDoor;
+            CustomDoor[] allCustomDoors = building.GetComponentsInChildren<CustomDoor>();
+            for (int i = 0; i < allCustomDoors.Length; i++)
+            {
+                if ((allCustomDoors[i].DoorType != null) && (allCustomDoors[i].DoorType != DoorTypes.None))
```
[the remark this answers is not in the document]
XJDHDR: Fixed.

Assets/Scripts/Utility/AssetInjection/Components/CustomDoor.cs
```diff
@@ -53,9 +56,13 @@ public static void InitDoors(GameObject building, StaticDoor[] staticDoors, int
...
+                    staticDoor.DoorType = allCustomDoors[i].DoorType;
```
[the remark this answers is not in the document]
XJDHDR: Fixed this as well.

Assets/Scripts/Utility/AssetInjection/Components/CustomDoor.cs
XJDHDR: Please take a look at line 95 to see how I use this variable. I check if this variable is set to anything other than `None` and if so, the trigger door's type is set to whatever value the modder set. If it is set to `None`, the DoorType doesn't get touched and it will remain what it was set to by the aforementioned GetModelData function. There is no possibility for my changes to set the door type to `None`. If you're worried about a modder choosing the wrong DoorType for their model of the three valid ones, I think it's their responsibility to either make sure it's correct or just keep it set to `None` and let the game decide what type they want.

As for the intended use case: In general, I want the ability to override the GetModelData function's decision about what type of door I'm placing. A specific example is the scenario I said in my other comment. Another one I can think of is if the modder is adding doors to a model that didn't originally have doors on it. Unless the game already has some failsafes in place, I think the DoorType of the door written into staticDoor will be `None` unless the modder uses my override.

Assets/Scripts/Game/PlayerActivate.cs
XJDHDR: That is correct. If there is an overlap between the two, I want CustomDoors to have priority. 

My other comment explains how this solves an issue I found.

TheLacus:
An instance of `CustomDoor` component defines an individual door, corresponding to an individual trigger. Its purpose is to add a new door, it was never intended to replace or affect in any way the original doors. In fact, building replacements must necessarily preserve general shape, including original doors position; this is vital for correct town layout.

There are two different topics addressed here:

- A building can have as many doors as you want, even on different levels of hierarchy. If you have more than one door on the same gameobject you need to make sure that each trigger is assigned to the right door in the inspector. I'll try to replicate the issues reported for multiple doors and door bashing and then we can continue from there.
- If there is a mistake in classic game data, it's better fixed via json data overrides or even directly in core (as i belive happened in a few other specific situations). Door data is provided by internal game logic and i think it should continue to be so. There is currently no support for custom buildings (meaning a new "link" from exterior to interior). If this is added in the future, `CustomDoor` would probably take a part in it, but it would require more than just exposing door type, so it's not something i'd consider here.

TheLacus:
I tested again and can confirm what i said above. If you have multiple doors on different gameobjects (i.e. different submeshes) there are no issues at all, the trigger is added and seeked automatically. If they are on the same gameobject you need to drag the collider into the inspector field for the corresponding door.

customDoor

I think is unlikely to have the need to handle more than two or three doors on the same gameobject (and probably only one in most cases) so i'm happy with the current approach.

XJDHDR:
> Its purpose is to add a new door, it was never intended to replace or affect in any way the original doors.
> including original doors position
> There is currently no support for custom buildings (meaning a new "link" from exterior to interior)

Okay, I'm confused. How is one supposed to use this script if both new links and replacing existing doors are not supported? I can't think of a scenario where using CustomDoors wouldn't trample on at least one of these three points.

> building replacements must necessarily preserve general shape, including original doors position; this is vital for correct town layout.

I don't agree. I don't believe there is anything in the game that requires buildings to retain their original shape. NPC pathing is probably designed to account for the positions of models in their area. And if it isn't, a model replacement won't cause any problems that aren't already in the game. Automaps can be re-painted to reflect the new building's shape. Ground retexures can likewise be altered to reflect the new model. What aspects of town layout requires that building models remain the same shape as vanilla that can't be resolved by the modder?

> fixed via json data overrides

Someone would have to implement such a method because one doesn't exist right now.

> even directly in core (as i belive happened in a few other specific situations)

Except Ferital is the only one who knows how to do those core-based fixes. Or at least, he is the only contributor to Arch3dPatch. I once looked into this myself but gave up when an hour later, all I had accomplished was mapping out where the start of the 4 records for each of the first 12 regions in MAPS.BSA were. And this is assuming that the bug in question can be fixed by hex-editing ARCH3D in the first place, which isn't necessarily the case.

> There is currently no support for custom buildings (meaning a new "link" from exterior to interior).

Agree and disagree, mostly the latter. My code doesn't change anything that your script can't already do in this regard. It is possible for someone to add a new door using this script to whatever model with doors they like, regardless of whether my code get merged or not. The only thing I've changed in this regard is giving the modder a way to choose what type of door their new addition is.

As far as dungeons are concerned, a new door would simply behave the way the game already handles multiple dungeon doors. All exterior doors link to the same entrance and vice versa. A new custom door in either place will simply link into that already existing paradigm.

For settlements, my understanding from reading your script is that any custom doors will copy the data for the first vanilla door in the staticDoors array. Which means that the new door will simply behave the same way, including linking to the same destination, as the first. My code doesn't change anything about this except what I said above.

> exposing door type, so it's not something i'd consider here.

I'm genuinely curious: What exactly is the problem caused by exposing this override?

> If you have multiple doors on different gameobjects (i.e. different submeshes) there are no issues at all, the trigger is added and seeked automatically.

Fair enough. If the intention is to support multiple doors through multiple scripts, I can revert that change.

TheLacus:
There are too many topics raised here, let's make a summary and then focus on one thing at a time, starting from door bashing. I'm sure we will understand each other better that way :)

- CustomDoor component allows to add (not replace) one or multiple new doors to an existing building (which already have doors). This was a feature requested on forums some time ago. All doors of a building are supposed to behave the same (this was confirmed by Interkarma at the time if i'm not misremembering) so static door data is copied only from first door (specific trigger bound values of classic doors are ignored by custom doors).
- Replacing a classic door with a custom door is currently unsupported. The reasoning of the time was that a new building mesh should be simililar to classic in shape to respect towns design, including original door position. This was the general agreement on the forums, and i personally find it hard to disagree. Said that, if there is a request to add support for it (i.e. ignore a classic door), it might be added by identifying doors with something unique, like door index value. For the being being I'm OK, as a workaround, to give custom doors priority over classic doors if there is an overlap of trigger volumes. However, classic doors shouldn't be disabled; the best approach (which preserves the original purpose of the component) is to implement functionality of replacing by door index during location layout.
- Adding doors to a model without classic doors (what i called a _a new "link" from exterior to interior_) is not supported. This is something you initially talked about as a reason to expose door kind, and i only wanted to suggest that is a bit out of scope here.
- I'm OK exposing ALL static door data (including the property needed for specific bug fix) via virtual method or event. First, i'd like @ajrb 's opinion on an alternative approach based on json override so we can make an informed decision together. This is not meant to remove value from your efforts, and i apologise if it might have seemed to be the case.

ajrb:
@TheLacus & @XJDHDR I need to admit to you both that I have no idea what you're talking about here because I've never dived into the doors of DFU much, and don't really understand why there's any restrictions. I think @BadLuckBurt looked into it for dungeon doors at least, so I'll see if he can explain it to me on discord. I will try to get up to speed enough to understand all this soon.

TheLacus:
Thanks mate! I'll try to explain it better, at least what i understand of it.

When you replace a building mesh with a mod, you need to make sure doors are in the same position of original mesh. Mod authors don't need to worry about internal door information (i.e. where this is going to teleport, etc.), only about position. Using any custom material has no effect on this. The component discussed here allows to defines new door trigger volumes, but this doesn't change the core of the issue.

My question is if overriding this kind of building data (not individual door positions on model, but the internal data used for exterior/interior transition) could make sense to be part of your location customization features via json files. Don't feel obliged to do anything, i just thought it was worth to let you know of this topic.

The alternative is to expose it via asset replacement of building models, which is fine but i believe it would be better to keep _transition data_ and _artistic_ side separated if possible, so i'd like the opinion of other devs first.

XJDHDR:
@TheLacus 
Based on your feedback, I have made the following changes to this PR:

- I've removed my addition of multiple Box Collider support and put it back the way it was.
- Instead of a Static Doors check being blocked in PlayerActivate based on whether CustomDoors is present or not, I've added a checkbox to the script that will only block this activation if ticked.
- I've removed my exposing the ability to change the Custom door's type and instead, added the ability to pick which of the doors in the model's StaticDoors array will be used to create the Custom door.

These should resolve most of our disagreement so far while still giving me the ability to do what I want to do.


> All the modeller needs to do is ensure door positions are in same position in replacement model and the virtual collider should overlay.

The problem I have with StaticDoors is that it doesn't (and can't) construct it's info from modded building models. As a result, my hope is something that can be used to overcome this limitation and my belief is that extending CustomDoors is the easiest way to do so.


> The idea of removing/overriding a classic, functional, door was not contemplated at the time because it was simply outside of the scope of the request.

I guess the big question that needs to be answered is whether modders should be barred from doing this through CustomDoors, for now at least. Should the scope be increased to include this?


> static door data is copied only from first door (specific trigger bound values of classic doors are ignored by custom doors).

In light of my recent changes, does it have to be the first door?


> Of course, if there is a request to add support for it (i.e. ignore a classic door) it can be added,

I vote for this.


> I'll think about your idea of a checkbox to ignore all classic doors, but i wonder how well it plays with multiple components on parent/children.

In the meantime, I've decided that this is a better choice than my hard block on Static Door activation if CustomDoors is attached. 
What do you mean by "multiple components on parent/children"? If you're talking about multiple DaggerfallStaticDoors scripts, the game is only programmed to recognise a single script per model AFAIK. If you're talking about multiple CustomDoors scripts, things are current set such that a single script having that tickbox checked will block Static Door activation for the entire model. As a result, someone ticking that box would have to add a Custom door for each Static door on the model. Blocking only specific doors in the array might be possible but I would have to investigate it some more. 

In my case, I won't be using that tickbox. I will simply place a single Custom door over the bugged Static door and let DaggerfallStaticDoors handle the two working ones.


> This is something you initially talked about as a reason to expose door kind, and i only wanted to suggest that this would require several other changes and we should better focus on the other topic for the moment.

My mistake. Neither of us have done anything to make CustomDoors support models that don't have Static doors present.


> I'm OK exposing ALL static door data (including the property needed for specific bug fix) via virtual method or event.

I understand now. You want to open this up all at once, not in a piecemeal fashion.


> This is not meant to remove value from your efforts, and i apologise if it might have seemed to be the case. :)

Not at all! Just keep in mind that as the one who made these changes, I am biased in favour of keeping them.

And one more thing I didn't really cover:
> json override (the latter being my first choice since it wouldn't require to replace the entire building mesh)

My main disagreement is that I can't really think of a situation where someone would want to mod Static door info without also changing the model. Even my use case involved editing the door's texture to be one that is valid for dungeon entrances. I only turned to CustomDoors when I found that this wasn't enough. I suppose it is technically possible to do what I want by just editing door data but that, IMO, is a very rare circumstance. Overall, I'm not sure it makes sense to expose a method to just edit door data alone when I think the majority of the time, that editing would be in addition to changing the model anyway.

@ajrb 
Is there anything in particular you need help with or are you completely lost?
The way I figured out what I know is by going to the ActivateStaticDoor function on line 462 of PlayerActivate and then following the code forwards and backwards. IN particular, here's a message I posted on Discord for BadLuckBurt while I was investigating this:
> It looks like you're right. The properties of the DaggerfallStaticDoors script are populated in this case by the RMBLayout script. That ultimately uses MeshReader's GetModelData() on line 739 to get this info. And that is indeed hardcoded to look for models in ARCH3D.BSA

TheLacus:
> > I'll think about your idea of a checkbox to ignore all classic doors, but i wonder how well it plays with multiple components on parent/children.
> 
> In the meantime, I've decided that this is a better choice than my hard block on Static Door activation if CustomDoors is attached.
> What do you mean by "multiple components on parent/children"? If you're talking about multiple DaggerfallStaticDoors scripts, the game is only programmed to recognise a single script per model AFAIK. If you're talking about multiple CustomDoors scripts, things are currently set such that a single script having that tickbox checked will block Static Door activation for the entire model. As a result, someone ticking that box would have to add a Custom door for each Static door on the model. Blocking only specific doors in the array might be possible but I would have to investigate it some more.


One instance of `DaggerfallStaticDoors` covers and entire block of buildings (which, at least for non-custom models, are batched in a single mesh). `CustomDoor` component is added by the mod author to the individual building gameobject, at any level of the gameobject hierarchy (i.e. on door submesh). If a raycast hits a trigger associated to `CustomDoor` instance it means there is a door. If there isn't a hit, it doesn't mean there isn't another door (with enabled checkbox) on another branch of hierarchy.

I haven't had time to test it yet, but i want to be sure that the checkbox you added actually affect the entire building. If that's not the case, the only reliable way to do it is at layout time. I'll provide more details when i can find some time to test all your changes. :)

Interkarma:
Hi @rossipaolo and @XJDHDR. Let me know when you guys are happy enough with this one for me to merge.

I'm planning to release weekly updates over next 2-3 weekends until I'm caught up on PRs, and happy to include this one once you feel it's ready.

XJDHDR:
I'm pretty much finished with this unless TheLacus has anything else to offer. I feel that my other PRs are more important than this one though and would rather see them merged first.

rossipaolo:
Hi! After further testing i can confim the issue i raised above was correct, this approach only disable static doors on the gameobject where component is added, not for other gameobjects related to the same building. This is how to reproduce:

1. Make a parent gameobject for the building.
2. Add a child gameobject for custom door (and choose option to suppress classic doors).
3. Classic door is still activable due to parent (or another sibling) collider.

My suggestion was to avoid adding (or to mark as deactivated) classic doors when performing town layout, instead of as part of hit check, but if current solution works fine for current needs i'd say to go for it :)

Assets/Scripts/Utility/AssetInjection/Components/CustomDoor.cs
```diff
@@ -25,18 +25,35 @@ namespace DaggerfallWorkshop.Utility.AssetInjection
...
+            if (!DoorTrigger)
+            {
+                if (name == "default")
+                    objectName = transform.parent.name + "\\default";
+                else
+                    objectName = name;
+                Debug.LogErrorFormat("No Box Colliders are attached to {0}. Did you forget to add one?", objectName);
+                return;
+            }
+
```
rossipaolo, line 56: `[RequireComponent(typeof(BoxCollider))]` should actually enforce that a collider is always present.

Interkarma:
Thank you for review @rossipaolo!

@XJDHDR would you like to add these changes? Otherwise I'm happy to add @rossipaolo's suggestion after merging if you don't have any issues with that?

XJDHDR:
@Interkarma 
Sure, I'll see what I can do.

XJDHDR:
@rossipaolo @Interkarma 
Sorry for the taking so long but I recently became busy with [a new project](https://github.com/XJDHDR/BethBryo_for_Unity).

Anyway, based on TheLacus' suggestions, I've made the following changes:
- Removed the check for a BoxCollider since, as he said, the RequireComponent declarative already ensures this.
- Instead of blocking activation, the `NoStaticDoorCheck` tickbox will, instead, stop those static doors from being created in the first place.

If TheLacus is happy with what I have so far, I'm happy for this PR to be merged.

Interkarma:
@XJDHDR Thank you for the update. Good luck with your new project! Don't hesitate to post progress to the forums if you ever want to. :)

rossipaolo:
I believe `NoStaticDoorCheck` now affects the entire RMB block, not just the building. Might be better to suppress only individual buildings here https://github.com/Interkarma/daggerfall-unity/blob/bd2c56af23d271a2895c42c10bb6920b88615273/Assets/Scripts/Utility/RMBLayout.cs#L747-L753

rossipaolo (approved):

Interkarma:
Thank you for review and merging @rossipaolo 👍

XJDHDR:
And thank you for fixing that remaining issue too. And sorry about that. I guess I didn't test things properly.

PR #2008, opened 2021-01-29
Bugfix of wrong choice for Nature Billboards at Climate borders

Bl4ckh34d:
Solved a bug where the PlayerGPS climate was used when picking nature billboards for new terrains instead of the terrains climate itself. This solves the bug at climate transitions, where terrains got nature billboards assigned, that were based on the players current climate location, not their actual climate.

### DIFF
--- a/Assets/Scripts/Terrain/StreamingWorld.cs
+++ b/Assets/Scripts/Terrain/StreamingWorld.cs
@@ -1250,7 +1250,8 @@ namespace DaggerfallWorkshop
             if (dfTerrain && dfBillboardBatch)
             {
                 // Get current climate and nature archive and terrain distance
-                int natureArchive = ClimateSwaps.GetNatureArchive(dfTerrain.MapData.worldClimate, dfUnity.WorldTime.Now.SeasonValue);
+                DFLocation.ClimateSettings worldClimateSettings = MapsFile.GetWorldClimateSettings(dfTerrain.MapData.worldClimate);
+                int natureArchive = ClimateSwaps.GetNatureArchive(worldClimateSettings.NatureSet, dfUnity.WorldTime.Now.SeasonValue);
                 dfBillboardBatch.SetMaterial(natureArchive);
                 int terrainDist = GetTerrainDist(LocalPlayerGPS.CurrentMapPixel, dfTerrain.MapPixelX, dfTerrain.MapPixelY);
                 dfUnity.TerrainNature.LayoutNature(dfTerrain, dfBillboardBatch, TerrainScale, terrainDist);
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Terrain/StreamingWorld.cs
```diff
@@ -1250,7 +1250,7 @@ public void UpdateTerrainNature(TerrainDesc terrainDesc)
             if (dfTerrain && dfBillboardBatch)
             {
                 // Get current climate and nature archive and terrain distance
-                int natureArchive = ClimateSwaps.GetNatureArchive(LocalPlayerGPS.ClimateSettings.NatureSet, dfUnity.WorldTime.Now.SeasonValue);
+                int natureArchive = ClimateSwaps.GetNatureArchive(dfTerrain.MapData.worldClimate, dfUnity.WorldTime.Now.SeasonValue);
```
Interkarma, line 1253: Good catch. :) But can you please use MapsFile.GetWorldClimateSettings(worldClimate) here instead of a custom method. Like so:

    DFLocation.ClimateSettings worldClimateSettings = MapsFile.GetWorldClimateSettings(dfTerrain.MapData.worldClimate);
    int natureArchive = ClimateSwaps.GetNatureArchive(worldClimateSettings.NatureSet, dfUnity.WorldTime.Now.SeasonValue);

GetWorldClimateSettings() returns a DFLocation.ClimateSettings package from which you can then get nature archive. This keeps the conversation of climate index to nature archive all in the one place so changes won't become out of sync. It also simplifies the PR a bit. Cheers!

Bl4ckh34d: Hey, I am new to Github and for me that's it. I am done with it an never want to work with it ever again. I tried it for one hour now and couldn't figure out how to change this PR. I think I closed the branch already by accident but who knows what branches, commits, detached heads and all this other crap really are...

Maybe you can just copy those two lines into your original code by yourself and delete this PR?
I just reached my threshold of tolerance and time I want to waste on this.
I am so done with Github...

Bl4ckh34d (commented):
I have no idea what I am doing here. I hate Github from the depth of my heart and wasted already too much time on it, trying to understand their secret language and twisted system.

Please, just add those two lines you suggested to your code, if you deem them correct. No need to mention me on your page or anything. Let's just get rid of this small bug.
This way it will save me countless of hours dealing with this pain of a platform and a huge part of my sanity.

Interkarma:
Hey, don't worry about it. :) I'll merge change on my end and credit you for bug fix. All the best.

Bl4ckh34d:
Thank you! <3
Git really got me to my limits.
I really enjoy coding as a hobby, but versioning is my personal hell.

XJDHDR:
For future reference, though you may have figured this out already, you can modify a PR by pushing additional commits to the branch you used to make this PR (nature-billboard-fix). The things you appear to be having trouble with are not unique to GitHub but are the way Git itself works. If you go to any other place that uses Git (such as GitLab, Bitbucket or SourceForge), you will see the same thing.

All I can recommend is to try look up some tutorials on how Git works. I myself stuck with SVN for a long time because I couldn't understand Git.

PR #2010, opened 2021-01-30
Move Stealth skillcheck to FormulaHelper

Ralzar81:
I am working on a Thief overhaul mod and to make any headway I need the ability to override the stealth check.

As far as I can see, only EnemySenses.cs checks this, so I have made a PR that moves this calculation to FormulaHelper, making it possible for mods to easily change how stealth is checked.

### DIFF
--- a/Assets/Scripts/Game/Formulas/FormulaHelper.cs
+++ b/Assets/Scripts/Game/Formulas/FormulaHelper.cs
@@ -277,7 +277,7 @@ namespace DaggerfallWorkshop.Game.Formulas
         }
         
         // Calculate chance of stealth skill hiding the user.
-        public static int CalculateStealthChance(float distanceToTarget, float globalScale, DaggerfallEntityBehaviour target)
+        public static int CalculateStealthChance(float distanceToTarget, DaggerfallEntityBehaviour target)
         {
             Func<float, float, DaggerfallEntityBehaviour, int> del;
             if (TryGetOverride("CalculateStealthChance", out del))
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Game/Formulas/FormulaHelper.cs
```diff
@@ -275,6 +275,17 @@ public static int CalculateShopliftingChance(PlayerEntity player, int shopQualit
             chance += shopQuality + weightAndNumItems;
             return Mathf.Clamp(chance, 5, 95);
         }
+        
+        // Calculate chance of stealth skill hiding the user.
+        public static int CalculateStealthChance(float distanceToTarget, float globalScale, DaggerfallEntityBehaviour target)
```
ajrb, line 280: Are both float params really necessary? It could simply be a single param which is already divided by global scale or the formula could just use the const directly. Either way I don't see the use of this second param for custom formula.

Ralzar81: True, I went a bit back and forth on this. But decided to take both, since anyone making a mod for stealth in the future might like the ability to change the distance calculation which is done by using those two floats.

ajrb: Yes, okay. My point is that even so the param being supplied is actually a public constant which can be used from any mod code so the param is redundant.

I think it should be removed.

Ralzar81: Good point. I changed this so it just reads MeshReader.Globalscale dirctly instead.

Interkarma: Thank you, I definitely prefer to read the constant. The whole world is constructed on that scale value, it's like a constant of the universe.

PR #2011, opened 2021-01-31
fix potion maker's stackCounts

petchema:
Keep track of parent stack of splitted stacks;

Use it to enforce a new invariant: if a parent stack is in cauldron, none of its children must be in the ingredients 
(=> move parent to cauldron last, but also move parent to ingredients first).

Forums: https://forums.dfworkshop.net/viewtopic.php?f=31&t=4507


ajrb:
Not very keen on this, I would prefer that the scheme you described on forums was implemented. It will make code much easier to understand that way.

petchema:
> Not very keen on this, I would prefer that the scheme you described on forums was implemented. It will make code much easier to understand that way.

I created #2020 implementing the original idea

Interkarma:
Hi @petchema. Thank you so much for the two PRs. I'll review these for 0.11.2 due in March sometime.

Interkarma:
Hey @petchema. I think I'll roll with #2020 fix. Will close this one now and work my way up to that PR. :)

PR #2013, opened 2021-01-31
Change removal of Oghma to be handled by effect manager

ajrb:
Suggested fix only - untested, so will need testing first. I just figured this would be the correct way to fix the issue rather than adding remove from wagon into Oghma effect class. Feel free to close if you don't like it.


Interkarma:
This is a nice solution. Much neater than what I was considering. Thank you!

Interkarma:
This works perfectly. Thanks for the fix @ajrb! 👍

PR #2014, opened 2021-02-01
Add thrust to click to attack

AsesinoBlade:
Hi all,
I saw this as a bug, but it may have been intentional.  If so, please reject this PR.
I noticed that the click to attack never selects Thrust forward.  It turns out that the random selection doesn't have thrust as an option.  I changed it to now include upRight (enum #3 instead of starting with Left Enum #4).

I did have to change FPSWeapon as OnAttackDirection was only looking for Up instead of also looking for UpLeft and UpRight.

Thanks 
a


AsesinoBlade:
Will resubmit with cleaner history

PR #2015, opened 2021-02-01
Improve support of mod localization

rossipaolo:
I have been looking at improving support for localization of custom text used by mods. Ideally the process should be similar to core translation but it doesn't look like new databases can be discovered at runtime, so this is still based on old csv tables, at least until we find a better solution.

A mod with custom text should have a main table named **textdatabase.txt** for fallback (this was already supported), while built-in translations now can be provided with language identifiers (for example **[en]textdatabase.txt**, **[fr]textdatabase.txt**, ...). Other mods can supply missing translations via GUID (for example **[fr(7e0510a9-be7d-4fd4-9c89-ed2f2b2e5fb3)]textdatabase.txt**).

### DIFF
--- a/Assets/Game/Addons/ModSupport/Mod.cs
+++ b/Assets/Game/Addons/ModSupport/Mod.cs
@@ -19,6 +19,8 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using UnityEngine.Localization;
+using UnityEngine.Localization.Settings;
 using DaggerfallWorkshop.Utility;
 using FullSerializer;
 using DaggerfallWorkshop.Game.Utility.ModSupport.ModSettings;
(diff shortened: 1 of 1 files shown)


ajrb:
Nice. I took a look at the existing docs here https://www.dfworkshop.net/projects/daggerfall-unity/modding/features/#localization but couldn't see how these entries are actually accessed anywhere. Up to now I've not bothered to extract the text in my mods so I don't know how to use the existing system this is built on.

One question for the future is whether tokenised text can be supported somehow? Each line can be a different entry easily enough so it's not a must have but thought I'd mention it in case you had some nice easy way to provide it.

TheLacus:
It's written in that section (after the screenshot) how to use it for settings and custom text. For convenience, this is how you use it from mod code:

```cs
string text = mod.Localize("textKey");
```

where `textKey` is a key in the table.

```csv
schema: *key,text

textKey,                 Example text.
```

I don't think there is a chance for tokenised text with this approach. Maybe it doesn't really answer your question, but if you want to store custom formatted text (or data) you could use the localization system to "translate" an asset name:

```cs
ModManager.Instance.TryGetAsset(mod.Localize("example.json"), out TextAsset textAsset));
```

```csv
schema: *key,text

example.json,                 example_fr.json
```

Interkarma and I talked about a more integrated support for localized assets some time ago, but this should already work for now as a workaround.

ajrb:
Okay thanks. Maybe move the settings section below the string section... or maybe it's just me that didn't see it cos I am old. :-)

Interkarma:
Thank you for this one. I'll merge now so we can start community testing from next release.

PR #2016, opened 2021-02-01
Added Forward Thrust as an option

AsesinoBlade:
Hi all,
I saw this as a bug, but it may have been intentional. If so, please reject this PR.
I noticed that the click to attack never selects Thrust forward. It turns out that the random selection doesn't have thrust as an option. I changed it to now include upRight (enum #3 instead of starting with Left Enum #4).

I did have to change FPSWeapon as OnAttackDirection was only looking for Up instead of also looking for UpLeft and UpRight.

Thanks
a

### DIFF
--- a/Assets/Scripts/Game/FPSWeapon.cs
+++ b/Assets/Scripts/Game/FPSWeapon.cs
@@ -172,6 +172,8 @@ namespace DaggerfallWorkshop.Game
                     state = WeaponStates.StrikeDownRight;
                     break;
                 case WeaponManager.MouseDirections.Up:
+                case WeaponManager.MouseDirections.UpLeft:
+                case WeaponManager.MouseDirections.UpRight:
                     state = WeaponStates.StrikeUp;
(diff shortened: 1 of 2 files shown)


Interkarma:
I think the bug here is  that the Random.Range didn't include the full enum range. Good catch.

UpLeft and UpRight aren't used for forward thrust that I'm aware of, but I'm thinking this would make thrust gesture easier to execute in any case. Will play with this and accept as-is unless there's some issue. Thanks for fixes. :)

gsaeed:
I was thinking to include the enum but then worried that it would make thrust forward occur 3 out of 8 times.   For that reason, i only used UpRight as that was enum #4 so it would choose thrust 1 out 6 times.

I thought to move “Up” in the enum class but worried that if anyone had used that enum class in a precompiled mod, it would no longer work.  (Although I am not certain if that is a legitimate concern.)

Thanks
A

PR #2018, opened 2021-02-02
Fix %pcn macro written as %pfn and add missing apostrophes

df-ferital:
Fix a bug reported on the forums: http://forums.dfworkshop.net/viewtopic.php?f=5&p=52162&sid=a9eea9ca6a9b8d58b2ded2b973db6856

I choosed to use %pcn (player's full name) here rather than %pcf (player's first name), as it seems more appropriate for a hostile reaction.

I also took this occasion to add some missing apostrophes.


Interkarma:
Thank you @df-ferital! Appreciate the fix on missing apostrophes too. :)

PR #2020, opened 2021-02-02
fix potion maker's stackCounts (take 2, alternative to #2011)

petchema:
- Replace ingredients with an ItemCollection of cloned ingredients, not shared with either player's inventory or wagon.
- Add/Remove ingredients to the cauldron using ItemCollection methods
- When the cauldron is mixed, find and remove ingredients from player's inventory first, and wagon as a fallback.

This is a more extensive rewrite than take 1, so I don't exclude bugs.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallPotionMakerWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallPotionMakerWindow.cs
@@ -124,6 +124,14 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             Refresh();
         }
 
+        public override void OnPop()
+        {
+            ingredients.Clear();
+            ingredientsList.Clear();
+            cauldron.Clear();
+            recipes.Clear();
+        }
+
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Game/UserInterfaceWindows/DaggerfallPotionMakerWindow.cs
```diff
@@ -122,11 +124,6 @@ public override void OnPush()
             Refresh();
         }
 
-        public override void OnPop()
-        {
-            ClearCauldron();
-        }
-
```
ajrb, line 129: I think OnPop() should still be here and it should clear the ingredients item list and allow the cloned items to be disposed of. Just do:

`ingredients = null;`

What do you think? (I'm theory crafting here)

petchema: Ah yes, because the window is reused instead of destroyed

Interkarma:
Thank you so much @petchema. This all works perfectly.

PR #2024, opened 2021-02-05
Automap: Add support for the discovery state of OBJ models to be saved and loaded

### LINKED ISSUE #1995: Replacement OBJ models invisible on dungeon map after loading save

XJDHDR:
This does two things:
1. Makes the UpdateMeshRendererDungeonState() function be called on all components of all children of the current transform model instead of only the first component in it's root. This fixes an issue where already discovered models on the dungeon map are not revealed after loading a save game if an OBJ file has been used to replace the model.
2. Solves an issue where the SaveStateAutomapDungeon function doesn't properly save the discovery state of OBJ models because it only checked the MeshRenderer component attached to the model's root and not those attached to children.

This PR is identical to #2002, just with all of my changes moved to a new branch. This PR fixes #1995.

### DIFF
--- a/Assets/Scripts/Game/Automap.cs
+++ b/Assets/Scripts/Game/Automap.cs
@@ -2264,16 +2264,20 @@ namespace DaggerfallWorkshop.Game
                     foreach (Transform currentTransformMesh in currentTransformModel.transform)
                     {
                         AutomapGeometryModelState model = new AutomapGeometryModelState();
-                        MeshRenderer meshRenderer = currentTransformMesh.GetComponent<MeshRenderer>();
-                        if ((meshRenderer) && (meshRenderer.enabled))
+                        MeshRenderer[] meshRenderers = currentTransformMesh.GetComponentsInChildren<MeshRenderer>();
+                        for (int i = 0; i < meshRenderers.Length; i++)
                         {
-                            model.discovered = true;
-                            model.visitedInThisRun = !meshRenderer.materials[0].IsKeywordEnabled("RENDER_IN_GRAYSCALE"); // all materials of model have the same setting - so just material[0] is tested
-                        }
-                        else
-                        {
-                            model.discovered = false;
-                            model.visitedInThisRun = false;
+                            if ((meshRenderers[i]) && (meshRenderers[i].enabled))
+                            {
+                                model.discovered = true;
+                                model.visitedInThisRun = !meshRenderers[i].materials[0].IsKeywordEnabled("RENDER_IN_GRAYSCALE"); // all materials of model have the same setting - so just material[0] is tested
+                                break;
+                            }
+                            else
+                            {
+                                model.discovered = false;
+                                model.visitedInThisRun = false;
+                            }
                         }
                         models.Add(model);
                     }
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Game/Automap.cs
```diff
@@ -2481,10 +2485,13 @@ private void RestoreStateAutomapDungeon(bool forceNotVisitedInThisRun = false)
...
 
-                        MeshRenderer meshRenderer = currentTransformModel.GetComponent<MeshRenderer>();
-                        if (meshRenderer)
+                        MeshRenderer[] meshRenderers = currentTransformModel.GetComponentsInChildren<MeshRenderer>();
+                        for (int i = 0; i < meshRenderers.Length; i++)
                         {
-                            UpdateMeshRendererDungeonState(ref meshRenderer, automapDungeonState, indexBlock, indexElement, indexModel, forceNotVisitedInThisRun);
+                            if (meshRenderers[i])
+                            {
+                                UpdateMeshRendererDungeonState(meshRenderers[i], automapDungeonState, indexBlock, indexElement, indexModel, forceNotVisitedInThisRun);
```
Interkarma, line 2493: This is just missing a "ref" for meshRenderers[i] when passed to UpdateMeshRendererState(). Otherwise no issues found, so I'll just merge and add the missing ref on my end.

XJDHDR: I removed it because @rossipaolo [mentoned](https://github.com/Interkarma/daggerfall-unity/pull/2002#discussion_r563820584) that it's not needed.

Interkarma: Well code doesn't compile without it. To remove it, we also need to change the corresponding ref in method signature for UpdateMeshRendererDungeonState().

But yeah @rossipaolo is correct - MeshRenderer is a class, so shouldn't be any issues with removing ref from both call and method. I'm happy to test and change on my side now merged.

PR #2028, opened 2021-02-11
PlayerEnterExit: Reduce delay filling Dungeon variable

XJDHDR:
Previously, GameObjectHelper ran the SetDungeon function in DaggerfallDungeon before returning it's results. Since PlayerEnterExit uses the value of this returned result to fill it's Dungeon variable, this meant that there was a short period where Dungeon was equal to Null even though the DaggerfallDungeon object existed. This moves the SetDungeon function so that it runs after the Dungeon variable is filled.

I encountered this problem while working on PR #2026 where I couldn't retrieve the value of the dungeon's ID because Dungeon was still Null while the ApplyMaterials function in RuntimeMaterials was running.


Interkarma:
Looks good, thanks for the fix. :)

PR #2030, opened 2021-02-13
Constant horse jumping capability

petchema:
Make horse jumping capability independant of player's capability, so you can always jump over hedges with the horse.
I used a jumpSpeedMultiplier value higher than strict minimum so horse doesn't "stumble" as much on hedges (unless you have perfect timing); But best value is debatable.

Forums: https://forums.dfworkshop.net/viewtopic.php?p=51147#p51147


Interkarma:
Thank you for this. It's a nice change and I'm happy to have a way to jump those hedges again! :)

PR #2031, opened 2021-02-13
BookFile: Added try-catch check to BinaryReader portion

XJDHDR:
On a few occassions, I've loaded a save only to get the HUD overlaying a black screen. The game still appeared to be running but the game didn't seem to react to my inputs. I couldn't open the main menu either and had to kill the game. On this occasion, I caught this happening in the editor which reported an `EndOfStreamException` in the line `header.PageOffsets[i] = reader.ReadUInt32()`. As a result, I added a try-catch check to gracefully handle this exception in future for any part of the BookFile's BinaryReader code.

### DIFF
--- a/Assets/Scripts/API/BookFile.cs
+++ b/Assets/Scripts/API/BookFile.cs
@@ -30,6 +30,8 @@ namespace DaggerfallConnect.Arena2
         readonly FileProxy bookFile = new FileProxy();
         BookHeader header = new BookHeader();
 
+        internal string lastSuccessfulBookHeaderRead = "";
+
         #endregion
 
         #region Properties
@@ -96,8 +98,14 @@ namespace DaggerfallConnect.Arena2
                 return false;
 
             // Read book
-            ReadHeader();
-
+            try { ReadHeader(); }
+            catch (EndOfStreamException e)
+            {
+                Debug.LogErrorFormat($"EndOfStreamException encountered for book {name}.\n" +
+                    $"The last piece of data that was successfully read was {lastSuccessfulBookHeaderRead}.\n" +
+                    $"{e.ToString()}");
+                return false;
+            }
             return true;
         }
 
@@ -106,10 +114,18 @@ namespace DaggerfallConnect.Arena2
         /// </summary>
         /// <param name="data">Byte array to parse as a book file.</param>
         /// <param name="name">Name to describe book.</param>
-        public void OpenBook(byte[] data, string name)
+        public bool OpenBook(byte[] data, string name)
         {
             bookFile.Load(data, name);
-            ReadHeader(name.StartsWith("BOK", StringComparison.OrdinalIgnoreCase));
+            try { ReadHeader(name.StartsWith("BOK", StringComparison.OrdinalIgnoreCase)); }
+            catch (EndOfStreamException e)
+            {
+                Debug.LogErrorFormat($"EndOfStreamException encountered for modded book {name}.\n" +
+                    $"The last piece of data that was successfully read was {lastSuccessfulBookHeaderRead}.\n" +
+                    $"{e.ToString()}");
+                return false;
+            }
+            return true;
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


petchema:
If you hit the end of file while reading, then all further reads will also fail, so handling of the exception could be done at a higher level. In fact, I assume pages come after the header in the same file, so reading pages will also fail.
I think the caller should be made aware that the book is unreadable; The result of just logging exceptions and trying to move on with a partially initialized datastructure is too unpredictable for my taste...

XJDHDR:
Good points. I should nest those two try-catch blocks. I assumed you could not nest them but I see that you actually can. Another thing I'm assuming is that as soon as one of the functions in a try block triggers an exception, the rest of the try block does not run. Is this correct?

I'll see what I can do about telling the caller about the read failure.

TheLacus:
You tipically only catch an exception if you can handle the problematic situation. Exceptions are used specifically when you can't do it, so catching the exception and log an error message is not really helpful.

I don't think @petchema was suggesting to nest exceptions, rather to let the caller (i.e. who uses the book) to be aware that read operation failed. For example return false from `OpenBook()`, like it already does for other issues.

```cs
public bool OpenBook(string arena2, string name, FileUsage fileUsage = FileUsage.UseMemory, bool readOnly = true)
{
    // Validate filename
    if (!name.StartsWith("BOK", StringComparison.InvariantCultureIgnoreCase) ||
        !name.EndsWith(".TXT", StringComparison.InvariantCultureIgnoreCase))
    {
        return false;
    }

    // Load book file
    string booksPath = Path.Combine(arena2, books);
    if (!bookFile.Load(Path.Combine(booksPath, name), fileUsage, readOnly))
        return false;

    // Read book
    try
    {
         ReadHeader();
         return true;
    }
    catch (EndOfStreamException)
    {
        return false;
    }
 }
```

Said that, i would suggest to find out what caused the issue you experienced.

XJDHDR:
My two goals for this PR are:

- **Give more info in the error log:** The one I received that prompted this PR did not, IMO, provide enough. The only info I could glean from it was the line that triggered the exception. I would like to add some data about which book the binary reader was trying to read when the error occurred. Maybe even the offset and file as well. I admit that what I've got so far doesn't do enough, though an improvement.
- **Handle the exception more gracefully:** As I wrote, this error's appearance was accompanied by a black screen and unresponsive game. While this doesn't prove a correlation, I think it strongly suggests one. And if this exception is the cause, I think that the game semi-hard locking because it couldn't read a book is extreme. I would say some sort of error message popping up would be more appropriate. 

Moving the exception handling to the two `OpenBook` functions that call it may be an option. It mainly depends on which route gives the most information I can include in the log. Taking a second look, maybe the info in those functions' local variables might be a good idea to include.

As for finding out what caused my issue, the main problem is that it is extremely intermittent. 99% of the time, my saves would load without issue. Then out of the blue, I load to an unresponsive black screen. I can then restart my game and load the same save. Only this time, it would load without issue.

Interkarma:
I'm happy to have exception handling or other checks to prevent a crash state and log more information. Since we're potentially dealing with random sources of gamedata on user end, this is actually very helpful. But I'd also like to dig into the base problem a little more before merging.

If a book is failing to read here, it means the data format is incorrect/unexpected to reach end of stream prematurely. Unfortunately, I haven't once encountered this black screen or exception reading using book data on my end. Can you confirm your source of game data, and ideally which book was being read at the time of crash?

XJDHDR:
The files came from DaggerfallSetup with all it's fixes activated. It was whichever was the latest version around 2 years ago. I know that the recommended way of playing is with DaggerfallGameFiles. I probably should switch over while I'm thinking about it. Unless keeping the files is a good idea to debug?

As for the book, the funny thing is that I wasn't reading any book at the time. I just loaded my one save in a dungeon and it loaded to a black screen. I don;t have any books in my inventory; just 3 Letters of Credit and two quest letters for Morgiah and Prince Lhotum. There are 3 books in my wagon.
This only happened once though, I have loaded that same save a number of times before and afterwards and didn't have any issues those times.

Interkarma:
DaggerfallSetup+fixes should be fine. I wonder why it was even loading a book file at that point in order to throw exception? Just loading a save doesn't trigger this code (just tested then to be sure). BookFile.ReadHeader isn't called until player actually has some touch on a book - e.g. tooltip, reading, buying.

Off the top of my head, I can't explain why this code would be executed when just loading a save. Understanding that might give us some clue to why exception is thrown, and why I've never seen it happen. It feels like there's a mod or file change somewhere in the mix here contributing to problem.

XJDHDR:
If it was a mod that did this, I think it would have had to have been a mod that left some data behind after uninstalling. In the Editor the only mod I had installed was my Unofficial Block, Location and Model Fixes mod, which doesn't touch books or have any code to execute.
In terms of code changes, I was working on SoundReplacement as I described in [this thread](https://forums.dfworkshop.net/viewtopic.php?f=23&t=4568). I don't think anything I changed in that script would have made that book code run,

I did compare my previous Arena2 folder to the one from DaggerfallGameFiles and WinMerge reported that there were some books that had binary differences. Maybe one of them was the cause of this problem? I've attached those books here: 
[Different books.zip](https://github.com/Interkarma/daggerfall-unity/files/5991286/Different.books.zip)

Interkarma:
Thank you. I'm just confused why the BookFile code would even execute when loading a save. There's got to be an edge case of some kind at work. Whether that's related to exception or not is the question.

Here's another way to look at it. Books will soon be moved to localisation string tables, and BookFile will only be used to import data for first time. Any data issues will be uncovered then as all books will be processed. Your exception handling will help make this more robust and work out any problem books. It also saves a random crash that we aren't sure how to reproduce yet. Sounds like a good change to me, so will happily merge this one soon.

Thanks for your time on this one.

XJDHDR:
Don't merge it yet. I still need to see what info I can add to the log to narrow down the cause of any exceptions.

TheLacus:
The stack trace would probably explain why a book is read when loading a save. @XJDHDR You can preserve it after exception is catched if you use `Debug.LogException()` or log `Exception.StackTrace` property.

XJDHDR:
From what I've read, `e.ToString()` includes the contents of `e.StackTrace`. I'll probably use that instead.

XJDHDR:
@Interkarma 
I think I'm pretty much done with this PR. The only thing I would add is a simple messagebox that tells the user that this exception occurred but I don't know how to do that.

Interkarma:
This looks OK to me, thank you for the detailed exception logging. 👍

PR #2032, opened 2021-02-13
Implement right clicks as described in Daggerfall manual

ajrb:
See page 53 of the original Daggerfall manual. Variant changing was added to right clicks, this is now on middle mouse click.

As discussed here: https://forums.dfworkshop.net/viewtopic.php?p=52256#p52256

@Interkarma This is a candidate change which you can veto if you wish. 

Note that it may break mods that override inventory or trade UI windows until they are updated.

@petchema Would you mind casting your eye for detail over this one and test it to see if I missed any interactions.


Interkarma:
This works as described and matches classic. I have no problems with change, but noted it could break mods overriding inventory and trade windows.

Thank you!

PR #2033, opened 2021-02-14
Add joystick middle-click keybind and set Joystick UI controls to protected

jefetienne:
Inserts a new middle-click keybind accessible through the Joystick controls window, defaulted to `JoystickButton2`. Added in part by #2032 but would also benefit existing in the first place for mods that might make use of the middle mouse button.

Set the UI controls to protected so they can be accessible to subclasses created by mods.


Interkarma:
Awesome - thank you! :)

PR #2035, opened 2021-02-14
Prevent player location being selected for remote place

ajrb:
It happens occasionally and breaks the rest of one of the Archs guild quests Jay wrote when it does occur.

Rather hard to reproduce as it's based on random number selecting the current location.


petchema:
(Just a note, this fixes randomly picking the local place out of all region places, which is a different issue from
https://forums.dfworkshop.net/viewtopic.php?f=31&t=1879
which always picks the local place, when it doesn't look like this was the quest intent)

Interkarma:
Short and sweet, a good fix. Thank you. :)

Cheers @petchema for confirming unrelated to that forum issue. I need to start working down through extant bugs again someday once my plate is cleared.

PR #2036, opened 2021-02-15
Automap: Check CombineRDB state before resetting

XJDHDR:
While building the dungeon's Automap, the script's CreateDungeonGeometryForAutomap function disables the CombineRDB option to create individual models in the map instead of a combined model. After generating these models, it then re-enables this option. The problem is that if the user has disabled this option through the Unity Editor, this script re-enables it after generating the automap, forcing the user to re-disable it every time they load or enter a dungeon.

This commit adds a check to the setting manipulation so that at the end of automap generation, the CombineRDB option will only be re-enabled if it was originally enabled at the start of generation.


Interkarma:
This is a good catch, thank you.

Nystul-the-Magician:
thx for catching!

PR #2040, opened 2021-02-16
Import mod settings after changes

rossipaolo:
Automatically import mod settings and presets files after creation or changes from editor window.


Interkarma:
Thank you!

PR #2043, opened 2021-02-16
Fix cache issue with climate/season model variants

### LINKED ISSUE #2042: "triedModels" cache can break location/climate model variants
**Describe the bug**
#1148 by @ajrb  introduced a sort of cache to avoid additional seeks for models that are known not to be available. Unfortunately i didn't notice at the time that it can cause false positives when location/climate model variants are involved. Specifically this happens if a variant is available but the base prefab is not; this is probably a rare situation but i think it should still be supported properly.

**To Reproduce**

1. Make a mod containing prefab named `41116_Desert`. This is the variant only used for desert climate; base variant `41116` is not provided.
2. Load save in desert. Desert variant is used as expected.
3. Load save in a different climate than desert. No matching variant is found, so original asset is used. This is intended behaviour.
4. Load save in desert again. Desert variant is not used because model id was marked as missing.

**Expected behavior**
Climate/season variant should always be used if available, even if base variant is not provided. Behaviour shouldn't change depending on which climate/season is met first. 

**Desktop:**
 - Version 0.11.1

**Additional context**
Cache shouldn't only contain model uid, but also information on climate/season to allow proper invalidation.

[Saves.zip](https://github.com/Interkarma/daggerfall-unity/files/5990561/Saves.zip)
[Example.zip](https://github.com/Interkarma/daggerfall-unity/files/5990575/Example.zip)

rossipaolo:
Fixes #2042

### DIFF
--- a/Assets/Scripts/Utility/AssetInjection/MeshReplacement.cs
+++ b/Assets/Scripts/Utility/AssetInjection/MeshReplacement.cs
@@ -37,8 +37,8 @@ namespace DaggerfallWorkshop.Utility.AssetInjection
         static Func<Color32> getTreeColorCallback = () => Color.Lerp(Color.white, Color.grey, Random.value);
         static Action<Terrain> setTreesSettingsCallback = SetTreesSettings;
 
-        static HashSet<Vector2Int> triedBillboards = new HashSet<Vector2Int>();
-        static HashSet<uint> triedModels = new HashSet<uint>();
+        static readonly HashSet<Vector2Int> triedBillboards = new HashSet<Vector2Int>();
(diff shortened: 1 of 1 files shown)


Interkarma:
Just checking if this one is ready to go. I note it's still in draft mode and wasn't sure if it was ready for me yet. :)

rossipaolo:
@Interkarma Sorry for late reply, I'll test again and confirm is ready as soon as i can.

Interkarma:
Thank you :)

rossipaolo:
@Interkarma  It's ready for review now.

Interkarma:
Cheers! Will aim to get this merged for next release.

Interkarma:
Thank you @rossipaolo! Happy to finally merge this one.

PR #2045, opened 2021-02-17
Fix mod asset loading on turkish culture

rossipaolo:
[Win 10 - When I set "region format" to Turkish (Turkey) 90% of the DREAM mod gets compromised](https://forums.dfworkshop.net/viewtopic.php?f=31&t=4526&p=52693#p52693)


Interkarma:
Thanks for fixing, TheLacus!

Interkarma:
Thank you for this. I'll need to make the same change in our implementation of softsynth as well.

PR #2047, opened 2021-02-21
Fixed misnamed formula overrides

KABoissonneault:
Fixed misnamed formula overrides for CalculateFatigueRecoveryRate and CalculateSpellPointsRecoveryRate

Every other formula in that file has an override named after the function name. These two seem to be a copy-paste error. The function signature is not even consistent, so it wouldn't work as intentional design.


ajrb:
It's a copy paste derp. Thanks for spotting it,

PR #2049, opened 2021-02-24
GameObjectHelper: Set static flags while tagging Static Geometry

XJDHDR:
In addition to adding the staticGeometryTag to any GameObject run through the TagStaticGeometry function, this PR makes it also enable all of the [StaticEditorFlags](https://docs.unity3d.com/Manual/StaticObjects.html) for that object. This might help with performance.


Interkarma:
Scene objects can't become static at runtime (this is editor only) and it has no effect on performance. This is actually how things were until I realised it didn't work, and why the static geometry tag was added to identify them.

https://github.com/Interkarma/daggerfall-unity/commit/11ef17793f9631ebb4918b0c67ef101cfe50f442

I removed the line to set `go.isStatic = true` in above commit as it has no effect. :)

PR #2050, opened 2021-02-24
Adding formula overrides for spell effect cost formulas

KABoissonneault:
This pull request is for overriding CalculateTotalEffectCosts, CalculateEffectCosts, ApplyTargetCostMultiplier. Since C#'s `System.Func` does not handle `out` parameters, I've replaced the gold cost and spell point cost out parameters with a single return type containing both values, a struct inside `FormulaHelper` called `SpellCost`. I thought it would be preferable to having a generic `Tuple<int, int>`. Tell me if there's a better place to put this structure.

The branch has no impact on gameplay, but allows mods to change how spell effect costs are calculated. 
Example: https://github.com/KABoissonneault/DFU-BalancedSpellCosts/blob/main/Scripts/BalancedSpellCostsMod.cs#L41
This mod keeps the same overall formula, but removes the truncating in the "Per-level" cost calculate and in the average magnitude calculation

In the future, I plan to use this to plug new spell effect settings inside the formula to change spell costs without having to replace all the spell effects.

### DIFF
--- a/Assets/Scripts/Game/Entities/DaggerfallEntity.cs
+++ b/Assets/Scripts/Game/Entities/DaggerfallEntity.cs
@@ -743,8 +743,7 @@ namespace DaggerfallWorkshop.Game.Entity
             List<EffectBundleSettings> sortedSpellbook = spellbook
                 .OrderBy((EffectBundleSettings spell) =>
                 {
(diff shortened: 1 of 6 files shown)


Interkarma:
Awesome work! Thank you. :)

PR #2051, opened 2021-02-24
Automap: Move Mesh and Material update to separate script

XJDHDR:
This moves the contents of the InjectMeshAndMaterialProperties and UpdateMaterialsOfMeshRenderer functions into the AutomapModel script. This script is attached to every model that is used to construct the Automap. Then, instead of the Automap script iterating through the Automap tree to run these function on every model it finds, it raises an event that causes each model to run it's copy of the AutomapModel script functions.

I also made one or two optimisations to the existing code, resulting in a 55-60% reduction in the time it takes. Here are some numbers I got for the time taken for the InjectMeshAndMaterialProperties loop/event to run on loading or entering these locations on my desktop PC. For some reason, the function appears to run twice for dungeons and once for castles. I included both times for the dungeons:

Current code:
- Daggerfall Castle:
	- 543.5694ms
- Privateers Hold:
	- 161.139ms and 160.9252ms
- Castle Buckington:
 	- 351.8485ms and 348.7866ms
	
My PR:
- Daggerfall Castle:
	- 226.1401ms
- Privateers Hold:
	- 65.6735ms and 66.3357ms
- Castle Buckington:
	- 146.2152ms and 143.3541ms

I'm not sure if you would be interested in this PR. I was working on this while doing some experiments with multithreading Automap creation. Unfortunately, Unity has blocked the ability to run any part of these functions in anything but the main thread. Maybe what did work will be useful though? I think this event format makes the code look a bit neater.

### DIFF
--- a/Assets/Scripts/Game/Automap.cs
+++ b/Assets/Scripts/Game/Automap.cs
@@ -1296,115 +1296,6 @@ namespace DaggerfallWorkshop.Game
             Shader.SetGlobalFloat("_SclicingPositionY", slicingPositionY);
         }
 
-        private void UpdateMaterialsOfMeshRenderer(MeshRenderer meshRenderer, bool visitedInThisEntering = false)
-        {
-            Vector3 playerAdvancedPos = gameObjectPlayerAdvanced.transform.position;
-            Material[] newMaterials = new Material[meshRenderer.materials.Length];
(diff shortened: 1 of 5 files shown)


Interkarma:
This is excellent! I like the use of events to drive updating automap properties and optimisations are always a bonus. Great work. :)

PR #2053, opened 2021-02-25
Fix issues where dungeon Automap models are created using the wrong texture table

### LINKED ISSUE #1990: Various issues in Runtime Materials

XJDHDR:
I noticed some issues with the way dungeon Automap models are created which caused them to use the wrong texture tables when creating their materials. Both the vanilla models and RuntimeMaterials had this problem.

For the vanilla models, the script was passing a null reference for the texture table, causing the CreateBaseGameObject method to just use the default table. For RuntimeMaterials, the script was getting the texture table from an event that never fired.

It has been reported in the forum [here](https://forums.dfworkshop.net/viewtopic.php?f=31&t=4412).

If merged, this PR will fix the second problem mentioned in issue #1990.

@TheLacus - Maybe you should review the RuntimeMaterials portion of this PR?

Edit: Merging this PR will fix #1990 as this is the last part of that issue that needs to be fixed.

### DIFF
--- a/Assets/Scripts/Game/Automap.cs
+++ b/Assets/Scripts/Game/Automap.cs
@@ -2041,6 +2041,8 @@ namespace DaggerfallWorkshop.Game
 
                         DFBlock blockData;
                         int[] textureTable = null;
+                        if (GameManager.Instance.PlayerEnterExit.Dungeon != null)
+                            textureTable = GameManager.Instance.PlayerEnterExit.Dungeon.DungeonTextureTable;
                         GameObject gameobjectBlock = RDBLayout.CreateBaseGameObject(block.BlockName, null, out blockData, textureTable, true, null, false);
                         gameobjectBlock.transform.position = new Vector3(block.X * RDBLayout.RDBSide, 0, block.Z * RDBLayout.RDBSide);
 
--- a/Assets/Scripts/Game/Automap.cs
+++ b/Assets/Scripts/Game/Automap.cs
@@ -2047,9 +2047,7 @@ namespace DaggerfallWorkshop.Game
                         }
 
                         DFBlock blockData;
-                        int[] textureTable = null;
-                        if (GameManager.Instance.PlayerEnterExit.Dungeon != null)
-                            textureTable = GameManager.Instance.PlayerEnterExit.Dungeon.DungeonTextureTable;
+                        int[] textureTable = GameManager.Instance.PlayerEnterExit?.Dungeon?.DungeonTextureTable;
 
                         Automap.isCreatingDungeonAutomapBaseGameObjects = true;
                         GameObject gameobjectBlock = RDBLayout.CreateBaseGameObject(block.BlockName, null, out blockData, textureTable, true, null, false);
--- a/Assets/Scripts/Game/Automap.cs
+++ b/Assets/Scripts/Game/Automap.cs
@@ -269,6 +269,8 @@ namespace DaggerfallWorkshop.Game
 
         bool iTweenCameraAnimationIsRunning = false; // indicates if iTween camera animation is running (i.e. teleporter portal jump animation plays)
 
+        static bool isCreatingDungeonAutomapBaseGameObjects = false; // Indicates that the Automap is creating the models for a Castle's or Dungeon's Automap
+
         #endregion
 
         #region Properties
@@ -359,6 +361,11 @@ namespace DaggerfallWorkshop.Game
         public bool ITweenCameraAnimationIsRunning
         {
             get { return iTweenCameraAnimationIsRunning; }
+        }
+
+        public static bool IsCreatingDungeonAutomapBaseGameObjects
+        {
+            get { return isCreatingDungeonAutomapBaseGameObjects; }
         }
 
         #endregion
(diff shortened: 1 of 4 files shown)


XJDHDR:
Here are some screenshots of this PR in action. 

First, before and after images for vanilla models:
Automap textures beforeAutomap textures after


Next, before and after images for Runtime Materials:
Runtime Materials beforeRuntime Materials after

Interkarma:
Thank you so much for this fix. It's wonderful to see the correct texture layout in automap at last. 👍

XJDHDR:
You're welcome.

Interkarma:
I just fixed a minor conflict with other PR and will merge this one now.

Interkarma:
I just noticed - after merging #2051 and #2053 the automap models are now extra bright. This didn't happen when each PR was tested in isolation (or at least, I didn't notice it when testing). Do you know what might be causing this?

image

Here's how it looked before changes.

image

XJDHDR:
There's two possibilities I can think of. My optimisation in 2051 involved creating the Automap material once and then Instantiating it each time another one was needed, whereas the previous code just created a new Material each time. It's possible that Instantiating creates some differences in the material verses creating it from scratch.

The other is some strange incompatibility between those PRs. I'll need to check a few things.

XJDHDR:
It looks like it is this PR (2053) that is causing the problem. 2051 is fine.

Interkarma:
Thank you. My apologies if conflict edit created this. It seemed simple enough - the only conflict was a couple of overlapping lines in Automap.cs starting from line 1941. The rest came straight from master, which already had a few other PRs merged, including 2051.

https://github.com/Interkarma/daggerfall-unity/pull/2053/commits/d199de3dbacce019f0bdf773278617cb378a1eaa

XJDHDR:
I just re-created this PR basing it off the code as of [Deprioritise System Locale Selector ](https://github.com/Interkarma/daggerfall-unity/commit/dea6c1a4e668d80ba015ee53b6d1398ae83a7795) and it works fine. It seems like something added after that point is interfering here.

Interkarma:
Found the problem. 2051 changed the following call to add a "true" as final param.

`GameObject gameobjectBlock = RDBLayout.CreateBaseGameObject(block.BlockName, null, out blockData, textureTable, true, null, false, true);`

When I resolved conflict, I didn't keep the correct version of call. Have fixed this now.

My apologies for that error and thank you for your help isolating. That was a somewhat subtle difference I should have spotted. :)

XJDHDR:
No problem and now it makes sense. 
Since the final boolean wasn't being set, the default value of False was used. This boolean ([isAutomapRun](https://github.com/Interkarma/daggerfall-unity/blob/master/Assets/Scripts/Utility/RDBLayout.cs#L101)) being false meant the [AutomapModel script wasn't added](https://github.com/Interkarma/daggerfall-unity/blob/master/Assets/Scripts/Utility/RDBLayout.cs#L726) to the Automap models which meant the Automap shader wasn't applied to those models.

PR #2054, opened 2021-02-26
Fixed torch burning ticks progressing while the game is paused

KABoissonneault:
Every frame, if the game is not paused, the torch checks if more than 20 seconds has elapsed since the last tick, and if so, reduces condition. This check is based on real time­. Therefore, if a tick occurs and the player pauses immediately for at least 20 seconds, once they come back to gameplay, the torch will burn again immediately. No game mechanic should progress while the game is paused.

This implementation instead uses a time buffer that increases in time only while the game is not paused (and the torch is active of course). One potential issue is that the timer does not reset when unequipping the torch, but it might be a good thing, since it prevents "exploits" based on unequipping the torch every 19 seconds.

### DIFF
--- a/Assets/Scripts/Game/EnablePlayerTorch.cs
+++ b/Assets/Scripts/Game/EnablePlayerTorch.cs
@@ -29,7 +29,7 @@ namespace DaggerfallWorkshop.Game
         Light torchLight;
         float torchIntensity;
         float intensityMod = 1f;
-        float lastTickTime;
+        float tickTimeBuffer = 0f;
         float tickTimeInterval = 20f;
         float guttering = 0;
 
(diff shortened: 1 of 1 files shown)


ajrb:
I don't agree that there's a bug here that needs to be fixed. The Update() method checks the game is not paused and if it is it simply returns early so I don't see how this bug you're describing could occur. 

How did you prove that this was happening?

KABoissonneault:
Copying the message I sent you
"Read it again. Yes, the "tick" check is not done while paused, but realtimeSinceStartup still goes up. This means that if you pause right after a tick, once you unpause, you'll get another tick immediately. In the worst case, a player might walk for 3 seconds, look at their map for 17 seconds, and burn a torch tick every time they repeat this. With my fix, the torch only burns for 20 seconds of gameplay, not 20 seconds of realtime"

I'll post repro steps just so we're clear:
1. Equip a torch with 1 hp
2. Do 1 second of gameplay
3. Pause the game in any way (Inventory, Escape, debugging breakpoint, etc...) for at least 20 seconds
4. Go back to gameplay

Observed behavior: The torch will extinguish immediately
Expected behavior: A torch with a single condition point should extinguish after 20 seconds of gameplay

ajrb:
Looks good, thanks. Wanted to wait until my PC was fully fixed before I worked through this one.

Interkarma:
Good catch, thank you. 👍

PR #2055, opened 2021-03-02
Fixed date lines in journal entries of quests

deepfighter:
In some quests the text of the journal entry starts directly after the variable _%qdt:_. 

I changed those entries and reformatted them, so that now all the text is on a new line, which is the preferred way, as seen in all the other instances.

(btw. first pull request (ever)! :))


Interkarma:
This is good work, thank you @deepfighter :)

PR #2057, opened 2021-03-05
Fix several issues with conjured arrows

petchema:
- don't stack conjured arrows with different TTLs;
- shoot conjured arrows first, starting with the ones with lowest TTL;
- change color of arrow counter to tell whether you're about to shoot a conjured or real arrow;
- prevent looting conjured arrows (currently they're converted to real arrows in enemies inventory).

Forums: https://forums.dfworkshop.net/viewtopic.php?f=5&t=4629

### DIFF
--- a/Assets/Scripts/Game/DaggerfallMissile.cs
+++ b/Assets/Scripts/Game/DaggerfallMissile.cs
@@ -83,6 +83,7 @@ namespace DaggerfallWorkshop.Game
         float initialIntensity;
         EntityEffectBundle payload;
         bool isArrow = false;
+        bool isArrowSummoned = false;
         GameObject goModel = null;
         EnemySenses enemySenses;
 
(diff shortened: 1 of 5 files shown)


Interkarma:
I do like this change. Thank you! :)

What are your thoughts about my question in forum post?

> I wonder if rather than multiple stack of arrows with different TTLs, if the conjured arrows should instead always stack with each other and reset the stack TTL with latest cast. This saves the inventory being cluttered with multiple arrow stacks, and user is just "topping up" their stack occasionally.

This feels like it would make the recast behaviour work more in line with how users top up other spells, and avoid having multiple stacks of conjured arrows with different TTLs. It would also remove the need to choose lowest TTL first and just select conjured arrow stack first.

petchema:
Forgot about that one, answered in the forums

PR #2058, opened 2021-03-06
Fixed spells with multiple effects ending too early

KABoissonneault:
If a spell has multiple effects, the entity effect manager will remove its bundle when the last effect expires, rather than when all effects expire.

To reproduce:
1. Create a spell with at least two effects. Make the first effect last at least 2 rounds, and make the last effect last 1 round (instant counts as 1)
2. Cast spell
3. Wait 1 round
4. See spell being immediately removed, even though the first effect still had another round

The bug still occurs even if the first effect has 10+ rounds remaining.

I'm not sure if this is a classic issue, but even if it is, it's worth fixing in my opinion. The bug can be worked around by having the longest effect be the last in the list.


Interkarma:
This is a clever and straightforward fix for that problem. Thank you!

PR #2059, opened 2021-03-06
Toggle large HUD with F10 like classic

petchema:
Minimum implementation.

By the way, while implementing this, I noticed what looks like an unrelated bug: https://forums.dfworkshop.net/viewtopic.php?f=5&t=4630


Interkarma:
I don't believe I ever knew F10 did this in classic. Thank you for catch and fixing it. :)

PR #2060, opened 2021-03-07
Only interact with large HUD when cursor active on Linux

petchema:
Doesn't prevent mouse cursor from moving while cursor is disabled though

Forums: https://forums.dfworkshop.net/viewtopic.php?f=31&t=4630


Interkarma:
This looks good, thank you! :)

PR #2061, opened 2021-03-11
Prevent detecting accidental axis button "presses" from movement and camera axes

jefetienne:
If the player accidentally moves their movement or camera axis in a cutscene sequence, it will recognize it as a button press in the direction it went and skip the video, since it was looking for "any key down". Similarly, a player might accidentally bind either axis while the control window is waiting for a key press (especially when trying to bind the controller buttons commonly known as L3 and R3, where you press down on the joysticks.) Those axis buttons are not needed for any practical reason since they are already in use for moving the character and camera.

These changes will simply ignore those detected axis buttons if they are a part of the movement or camera axes. I also added two new boolean property getters `AnyKey(Down/Up)IgnoreAxisBinds` and four new public `KeyCode` methods to return the specific key of the `AnyKey` methods. Some of these might be unused in the codebase, but they can be helpful for modding.

### DIFF
--- a/Assets/Scripts/Game/InputManager.cs
+++ b/Assets/Scripts/Game/InputManager.cs
@@ -930,6 +930,17 @@ namespace DaggerfallWorkshop.Game
             axisActionInvertDict[(int)axis] = invert;
         }
 
+        public bool IsUsedInAxisBinding(KeyCode code)
+        {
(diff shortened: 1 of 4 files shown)


Interkarma:
Great fix, thank so much! :)

PR #2064, opened 2021-03-18
Fixed some editor warnings

XJDHDR:
This fixes three of the warnings that appear in the editor every time scripts are re-compiled by:

- Adding "new" keyword to ViewportChanger's Camera variable.
- Deleting line in DaggerfallTerrain that sets materialType as materialTemplate is now set directly.

The warnings related to WWW being obsolete are still present because, [as I discovered](https://forums.dfworkshop.net/viewtopic.php?p=52520#p52520), UnityWebRequest is not a straight drop-in replacement for DFU's current code.


Interkarma:
Thank you. Nice to knock over a couple more of these.

PR #2065, opened 2021-03-19
Added the ability for mods to add new Biography templates

KABoissonneault:
I felt like investigating the biography and History system in Daggerfall, and noticed that it wouldn't be too hard for DFU to give mods the ability to introduce alternative biographies for character.

With these changes, after choosing a class (ex: Mage aka 0), the game looks for all files that match BIOG00T([0-9]+).txt in the BIOG source folder. If multiple files are found, one is selected at random (note: could be extended later to offer a list to choose from).

In addition, an extension was added to the BIOG file format. Before the first question, if a line starts with #, the rest of the line will be interpreted as a string id, and will be used for the backstory. With a custom ITextProvider, mods can make use of unused string values to provide new backstory text.

For example, if a mod adds a BIOG00T1.txt file, when a player creates a Mage (or similar custom class), the game will randomly pick between BIOG00T0 and BIOG00T1, with possibility for a new set of questions and/or a different backstory.

I've asked around, and it seems like it would be a popular feature.

### DIFF
--- a/Assets/Scripts/API/BiogFile.cs
+++ b/Assets/Scripts/API/BiogFile.cs
@@ -27,6 +27,7 @@ namespace DaggerfallConnect.Arena2
     {
         const int questionCount = 12;
         const int socialGroupCount = 5;
+        const int defaultBackstoriesStart = 4116;
 
         // Folder names constants
         const string biogSourceFolderName = "BIOGs";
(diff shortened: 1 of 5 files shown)


Interkarma:
I'm happy to merge this, will be interesting to see where it goes. Cheers!

PR #2067, opened 2021-03-21
I found a bug in the code that I had submitted as part of PR 2056.

AsesinoBlade:
I found an issue in the code that I had submitted in PR # 2056.

In DaggerfallUI, I check for uiManager.TopWindow.FocusControl not to be null but I just hit a situation where uiManager.TopWindow was null so this failed on nullException.
I added a check for TopWindow as well now to correct that issue.
Sorry for this, the issue.

Thanks 
A

### DIFF
--- a/Assets/Scripts/Game/DaggerfallUI.cs
+++ b/Assets/Scripts/Game/DaggerfallUI.cs
@@ -393,7 +393,7 @@ namespace DaggerfallWorkshop.Game
                 processHotkeys = false;
                 HotkeySequenceProcessed = ProcessHotKeySequences();
             }
-            else if (uiManager.TopWindow.FocusControl != null && uiManager.TopWindow.FocusControl.OverridesHotkeySequences)
+            else if (uiManager?.TopWindow?.FocusControl?.OverridesHotkeySequences == true)
                 HotkeySequenceProcessed = HotkeySequence.HotkeySequenceProcessStatus.Disabled;
             else
                 HotkeySequenceProcessed = HotkeySequence.HotkeySequenceProcessStatus.NotFound;


Assets/Scripts/Game/DaggerfallUI.cs
```diff
@@ -393,7 +393,7 @@ void OnGUI()
                 processHotkeys = false;
                 HotkeySequenceProcessed = ProcessHotKeySequences();
             }
-            else if (uiManager.TopWindow.FocusControl != null && uiManager.TopWindow.FocusControl.OverridesHotkeySequences)
+            else if (uiManager.TopWindow != null && uiManager.TopWindow.FocusControl != null && uiManager.TopWindow.FocusControl.OverridesHotkeySequences)
```
XJDHDR, line 396: As a suggestion, you could shorten this to the following that does the same thing:
```c#
else if (uiManager.TopWindow?.FocusControl?.OverridesHotkeySequences)
```

AsesinoBlade: Thank you for looking at this XJDHDR.  I had thought that I could do it that way but the compiler complained that I was converting from Bool? to Bool and I didn't want to take any chances.    I need to read up to see if I can safely do that with an explicit conversion.

Thanks
A

XJDHDR: I'm assuming you wrote `OverridesHotkeySequences?`. If so, that's because a bool is either `true` or `false`. I don't think it can be equal to `null`, so the backend wasn't designed to handle attaching a null conditional operator to one.

AsesinoBlade: Ah I see what you are saying.   So if I had used ?? Instead, it would have returned a bool but since I used ? It returns a potential null.  I just haven’t played around enough with ?? And ? To be confident on how to apply them.

Thanks 
A

XJDHDR: I think I misunderstood things there. That would work but I like the "add `== true`" answer more, like so:
```c#
else if (uiManager.TopWindow?.FocusControl?.OverridesHotkeySequences == true)
```

And you're welcome.

AsesinoBlade: Thank you so much. That code is much cleaner, and I learned something new. I really appreciate the help.

thanks
A

XJDHDR: You're welcome.

Interkarma:
Thanks for the fix! I'll get this one merged now as I'm about to spin up new builds.

@XJDHDR Thank you for the assist. :)

PR #2068, opened 2021-03-25
Change journal UI window permissions to allow overriding by mods

ajrb:


Interkarma:
Thank you mate, all good with these changes.

PR #2069, opened 2021-03-26
ActionDoor: Add support for MeshCollider

XJDHDR:
This PR adds the option of using a MeshCollider to provide a door's collision instead of a BoxCollider. To do so, you have to disable the door's BoxCollider and add a MeshCollider.

This allows modders to design doors that are not box shaped yet have accurate collision. In fact, there are at least 2 such doors in the base game (specifically, 55007 and 55008).


Interkarma:
Thank you. :)

PR #2070, opened 2021-03-26
Bugfix for PR 2053

XJDHDR:
This fixes a bug in PR #2053 where Runtime Materials only detected that it was being created for a dungeon's or castle's interior if attached to a Base Game Object and not if attached to anything else that is inside a dungeon, such as Action Doors. This meant that modded secret doors received the wrong textures. The script will now correctly detect a dungeon/castle interior creation if attached to other objects.

To test this PR, you can download my mod [Unofficial Block Location and Model fixes](https://www.nexusmods.com/daggerfallunity/mods/100?tab=files) then load the attached save game. Without this PR, the secret door in front of you will have the wrong texture. With this PR, the texture will be correct.
[Secret door test.zip](https://github.com/Interkarma/daggerfall-unity/files/6209795/Secret.door.test.zip)

### DIFF
--- a/Assets/Scripts/Game/PlayerEnterExit.cs
+++ b/Assets/Scripts/Game/PlayerEnterExit.cs
@@ -35,7 +35,7 @@ namespace DaggerfallWorkshop.Game
         UnderwaterFog underwaterFog;
         DaggerfallUnity dfUnity;
         CharacterController controller;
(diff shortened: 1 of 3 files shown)


Interkarma:
Thank you for fix! Will merge this for next release.

PR #2075, opened 2021-04-10
Fix regression in PR 2051

XJDHDR:
When I moved the InjectMeshAndMaterialProperties and UpdateMaterialsOfMeshRenderer functions into a separate script, the intention was for each of these scripts to do the same thing as the previous code. However, PR #2051 has a regression in it where this script is only added to vanilla models generated from Arch3D.bsa and is not applied to mod added models. This PR fixes this so that mod added models will now also receive the script and have the Automap shader and textures applied.

At the same time, I also removed some comments I accidentally left in RDBLayout in one of my other PRs. Unfortunately, Unity does not allow manipulating a Game Object's Transform from anything except the main thread.


Interkarma:
Thanks for fixes, again. :)

PR #2078, opened 2021-04-13
Fixed issues while parsing Deepfighter's BIOG files

KABoissonneault:
1) Some lines (mostly the ones with IT) use tabs instead of spaces to split token­. This caused BiogFile.cs to fail parsing those lines, failing to give the character the intended item. BiogFile.cs now splits tokens with any whitespace
2) Almost all BIOG files had the same copy-pasted "&" line, which caused BiogFile.cs to throw an error on invalid command. Removed that line from all BIOG files, as this was likely not intended

Repro case:
1. Start DFU without any mods
2. Start New Game
3. Create a Warrior character, and answer "Axe fighting" to the first question. The rest can be answered randomly
4. Open Inventory

Expected:
In BIOG16T0.txt, "Axe fighting" to question 1 should give you IT 3 15 0, aka Iron War-axe

Actual result:
No axe in inventory

The error logs can be seen when creating the character from the Unity Editor

### DIFF
--- a/Assets/Scripts/API/BiogFile.cs
+++ b/Assets/Scripts/API/BiogFile.cs
@@ -223,7 +223,7 @@ namespace DaggerfallConnect.Arena2
 
         private static void ApplyPlayerEffect(PlayerEntity playerEntity, string effect)
         {
-            string[] tokens = effect.Split(' ');
+            string[] tokens = effect.Split(null);
             int parseResult;
 
             // Skill modifier effect
--- a/Assets/StreamingAssets/BIOGs/BIOG01T0.TXT
+++ b/Assets/StreamingAssets/BIOGs/BIOG01T0.TXT
@@ -215,7 +215,6 @@ a.	Resisting poisons
 b.	Staying awake and alert
 	FT -5
 c.	Getting along with others
-	&
 	RR -5
 d.	Avoiding diseases
 	RD -5
--- a/Assets/StreamingAssets/BIOGs/BIOG02T0.TXT
+++ b/Assets/StreamingAssets/BIOGs/BIOG02T0.TXT
@@ -200,7 +200,6 @@ a.	Resisting poisons
 b.	Staying awake and alert
 	FT -5
 c.	Getting along with others
-	&
 	RR -5
 d.	Avoiding diseases
 	RD -5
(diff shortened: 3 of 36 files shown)


deepfighter:
Thank you for finding and fixing! :) I just want to clarify that both of your points are already in the original BIOG files. I haven't changed anything in this respect. I also can't see any other tabs than those at the beginning of each line to bring every line "in line" (woah! :-D). Any they are always there, not only on lines with IT. Do you mean those or do I miss anything else?

The same goes for the "&", which I kept in the files, as I thought this was intended and needed to get the line working. As this specific answer is not visible while playing (like getting an item), but instead changes an value of the game mechanic in the background, I couldn't investigate further, as this (honestly) exceeds my knowledge. Good that you checked on that! :) With that knowledge, I will also update my fix for Classic accordingly, as it seems that the & is just another error of the original game.

KABoissonneault:
>I also can't see any other tabs than those at the beginning of each line to bring every line "in line"

It's hard to see because in every case where it happens, the tab has the width of 1 space. But here's an example for BIOG15T0.txt, where I replaced every tab that was causing a parser issue with "\t"

```
1.	You are most fond of using ___.
a.	Short-bladed weapons
	#4200
	!4701
	28 +12
	IT 3 3 0
b.	Hand-to-hand combat
	#4201
	!4702
	30 +12
	IT 2 1 0
c.	Axe fighting
	#4202
	!4703
	31 +12
	IT 3 15\t0
d.	Blunt weapons
	#4203
	!4704
	32 +12
	IT 3 13\t0
e.	Archery
	#4204
	!4705
	33 +12
	IT 3 16\t0
	IT 3 18\t1
	IT 3 18\t1
	IT 3 18\t2
	IT 3 18 1
	IT 3 18 1
	IT 3 18 1
f.	Longblade
	#4205
	!4706
	29 +12
	IT 3 9 0
```

You can see that it affects quite a few lines, and that's just one question from one BIOG.

Perhaps those were always there in the BIOG files and DFU always had this regression, who knows. I'm pretty sure your files aren't an issue in classic DF though.

Interkarma:
Thank you for fixes! :) I must admit, this is an area my understanding is lacking also. I hadn't noticed that exception before.

PR #2079, opened 2021-04-14
Do not draw TextBox defaultText if null

rossipaolo:


Interkarma:
Always happy to have a null check. :)

PR #2083, opened 2021-04-19
Update PlayerSpeedChanger

l3lessed:
- PlayerSpeedChanger: Added properties to add sequential speed modifiers for walk and run speed. These can be accessed and removed using the AddRunSpeed, AddWalkSpeed, RemoveSpeed properties. Individual speed modifications are stored using the output UID string for the Add properties.
- Console command updated to allow addition of speed modifications. Will output the custom UID string value into the console for player to reference when removing speed modification.
- Changed one line of code to reference the proper property for grabbing updated runspeed from PlayerSpeedChanger.
- Examples of how this works in actions can be found here: https://forums.dfworkshop.net/viewtopic.php?f=23&t=4633


l3lessed:
Accidently added older commits

PR #2084, opened 2021-04-19
SpeedChanger Updates

l3lessed:
Developers can now hook into and update player walk and run speed using the player speed changer. This is done by sequentially adding the change values to a dictionary and tying them to a UID out string property. The developer stores the UID out so they can remove/manipulate their speed changes at a later time in code execution.

**Currently we have these public properties:**

- Bool walkSpeedOverride: turns on and off using walk modifier list to change player movement speed. Does not clear list, merely ignores it.
- Bool runSpeedOverride: turns on and off using run modifier list to change player movement speed. Does not clear list, merely ignores it.
- Bool updateWalkSpeed: turn true to recompute the walk speed; will flip false once done on next frame.
- Bool updateRunSpeed: turn true to recompute the run speed; will flip false once done on next frame.

**Currently we have these public routines:**

- Bool AddWalkSpeedMod(out string UID, float walkSpeedModifier = 0, bool refreshWalkSpeed = true): Add a percentage multiplier to the walk speed modifier list. Returns true if the unique ID is output as an out string value to refer to/manipulate modifier. Returns false if it couldn't add it.
- Bool AddRunSpeedMod(out string UID, float speedModifier = 0, bool refreshRunSpeed = true): Add a percentage multiplier to the runspeed modifier list. Returns true if the unique ID is output as an out string value to refer to/manipulate modifier. Returns false if it couldn't add it.
- Bool RemoveSpeedMod(string UID, bool removeRunSpeed = false, bool refresSpeed = true): Remove a speed modifier using previously generated unique ID string value. Allows you to select if your removing running or walking speed modifiers. Returns true or false based on if modifier was removed or not.
- Bool ResetSpeed(bool walkSpeedReset = true, bool runSpeedReset = true): Remove a speed modifier using previously generated unique ID string value. Allows you to select if your removing running or walking speed modifiers. Returns true or false based on if modifier was removed or not.
- Float RefreshWalkSpeed(): Updates the players walk speed using for loop and dictionary values to ensure proper sequential processing to get proper end speed and returns updated walk speed value. Processing of modifiers is processed by their addition order. First added by modder is multiplied first, and so on.
- Float RefreshRunSpeed(): Updates the players run speed using for loop and dictionary values to ensure proper sequential processing to get proper end speed and returns updated run speed value. Processing of modifiers is processed by their addition order. First added by modder is multiplied first, and so on.

**Current console command settings are. These apply the same to set_walkspeed console command:**

- Set_runspeed -1: Enable/Disable run speed sequential modifier. Does not clear modifier list, just ignores it and returns base run speed. Returns current run speed value based on if sequential modifier is enabled or not.
- Set_runspeed -2: Run speed modifiers cleared. Clears out current run speed modifier list. Returns run speed value after all modifers have been removed from modifier list.
- Set_runspeed #: Adds percentage multiplier to run speed modifier list for sequential calculation; tells user UID for added modifier so they can remove it without clearing list if they want. Users cannot insert negative numbers, other than -1 and -2. Returns run speed after addition of modifer value.
- Set_runspeed UID: Checks to see if dictionary modifier list contains UID modifer. If it does, will remove the added modifier. If not will return normal error message saying invalid input. Returns run speed after removal of modifier.
- Set_runspeed : Returns basic error and current updated run speed.

### DIFF
--- a/Assets/Scripts/Game/Player/PlayerSpeedChanger.cs
+++ b/Assets/Scripts/Game/Player/PlayerSpeedChanger.cs
@@ -34,11 +34,11 @@ namespace DaggerfallWorkshop.Game
         private const float dfCartBase = dfWalkBase + 100f;
 
         public bool walkSpeedOverride = true;
-        public float currentWalkSpeed { get; private set; } = 0;
+        private float currentWalkSpeed = 0;
         private Dictionary<string, float> walkSpeedModifierList = new Dictionary<string, float>();
 
         public bool runSpeedOverride = true;
-        public float currentRunSpeed { get; private set; } = 0;
+        private float currentRunSpeed = 0;
         private Dictionary<string, float> runSpeedModifierList = new Dictionary<string, float>();
 
         public delegate bool CanPlayerRun();
@@ -115,7 +115,7 @@ namespace DaggerfallWorkshop.Game
 
             if (isRunning)
             {
-                speed = RefreshRunspeed();
+                speed = RefreshRunSpeed();
 
                 //switch sneaking off if was previously sneaking
                 sneakingMode = false;
--- a/Assets/Scripts/Game/Player/PlayerSpeedChanger.cs
+++ b/Assets/Scripts/Game/Player/PlayerSpeedChanger.cs
@@ -281,7 +281,7 @@ namespace DaggerfallWorkshop.Game
         /// Updates the players walk speed using for loop and dictionary values to ensure proper sequential processing to get proper end speed.
         /// Processing of modifiers is processed by their addition order. First added by modder is multiplied first, and so on.
         /// </summary>
-        private float RefreshWalkSpeed()
+        public float RefreshWalkSpeed()
         {
             //setup and grab needed base values for computing end speed.
             float baseWalkSpeed = GetWalkSpeed(GameManager.Instance.PlayerEntity);
@@ -331,7 +331,7 @@ namespace DaggerfallWorkshop.Game
         /// Updates the players walk speed using for loop and dictionary values to ensure proper sequential processing to get proper end speed.
         /// Processing of modifiers is processed by their addition order. First added by modder is multiplied first, and so on.
         /// </summary>
-        private float RefreshRunspeed()
+        public float RefreshRunSpeed()
         {
             //setup and grab needed base values for computing end speed.
             float baseRunSpeed = GetRunSpeed();
--- a/Assets/Scripts/Game/Player/PlayerSpeedChanger.cs
+++ b/Assets/Scripts/Game/Player/PlayerSpeedChanger.cs
@@ -53,7 +53,6 @@ namespace DaggerfallWorkshop.Game
         public bool updateRunSpeed;
         private float previousBaseWalkSpeed;
         private float previousBaseRunSpeed;
-        private float baseSpeed = 0;
 
         private void Start()
         {
@@ -143,6 +142,7 @@ namespace DaggerfallWorkshop.Game
         public float GetBaseSpeed()
         {
             Entity.PlayerEntity player = GameManager.Instance.PlayerEntity;
+            float baseSpeed = 0;
             float playerSpeed = player.Stats.LiveSpeed;
             if (playerMotor == null) // fixes null reference bug.
                 playerMotor = GameManager.Instance.PlayerMotor;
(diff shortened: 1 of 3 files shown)


l3lessed:
Sorry about that. I updated my df build to the newest release without a compile and test. I'll fix and update the request.

l3lessed:
Submitted the fix. I'll keep doing testing to ensure I didn't overlook any bugs, but as of last tests, I didn't see any issues cropping up; it seems to be working as intended.

Interkarma:
Thank you! I'll loop back and review again soon. :)

l3lessed:
- Fixed basespeed error accidently reintroduced.
- Cleaned up console output to ensure it clearly explains how console commands work and provides UID modification property in console output so users can also add and remove speed changes straight through console for testing purposes.

l3lessed:
I believe last lose coding ends are officially finished. Sorry about the sloppy submit on the front end. Codes cleaned, notes are updated and clarified, console is updated with better output notes, and speed modifiers are reset on every game load to stop any values carrying over and bugging loaded games.

Interkarma:
Thank you for fixes to PR. Working through this one now. :)

Everything so far works as described and it doesn't break anything in base game speed-wise. I can see how this would be a useful addition that allows multiple mods to assign incremental speed multipliers.

I only have a one minor feedback item so far. The debug outputs after using "set_runspeed/set_walkspeed" don't print the correct current speed values, as this is only updated fully outside of the console when player moves in world. It would be ideal to always display the current speed values in debug output. While it's nothing players see in game, it would be confusing for modders trying out different speed multiplier combos without realising they need to move in the world each time. But yeah, it's a minor thing.

This is really great work, and I can see the time and effort you've put into this. I don't have any hesitations adding this to the game. 👍

Assets/Scripts/Game/Player/PlayerSpeedChanger.cs
```diff
@@ -145,41 +155,249 @@ public float GetBaseSpeed()
...
+
+            //return the final modified walk speed.
+            return currentWalkSpeed;
+        }
+
+        /// <summary>
+        /// Updates the players walk speed using for loop and dictionary values to ensure proper sequential processing to get proper end speed.
+        /// Processing of modifiers is processed by their addition order. First added by modder is multiplied first, and so on.
+        /// </summary>
+        private float RefreshRunspeed()
```
Interkarma, line 334: Casing typo - should be RefreshRunSpeed() :)

l3lessed:
Thanks for the encouraging words. It means allot coming from such a skilled coder. Oh, me and my typos. And to think, I'm a composition major. I'll fix those recommendations and resubmit. Again, thank you for all the work you have put into DFU.

l3lessed:
Code updated. Typo fixed, made Refresh float speed properties public so proper current speed values, after full sequential processing, for run and walk could be pulled. One last question. For my interactive terrain mod, I'm thinking about trying to lerp the speed change, so as players move into snow or sand they slow down over a small amount of time to make it more immersive/realistic. Do you think this would be better to just build into the speed changer script for other modders? I'm thinking it may be better to leave this to individual modders to do on their script side, so they have more control, even though it would require more knowledge and coding on their end.

l3lessed:
Pushed small error oversite and removed currentRunSpeed and currentWalkSpeed from public to private because of how they are being used by the scripts.

Interkarma:
Thanks so much! I'll do a final check and merge shortly. Good catch on the final fix to distance threshold, I didn't catch that one.

For lerping the changes, I do agree this is better done in a mod based on its own needs. Anyone already coding up a speed changer mod are more than likely going to be capable of this, so best not to force it in core.

This is really good work. I've found no issues other than the really minor things you've now solved. Looking forward to seeing more about your mod after next release. :)

ajrb:
I just tried setting runspeed to speed up testing as I've been doing for almost 4 years and it didn't do what it used to at all. I've no idea if this is intended or not, and I've not read through this thread or the code changes to try and figure it out. (and don't plan to)

Here's a screenshot of my setting runspeed in console:
image

Interkarma:
It's now a multiplier rather than a flat value. If you used to "set_runspeed 40" to zip around town, now use "set_runspeed 4" instead. 

Each time you run the command, it adds a new multiplier. Use "set_runspeed -1" to toggle on/off and keep speed modifiers, or -2 to clear modifiers. It will also clear on load.

The idea is that mods can register their own speed changer and get back a GUID. Then use that GUID to finely control speed multiplier based on some condition (e.g. walking on snow). It also means more than one mod can influence speed multipliers without conflict or load order being an issue.

A lot of thought has gone into the changes here and I support the work. But always open to feedback though, this is a very community project. :)

PR #2085, opened 2021-04-19
Change NPC knowledge in Talk Manager to match classic

df-ferital:
Change made after having reverse engineered classic Daggerfall function used to check if an NPC is able to answer a question or not. NPCs will now use an array of modifiers corresponding to their social class and to the kind of question they are asked in order to get the roll to beat to check if they are able to answer or not.

This change also makes NPC keep their previous knowledge after closing the chat window, as in classic.


Interkarma:
Very happy to see this one. Excellent work! :)

df-ferital:
Thanks :) I added a check so NPCs belonging to a faction will always have knowledge about it. This prevents weird situations like this one:

TG_member_does_not_know_TG

Interkarma:
Thank you again for this one. I ran through several quests at random and haven't been able to identify any regressions. Everything looks to be working beautifully. I'll merge this one for next general release.

PR #2087, opened 2021-04-21
Added two event handlers for loot spawning

KABoissonneault:
There's currently many ways to modify what loot is found in enemy loot piles, and the loot piles found in dungeons and interiors.

When mods register custom items in a given category, it can replace the default loot of that category. 
Mods can replace the default loot tables found in LootTables.
Mods can use the "ModifyFoundLootItems" formula, now fixed for 0.11.3
Mods can also register to the EnemyDeath.OnEnemyDeath event handler and modify the entity's items just before it creates the loot pile.

For some situations, all these solutions have issues.

First, the loot tables in LootTables and the "ModifyFoundLootItems" formula can only be changed by one mod. If multiple mods take this approach, they will conflict.
Second, registering items in a category used by loot tables does not work if you want to add a new kind of drop entirely. For example, a hunting mod might want to add meat on beast enemies. 

For the EnemyDeath approach, it just doesn't feel right. You need to do a lot of work to access the data you need (ex: [Climate&Calories](https://github.com/Ralzar81/Climates-Calories/blob/034ca75f4a9ef9d2ebf23ff25d01d8aa6e4af1bc/Climates%20%26%20Calories/Hunting.cs#L41)). Also, unlike the rest of the loot, this is generated only when the enemy dies, and won't remain consistent if the player reloads the game and kills the same enemy.

This PR therefore adds two new event handlers for enemy loot piles and "location" piles, each with all the information a mod could need, and called as soon as loot is generated. Event handlers can be registered by multiple mods without issues.

I have an example of how this would be used [here](https://github.com/KABoissonneault/DFU-Mod_Repair-Tools/blob/feat/leveled-repairs/Scripts/RepairTools.cs#L222).

### DIFF
--- a/Assets/Scripts/Game/Entities/EnemyEntity.cs
+++ b/Assets/Scripts/Game/Entities/EnemyEntity.cs
@@ -22,11 +22,34 @@ using DaggerfallWorkshop.Game.Items;
 namespace DaggerfallWorkshop.Game.Entity
 {
     /// <summary>
+    /// The parameters involved in creating an enemy loot pile
+    /// </summary>
+    public class EnemyLootSpawnedEventArgs : System.EventArgs
+    {
+        /// <summary>
+        /// The Mobile object used for the enemy
+        /// </summary>
+        public MobileEnemy MobileEnemy { get; set; }
+
+        /// <summary>
+        /// The Career template of the enemy
+        /// </summary>
+        public DFCareer EnemyCareer { get; set; }
+
+        /// <summary>
+        /// The collection containing all the items of the loot pile. New items can be added
+        /// </summary>
+        public ItemCollection Items { get; set; }
+    }
+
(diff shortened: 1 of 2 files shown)


ajrb:
This is great as far as I am concerned, much better than what I did in adding a formula.. at the time I was after a quick way to change the condition of spawned items. Once the mod was done, I never gave it any more thought. :)

I'd like to propose that the ModifyFoundLootItems formula is removed in the release after this is released. I think it's only used by my mod, but that will give time for the mods to be updated. It was only an artificial formula anyway, so I think it should get removed.

Interkarma:
Thank you @KABoissonneault! Happy to merge this.

@ajrb Thank you for help with review. I'll get this one merged for 0.11.4, and we can remove your other changes whenever you're ready.

PR #2090, opened 2021-04-23
Add %mn and %fn macros support to TalkManagerMCP

df-ferital:
This PR fixes this:

bad_%fn_expansion

The %mn macro is not used in conversations but I added it for the sake of consistency. Note that the random seed for an NPC is not fixed for now so what I do in `MaleName` to differentiate the male name random seed from the female name one is uselesss until PR #2085 is merged.


Interkarma:
Thank you Ferital! :)

Issue #1974, opened 2020-11-28
Special door in Wayrest Castle opens when it shouldn't


wheybags:
**Describe the bug**
Special door in Wayrest Castle (the kind that shows the "Only castle staff and trusted guests are allowed beyond these doors" message when you try to open it) opens when it shouldn't.

**To Reproduce**
1. Load the attached save.
2. Open the door in front of you
3. The "Only castle staff and trusted guests are allowed beyond these doors" message will show, but the door will open anyway.
4. If you walk through the door into the corridoor, all the guards will attack you.

**Expected behavior**
The door should not open. There are similar doors in the main foyer just to the right and left of the main entrance that show the same message, but behave correctly and do not open.

**Desktop (please complete the following information):**
 - OS: Linux (Ubuntu 20.10)
 - Version: 0.10.27

**Additional context**
Save file: [broken_door_bug.zip](https://github.com/Interkarma/daggerfall-unity/files/5611288/broken_door_bug.zip)

PR #2092, opened 2021-04-24
Fix Castle Wayrest backroom door messages not showing up (issue #1974)

petchema:
Some doors in Castle Wayrest do not show their "warning" messages on first click because their actions are flagged MultiTrigger and only Door and Direct flags were handled in DaggerfallActionDoor.Open().

However, I noticed that some MultiTrigger doors (like the large double doors of Castle Wayrest, no need to search very far) are flagged as having some DoorText yet have no text to display, so if I just add MultiTrigger to the handled cases, those doors suddenly require 2 clicks to open.
So, I also added an action.Index > 0 check to try to make sure the DoorText flag is legit (I assume there's no text with index TYPE_11_TEXT_INDEX aka 8600?)

Fixes issue #1974


Interkarma:
Thank you Pango! :)

PR #2093, opened 2021-04-24
Add Large HUD offset logic to FPSSpellCasting like in FPSWeapon

jefetienne:
I received a [comment ](https://forums.dfworkshop.net/viewtopic.php?p=53906#p53906)on my Readied Spellcasting Hands mod to fix the offset for the Large HUD. When I went to investigate, I found out that the FPSSpellCasting class does not have the offset fix, either. While it seems unnoticeable because of how fast the animation is, it would probably be noticed if someone was using my mod with the fix but not DFU. The readied hands would be offset properly, but then the casting animation would suddenly move them downwards.

This PR resolves that issue by adding the offset logic like in FPSWeapon. It also will be affected by the "LargeHUDUndockedOffsetWeapon" option in the settings.ini, which I think is okay since hands and weapons are viewed in a similar vein.


Interkarma:
Very nice, thank you!

PR #2094, opened 2021-04-28
Quest and enchanted items exclusions

petchema:
Sometimes quest items and enchanted items deserve to not be used as standard items:
- (subreddit below) arrow quest item should not be picked as missile;
- enchanted ingredients should not be used to mix potions (regression in my refactoring of potion maker);
- I assume quest spell books could exist, and should be given a normal one as needed;
- I assume quest lanterns could exist, and should be excluded;
- I assume no quest or enchanted wagon/horse, sounds silly;
- I didn't consider quest or enchanted currency either. Currencies are hard enough (and a bit of a mess already), and also currencies are a good example of commodity so I don't think quest or enchanted money would make a great feature.

I also considered using a predicate parameter instead of flags to filter the result of GetItem() search, it'd be more general but it's also a matter of taste.

There may be other places where quest and enchanted items could benefit from exclusions, so I consider this a RFC for the moment.

Subreddit:
https://www.reddit.com/r/daggerfallunity/comments/n07718/til_that_if_your_quest_item_is_an_arrow_using/

### DIFF
--- a/Assets/Scripts/Game/Items/ItemCollection.cs
+++ b/Assets/Scripts/Game/Items/ItemCollection.cs
@@ -348,14 +348,33 @@ namespace DaggerfallWorkshop.Game.Items
         /// </summary>
         /// <param name="itemGroup">Item group.</param>
         /// <param name="itemIndex">Template index.</param>
-        /// <param name="allowEnchantedItem">Include enchanted items.</param>
-        /// <param name="allowQuestItem">Include quest items.</param>
-        /// <param name="preferConjured">Prefer (short lived) conjured items.</param>
         /// <returns>An item of this type, or null if none found.</returns>
-        public DaggerfallUnityItem GetItem(ItemGroups itemGroup, int itemIndex, bool allowEnchantedItem = true, bool allowQuestItem = true, bool preferConjured = false)
+
+        public DaggerfallUnityItem GetItem(ItemGroups itemGroup, int itemIndex)
         {
             int groupIndex = DaggerfallUnity.Instance.ItemHelper.GetGroupIndex(itemGroup, itemIndex);
-            if (!preferConjured)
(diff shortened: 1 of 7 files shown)


Assets/Scripts/Game/Items/ItemCollection.cs
```diff
@@ -348,16 +348,18 @@ public DaggerfallUnityItem GetItem(int index)
         /// </summary>
         /// <param name="itemGroup">Item group.</param>
         /// <param name="itemIndex">Template index.</param>
-        /// <param name="priorityToConjured">Prefer (short lived) conjured items.</param>
+        /// <param name="allowEnchantedItem">Include enchanted items.</param>
+        /// <param name="allowQuestItem">Include quest items.</param>
+        /// <param name="preferConjured">Prefer (short lived) conjured items.</param>
         /// <returns>An item of this type, or null if none found.</returns>
-        public DaggerfallUnityItem GetItem(ItemGroups itemGroup, int itemIndex, bool priorityToConjured = false)
+        public DaggerfallUnityItem GetItem(ItemGroups itemGroup, int itemIndex, bool allowEnchantedItem = true, bool allowQuestItem = true, bool preferConjured = false)
```
rossipaolo, line 355: May i suggest to use an overload instead, to preserve compatibility with mods?

petchema: Would that do?
C# didn't let me create a GetItem(ItemGroups, int, bool) next to my new method because of the ambiguity it would cause...

petchema: Ah, I see, interesting implementation, and minimal runtime overhead. I'll probably need some more time to totally wrap my head around the consequences.
In any case, thank you for the investigation and provided solutions!

Interkarma:
This is an excellent change. Thank you! :)

PR #2095, opened 2021-04-28
Fix: Show biography skill bonuses during character creation

KABoissonneault:
I noticed the original TES2 shows the Biography skill bonuses during the "Add Bonus Skills" and "Summary" phase of character creation. The bonuses themselves are bugged in TES2, but the ones you see during creation are the ones you get when you start the game.

So, I did this for consistency

### DIFF
--- a/Assets/Scripts/API/BiogFile.cs
+++ b/Assets/Scripts/API/BiogFile.cs
@@ -444,7 +444,7 @@ namespace DaggerfallConnect.Arena2
             }
         }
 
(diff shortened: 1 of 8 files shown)


KABoissonneault:
Also, I notice a few mods currently modify stats on character creation. I'm thinking it'd be more convenient for users of these mods if they could show those changes during character creation.

Maybe another time.

Interkarma:
This is a good change and works well. Thank you. :)

Issue #2096, opened 2021-04-28
Desynchronized vital indicators when toggling large HUD


petchema:
**Describe the bug**
If the vital indicators change, and you toggle the display of the large HUD, the indicators displayed won't correctly reflect your character current stats.

**To Reproduce**
1. Load a game save with health maxed out. You can toggle the large HUD to verify that it's correctly displayed in both modes
2. Hurt yourself (fall damage, or anything else not deadly)
3. Observe health indicator lowering
4. Toggle large HUD and observe health indicator still maxed out

**Expected behavior**
Have consistent vital indicators between full screen and large HUD modes.

**Desktop:**
 - OS: Linux
 - Version 0.11.3

**Additional context**
Hypothesis: Both full screen and the large HUD have their own instance of HUDVitals, but only one is enabled at any time, and only the Update() method of enabled instances calls display adjusting methods; Disabled instances can then become out of sync.

PR #2097, opened 2021-04-28
Resynchronize vital indicators after toggling large HUD

petchema:
Fix for issue #2096

Vital indicators logic is already split in two parts:
- update methods deal with the inertia/convergence usual behavior;
- events can force an immediate adjustment of indicators to current value

This PR add an event triggered right after the large HUD is toggled, and then subscribe to that event to force an immediate update of the indicators.


Interkarma:
Nice one. Thanks for the fix! :)

PR #2098, opened 2021-04-30
Modding the character sheet's Skills dialog

KABoissonneault:
Mods may add new benefits to skills. For example, my [Skilled Spells](https://www.nexusmods.com/daggerfallunity/mods/186) mod bases the caster level for a magical effect on the skill associated with the effect.

This is currently hard to communicate to the player, without telling them to just enter console commands. 

I found that having a mod replacing the entire character sheet window just to change the skills dialog was too prone to conflicts, so I took another approach.

The skill summary code in DaggerfallCharacterSheetWindow.cs was simplified - without changing how it looks - and moved to TextProvider, in a new GetSkillSummary method. The text provider only has to handle the "starting position" for the tokens due to the Misc skills window's two columns. Otherwise, it can show whatever it wants.

Here's an example of a mod overriding the skill summary string with these changes: https://github.com/KABoissonneault/DFU-SkilledSpells/blob/feat/skill-summary-level/Scripts/SkillsSpellsTextProvider.cs

Which results in
image

Again, without any mods, the appearance does not change from the existing implementation.

As a bonus, while dealing with mods that replace text providers, I found having a "fallback" text provider base implementation useful for ensuring no conflicts between mods, and avoiding redundant work. I thought it would be better to provide this class in DFU itself, so all mods can use this approach. See the comments in the code itself, and consult the reference implementation linked above.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallCharacterSheetWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallCharacterSheetWindow.cs
@@ -270,63 +270,6 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             button.OnMouseClick += StatButton_OnMouseClick;
         }
 
-        // Creates formatting tokens for skill popups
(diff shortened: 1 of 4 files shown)


Interkarma:
Thank you!

PR #2100, opened 2021-05-06
Cache RMB blocks

df-ferital:
While working on new PRs, I noticed again that `RMBLayout.GetLocationBuildingData` takes a long time for big cities when a replacement mod such as _Taverns Redone_ is used (around 1.5s for Daggerfall city on my computer, using a stopwatch). Furthermore, the method is called at least three times when loading a savegame. The more problematic issue is that it is also called in several places like when opening the talking window or selecting a random place for a quest.

Some time ago, I made a first attempt at fixing this issue in PR #1541, however, it was not satisfactory and since I didn't have time to work on it anymore, I closed it.

For this new attempt, I decided to cache RMB blocks as a static field in RMBLayout itself, with a size of 12 locations at most. This works quite well, reduces loading and travelling times a little bit and, most importantly, allow the dialogue window to be displayed almost immediately.

I also took the opportunity to use some very useful new features from C# 7.2 such as `in` and `ref readonly` in a few places, where big structures such as `DFBlock` or `DFLocation` are used.

### DIFF
--- a/Assets/Scripts/Terrain/StreamingWorld.cs
+++ b/Assets/Scripts/Terrain/StreamingWorld.cs
@@ -101,7 +101,7 @@ namespace DaggerfallWorkshop
 
         bool init;
         bool terrainUpdateRunning;
-        bool updateLocatations;
+        bool updateLocations;
 
         DaggerfallLocation currentPlayerLocationObject;
         int playerTilemapIndex = -1;
@@ -251,7 +251,7 @@ namespace DaggerfallWorkshop
                 UpdateWorld();
                 InitPlayerTerrain();
                 StartCoroutine(UpdateTerrains());
-                updateLocatations = true;
+                updateLocations = true;
                 init = false;
             }
 
@@ -259,16 +259,16 @@ namespace DaggerfallWorkshop
             // Raise location update flag any time terrain updates
             if (terrainUpdateRunning)
             {
-                updateLocatations = true;
+                updateLocations = true;
                 return;
             }
 
             // Update locations when terrain data finished
             // This flag is raised during terrain update
-            if (updateLocatations)
+            if (updateLocations)
             {
                 UpdateLocations();
-                updateLocatations = false;
+                updateLocations = false;
             }
 
             // Reposition player
@@ -721,7 +721,9 @@ namespace DaggerfallWorkshop
                 {
                     // Add building directory to location game object
                     BuildingDirectory buildingDirectory = locationObject.AddComponent<BuildingDirectory>();
-                    buildingDirectory.SetLocation(location);
+
+                    DFBlock[] blocks;
+                    buildingDirectory.SetLocation(location, out blocks);
 
                     // Add location to loose object list
                     LooseObjectDesc looseObject = new LooseObjectDesc();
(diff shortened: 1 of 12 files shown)


Interkarma:
Thank you! This performs like a very positive update and no issues found so far. It's a complex-feeling PR with new language features and some minor style changes, so I've come back to it a few times over the last week. I've not hit any issues and think that I've internalised the changes.

I've tested everything I can think of that would be impacted by this: world loading, automap, talk, asking about work, quest placement, building info, discovery, etc. and everything working OK so far as I can see. Can I please open up this one up to @ajrb, @rossipaolo, @petchema for feedback if possible, and if you can spot any gaps I might have missed in review. If nobody has any comments, I'll merge ahead of 0.11.4 on weekend.

Thank you all. :)

KABoissonneault (approved):

Assets/Scripts/Terrain/StreamingWorld.cs
```diff
@@ -772,10 +774,10 @@ private IEnumerator UpdateLocation(int index, bool allowYield)
                     }
 
                     // Perform layout and yield after each block is placed
-                    ContentReader contentReader = DaggerfallUnity.Instance.ContentReader;
-                    for (int y = 0; y < height; y++)
+                    int blockIndex = 0;
+                    for (int y = 0; y < height; ++y)
```
ajrb, line 778: Why change the for loop to pre-increment y? I don't think it makes any difference to how the loop functions, but it's strange and seems an odd thing to change here. Is there a reason I am missing?

Assets/Scripts/Utility/RMBLayout.cs
```diff
@@ -36,6 +40,11 @@ public static class RMBLayout
         public const uint CityGateClosedModelID = 447;
         public const uint BulletinBoardModelID = 41739;
 
+        private static int maxLocationCacheSize = 12;
```
ajrb, line 43: How did you arrive at 12 for the cache size?

ajrb:
The performance improvements are amazing, nice work!

ajrb (approved):

Assets/Scripts/Terrain/StreamingWorld.cs
df-ferital: It's a minor style change, the main change there is to increment blockIndex on each step instead of calculating it every time. And I always use pre-increment in loops since it is slightly more efficient and recommended. But I can revert it to post-increment to remain consistent with the rest of the DFU codebase if that's the coding convention, no problem on my side.

Assets/Scripts/Utility/RMBLayout.cs
df-ferital: Indeed, I chose this number since it covers all locations currently loaded at a given position on the world map, as well as a dungeon and locations near it on another point. So if player gets a quest that sends them to a dungeon, and then returns the questgiver town, no loading required.

df-ferital:
Thanks for you review @ajrb, I really appreciate it!

Interkarma:
Thank you @ajrb. :) I'll merge this one for 0.11.4 now.

petchema:
When I go from town to town without using fast travel, the town map of the destination town is empty.
Game save in Broadtry, Daggerfall; there's another town 2 zones north of it. I'm using a Terrain Distance of 4.
[SAVE937.zip](https://github.com/Interkarma/daggerfall-unity/files/6533771/SAVE937.zip)
Git bisect leads to commit 0cb8c1a2a207c69a5b5449e0c135e4646a65fedf

petchema:
Tried different Terrain Distances, same result. Same for switching town roles and traveling south.
Town map is immediately fixed by saving then reloading a game save.

df-ferital:
Thank you @petchema for noticing this! I fixed it in #2101.

ajrb:
I think that's not the PR you meant there @df-ferital

df-ferital:
Yes, sorry, updated!

Interkarma:
Hey @df-ferital! :)

This change appears to be causing duplicate building name assignment when the same block is present twice in the one location.

https://forums.dfworkshop.net/viewtopic.php?f=31&t=5060

I've confirmed not present in 0.11.3 and earlier. Regression appears in 0.11.4 after this change. We'll need to somehow exclude building name seeds from cache when block used more than once in same location.

PR #2101, opened 2021-05-07
Added the ability for mods to specify an item type as "noRepairs"

KABoissonneault:
As discussed in this thread: https://forums.dfworkshop.net/viewtopic.php?f=23&t=4794

Comment from IK:
>I'd like this to be done in a way that doesn't break if a mod adding these properties is removed later and the item remains

I tested my solution, and if the mod is removed, the item reverts to the default value (false). This means that if a mod makes ingredients noRepairs, when it is removed, those ingredients go back to the default behavior of allowing repairs (if damaged).

I added a new string, so that users aren't told their noRepairs damaged item "does not need repairs".


magicono43:
From Hazelnut: "I'd go for isNotRepairable  myself. noRepairs is far to ambiguous IMO"

KABoissonneault:
Also, I thought at first arrows didn't need to be excluded because nothing damages them, but I just saw in code that they're spawned with 0 condition, so my change would accidentally make them repairable

Issue #2052, opened 2021-02-24
Missing setting to disable shadows from player light sources


KABoissonneault:
**Describe the bug**
Players looking for a more retro style can disable shadows from light sources in dungeon, interiors, and exteriors. However, lights from the player torch and the MagicCandle prefab used by the Light spell effect always use soft shadows - there's no setting to make their shadows "none".

**To Reproduce**
1. Go in settings
2. Disable all shadows
3. Go in game
4. Find a mobile NPC (ex: in a dungeon)
5. Use a Torch or a Light spell
6. See soft shadow from the NPC

**Expected behavior**
There should be a way to disable those shadows too.

**Screenshots**
![image](https://user-images.githubusercontent.com/5789925/109074804-37668580-76c6-11eb-9b63-2ec53ea1188d.png)

Interkarma:
The magic candle should probably follow the "spell shadows" setting.

The standard default player torch casts no shadows. I believe the light source for "item based player torch" enables shadows. I agree there should be a way to switch this off.

ajrb:
So if the player based torch shadows reflected the spell shadows setting then this is resolved? I'll do that, it was my oversight.

Interkarma:
Thank you for PR! We'll also need to assign spell shadow settings to magic candle from standard Light effect. I think shadows on that prefab are always enabled.

ajrb:
Will leave the prefab to you. 

FYI I tried switching on shadows for magelight in same way as player torch as KAB suggested on the PR and it did nothing as far as I can see. Not sure that it's needed really since magelight is kinda an magic ambient light spell which you did just as a demo IIRC.

ajrb:
#2106 Fixed this for player carried lights. Magic candle and standard light are still outstanding.

PR #2106, opened 2021-05-09
Enable player torch shadows based on spell shadows setting

ajrb:
Partially fixes issue #2052


KABoissonneault:
As commented by IK in #2052, you should also cover the LightNormal and MageLight magic effects to fully fix the issue.

PR #2108, opened 2021-05-10
Added "resistance" formula overrides. 

KABoissonneault:
Added formula overrides for the following formulas:
- InflictPoison
- InflictDisease
- SavingThrow

Also fixed OnMonsterHit formula not returning after invoking the formula override.
For "formulas" that don't return a value, it's better to use Action than have a Func with an unused return value.

### DIFF
--- a/Assets/Scripts/Game/Formulas/FormulaHelper.cs
+++ b/Assets/Scripts/Game/Formulas/FormulaHelper.cs
@@ -685,7 +685,7 @@ namespace DaggerfallWorkshop.Game.Formulas
                 // Handle poisoned weapons
                 if (damage > 0 && weapon.poisonType != Poisons.None)
                 {
-                    InflictPoison(target, weapon.poisonType, false);
+                    InflictPoison(attacker, target, weapon.poisonType, false);
                     weapon.poisonType = Poisons.None;
                 }
             }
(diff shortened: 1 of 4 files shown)


Assets/Scripts/Game/Formulas/FormulaHelper.cs
```diff
@@ -685,7 +685,7 @@ public static int CalculateAttackDamage(DaggerfallEntity attacker, DaggerfallEnt
                 // Handle poisoned weapons
                 if (damage > 0 && weapon.poisonType != Poisons.None)
                 {
-                    InflictPoison(target, weapon.poisonType, false);
+                    InflictPoison(attacker, target, weapon.poisonType, false);
```
Interkarma, line 688: Why was this changed to attacker? Poison weapons assign their payload to the target hit by the weapon, not to the wielder.

KABoissonneault: Yeah, I've only added extra info for mods. I thought adding the parameter there would be consistent with the other formulas. No change in behavior for DFU itself

Interkarma: All good, sorry about that. Still working through, but looks good so far. :)

Interkarma:
Thank you! 👍

PR #2109, opened 2021-05-13
Expose the SpellMaker and its subwindows to mods

KABoissonneault:
Sparkled some "protected" and "virtual" all over.

Added SpellEffectEditor and SpellIconPicker to UIWindowFactory.


KABoissonneault:
I just realized that adding an enumerator value in the middle of UIWindowType like this breaks when using precompiled mods that register custom windows.

Do we consider this an issue? It seems to work fine if I add the enum at the end of the list, but then it won't be sorted alphabetically

rossipaolo:
I'm in favor of preserving compatibility. If you want to keep the alphabetical order you should be able to set an explicit value:

```cs
Rest,
SpellBook,
SpellIconPicker = 39,
SpellMaker = 24,
StartNewGameWizard,
Start,
```

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallEffectSettingsEditorWindow.cs
```diff
@@ -242,14 +246,14 @@ void SetupSpinners()
...
-        void SetupButtons()
+        protected virtual void SetupButtons()
         {
             // Exit button
             exitButton = DaggerfallUI.AddButton(exitButtonRect, NativePanel);
             exitButton.OnMouseClick += ExitButton_OnMouseClick;
         }
 
-        void InitControlState()
+        protected void InitControlState()
```
ajrb, line 256: Why no virtual here? I can't figure out the reason.

Assets/Scripts/Game/UserInterface/UpDownSpinner.cs
```diff
@@ -174,7 +174,7 @@ public void SetRange(int minValue, int maxValue)
         public void SetValue(int value)
         {
             // Clamp values if ranges specified, otherwise accept whatever owner wants
-            if (minValue != 0 && maxValue != 0)
+            if (minValue != 0 || maxValue != 0)
```
ajrb, line 177: Why have you changed this? Are you sure that this doesn't change behaviour of any other UI window? This component is used in a few places.

ajrb:
@rossipaolo So I also have a PR which adds enum values. I was not going to worry about breaking a few pre-compiled mods pre v1.0 release since they simply need re-compiling to work. However there's no point KAB adding the explicit values to avoid it if I don't also do this in my PR.

So, shall we sync our PRs or just let a few pre-compiled mods break?

Assets/Scripts/Game/UserInterface/UpDownSpinner.cs
KABoissonneault: > Why have you changed this? Are you sure that this doesn't change behaviour of any other UI window? This component is used in a few places.

As the comment specifies, this code is meant to prevent the spinner from working if no range was set. If _either_ of the values have been set to non-zero, then it's been set. The way the code was written, a range like "0 to 60" would be considered "range not specified".

I did this so custom spellmaker windows (ex: my mod) could allow 0 as the minimum value, instead of 1.

I doubt this would break anything. This would only break a spinner that sets a non-zero maximum value yet somehow expects the spinner to be disabled.

From my look at the code, the only other place it's used is in StatsRollout.cs, and it sets no range, so both minValue and maxValue will be 0, which means it still works the same as before

rossipaolo:
@ajrb I think we should aim not to break mods, at least when it doesn't take much to avoid it.

ajrb:
Fair enough, I have complied. @KABoissonneault you will need to update your PR to use different values, I did try the same as you with `= WitchesCovenPopup + 2;` but VS complained about circular references, and I also think this is less clear from a code readability standpoint so I went with explicit values.

Assets/Scripts/Game/UserInterfaceWindows/UIWindowFactory.cs
```diff
@@ -42,7 +42,8 @@ public enum UIWindowType
         QuestOffer,
         Rest,
         SpellBook,
-        SpellMaker,
+        SpellIconPicker = WitchesCovenPopup + 1, // new enum: setting this to the value after WitchesCovenPopup to avoid breaking precompiled mods
+        SpellMaker = SpellBook + 1,
```
ajrb, line 46: Please could you change this to:
`
SpellIconPicker = 42,
SpellMaker = 24,
`

Thanks.

PR #2110, opened 2021-05-14
Added new time-related text macros

KABoissonneault:
- %day: Current day of the month as number
- %dayn: Name of the day of the week, in TES terminology (ex: Morndas)
- %days: Current day of the month as number, with suffix (ex: 1rst, 2nd, ..., 30th)
- %mon: Month as number
- %monn: Month as name, in TES terminology (ex: Hearthfire)
- %year: Year as number (ex: 405, not 3E405)
- %min: Current minute on the clock
- %hour: Current hour on the clock (24 hour mode)
- %sign: Sign of the current month. These aren't lore in TES2, but they're a staple of the series right after that, and it was already in the code...
- %sea: Current season name

Fixed an issue which prevented sequences of characters like "%day/%mon/%year" from properly expanding as 3 separate macros.

Before fix:
image

After fix:
image

These macros were requested by a quest modder. I think they would be useful for many people

### DIFF
--- a/Assets/Scripts/Utility/MacroHelper.cs
+++ b/Assets/Scripts/Utility/MacroHelper.cs
@@ -17,6 +17,7 @@ using DaggerfallWorkshop.Game.UserInterfaceWindows;
 using UnityEngine;
 using DaggerfallWorkshop.Game.Utility;
 using DaggerfallWorkshop.Game.MagicAndEffects.MagicEffects;
+using System.Text;
 
 namespace DaggerfallWorkshop.Utility
 {
@@ -390,8 +391,8 @@ namespace DaggerfallWorkshop.Utility
 
         #region Macro Expansion Code
 
-        // Any punctuation characters that can be on the end of a macro symbol need adding here.
-        static char[] PUNCTUATION = { '.', ',', '\'', '?', '!', '/', '(', ')', '{', '}', '[', ']', '\"', ';', ':' };
+        // Any non-alpha characters that can be on the end of a macro symbol need adding here.
+        static readonly char[] MACRO_TERMINATORS = { ' ', '%', '.', ',', '\'', '?', '!', '/', '(', ')', '{', '}', '[', ']', '\"', ';', ':' };
 
         /// <summary>
         /// Expands any macros in the textfile tokens.
@@ -404,6 +405,7 @@ namespace DaggerfallWorkshop.Utility
             string tokenText;
             int multilineIdx = 0;
             TextFile.Token[] multilineTokens = null;
+
             // Initialise macro cache - used to ensure macros are only evaluated once per ExpandMacros() call.
             // Important since some macros evaluate differently each time. (e.g. macros with random generated names like %fx1 & %fx2)
             Dictionary<string, string> macroCache = new Dictionary<string, string>();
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Utility/MacroHelper.cs
```diff
@@ -228,12 +228,22 @@ public static class MacroHelper
             { "%wth", Worth }, // Worth
         // DF Unity - new macros:
             { "%pg", PlayerPronoun },   // He/She (player)
-            { "%pg1", PlayerPronoun },  // His/Her (player)
+            { "%pg1", PlayerPronoun },  // He/She (player)
```
petchema, line 231: Why this change? (and why in this unrelated PR)
`Assets/StreamingAssets/Quests/L0A01L00.txt
24:<ce>               gold coin for %pg1 first sanctioned death.
`

ajrb:
You've changed the macro processing quite a lot here to enable that specific sequence of separate data info. I am concerned about regressions here, as this code has been pretty stable for a couple of years now and I don't think this is what Interkarma meant by submit changes before the release freeze date.

A simpler approach would simply to have a macro for the complete date string. day/mon/year but that would be less flexible I realise. How important is this?

Assets/Scripts/Utility/MacroHelper.cs
```diff
@@ -767,14 +818,61 @@ private static string PlayerLastname(IMacroContextProvider mcp)
...
+        }
+
+        private static string CurrentSeason(IMacroContextProvider mcp)
+        {   // %sea
+            return DaggerfallUnity.Instance.WorldTime.Now.SeasonName;
+        }
+
         private static string PlayerPronoun(IMacroContextProvider mcp)
-        {   // %pg
+        {   // %pg & %pg1
```
ajrb, line 872: Did you check this was not a mistake in transferring hard strings to unity localised strings?

Assets/Scripts/Utility/MacroHelper.cs
KABoissonneault: > Why this change? (and why in this unrelated PR)

Look carefully, I only changed comments, nothing functional

Assets/Scripts/Utility/MacroHelper.cs
KABoissonneault: It's consistent with how %g and %g1 works. As pointed out in the other comment, this is only a change in comments so maintainers aren't confused

KABoissonneault:
>I am concerned about regressions here, as this code has been pretty stable for a couple of years now 

**The algorithm only makes currently invalid cases valid**. If someone attempted to make a macro followed by another macro before (ex: `%g%g`), they would have gotten `%g%g[unhandled]` before, while my changes make it `hehe`. Same for a macro followed by punctuation, it turns an error into a feature.  

It's true that there could be bugs, but ideally I'd rather we have a preview release where people can point out any issues than avoid touching any code.

KABoissonneault:
I've added some comments, maybe this will help seeing what changes I made, and for what reason. The code is not actually that different from the previous code, in many cases it's minor improvements like avoiding a double dictionary lookup (ex: use `macroCache.TryGetValue(key, out symbolStr)` instead of `macroCache.Contains(key)` followed by `symbolStr = macroCache[key]`)

There's 3 significant changes compared to the previous code:
- The macro is evaluated in a loop, in case there's multiple macros in a sequence without any spaces (ex: "%day/%mon/%year")
- A macro followed by punctuation (ex: "%pg3.") now uses the macro cache, instead of re-evaluating it again. This might fix a bug, but otherwise, it's still a performance improvement, probably
- Having multiple macros next to each other is now supported (ex: "%pg1%pg1")

If we want to minimize the change, I'm willing to give up on that third point, as I don't see the utility myself - I just did it for the sake of completeness / to avoid surprises if someone finds a use for it someday.

Assets/Scripts/Utility/MacroHelper.cs
petchema: Even if they're just comments they're unrelated to this PR
And the only occurence of %pg1 I could find was introduced by @JayH2971  
```
commit 8e03008fc7bf6a5cba04ea419212934da8f68144
Author: JayH2971 <37541932+JayH2971@users.noreply.github.com>
Date:   Fri Jul 3 16:26:17 2020 -0700

    DB, Knight, MG, FG typos

diff --git a/Assets/StreamingAssets/Quests/L0A01L00.txt b/Assets/StreamingAssets/Quests/L0A01L00.txt
index 08744a069..b380f72b3 100644
--- a/Assets/StreamingAssets/Quests/L0A01L00.txt
+++ b/Assets/StreamingAssets/Quests/L0A01L00.txt
@@ -22,7 +21,7 @@ QuestComplete:  [1004]
 <ce>               town, you will know if there are any guild
 <ce>                  buildings there. It is customary for
 <ce>               the novice to receive payment of a single
-<ce>               gold coin for his first sanctioned death.
+<ce>               gold coin for %pg1 first sanctioned death.
 <ce>                             Here is yours.
 
 RumorsDuringQuest:  [1005]

```

So maybe a better fix would be to drop %pg1 handling altogether

petchema:
The original code was unnecessarily complex (why split into words, handle each word separately then concatenate everything back together?), and this PR adds some more complexity.
If I had to rewrite it, I'd go like:
https://github.com/KABoissonneault/daggerfall-unity/compare/feat/time-macro...petchema:simplified-expandmacros
Then again, are we still okay with such rewrites? We really lack unit tests to protect against regressions...

KABoissonneault:
> The original code was unnecessarily complex (why split into words, handle each word separately then concatenate everything back together?), and this PR adds some more complexity.
> If I had to rewrite it, I'd go like:
> [KABoissonneault/daggerfall-unity@feat/time-macro...petchema:simplified-expandmacros]

I like it. It's shorter, simpler, uses StringBuilder, and keeps the benefits of my version.

I really don't think the risk is that high, to be honest. The existing text had to be written very carefully for the non-permissive TES2 macro system, so the moment we have "macro followed by space" and "macro followed by punctuation" covered, it should be good. Everything else was already an error

ajrb:
> The original code was unnecessarily complex (why split into words, handle each word separately then concatenate everything back together?), and this PR adds some more complexity.

It's three years ago now so it's hard to remember now honestly. Could be as simple as not having the understanding of macros and how they're used in Daggerfall that we do now.

I like the simplified approach @petchema, though not sure it handles % that are not macro prefixes. (i.e. followed by a space)

Agree with comments about risk of regression given we have no automated tests as you say. I think that's a call for @Interkarma to answer as to his appetite for risk this close to the release freeze. I don't mind either way.

Assets/Scripts/Utility/MacroHelper.cs
ajrb: I've traced this down now. Seems I made a mistake when I added thse for Jay H almost 3 years ago and he never noticed. Please can you remove this change and make the change I suggest in my next review comment. Ta

Assets/Scripts/Utility/MacroHelper.cs
```diff
@@ -228,12 +229,22 @@ public static class MacroHelper
             { "%wth", Worth }, // Worth
         // DF Unity - new macros:
             { "%pg", PlayerPronoun },   // He/She (player)
-            { "%pg1", PlayerPronoun },  // His/Her (player)
+            { "%pg1", PlayerPronoun },  // He/She (player)
```
ajrb, line 232: This line should be changed to:

`{ "%pg1", PlayerPronoun1 },  // He/She (player)`

and the corresponding method not removed further down. Careless typo by me from Sept 2017.

Alternatively just leave both alone and I'll submit a tiny fix PR.

Assets/Scripts/Utility/MacroHelper.cs
ajrb: Please see my comments. It was a typo by me. Sorry. :(

petchema:
> I like the simplified approach @petchema, though not sure it handles % that are not macro prefixes. (i.e. followed by a space)

No, it doesn't, it seems they're no case of % at the end of a word in the existing quests. And if it was a special case left on purpose (say, in case we want to be able to display a % in a text) then it's not great because it would only allow to use % at the end of words. If such need arise, I'd go with say doubling % (%%). Until then I just removed the special case.

ajrb:
> 
> No, it doesn't, it seems they're no case of % at the end of a word in the existing quests. And if it was a special case left on purpose (say, in case we want to be able to display a % in a text) then it's not great because it would only allow to use % at the end of words. If such need arise, I'd go with say doubling % (%% handled specifically to mean a single %). Until then I just removed the special case.

As discussed on Discord you're correct. I must be remembering an assumtion that was faulty in the first place.

Personally I'd go for adding a %perc macro so it works like all the others, but %% is fine too. Don't care either way.

ajrb:
Looks good to me, though I've done no actual testing myself. I've now submitted the small fix for player pronouns in a separate PR #2112

KABoissonneault:
A modder requested a macro for the current building, I decided to slip it through in here

PR #2115, opened 2021-05-16
Prevent levitating and swimming diagonally from being 40% faster

petchema:
Also fix forward taking priority over backward, and right taking priority over left when pressing both controls simultaneously

### DIFF
--- a/Assets/Scripts/Game/LevitateMotor.cs
+++ b/Assets/Scripts/Game/LevitateMotor.cs
@@ -68,21 +68,25 @@ namespace DaggerfallWorkshop.Game
             if (GameManager.Instance.PlayerEntity.IsParalyzed)
                 return;
 
-            Vector3 horizontalVector = new Vector3 (0, 0, 0);
+            float inputX = 0f;
+            float inputY = 0f;
             // Forward/backwards
             if (InputManager.Instance.HasAction(InputManager.Actions.MoveForwards))
-                horizontalVector += playerCamera.transform.forward;
+                inputY++;
             if (InputManager.Instance.HasAction(InputManager.Actions.MoveBackwards))
-                horizontalVector -= playerCamera.transform.forward;
+                inputY--;
 
             // Right/left
             if (InputManager.Instance.HasAction(InputManager.Actions.MoveRight))
-                horizontalVector += playerCamera.transform.right;
+                inputX++;
             if (InputManager.Instance.HasAction(InputManager.Actions.MoveLeft))
-                horizontalVector -= playerCamera.transform.right;
+                inputX--;
 
-            if (horizontalVector != Vector3.zero)
-                AddMovement(horizontalVector.normalized);
+            if (inputX != 0.0f || inputY != 0.0f)
+            {
+                float inputModifyFactor = (inputX != 0.0f && inputY != 0.0f && playerMotor.limitDiagonalSpeed) ? .7071f : 1.0f;
+                AddMovement(playerCamera.transform.TransformDirection(new Vector3(inputX * inputModifyFactor, 0, inputY * inputModifyFactor)));
+            }
 
             // Up/down
             Vector3 upDownVector = new Vector3 (0, 0, 0);
--- a/Assets/Scripts/Game/LevitateMotor.cs
+++ b/Assets/Scripts/Game/LevitateMotor.cs
@@ -68,17 +68,21 @@ namespace DaggerfallWorkshop.Game
             if (GameManager.Instance.PlayerEntity.IsParalyzed)
                 return;
 
+            Vector3 horizontalVector = new Vector3 (0, 0, 0);
             // Forward/backwards
             if (InputManager.Instance.HasAction(InputManager.Actions.MoveForwards))
-                AddMovement(playerCamera.transform.forward);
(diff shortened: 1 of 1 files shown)


petchema:
@jefetienne Does it look okay? Could it be simplified or improved?

Assets/Scripts/Game/LevitateMotor.cs
```diff
@@ -68,17 +68,25 @@ private void Update()
...
                 return;
 
+            float inputX = 0f;
+            float inputY = 0f;
             // Forward/backwards
             if (InputManager.Instance.HasAction(InputManager.Actions.MoveForwards))
-                AddMovement(playerCamera.transform.forward);
-            else if (InputManager.Instance.HasAction(InputManager.Actions.MoveBackwards))
-                AddMovement(-playerCamera.transform.forward);
+                inputY++;
```
KABoissonneault, line 75: Not sure about using ++ on float, to be honest. Minor issue though

Assets/Scripts/Game/LevitateMotor.cs
```diff
@@ -68,17 +68,25 @@ private void Update()
...
-                AddMovement(playerCamera.transform.right);
-            else if (InputManager.Instance.HasAction(InputManager.Actions.MoveLeft))
-                AddMovement(-playerCamera.transform.right);
+                inputX++;
+            if (InputManager.Instance.HasAction(InputManager.Actions.MoveLeft))
+                inputX--;
+
+            if (inputX != 0.0f || inputY != 0.0f)
+            {
+                float inputModifyFactor = (inputX != 0.0f && inputY != 0.0f && playerMotor.limitDiagonalSpeed) ? .7071f : 1.0f;
```
KABoissonneault, line 87: Consider putting both components in a 2d vector and normalizing it. If we ever want to add joystick support

Assets/Scripts/Game/LevitateMotor.cs
petchema: That's what was suggested by VS Code...

Assets/Scripts/Game/LevitateMotor.cs
petchema: Normalisation is what I used in the first commit, then I aligned implementation with the other "motor" classes.
I know nothing about controllers support, that's why I asked feedback from jfetienne.

Assets/Scripts/Game/LevitateMotor.cs
```diff
@@ -68,17 +68,25 @@ private void Update()
             if (GameManager.Instance.PlayerEntity.IsParalyzed)
                 return;
 
+            float inputX = 0f;
```
jefetienne, line 71: Hey, sorry for the wait! I have been traveling and I finally got a chance to take a look at this.

After looking at the FrictionMotor, what it does is simply set the inputX and inputY to
```
float inputX = InputManager.Instance.Horizontal;
float inputY = InputManager.Instance.Vertical;
```
I tested the code using this instead of that chunk of code dealing with HasAction() and it looks like it works perfectly, so I think you can remove that entire chunk and use that instead.

petchema: Thanks for the feedback!
I have a concern though, InputManager.Instance.Horizontal is not bound by -1 .. -1 but by -NegHorizontalLimit .. PosHorizontalLimit, etc. so shouldn't they be scaled before use, something like
```
InputManager IM = InputManager.Instance;
float inputX = (IM.Horizontal + IM.NegHorizontalLimit) / (IM.PosHorizontalLimit + IM.NegHorizontalLimit);
float inputY = (IM.Vertical + IM.NegVerticalLimit) / (IM.PosVerticalLimit + IM.NegVerticalLimit);
```

Assets/Scripts/Game/LevitateMotor.cs
petchema: Removed by the simplification suggestion by @jefetienne

KABoissonneault (approved):

Assets/Scripts/Game/LevitateMotor.cs
jefetienne: I might be wrong but I think it may have been intended to limit the top speed of the player, so we might not need to have scaling. @ajrb wrote the limits so I think he can answer that.

ajrb: The ability to change the limit of maximum movement in each direction was added to enable different speeds in various directions by mods. It's used in the enhanced riding module of R&R mod to limit the speed you can move sideways or backwards when riding a horse. It always seemed odd to be able to move backwards as fast as you could forward on a horse.

I think when levitation is involved, in my use case of the limits at least, there's no reason to limit movement in any direction. I think it would actually be bad for the normal grounded movement limits to also affect magical flying.

petchema: Shouldn't it be up to the mods to decide what limits should be applied during walk/swimming/riding/levitation/whatever?
Are you okay with with PR as-is?

jefetienne: I think it looks good!

ajrb: I did what I needed for my mod, and havn't see any calls for a more complex system. I am happy with the PR as-is w/o any levitation limiting.

PR #2117, opened 2021-05-16
Immediately interrupt resting if you die (RFC)

petchema:
Currently if you die while resting (from disease or poisoning), game plays death moan but waits until you're fully rested before going on with the death video (which can take a while because healing has to overcome disease or poison effect).

With this change game immediately interrupts resting with "You never awaken." message. I would also be okay if no message was displayed (given it's not in classic), but currently preventedRestMessage = null and preventedRestMessage = "" show inappropriate messages instead.


petchema:
Already poisoned character, for testing purposes
[SAVE201.zip](https://github.com/Interkarma/daggerfall-unity/files/6490171/SAVE201.zip)

KABoissonneault (approved):

PR #2118, opened 2021-05-17
Use classic hack to fix Orsinium overlapping blocks

df-ferital:
Instead of discarding one block as DFU does, classic simply moves one of the block south of the other.


KABoissonneault (approved):

Interkarma:
A couple of quick question on this one @df-ferital. If the block is being moved rather than disabled, is it still accessible from the main areas of map, or is it isolated to other side of a border block? If it's isolated, wouldn't culling it still be the best approach for performance reasons, considering it will be rendered but never used with this approach? And if player can't reach it, then it should still be excluded from quest system.

df-ferital:
It's the border block which is moved to the south, sorry I should have put this information in comments (and will do it).

Interkarma:
Ah, I see - thank you! That is indeed a better approach. :)

PR #2119, opened 2021-05-17
Added "weather", "season", and "climate" quest trigger conditions

KABoissonneault:
Added "weather", "season", and "climate" quest trigger conditions, and "train pc" quest action.

I was originally going to keep this as a mod since I made them for a modder, but after he shared his results, other people said they wanted those quest actions too, so might as well share it here.

The syntaxes are:
- weather sunny|cloudy|overcast|fog|rain|thunder|snow
- season fall|summer|spring|winter
- climate desert|desert2|mountain|mountainwoods|rainforest|ocean|swamp|subtropical|woodlands|hauntedwoodlands
- climate base desert|mountain|temperate|swamp
- train pc <skillName>

Note: "train pc" uses the same formula as the default training window, except it's free. It works like quest rewards, and therefore works after ending the quest (with the quest end message).

@rossipaolo Are you the one maintaining the VS Code plugin for quests? You might want to keep an eye on this

### DIFF
--- a/Assets/Scripts/Game/Questing/QuestMachine.cs
+++ b/Assets/Scripts/Game/Questing/QuestMachine.cs
@@ -346,6 +346,9 @@ namespace DaggerfallWorkshop.Game.Questing
             RegisterAction(new TotingItemAndClickedNpc(null));
             RegisterAction(new DailyFrom(null));
             RegisterAction(new DroppedItemAtPlace(null));
+            RegisterAction(new Season(null));
+            RegisterAction(new Actions.Weather(null));
+            RegisterAction(new Climate(null));
 
             // Register default actions
             RegisterAction(new EndQuest(null));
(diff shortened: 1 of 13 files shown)


Assets/Scripts/Game/Questing/QuestMachine.cs
```diff
@@ -346,6 +346,9 @@ void RegisterActionTemplates()
             RegisterAction(new TotingItemAndClickedNpc(null));
             RegisterAction(new DailyFrom(null));
             RegisterAction(new DroppedItemAtPlace(null));
+            RegisterAction(new Season(null));
+            RegisterAction(new Actions.Weather(null));
```
KABoissonneault, line 350: It wouldn't compile without that prefix, since "Weather" is a namespace

KABoissonneault:
Another modder thought it would be useful to be able to specify any building of a certain type for "pc at" instead of just one specific building. I agree that it would probably be useful to many people. I just wasn't sure if supporting other places would be worth it. We already have "when player enters" for exterior places

rossipaolo:
@KABoissonneault Would you be interested in making a PR yourself? If you are, [this is the file to change](https://github.com/rossipaolo/vscode-dftemplate/blob/master/modules/DaggerfallUnity.dfmodule.json) and [this is the list of types](https://github.com/rossipaolo/vscode-dftemplate/blob/66ba84a954b7bb9d4f43e331832c38a934801eee/src/language/static/parameterTypes.ts#L11).

KABoissonneault:
@rossipaolo Sure, that's fine

PR #2121, opened 2021-05-19
Added a TextAssetReader Merge for SPELLS.STD spell records

KABoissonneault:
Summary:
- Added a TextAssetReader Merge for SPELLS.STD spell records through a "SpellRecords.json" file, similar to "ItemTemplates.json". 
- Ensured the DaggerfallSpellBookWindow uses the modded standard spells rather than reading SPELLS.STD directly.

So, as part of my magic modding, I started feeling the need to replace the "classic" (aka circinate) spells. Those spells are sold by spell vendors, and are part of potions and enchanted items. As one starts tweaking spell costs for spell effects, they suddenly find the need to tweak those spell bundles too.

I looked into the existing work, and I found only the "custom bundle" in EntityEffectBroker. However, this can only add new spells, and the work seems incomplete (ex: see TODO in CastWhenUsed.EnchantmentPayloadCallback).

I thought the TextAssetReader Merge used for ItemTemplates was simple enough and essentially only requires 1 line of code to work.

However, there's some points to consider:
- Changing the `effects` field is an all-or-nothing kind of deal. You can't just change the `magnitudeBaseLow` of one effect, you have to copy-paste the entire array of effects with all its relevant fields. However, few spells have more than one effect, so it's not that much of a problem for most spells
- EffectRecordData fields that aren't specified in the JSON are going to default to 0, rather than 1 or -1 as expected for certain fields. I had to go and ensure modders wouldn't accidentally cause issues in the code by leaving a field as 0. `descriptionTextIndex` and `spellMakerTextIndex` are unused, so not an issue, The effect attributes (duration, chance, magnitude) can be at 0 just fine, except I ensured the "per level" value sent to EffectSettings was at least 1 to avoid divisions by 0. The only one that concerns me is `subType` for effect types that don't have variants: for the most part, 0 should have the same effect as -1, but I hope I'm not forgetting anything
- Changing the length of `effects` from 3 to 1 or 2 does not cause any issues from my testing, as all places that use the effects of SpellRecordData use `spellRecordData.effects.Length` rather than assuming a size of 3.
- This approach "accidentally" lets users create new spells, by using an unused index. I'm only pointing this out because there's already a system to add new spells, the "custom bundle" I mentioned above. Spells created this way properly appear in the spell vendor, and don't appear in the item maker.
- A mod that changes a spell won't change a spell already written in the player's Spellbook. I'm not sure it's possible to fix this, since spellbook entries don't track the spell record they were created from, I believe?

Anyway, I believe letting mods tweak classic spells this way will be useful in general. For example, many players find poisons way too strong, and this change would allow a modder to change its effects.

### DIFF
--- a/Assets/Scripts/Game/Entities/DaggerfallEntity.cs
+++ b/Assets/Scripts/Game/Entities/DaggerfallEntity.cs
@@ -16,6 +16,7 @@ using System.Linq;
 using System.Collections.Generic;
 using DaggerfallConnect;
 using DaggerfallConnect.Arena2;
+using DaggerfallConnect.Save;
 using DaggerfallWorkshop.Game.Formulas;
 using DaggerfallWorkshop.Game.Items;
 using DaggerfallWorkshop.Game.Serialization;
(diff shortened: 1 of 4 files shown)


Interkarma:
This is an interesting approach! My main concern is the one you raise at the end - that spellbook will store copies of spells that could differ from classic settings and user might not be aware this persists after removing mod. That said, it's a simple thing for user to delete from spellbook and buy again once they know to, either when adding a mod or to revert back to classic after removing a mod. I'll give this one another pass and merge when I can.

Also, thank you for all your work on new PRs and helping with review. Your efforts are seen and very much appreciated. Cheers! :)

KABoissonneault:
I was thinking, maybe we could update EffectBundleSettings and the rest to add an optional reference to the original standard spell record it was created from? When the game is loaded, you check for mismatch between the spell's settings and the current spell record associated with it, and adapt the stats in consequence.
It won't help current playthroughs, but it would allow future playthroughs to enable/disable mods that change spell records without having to refill their spellbook

Interkarma:
Yep, that's a fair workaround. Thank you. :)

KABoissonneault:
Seems like people are looking into modding SPELLS.STD itself. Someone was talking about on the Discord, and then JayH releases this

https://forums.dfworkshop.net/viewtopic.php?f=14&t=4861&p=54958

I think having DFU support is becoming necessary, if we want some sort of mod compatibility

Interkarma:
I'm happy with what you're doing here. Will get it merged soon, I haven't had time to look at new changes yet (I trust you of course, I just like to try and internalise everything before merging).

My preference would be for modders to use your feature here to mod circinate spells rather than edit SPELLS.STD directly. My philosophy is that mods should always be in StreamingAssets or a .dfmod and never change gamedata. I also don't think it's our responsibility to support altered gamedata.

Interkarma:
I found the changes to DeserializeSpellbook() will overwrite any spellpack icons the player is using on load. Another approach is needed for resetting spell data.

Now that I more think this - it's really impossible to predict all the ways spellbook data might change in future, or what a mod might do with it. I think it's best not to mess with the player's spellbook data by naively copying back template data. I'm more comfortable with player being in control of this and simply deleting/rebuying spells as needed.

KABoissonneault:
So would you rather revert that last changelist?

Interkarma:
Yes please, just revert that last change. Otherwise no problem found here with the changes and I'm happy to merge for next round of builds.

KABoissonneault:
As I wrote in the commit, I decided to keep the `StandardSpellIndex` in the spell settings, as this can be used by mods to identify a circinate spell. For example, my spell cost mod could use this to give a spell cost that's different from an equivalent custom spell.

You didn't mention an objection to that part, so I went ahead and kept it :)

Interkarma:
That's fine, cheers mate. :)

PR #2126, opened 2021-05-21
Connect N0000055.RDB disconnected parts

df-ferital:
Fix potential combination issues with this block, for example in Ruins of Kinghouse Grange, Glenpoint, as reported by @petchema 
: http://forums.dfworkshop.net/viewtopic.php?p=52821&sid=5a152a9109cc38760fd7b4ecb2b34853#p52821

I also moved a random monster to another corner.

Before the fix:
n0000055-before

After:
n0000055-after


df-ferital:
N0000055.RDB is not faulty, my assumption was based on the particular configuration of Ruins of Kinghouse Grange, Glenpoint. But the bugged block is N0000071.RDB, which is fixed in #2130.

PR #2128, opened 2021-05-22
Stack identical books together

petchema:
If you ever tried to collect all existing books, you probably know the pain of checking manually if you already have one book.
Having identical books stack together makes that easier.


Interkarma:
When selling books, the stack count doesn't factor into the sale price, i.e. a stack 2x Real Barenziah will sell for the same price as 1x Real Barenziah.

petchema:
Yeah, displayed stack price is total price, to be consistent with total weight. Sometimes it can get confusing, I guess that was the issue

PR #2125, opened 2021-05-20
Reposition W0000009.RDB exit

df-ferital:
Fix the exit by replacing a nearby corridor corner with one having an exit and moving the dungeon exit quad there. This is better than the dirty hack done by classic in my opinion. I also added a brick wall in case the block is not the starting block, which is what is done in similare cases like with W00000012.RDB.

Before the fix:
w0000009-before

After the fix:
w0000009-after

The following dungeons use this block as a starting block and can be verified:
The Troglobith Web, Alik'r Desert
Ruins of The Citadel of Char, Dragontail Mountains
Ruins of The Slajrn Farmstead, Dragontail Mountains
Ruins of Tower Moorhouse, Dwynnen
Ruins of Old Lysyna's Farm, Dwynnen
Ruins of Castle Churg, Dak'fron
The Tower of Hearthhouse, Wrothgarian Mountains
Ruins of Greenhart Hall, Wrothgarian Mountains
The Andard Quarry, Wrothgarian Mountains
The Tower of Hearthhart, Wrothgarian Mountains
Ruins of The Old Wickham Shack, Wrothgarian Mountains
Tower Hearthton, Wrothgarian Mountains
The Tower of Mordyctor, Glenpoint
The Citadel of Wicksley, Glenpoint
The Laboratory of Gondastyr, Anticlere
Balex Hollow, Anticlere
The Abbey of Asmeusunah, Anticlere
Darkenaus Hall, Anticlere
The Hold of Gav'kern, Lainlyn
Castle Kingcroft, Orsinium Area
Ruins of Old Barbara's Hovel, Orsinium Area
Ruins of Yeomsley Court, Northmoor
The Web of Anirugoth, Northmoor
Morgyna Cultus, Northmoor
The Tower of Ashham, Alcaire
Ruins of Hawkston Hall, Kambria
The Hold of Woodfield, Kambria
The Citadel of Ashston, Phrygias
The Citadel of Woodfield, Urvaius
Ruins of Old Elyzanna's Place, Ykalon
The Stronghold of Hearthcroft, Ykalon
Hawkfield's Guard, Ykalon
The Laboratory of Tristywyr, Ykalon
Castle Garomyth, Abibon-Gora
Castle M'arhtem, Tigonus
The Tower of Nhorc, Totambu
Ruins of Castle Stullten, Totambu
Castle Shrertke, Totambu
The Citadel of T'irge, Totambu
The Stronghold of Sarg, Totambu
The Stronghold of Nesh, Ephesus
[...]

PR #2130, opened 2021-05-23
Add a runtime method to fix invalid RDB blocks

df-ferital:
This fixes the following blocks:
N0000071.RDB: connect the lower western dead-end to the rest of the block, can be checked in Ruins of Kinghouse Grange, Glenpoint
W0000009.RDB: move the exit to another corner (classic applies another but less clean trick), can be checked in The Tower of Hearthhouse, Wrothgarian Mountains
W0000018.RDB: connect the upper rooms together and fix a wrong door model, can be checked in the The Roost of Grumerus, Wayrest
W0000020.RDB: connect the lower western dead-end to the rest of the block, can also be checked in The Roost of Grumerus, Wayrest

Supersedes #2125 and #2126

### DIFF
--- a/Assets/Scripts/API/BlocksFile.cs
+++ b/Assets/Scripts/API/BlocksFile.cs
@@ -494,6 +494,130 @@ namespace DaggerfallConnect.Arena2
             return found;
         }
 
+        /// <summary>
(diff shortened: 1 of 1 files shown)


Interkarma:
This is amazing work. Thank you @df-ferital! :)

PR #2131, opened 2021-05-24
Fix automap update not triggering on some occasions

df-ferital:
Fix a regression introduced in #2100 and fix a historical issue that could happen when fast travelling to a location with the same name as the current one (for example, from Ashford Manor, Anticlere to Ashford Manor, Dwynnen).


KABoissonneault (approved):

petchema:
Fixed the issue mentioned in #2100 for me

Interkarma:
Thank you for fix. :)

PR #2133, opened 2021-05-30
Add formula for bank loans

ajrb:
Also exposes some for overriding that were not before.

### DIFF
--- a/Assets/Scripts/Game/DaggerfallBankManager.cs
+++ b/Assets/Scripts/Game/DaggerfallBankManager.cs
@@ -537,7 +537,7 @@ namespace DaggerfallWorkshop.Game.Banking
                 result = TransactionResult.LOAN_REQUEST_TOO_HIGH;
             else
             {
-                BankAccounts[regionIndex].loanTotal += FormulaHelper.CalculateBankLoanRepayment(amount);
+                BankAccounts[regionIndex].loanTotal += FormulaHelper.CalculateBankLoanRepayment(amount, regionIndex);
                 BankAccounts[regionIndex].accountGold += amount;
                 bankAccounts[regionIndex].loanDueDate = DaggerfallUnity.Instance.WorldTime.DaggerfallDateTime.ToClassicDaggerfallTime() + loanRepayMinutes;
             }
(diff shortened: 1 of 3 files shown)


Assets/Scripts/Game/Formulas/FormulaHelper.cs
```diff
@@ -316,6 +317,10 @@ public static int CalculateClimbingChance(PlayerEntity player, int basePercentSu
         // Calculate how many uses a skill needs before its value will rise.
         public static int CalculateSkillUsesForAdvancement(int skillValue, int skillAdvancementMultiplier, float careerAdvancementMultiplier, int level)
         {
+            Func<int, int, float, int, int> del;
```
KABoissonneault, line 320: In my opinion, this formula would be more useful to mods if it just passed the `DFCareer.Skills` enum. 

In PlayerEntity.cs, the code does
```
int skillAdvancementMultiplier = DaggerfallSkills.GetAdvancementMultiplier((DFCareer.Skills)i);
                float careerAdvancementMultiplier = Career.AdvancementMultiplier;
                int usesNeededForAdvancement = FormulaHelper.CalculateSkillUsesForAdvancement(skills.GetPermanentSkillValue(i), skillAdvancementMultiplier, careerAdvancementMultiplier, level);
```

The permanent skill value, the skill advancement multiplier, and the career advancement multiplier, and the current level can all be acquired from a reference to the player entity, which can be found in the GameManager.

Suggested fix: remove the lines from `PlayerEntity.RaiseSkills`, put them inside the default formula implementation, change the signature to only take a `DFCareer.Skills` enum, and adjust the override in consequence

Assets/Scripts/Game/DaggerfallBankManager.cs
```diff
@@ -539,11 +533,11 @@ private static TransactionResult BorrowLoan(int amount, int regionIndex)
             TransactionResult result = TransactionResult.NONE;
             if (amount < 100)
                 result = TransactionResult.LOAN_REQUEST_TOO_LOW;
-            else if (amount > CalculateMaxLoan())
+            else if (amount > FormulaHelper.CalculateMaxBankLoan())
```
KABoissonneault, line 536: Would it be useful to pass the region index to the formula, or should mods just get the current region from the PlayerGPS?

Assets/Scripts/Game/Formulas/FormulaHelper.cs
ajrb: Going to leave this one as it is, I don't see much value in this change and I don't want to mess with this formula without good reason. It's been as it is for a long time and working as classic (we think)

Assets/Scripts/Game/DaggerfallBankManager.cs
ajrb: Yes, agree.

PR #2135, opened 2021-06-02
Preserve rumors line breaks on bulletin boards

petchema:
Keep the line breaks from rumors for display on town bulletinboards; It's not guaranteed to be perfectly formatted, but is better than current concatenated text on one line. It's also most likely a step in the right direction for text adjustment to display width, since it preserves text as text tokens.

By the way I had a hard time testing to PR, I found few towns with rumors

### DIFF
--- a/Assets/Scripts/Game/PlayerActivate.cs
+++ b/Assets/Scripts/Game/PlayerActivate.cs
@@ -665,15 +665,15 @@ namespace DaggerfallWorkshop.Game
             };
 
             // formatting message is split into 2 parts, depending whether we got any news or not.
-            if (bulletinBoardMessage != string.Empty)
+            if (bulletinBoardMessage != null)
             {
                 tokens.AddRange(new List<TextFile.Token>
                 {
                     new TextFile.Token(TextFile.Formatting.NewLineOffset, null),
                     new TextFile.Token(TextFile.Formatting.Text, string.Empty),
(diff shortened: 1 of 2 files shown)


KABoissonneault:
FWIW, I think the rumor system needs a redo to be on par with classic­. We did some tests, and found that there were major discrepancies between the kinds of rumors you get in classic vs DFU, But maybe I should raise that as a separate issue

Assets/Scripts/Game/PlayerActivate.cs
```diff
@@ -665,15 +665,15 @@ void ActivateBulletinBoard(RaycastHit hit, DaggerfallBulletinBoard bulletinBoard
             };
 
             // formatting message is split into 2 parts, depending whether we got any news or not.
-            if (bulletinBoardMessage != string.Empty)
+            if (bulletinBoardMessage != null)
```
KABoissonneault, line 668: What about `string.IsNullOrEmpty`?

petchema: that's no longer a string, that's a TextFile.Token[]

KABoissonneault: I see, I didn't scroll up and didn't think the type could have changed; because of `var`

KABoissonneault (approved):

Interkarma:
Thank you Pango. :)

Issue #2136, opened 2021-06-02
Rumor mismatch between DFU and classic


KABoissonneault:
There seems to be a mismatch between DFU and classic in terms of what shows up when you ask for "Any News?".

For example, we ran some tests in classic with two characters, asking for 15 rumors.

Here were the results
```
Fresh PC : 

Generic : 0
Region conditions : 7 (ex: regime changes, alliances, etc...)
Tutorial Quest: 3
Main Quest : 5 (ex: Don't walk the Daggerfall streets at night)

PC with 2 quests active and other recently done quests : 

Generic : 0
Main Quest : 1
Post-quest : 4
Current quests : 10
```

Meanwhile, in DFU, regardless of context, alliance changes seem to pop up more often than anything else.
I compared the RUMOR.DAT from both classic playthroughs above, and they definitely didn't generate as many rumors as one would find in ConversationData.txt in the equivalent DFU playthrough.

So I suspect we have two issues:
1) we generate conditions at a higher frequency than classic
2) we lack a priority for quest rumors

Does anyone have any info on the investigations / resources that were used for the implementation in TalkManager and PlayerEntity?

Interkarma:
This feature was implemented by a one-time contributor and just hangs off the conversation rumour mill with some basic filtering. If someone wants to perform the reverse engineering or refinement, we can incorporate that. Otherwise, I'm happy for it to remain in a "best effort" state. It's not like this is a critical gameplay feature or particularly useful in classic either. It's just a dash of flavour.

KABoissonneault:
Alright, then I'll give it a shot this month, before the code freeze

Interkarma:
That's OK, provided changes aren't too broad. My goal with notifying of upcoming freeze was for contributors to wrap up anything they already had in progress - not provoke a rush to fit in as many changes as possible before cutoff. :) I still want long-term stable code to undergo relatively few changes before the freeze starts next month. A large part of the final push to 1.0 is my confidence in existing systems.

So minor refinements to the notice boards specifically are fine, but a refactoring of the entire rumour system might not be depending on the scale you have in mind. Just so you know before committing too much time into this.

Interkarma:
Oh, I just realised you were speaking to the rumour mill in general, not just bulletin board notices. Not sure how I got that idea. Please disregard my previous comment about bulletin boards.

[...]

PR #2137, opened 2021-06-04
Quest rumor experimental setting and DaggerfallTalkWindow mod support

KABoissonneault:
In light of issue #2136, I suggest two changes that would be more appropriate for this stage of development.

1) An experimental setting which makes quest rumors appear at least once (even if just one variant) before falling back to the normal rumor mill
2) Mod support for the DaggerfallTalkWindow

For the setting, there should be no changes when the setting is left to false. When enabled, new quest rumors are added to a separate "quest rumor list". When asking for "Any News?", if the "quest rumor list" is not empty, it grabs one (at random), removes it from the "quest rumor list", and shows that one. Note that quest rumors still exist in the normal rumor mill, and may appear again after the quest rumor list has been exhausted.

To be safe, I chose to only add quest rumors to that list when the setting is enabled, but to always try to remove them, even when the setting is off. Ideally, this list should never grow too large.

I believe just this setting might satisfy fans of quest rumors, and if not, well, it's just an experimental setting.

For the mod support, we're kind of fortunate that rumors have little to no impact on the rest of the gameplay. By changing the Talk window itself, mods can basically do whatever they want with all those topics and rumors. The TalkManager itself has no mod hooks, but the window doesn't really have to interact with it if it doesn't want to. 

My changes should only be two things: adding `protected` and `virtual` to existing code, and "trivial refactors" in order to make the code a bit cleaner. This not only helps future maintainers, but also, modders like copy-pasting the window code into their own window, so this should reduce noise for them.

I also changed the order of the Setup code a bit, and refactored it all into functions. Hopefully, this should reduce the amount of copy-paste derived windows need to do.

The DaggerfallTalkWindow changes might be easier to understand by reviewing individual changelists rather than the full review

### DIFF
--- a/Assets/Scripts/Game/TalkManager.cs
+++ b/Assets/Scripts/Game/TalkManager.cs
@@ -193,7 +193,7 @@ namespace DaggerfallWorkshop.Game
         NPCData npcData;
 
         // type of npc talk partners for a conversion
-        enum NPCType
+        public enum NPCType
         {
             Static,
             Mobile,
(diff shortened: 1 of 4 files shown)


Assets/Scripts/Game/TalkManager.cs
```diff
@@ -1329,51 +1355,72 @@ public string GetNewsOrRumors()
             const int outOfNewsRecordIndex = 1457;
             if (npcData.numAnswersGivenTellMeAboutOrRumors < maxNumAnswersNpcGivesTellMeAboutOrRumors || npcData.isSpyMaster || consoleCommandFlag_npcsKnowEverything)
             {
-                string news = TextManager.Instance.GetLocalizedText("resolvingError");
```
KABoissonneault, line 1332: This code was simply moved to the "else" branch below. I had to do it this way because of C# variable shadowing rules...

Interkarma:
I'm not comfortable with the changes here. There's a lot more going on than discussed and it looks like a lot of code is being moved around just for the sake it.

I'm still willing to review the idea of prioritising quest rumours, but I'd like to a see a bare minimum implementation first.

PR #2138, opened 2021-06-05
Add mod script asset template

rossipaolo:
Adds menu item "Create > C# Mod Script" to quickly create new mod scripts.

createModScript


Interkarma:
A simple thing that should be very helpful to new modders. Good idea. 👍

PR #2139, opened 2021-06-06
Added new "PrioritizeQuestRumors" experimental setting

KABoissonneault:
Attempt at redoing PR #2137 but with less changes.

I removed the Talk window changes (I'll make a separate PR), and I changed the variable names to minimize the line count of the changes.

### DIFF
--- a/Assets/Resources/defaults.ini.txt
+++ b/Assets/Resources/defaults.ini.txt
@@ -120,6 +120,7 @@ TerrainHeightmapPixelError=5
 SmallerDungeons=False
 CustomBooksImport=True
 AssetCacheThreshold = 25
+PrioritizeQuestRumors=False
 
 [Enhancements]
 LypyL_GameConsole=True
--- a/Assets/Scripts/Game/TalkManager.cs
+++ b/Assets/Scripts/Game/TalkManager.cs
@@ -353,6 +353,7 @@ namespace DaggerfallWorkshop.Game
         }
         // list of rumors in rumor mill
         List<RumorMillEntry> listRumorMill = new List<RumorMillEntry>();
+        List<RumorMillEntry> listQuestRumorMill = new List<RumorMillEntry>(); // Quest rumors that we haven't seen yet
 
         // questor post quest message stuff (QuestorPostsuccess, QuestorPostfailure)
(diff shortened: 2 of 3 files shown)


Interkarma:
Thanks for splitting this out. It really helps with review and tracking any subtle problems if they emerge after release. :)

Interkarma:
I've found this setting does not discriminate between live quest rumours and post-failure / post-success quest rumours. For example, the following rumours will be prioritised after starting Azimuthal Vectors (N0B00Y17) despite quest still being active. These should only be displayed once quest has been completed/failed. You'll also note the macros aren't populated correctly.

image

image

AFAIK everything works OK in live master, including macro expansion (this is after ending quest)

image

Overall, I'm not too sold on this solution.  I understand the desire for higher priority to quest rumours, but perhaps the approach needs rethinking?

KABoissonneault:
Yeah, maybe I'll just mod the Talk window with the other PR after all to get the result I want.

Interkarma:
I was thinking about this one over the weekend. Another approach would be to replace the Random.Range() call in GetNewsOrRumors() to use a weighted random function instead - one that gives quest-based rumours a higher weighting than ambient rumours. This has the benefit of working fully within the talk system logic rather than diverting it.

The RumorMillEntry class that makes up validRumors carries the questID, so that alone could be used to give the quest rumours a higher weight in the weighted random implementation.

Having a look around, I found a few different ways to implement a weighted random selection - some more efficient than others. But as it's a generally small list with a simple weighting system (quest > ambient), and it's only generated intermittently, then even a basic weighted random setup should be OK.

It's OK if you want to implement another way via the other PR in next release. But I'm still happy to explore the concept of this PR further if you like. :)

KABoissonneault:
Yup, I like this solution. For sure, the algorithm will be inconsistent with classic, but in my opinion, the user experience will be closer to classic, so it's worth doing. 

Perhaps you could have `questRumorWeight` be derived from a settings.ini-only value, so users can tweak their preference? I doubt it's going to be easy to find the right value

Interkarma:
That's a good idea. In general the ambient rumours create a pool of about 100-120 give or take, so a starting value of 50-60 is a pretty good weighting for most cases. Then users can tweak up and down based on their preference.

I'll merge that other PR of yours and pop this in today. I'll make sure the weighted method is protected so you could even override it later if needed.

Cheers!

PR #2140, opened 2021-06-06
Exposed more functions and data in the Talk window 

KABoissonneault:
The Talk window is inside the UI window factory, and therefore can technically be overriden by mods, but there's not really anything useful they can change.

Changes minimized compared to the version in PR #2137.


Interkarma:
I don't see any issues with this one. Throwing open to @Nystul-the-Magician and @df-ferital in case they have any concerns. If not, happy to merge before next release.

Nystul-the-Magician:
I think this is safe to merge ;)

PR #2143, opened 2021-06-11
Clear location cache if world data variant is set

ajrb:
Also don't cache at all when running in unity editor in the same way the world data caching is disabled in editor. This allows mod developers to see their world data changes without needing to restart the game. The downside to this is to debug issues with the cache the directive needs to be modified.

### DIFF
--- a/Assets/Scripts/Utility/AssetInjection/WorldDataVariants.cs
+++ b/Assets/Scripts/Utility/AssetInjection/WorldDataVariants.cs
@@ -85,6 +85,7 @@ namespace DaggerfallWorkshop.Utility.AssetInjection
                 locationVariants[locationKey] = variant;
 
             Debug.LogFormat("Set variant \"{0}\" for the location index \"{1}\" in region {2}", variant, locationIndex, regionIndex);
+            RMBLayout.ClearLocationCache();
             return overwrite;
         }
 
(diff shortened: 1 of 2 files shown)


ajrb:
@df-ferital Are you okay with these changes? If you're happy I'll go ahead and merge.

df-ferital:
Hi @ajrb, I'm not familiar enough with world data variant, so I don't understand why the cache needs to be cleared in this case. Can you explain this?

Regarding the editor, I would prefer an option to disable location caching because I'm mostly playing DFU in the editor on my side, however I think this is OK for now, we can add such an option later.

ajrb:
> 
> 
> Hi @ajrb, I'm not familiar enough with world data variant, so I don't understand why the cache needs to be cleared in this case. Can you explain this?

Changing a variant switches the world data override being used which the data is potentially being cached in the location cache. This ensures that no stale data can be in the cache. It's a very occasional event, in fact currently only one can happen per game from my R&R mod. There can still be issues if the change is close to the player and the change is already rendered in the world, but that's just something modders need to bear in mind when using variants, and is nothing to do with the cache you added.

> Regarding the editor, I would prefer an option to disable location caching because I'm mostly playing DFU in the editor on my side, however I think this is OK for now, we can add such an option later.

You can comment out the directives if you need to test with it active. Not sure if there's a more refined way to switch it.

df-ferital:
Thank you for your explanation! I am perfectly fine with these changes then.

df-ferital (approved):

PR #2145, opened 2021-06-14
Match quest NPCs location scopes and genders to classic

df-ferital:
Quest NPCs location scope and gender information is available from QBN data, however it was not correctly understood until now. After having reverse engineered classic original function used to setup a quest NPC, I made a small program to extract the information from QBN data and patch DFU quests. This PR is the result of this patch, which also required to modify `Person.cs` in order to parse the new location scope information.

This should fix all quests reported by Tamalak in this post: http://forums.dfworkshop.net/viewtopic.php?p=52023&sid=9dce69880ec5c26f698316ab5c0a47d6#p52023

I also took this occasion to fix some incorrect quest variables names and to restore quest S0000988 to its original state, where several NPCs should be able to send a letter regarding Numidium. The DFQFIX patch which was integrated in DFU was intended for classic but DFU does not have the same engine issue, so the original quest should work.

Some other modifications I made:
50C00Y00: Set `_qgfriend_` to `female` even if the gender is random in classic, as it matches the description.
80C0XY00: Set `_qgfriend_` to `male` for the same reason.
B0B50Y11: Remove `_nobleman_` gender as there is no reason to use a specific gender to this NPC.

@rossipaolo This adds the "remote" and "local" attribute to a quest Person, so you may need to update your VS Code template when this PR gets merged.

### DIFF
--- a/Assets/Scripts/Game/Questing/Person.cs
+++ b/Assets/Scripts/Game/Questing/Person.cs
@@ -176,6 +176,7 @@ namespace DaggerfallWorkshop.Game.Questing
             string genderName = string.Empty;
             int faceIndex = -1;
             bool atHome = false;
+            string locationScopeName = string.Empty;
 
             base.SetResource(line);
 
@@ -187,6 +188,7 @@ namespace DaggerfallWorkshop.Game.Questing
                                      @"faction (?<factionAlliance>[a-zA-Z0-9'_.-]+)|" +
                                      @"group (?<careerAlliance>[a-zA-Z0-9'_.-]+)|" +
                                      @"(?<gender>female|male)|" +
+                                     @"(?<locationScope>local|remote)|" +
                                      @"(?<atHome>(atHome|athome))";
 
             // Try to match source line with pattern
@@ -230,6 +232,11 @@ namespace DaggerfallWorkshop.Game.Questing
                     if (genderGroup.Success)
                         genderName = genderGroup.Value;
 
+                    // Location
+                    Group locationGroup = option.Groups["locationScope"];
+                    if (locationGroup.Success)
+                        locationScopeName = locationGroup.Value;
+
                     // At home
                     Group atHomeGroup = option.Groups["atHome"];
                     atHome = atHomeGroup.Success;
@@ -262,7 +269,7 @@ namespace DaggerfallWorkshop.Game.Questing
                 AssignGender(genderName);
                 AssignHUDFace(faceIndex);
                 AssignDisplayName();
-                AssignHomeTown();
+                AssignHomeTown(locationScopeName);
 
                 // Is NPC at home?
                 isIndividualAtHome = atHome;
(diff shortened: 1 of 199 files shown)


Assets/Scripts/Game/Questing/Person.cs
```diff
@@ -620,44 +627,26 @@ void AssignHomeTown()
                 }
             }
 
-            // For other NPCs use default scope and building type
-            Place.Scopes scope = Place.Scopes.Remote;
-            string buildingTypeString = houseString;
+            // For other NPCs use the given scope if any, else generate it at random
+            if (string.IsNullOrEmpty(scopeString))
```
KABoissonneault, line 631: Should we treat Questor as always local?

Assets/StreamingAssets/Quests/$CUREVAM.txt
```diff
@@ -235,9 +235,9 @@ QBN:
 Item _letter_ letter used 1020
 Item _token_ bracelet
 
-Person _vamp1_ face 40 factiontype Vampire_Clan
-Person _vamp2_ face 4 factiontype Vampire_Clan
-Person _hunter_ face 1 factiontype Knightly_Guard anyInfo 1011 rumors 1012
+Person _vamp1_ face 40 factiontype Vampire_Clan remote
+Person _vamp2_ face 4 factiontype Vampire_Clan male remote
```
KABoissonneault, line 239: Any reason to make `_vamp2_` male? Message 1023 uses %g and %g2 to refer to them, so I assume they were intended to be either gender.

Assets/Scripts/Game/Questing/Person.cs
df-ferital: Questor is mostly local except when you get the quest through a letter as with the main quest or vampire quests

Assets/StreamingAssets/Quests/$CUREVAM.txt
df-ferital: I'm not sure, if it's set as male then it was _probably_ intended, but as always it can be an issue in the original quest. I only fixed those that were obvious flaws.

KABoissonneault: Oh, all those quest files have been regenerated from a tool? I guess I won't review them all, then

df-ferital: Yes all quests are updated using a dedicated tool of my own which extracts the necessary fields from QBN data and add them back to DFU in TEMPLATE quest format.

KABoissonneault (approved):

Interkarma:
This is incredibly impressive work @df-ferital! Thank you for digging into this and finding a way to correct DFU's behaviour. 

It's obviously not possible to re-test every quest in the time I have available, I have tested some at random and everything looks good so far. I also have faith in these changes considering your past experience with this and that you built a custom tool to patch these quests cleanly.

I'll merge this one now so that changes will be in the hands of users for last stages of beta. Thanks again! :)

PR #2146, opened 2021-06-14
AI pathfinding issues when getting shot at

KABoissonneault:
Fixed enemy AIs treating arrows and spell projectiles as obstacles, causing them to shortly initiate a detour every time they get shot at.

I started investigating as part of https://forums.dfworkshop.net/viewtopic.php?f=5&t=4890. As mentioned in the thread, I noticed that the CapsuleCast detects triggers, and therefore arrows were detected but not handled as "not an obstacle" in the EnemyMotor, so the AI treated them as an obstacle. Even if the CapsulaCast didn't detect triggers, spell projectiles are solid, and would be detected as an obstacle anyway.

Pango suggested I use the layer mask, as used for detecting a clear path for shooting. This worked well enough in my tests. 

Here's the save I used for testing. It's just a level 1 Archer at the start of Privateer's Hold.
[SAVE29.zip](https://github.com/Interkarma/daggerfall-unity/files/6650169/SAVE29.zip)

I go up the stairs, get to the bat and imp, wait for either of them to start chasing me, then run backward down the corridor I just came from while shooting (the imp was easier to test with since it doesn't die from Iron arrows).

Without the fix, the imp constantly moves left or right for a short time after every shot. 
With the fix, the imp only gets a slight knockback from the attack, but keeps moving in a straight line otherwise.

### DIFF
--- a/Assets/Scripts/Game/EnemyMotor.cs
+++ b/Assets/Scripts/Game/EnemyMotor.cs
@@ -67,6 +67,7 @@ namespace DaggerfallWorkshop.Game
         float strafeAngle;
         int searchMult;
         int ignoreMaskForShooting;
+        int ignoreMaskForObstacles;
         bool canAct;
         bool falls;
         bool flyerFalls;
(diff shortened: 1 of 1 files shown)


KABoissonneault:
Forgot to mention in the OP, but I'm not sure what DFU entities use the "Ignore Raycast" layer. I put it in the "ignore mask" for the obstacle check since it seemed appropriate, but it's possible it would have unintended side effects.

Feel free to weigh in if you know more about this.

Interkarma:
This works perfectly. Great work, and thank you! :)

I'm not sure the "Ignore Raycast" layer is actually used by anything. I'm not concerned about it being used here.

PR #2147, opened 2021-06-15
Colorblind accessibility settings

KABoissonneault:
A player mentioned that they had trouble seeing taverns and shops in the exterior automap, due to red-green color blindness. Nothing else in the game seems to cause any issues.

There's many kinds of color blindness, and various degrees of it, so I thought letting each person tweak colors to their preferences is best. DFU already has a color picker UI used in the Settings.

This PR adds a new Settings tab, and 4 new color settings used by the ExteriorAutomap.
image

For example, changing the taverns to yellow and brightening the residences looks like this:
image

Note how the caption at the bottom is updated with the selected colors.

I checked with the colorblind user, and this was much better for them. I hope others will find this useful too.

Side note, but perhaps DFU could use a "restore defaults" button for the settings. Right now, the only way to restore them is to delete `settings.ini`.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallExteriorAutomapWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallExteriorAutomapWindow.cs
@@ -111,6 +111,10 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
 
         ToolTip buttonToolTip = null;
 
+        protected Panel panelTempleColor;
+        protected Panel panelShopColor;
+        protected Panel panelTavernColor;
+
         // these boolean flags are used to indicate which mouse button was pressed over which gui button/element - these are set in the event callbacks
         bool leftMouseDownOnPanelAutomap = false;
         bool rightMouseDownOnPanelAutomap = false;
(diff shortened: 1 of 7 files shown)


Assets/Scripts/Game/UserInterfaceWindows/DaggerfallExteriorAutomapWindow.cs
```diff
@@ -453,6 +457,19 @@ protected override void Setup()
             isSetup = true;
         }
 
+        protected virtual void SetupCaptionColors(Texture2D textureCaption)
+        {
+            for(int i = 0; i < 5; ++i)
```
KABoissonneault, line 462: I've considered having colored panels drawn on top of `panelCaption` rather than overwrite the texture, but they ended up being rendered off by one pixel _after_ Unity adjusts for resolution. Seems like it was the only way to get it to work correctly in all cases

Interkarma: Something to consider is that modders could replace the UI visuals here and new textures might not be read-write in import settings. And the image resolution would be different so injecting colours straight into texture maybe not the best method in the end.

I'd stick with the overlay method, even if it's off by a pixel. I find this is more of an issue in the editor than in builds. You could probably slightly enlarge the overlay in this case (even by 0.5 pixels in all directions) and it should improve things.

Interkarma:
Thank you. Good work! :)

After this is merged, we could look at moving a few existing options to the Accessibility tab to free up space. The "outline regional map locations" is primarily for colour blind players. Left hand weapons and toggle sneak also fall loosely into this tab. And it's a good spot to drop any future options like this.

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallExteriorAutomapWindow.cs
KABoissonneault: How about I check for `!textureCaption.isReadable` instead? From looking at `TextureReplacement.TryImportTexture`, this seems to be what's used to tell if the mod asset is not read-write.

Also, I made `SetupCaptionColors` virtual, in case a mod wants to change how the overlay is made, in case of different resolution or other issues.

Interkarma: I honestly prefer the overlay method to asking modders to add a new script for this. I'm happy to take a look as well to assist with overlay alignment if slightly enlarging doesn't help or still looks bad. There's a couple other tricks we can use here to make the overlay work.

Interkarma:
Just tested this, it looks great and works well. Thank you. 👍 

You mentioned adding a "defaults" button and I think this is a nice addition. Do you think a single button to reset all 4 is appropriate or a single reset button per picker?

KABoissonneault:
Honestly, I feel the whole settings menu needs a Reset button. But maybe having one for this section in particular is more essential, since recovering the original value manually for color pickers is much harder.

I'll go with a single button for all 4 for now.

KABoissonneault:
Here's what it looks like now.
image

I'm not sure how to do the width of the reset button honestly. I did it so it fits with the retro font in English. 
With localization, I think we'll start feeling the next to have it auto-resize based on text content, like `TextBox.CalculateCurrentSize`

ajrb:
> Thank you. Good work! :)
> 
> After this is merged, we could look at moving a few existing options to the Accessibility tab to free up space. The "outline regional map locations" is primarily for colour blind players. Left hand weapons and toggle sneak also fall loosely into this tab. And it's a good spot to drop any future options like this.

I agree, it would be good to de-clutter the other tabs a touch! :)

Interkarma:
Happy with this one! Thank you for the awesome work! :)

PR #2149, opened 2021-06-17
Pass face record id into mobile person asset

ajrb:
This allows mods to use the assigned face to change the look of the mobile to match.

@rossipaolo If you want I will add backwards compatibility, but I haven't because I'm pretty sure no one has released a mod yet that uses the MobilePersonAsset override so it would feel like adding cruft for no reason. 
(BTW Kamers villager mod VIO overrides the PopulationManager as far as I can see)


ajrb:
This will also require the example in the modding tutorial to be updated.

ajrb:
I think 19 days is long enough for a review, going to merge this and get the fixes to face indexes in ASAP.

PR #2150, opened 2021-06-18
Fixed guards accidentally not being above the player level

KABoissonneault:
`careerIndex` is `18` for the City watch, but `MobileTypes.Knight_CityWatch` is `146`.

See https://github.com/Interkarma/daggerfall-unity/blob/master/Assets/Scripts/Game/Entities/EnemyEntity.cs#L186


ajrb:
Looks like a good fix to me. Will merge over weekend unless anyone chimes in with an objection.

ajrb (approved):

Interkarma:
Nice catch! Thanks for fix. :)

PR #2151, opened 2021-06-22
Layout building exterior flats for RMBs

ajrb:
This allows world data building overrides to have exterior flats without needing to override the entire RMB. In classic data there only appears to be editor flats define, but have filtered these out as they're not used by any DFU systems.

This should change nothing unless world data overrides are being used.


KABoissonneault (approved):

ajrb:
Will need updating after PR #2155 has changed the signatures.

PR #2153, opened 2021-06-24
Use generic IEnumerables for 'AddItems' methods

jefetienne:
Small change to prevent using a needless "ToArray()" method when using a non-array[] collection for the 'AddItems' methods. Parameter takes in an IEnumerable instead, which are the superclass of all collections, including regular arrays[], Lists, HashSets, etc. Checking for its length is no longer needed when using a foreach loop after confirming that the collection is not null.


KABoissonneault (approved):
Not sure if this breaks precompiled mods though.

Interkarma:
>Not sure if this breaks precompiled mods though.

That's a fair concern. Otherwise no problems here. I'll merge for next release and we'll see how it goes.

PR #2154, opened 2021-06-24
Use FLATS.CFG to set gender of exterior NPCs

df-ferital:
Fix a game data related bug by using FLATS.CFG information to set exterior NPCs gender (mostly, if not only, witches). The issue was reported by @Ralzar81 [on the forum](http://forums.dfworkshop.net/viewtopic.php?f=5&t=4914&sid=d6db9aca862954531d2f267f04f01ff0).

### DIFF
--- a/Assets/Scripts/Game/StaticNPC.cs
+++ b/Assets/Scripts/Game/StaticNPC.cs
@@ -159,10 +159,20 @@ namespace DaggerfallWorkshop.Game
         }
 
         /// <summary>
-        /// Sets NPC data from RMB layout flat record. (exterior NPCs)
+        /// Sets NPC data from RMB layout flat record (exterior NPCs).
         /// </summary>
         public void SetLayoutData(DFBlock.RmbBlockFlatObjectRecord obj)
         {
+            // Gender flag is invalid for RMB exterior NPCs: get it from FLATS.CFG instead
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Game/StaticNPC.cs
```diff
@@ -159,10 +159,19 @@ public void SetLayoutData(DFBlock.RmbBlockPeopleRecord obj, int buildingKey = 0)
...
         /// </summary>
         public void SetLayoutData(DFBlock.RmbBlockFlatObjectRecord obj)
         {
+            // Gender flag is invalid in game data for exterior NPCs: get it from FLATS.CFG instead
+            int flatID = FlatsFile.GetFlatID(obj.TextureArchive, obj.TextureRecord);
+            FlatsFile.FlatData flatCFG;
+            if (DaggerfallUnity.Instance.ContentReader.FlatsFileReader.GetFlatData(flatID, out flatCFG) &&
+                flatCFG.gender.Contains("2"))
+            {
+                obj.Flags |= 32;
```
KABoissonneault, line 172: Should we clear the flag if FLATS.CFG says "male", or is it always male by default?

df-ferital: It seems to be always male but you're right, it's better to clear the flag, I'll update the PR.

KABoissonneault (approved):

PR #2155, opened 2021-06-25
Fix exterior npc bug

ajrb:
See http://forums.dfworkshop.net/viewtopic.php?f=5&t=4768


KABoissonneault (approved):

Interkarma:
This is really fantastic. Thank you for tracking down and fixing!

PR #2156, opened 2021-06-26
Layout building exterior flats

ajrb:
This allows wold data building overrides to have exterior flats without needing to override the entire RMB. In classic data there only appears to be editor flats define, but have filtered these out as they're not used by any DFU systems.


ajrb:
had to create a new PR because the merge of master messed up

PR #2160, opened 2021-06-30
Modding new mobile textures

KABoissonneault:
As part of a greater effort to allow mods to add new enemies to DFU, this PR allows mods to add new "mobile" textures, and use them for MobileEnemy's MaleTexture/FemaleTexture.

On its own, the only use case I see would be for mods to give Class enemies distinct textures. For example, Mage and Healer both use 486/485, and one could be changed to use a new texture id beyond the 512 original Textures. It is my hope that eventually, this will be useful to add new MobileEnemy types too.

Here's an example
image

I'm not familiar with all the graphics features, but I tried to keep the emission textures working at least. Tell me if I'm forgetting something that's usually supported in mobile textures.

The extent of my testing is just adding a bunch of palette swaps of TEXTURE.486 as PNGs in the Textures folder of my test mod. Tell me if you have ideas for test cases I should consider.

### DIFF
--- a/Assets/Scripts/Utility/AssetInjection/TextureReplacement.cs
+++ b/Assets/Scripts/Utility/AssetInjection/TextureReplacement.cs
@@ -549,7 +549,7 @@ namespace DaggerfallWorkshop.Utility.AssetInjection
 
                 // If the archive has a Arena2 texture file, use it to get record and frame count
                 string fileName = TextureFile.IndexToFileName(archive);
-                var textureFile = new TextureFile();                
+                var textureFile = new TextureFile();
                 if (textureFile.Load(Path.Combine(DaggerfallUnity.Instance.Arena2Path, fileName), FileUsage.UseMemory, true))
                 {
                     // Import all textures in this archive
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Utility/AssetInjection/TextureReplacement.cs
```diff
@@ -547,32 +547,63 @@ public static Material GetMobileBillboardMaterial(int archive, MeshFilter meshFi
...
+
+                        importedTextures.Albedo[record] = frameTextures;
+                        if (importedTextures.EmissionMaps != null)
+                            importedTextures.EmissionMaps[record] = frameEmissionMaps;
+                    }
+                }
+                // Otherwise, check what files are available in the injected assets
+                else
+                {
+                    List<Texture2D[]> allAlbedo = new List<Texture2D[]>();
```
KABoissonneault, line 582: Unlike the branch above (ie: the existing code), I don't use the materialReader's cache. I wasn't sure how to do it without knowing the number of frames in the record ahead of time.
If anyone has any suggestions/ideas, do share them

Interkarma:
As I'm not heavily involved in the modding side of things, could I please ask for some help with review and feedback from @ajrb and @rossipaolo for this PR and #2163? Thank you.

KABoissonneault:
Gonna add some extra comments to help reviewers.

While certain features like item icons have allowed textures outside the values 0-511, mobiles have not, and this PR fixes that.
The biggest obstacle usually to getting other textures to work is when we rely on "Arena2" textures to get information on what we're loading, such as the number of records and number of frames per record. For mobiles, this is more important than for a single icon.

So first step is allowing DaggerfallMobileUnit to fail to get record sizes and frames from the Arena2 TextureFile, and wait for the TextureReplacement to load the modded achive. After this is done, we run the same routine, but using the archive we just loaded.

The second step is also letting `TextureReplacement.GetMobileBillboardMaterial` fail to load the TextureFile, and simply trying to load records and frames from the archive until it fails. Once a frame is missing, the record is done. Once a record is missing, we're done with the archive. The result should be exactly the same as if `TextureFile` was used to get the number of records and frames, except without the `MaterialReader` cache - except as noted in the comments, I haven't found how to get that working without knowing the number of frames per record ahead of time (ex: cache miss means frame does not exist, or frame should be loaded from disk?).

Assets/Scripts/Utility/AssetInjection/TextureReplacement.cs
Interkarma: @ajrb Mentioned it would be possible to use XML metadata for frame and record counts. That is a different approach however and I'd rather not delay this one any further at this stage.

Interkarma:
Happy to merge this one. It's a light touch again and your work is consistently good. Thank you. :)

I'm sure it can be iterated on some more after 1.0. For now, please just focus on fixing any issues with this PR if any come up.

PR #2161, opened 2021-06-30
Add frame speed divisor to abstract class

ajrb:
Otherwise mods adding custom mobile units get log spammed with null pointer exceptions. Not sure if this is the best correction but figured I'd take a stab at it.

Was broken by the fix for slow spells commit 47e2c588


KABoissonneault (approved):

rossipaolo (approved):

Interkarma:
Thank you mate. Sorry for making troubles here, and I appreciate the patch. This still works just as well for me.

PR #2162, opened 2021-07-01
Greatly reduce Lord Kavar spawns

JayH2971:
At current, there is no event more complained of than the Lord K'avar quest. This PR greatly reduces the spawn rate for Lord K'avar's soldiers.

### DIFF
--- a/Assets/StreamingAssets/Quests/M0B11Y18.txt
+++ b/Assets/StreamingAssets/Quests/M0B11Y18.txt
@@ -432,7 +432,7 @@ _S.12_ task:
 
 _conhouse_ task:
 	when _S.00_ and not _S.12_ 
-	create foe _F.03_ every 10 minutes 100 times with 10% success 
+	create foe _F.03_ every 30 minutes 5 times with 10% success 
 
 _S.14_ task:
 	injured _F.03_ 


ajrb:
This is a hefty nerf Jay, maybe a little too much? From potentially 300 archers in groups of 3 down to just 15 is quite a change. Maybe bump up the group size to 4 or 5 to compensate, or maybe increase success %? Note I've never actually played the quest so I don't know the context so stick to your guns if you feel it's right?

JayH2971:
I'm always happy to hear feedback, and I'll answer with how I've reasoned this so far :)

1. Where possible, quests should avoid spamming human enemies. Humans are the greatest source of wealth in the game, especially when DF is unmodded. A group of 10 enemies who carry anything above Mithril will likely deliver loot enough to equate to 200,000 gold or more, which essentially ends any material need for the rest of the game. 300 enemies for this quest was extremely unbalanced in this respect, and 15 is still high, but the essence of the quest should still remain.

2. The grouping of 3 archers is unique in Daggerfall, and I believe this is the only quest it's used. When you first encounter it, it's quite lethal and surprising. Once it starts repeating, the player gets the sense "Oh dear, I'm in something deep now." 3 or 4 of these attacks is sufficient to show that something very serious is happening. Increasing group size or encounter numbers would exacerbate the difficulty problem without increasing the theatrical "gravity" of it IMO. (We could increase the number once and again, or just make it two ambushes of 10 people, but by that point all survival goes out the window.) And returning to the point, if the first X groups of three archers failed to kill this adventurer, what sense is there in sending another Y-X groups of three archers?

3. As noted above, 3 archer enemies at the same time in Daggerfall can be extremely dangerous. They strike from afar, they appear out of sight due to the spawning system (thus precluding a lot of defensive actions), and they hit very hard. A player who chooses this quest unprepared will find the first attack to be a significant challenge. 5 encounters is enough to feel like a real victory for these players, while more than that seems like a more and more herculean task as it continues. (On that note, Daggerfall does well when it avoids epic, herculean tasks and just sticks to an ordinary life. Slaying armies and armies of people feels more Skyrim-ish than Daggerfall.)

4. I don't know that the purpose of this quest is to provide a significant challenge. It does show that we're pursuing a major noble who has significant connections to the underworld, but it's given at the second rank of the Fighters Guild, which means many unprepared players have stumbled into it unwillingly. There are a lot of ways to get into trouble in Daggerfall, but I think the player should have some sense of foreknowledge when that danger arrives. The Lord K'avar quest does not do this, but just throws the player into something far greater with no warning at all.

5. From 1997 Daggerfall players have always murmured about "that Lord K'avar quest" that we should always avoid unless we want a headache. While I've always heard people who are glad it's finally over, I haven't heard of anyone who feels like the quest is a triumph or a victory in any sense. Its narration is a flat line and filled with enemy spam, which interrupts the flow of the rest of the game full-stop (try training in the Mages Guild, there are the K'avar archers. Try going into a dungeon for the Dark Brotherhood, there are the K'avar archers.)

Gathering these thoughts and putting them in numerical form:

- 3 archers are generally very dangerous, probably more so for a rank 2 quest but still dangerous at any level.

- 5 attacks is enough to show that something serious is happening, without breaking the flow of the rest of the character's life. Usually by about 4 the player starts to ask "Okay, is this over now?" Ending it at 5 gives a good sense of finality, right as the question starts to spring up.

- Potentially, I could change the spawn to something a bit more firm like 120 minutes with 25% chance. However, making the timing too firm removes that element of surprise which assassins should be using. I'm satisfied with the spawn rate as submitted, since the unevenness of it shows that these are various packs of mercenaries arriving on their own schedules. 30 at 10% right now means the spawns will probably come in fairly quickly (a few days at most) and then be over. Were we to space out the spawns too much (like 3600 minutes at 50% chance), the player might forget about their purpose entirely and wonder "Why are these three enemies spawning?" (A player confusing enemies from one quest for another has been a significant issue in the past.)

ajrb:
Fair enough, that's pretty well thought through to say the least. I'd have just taken your word for it though you know, but I do appreciate the explanation of your thought process. :-)

Interkarma:
Thanks for the detailed comments Jay. I appreciate the patch and all the thought you've put into this. Cheers!

Issue #2004, opened 2021-01-27
Expose ability to change mod settings in game


rossipaolo:
The ability to change settings for supported mods in game is currently only available with the console command [change_modsettings](https://github.com/Interkarma/daggerfall-unity/blob/ca423ca3e19b3afd2110de419213bd6b61f72624/Assets/Game/Addons/UnityConsole/Console/Scripts/DefaultCommands.cs#L2675) and i wonder how many people know that this is possible at all. It should probably be made available from UI in the pause window, like control settings.

Interkarma:
Happy to put a button in somewhere. What about under Advanced control setting? It's a custom UI layout and has plenty of space for a new button without trying to cram it into the main controls UI.

jefetienne:
I [mentioned](https://github.com/Interkarma/daggerfall-unity/pull/1774#issuecomment-608610697) an idea to have a small drop-down menu in the corner of the pause menu in my "Added advanced controls window" PR but unfortunately I don't believe I ever got a response from it. What do you guys think?

I personally feel that if it is a setting window that isn't directly related to controls or input, that it would be better to keep it out of the navigation from Pause window -> "Controls" button.

Interkarma:
I'm sorry for not replying to that one @jefetienne. Yes, I'd be happy with that solution, and prefer it to placing under controls. :)

rossipaolo:
I also like it!

jefetienne:
Glad to hear! I might not be able to work on this the same speed I did last year, but I can try and get started on it.

ajrb:
Not sure how useful this will be for a lot of mod settings, as they're read once when the mod inits and configure it a certain way. We may need a flag or something indicating dynamic settings to make it visible which can be changed in game?

Interkarma:
>Not sure how useful this will be for a lot of mod settings, as they're read once when the mod inits and configure it to run a certain way. We may need a flag or something indicating dynamic settings to make it visible which can be changed in game?

That's an excellent point. Not all mods will support runtime changes, and giving modders the ability to limit this will be needed.

ajrb:
I think modders are in control, the issue is more messaging to players so they know which settings will make a difference in game. Generally this will be pretty obvious to those of us in the know, but we could get a lot of noise from confused players wondering why changing a setting in game does nothing.

rossipaolo:
What do you think of a flag for each group of settings?

ajrb:
[...]

PR #2165, opened 2021-07-01
Added dropdown options in pause menu

jefetienne:
(Very last minute but in EDT it isn't July yet! Hopefully I make it through the feature lock ;))

Implements #2004

Added a dropdown menu in the pause window that by default will include access to the mod settings menu. I simply took the same logic as the static class ChangeModSettings in DefaultCommands, so hopefully this should suffice. Shortcut to toggle it can can be used with 'D'.

Additional options (button text and button click event) can be easily added to the dropdown via a method from DaggerfallUI called RegisterPauseOptionToDropdown that takes in a string for the button text and an Action to implement the logic once it has been clicked.

I'm making this a draft because I have been having a hard time trying to get the dropdown toggle be [moved to the very edge of the screen](https://imgur.com/a/mNcSb6P). When I add it to ParentPanel, the controls become very tiny and I haven't figured out how to properly scale it with AutoSize, Scale = NativePanel.LocalScale, etc. I would appreciate any help to get this [working as I want it to](https://imgur.com/a/JiYlk75). Additionally, if there are any other improvements to what I have down, please feel free to let me know.


ajrb:
How will mods know that their settings have changed though? Most load settings at initialisation so changing them mid game will not change anything. There are some I can think of in my mods where they cause overriding of behaviour that currently has no way to undo it without restarting the game.

TheLacus suggested a flag per setting so mod authors could mark dynamically changeable settings. I think this PR will need synchronising with that work.

jefetienne:
> TheLacus suggested a flag per setting so mod authors could mark dynamically changeable settings. I think this PR will need synchronising with that work.

Right, I will try and implement that part. Would the flag essentially be used a visibility flag in the mod settings window?

> Could the standard vanilla arrows work for this, or are the new ones needed for some reason?

I actually wasn't aware that they existed - I'd be happy to use them. Which textures would those be?

ajrb:
the ones I used for the item list scroller component were these, though there may be others more applicable for what you're doing. I just thought I'd mention it so you could consider what's most appropriate. :)

`        const string greenArrowsTextureName = "INVE06I0.IMG";   // Green up/down arrows when more items available
        const string redArrowsTextureName = "INVE07I0.IMG";     // Red up/down arrows when no more items available
`

I think the flag would need to indicate that a mod setting was dynamically changeable at runtime and default to false. My thought is that settings would be 'greyed' out and non-interacting during runtime of DFU unless the flag had been set by mod author indicating that it can be changed.

jefetienne:
Considering that this PR is essentially a front-end to an already existing feature in the console, and that the mod setting changes might be more involved, I will set aside those commits for a separate PR. I am still trying to figure out a solution to positioning the dropdown to the top-left corner of the screen, but this PR is feature-complete otherwise.

jefetienne:
I finally found a solution to get the dropdown to move it to the ParentPanel and have it scale properly with the NativePanel. I had issues with the ListBox scaling properly to be in sync with what my mouse was hovering over, so I had to do some modifications. I don't believe it caused any regressions after testing the Save and Talk windows.

It is a very messy fix, but I am not aware of a better way to do it. If there is, please let me know.

Interkarma:
>I had issues with the ListBox scaling properly to be in sync with what my mouse was hovering over, so I had to do some modifications. I don't believe it caused any regressions after testing the Save and Talk windows.

Those changes should be OK. I never considering parenting ListBox outside of native panels, so something like this is really necessary in this case. I'll try to find any problems, but I think it will be fine.

rossipaolo (approved):

PR #2166, opened 2021-07-02
Fix billboard animation so frame 0 is rendered first

ajrb:
Mostly important for one shot animations. Is there anywhere else than may have the increment before the first frame?

As reported here: https://forums.dfworkshop.net/viewtopic.php?f=5&p=55561


KABoissonneault (approved):

Interkarma:
Thank you Hazelnut. :)

PR #2170, opened 2021-07-09
Fix default binding accidental overrides

jefetienne:
[Fixes this forum post issue](https://forums.dfworkshop.net/viewtopic.php?f=5&t=4943)

Default bindings now only get set when the default KeyCode is no longer used in either binding page, and that it isn't used as a modifier. This fix also lead me to find another bug where not all modifiers were being found, which has also been resolved.

Also integrated idea from @KABoissonneault to have a list of primary bindings specifically removed by the player. While the primary fix should suffice, this idea will help keep the binding behavior expected.

### DIFF
--- a/Assets/Scripts/Game/ControlsConfigManager.cs
+++ b/Assets/Scripts/Game/ControlsConfigManager.cs
@@ -311,9 +311,6 @@ namespace DaggerfallWorkshop.Game
                     button.Label.Text = KeyCode.None.ToString();
                     SetUnsavedBinding(action, button.Label.Text);
 
-                    if (UsingPrimary)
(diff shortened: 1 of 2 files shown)


Interkarma:
So quick! You're a legend, thank you jefetienne. :)

jefetienne:
I was just about to write up a comment on the other post before I saw this comment. :)

@KABoissonneault I think I'll also add your idea to this PR. I had a hesitancy about it regarding how we handle "unknown" actions but I've realized last minute that it should probably be fine, since the explicit "none" list are only added through the right-click removals.

Interkarma:
Works great! Thank you. :)

PR #2173, opened 2021-07-10
Ensure only new locations are picked up from mods

ajrb:
This corrects an issue where a location override is also loaded and applied as a new location when in a mod package rather than loose files. Bad mistake on my part, but thankfully I found it while developing an update for my guild mod.

----
One other thing I noticed while investigating this issue is that disabling the mod didn't prevent this bug, which made me realise that the use of FindAssets is returning assets from and mod (as I expected) including mods that are disabled - which is not what I expected. I'd not noticed before now also causes my dungeon tester mod to put a location pixel dot on the travel map that has no name and you can't travel to, this is also because my World Data replacement code is finding the new location asset even though the mod is disabled.

@rossipaolo Could you take a look at FindAssets (line 449 of ModManager.cs) and say whether this is working as you expect please? I'm really not sure whether this is a bug in that code or if it's intended to return matching assets regardless of mod being enabled or disabled. If it's meant to work this way then I'll need to do a second pass and load each asset individually by name or something like that,

### DIFF
--- a/Assets/Scripts/Utility/AssetInjection/WorldDataReplacement.cs
+++ b/Assets/Scripts/Utility/AssetInjection/WorldDataReplacement.cs
@@ -159,8 +159,11 @@ namespace DaggerfallWorkshop.Utility.AssetInjection
                 {
                     foreach (TextAsset locationReplacementJsonAsset in assets)
                     {
-                        DFLocation dfLocation = (DFLocation)SaveLoadManager.Deserialize(typeof(DFLocation), locationReplacementJsonAsset.text);
-                        newBlocksAssigned &= AddLocationToRegion(regionIndex, ref dfRegion, ref mapNames, ref mapTable, ref dfLocation);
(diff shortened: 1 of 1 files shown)


rossipaolo:
@ajrb disabled mods are unloaded when the game start (should be when the Game scene is loaded) but i can add an explicit check, i don't think it hurts to have it.

Are you sure those assets are seeked after you can't access mod window and toggle mods anymore?

ajrb:
Ah, maybe that's it - this happens really early in the DFU init process. When the map data is enumerated on startup in ContentReader.EnumerateMaps() method, which happens when the content readers are setup after the arena2 path is loaded/set.

This happens before the settings screen where you can enable / disable mods. Sorry the penny didn't drop until you replied, but at this point the code doesn't know which mods the player will enable or not. I'll take a look at what could be done to change the ordering of all this setup, but new locations do really need to be read when the map dictionary is created.

I'll merge this and start a thread on the forums once I've looked into it. Thanks.

PR #2176, opened 2021-07-12
Fix a loading issue when EnumerateMaps() is run before mods enable/disable state is set.

ajrb:
As discussed on PR #2173, the EnumerateMaps() is running before the mod settings have been confirmed and thus partially loading new locations regardless of whether a mod containing them is enabled or disabled. (since that state is undefined at that point) New locations have to be included and checked for collisions when the map dictionary is created.

This PR moves the setting up of the map dictionary to happen the first time it needed rather than on construction of the ContentReader. I felt this was less intrusive code change wise, but it could be done differently if you don't like the check method calls.


Interkarma:
I think this is OK. No issues that I can find.

PR #2177, opened 2021-07-13
Autorun fixes

jefetienne:
First fix from [this comment](https://forums.dfworkshop.net/viewtopic.php?p=55752#p55752)

 - Manually running and enabling AutoRun will now continue to let the player run
 - Prevent enabling Autorun while moving backwards, which causes unexpected behavior where the player is not moving forward, but is in a running state unless if they press the 'Run' button or move backwards again.

### DIFF
--- a/Assets/Scripts/Game/Player/PlayerSpeedChanger.cs
+++ b/Assets/Scripts/Game/Player/PlayerSpeedChanger.cs
@@ -64,8 +64,6 @@ namespace DaggerfallWorkshop.Game
             currentRunSpeed = GetRunSpeed();
         }
 
-
-
         /// <summary>
         /// Record player input for speed adjustment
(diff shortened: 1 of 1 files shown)


KABoissonneault:
I suspect some players may appreciate having access to an "autowalk" feature (known as "rp walk" aka "roleplay walking" in certain spheres). Why not have the "AutoRun" button preserve the current speed? Walking then pressing Auto has you autowalking, and running then pressing Auto has you autorunning.

jefetienne:
I suppose I'm okay with that if everybody else is. I would rather change the name to "Auto Move" in that case.

Interkarma:
I'm happy with this one as it is for now with the shift toggle. Thank you for the fix @jefetienne. :)

PR #2178, opened 2021-07-16
Adds failsafe spawn to A0C01Y01

JayH2971:
Adds a new spawn mechanism after 11 PM to guarantee that the enemies will spawn during the quest. Spawning "every 10 minutes 1 times with 25% success" still gives a tiny dangling chance that the enemy will fail to spawn for no good reason.


Interkarma:
This is a good fix, thank you Jay!

PR #2182, opened 2021-07-20
Fixed missing "OnLootSpawned" event for house/shop containers

KABoissonneault:
Sorry for breaking the 0 PR pending, but

#2087 was meant to cover all spawned loot, and Ralzar noticed that I missed containers, which are spawned from PlayerActivate.

PlayerActivate generates loot for containers, both for house/shop "private property" crates and shop shelves. Shop shelves do cover modded items in a more automatic manner, but I put the event anyway for consistency.

image

image


Interkarma:
This is OK with me, thank you for fix. :)

PR #2183, opened 2021-07-21
Fix code mistake in building variant getter

ajrb:
When I added the code in feb I made a small mistake. Doesn't affect most uses since lastLocationKey is usually the correct one anyway, but needs correcting for future uses.

So not essential for 0.11.5 release, but could go in.


Interkarma:
Thank you for fix mate. The 0.11.5 builds are already cooked and I'm in final tests and writing up the change notes for release.

Will merge now so it goes into whatever next release is.

Interkarma:
I haven't tagged 0.11.5 yet. Rerunning builds to include this change and will tag 0.11.5 here.

PR #2188, opened 2021-07-25
Allow to ask about individual NPCs in some quests

df-ferital:
Fix all Daedra quests as well as Lhotun's part of the main quest to allow asking about Daedra princes and The Underking. The only exception is Namira who has no dialogue for anyInfo.

Also fixed some artifacts dialogue messages which where incorrectly associated with the `used` command instead of `anyInfo`.


Interkarma:
Thank you! I'll merge this one ahead of next release also.

PR #2189, opened 2021-07-26
Implements Baltham Greyman questline

JayH2971:
Thoroughly playtested... but I'm sure bugs will get reported. Bring on the reports!

I had to change the place symbols L.00 and L.01 to other ones I created. After that, the quest loaded fine! No idea why those symbols caused problems. But this is done.

### DIFF
--- a/Assets/StreamingAssets/Quests/N0B11Y18.txt
+++ b/Assets/StreamingAssets/Quests/N0B11Y18.txt
@@ -53,10 +53,10 @@ Message:  1020
 Message:  1021
 %qdt:
  I agreed to spare _rebel_'s life.
- %g said I should just tell _questgiver_
+ He said I should just tell _questgiver_
  that I found no sign of _rebel_ in 
  ___rebelhouse_, except for a letter which
- %g gave me.
+ he gave me.
 
 Message:  1030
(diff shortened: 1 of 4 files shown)


Assets/StreamingAssets/Quests/N0B11Y18.txt
```diff
@@ -53,10 +53,10 @@ Message:  1020
 Message:  1021
 %qdt:
  I agreed to spare _rebel_'s life.
- %g said I should just tell _questgiver_
+ He said I should just tell _questgiver_
```
KABoissonneault, line 56: Can't Baltham appear as female?

JayH2971: I've never known there to be a female Baltham Greyman. The enemy sprite can be male or female, but the flat is always of a male (it's a pre-defined NPC, like Gothryd or Barenziah). I had to fix those pronouns because DFU was trying to assign a female gender to the male Baltham Greyman. One alternative is to fix him with the "male" attribute, but I remember doing that a long time ago and it caused weirdness to happen elsewhere.

Interkarma:
I'll merge this one for 0.12.1 so users can help test. Thank you @JayH2971! :)

Issue #2190, opened 2021-07-26
Quest fix needed: A0C00Y16, Missing Person Case


JayH2971:
I'm aware I'm "the quest person," but there are times when a job is really, really beyond me. Anyone is free to pick this up and try to fix it. The sheer number of variables and paths through this quest makes it gigantic to debug and repair properly.

I'm not going to give a thorough explanation of what's broken in the quest (if I could, I'd fix it and be done with it), but a few of the problems are NPCs who should appear in buildings where they don't, quest topics that should be made available but don't, and paths that depend on certain actions to happen but don't. Whether people choose to collaborate or work on it separately is up to them.

This issue is for workflow's sake, so we don't forget about it by the time 1.0 comes around.

mundifex:
Once #2194 #2197 and #2198 are in place, I think you will find that A0C00Y16 functions in a much better fashion, at least comparing to vanilla Daggerfall. #2199 polishes the presentation a bit.

Question for @df-ferital: does #2198 correct the issue where reading the log entry for the quest automatically adds _missingperson_ as a dialog link? I found that, depending on which "path" the quest selects, you can discern whether the missing person actually exists (i.e. path _S.00_) by reading the log entry immediately after the quest begins: if you end up with 4 people entries, the missing person exists and can be found; 3 entries implies the alternate paths where you will never physically locate the missing person (i.e. paths _S.01_, _S.02_, _S.03_).

For reference, quest flow for each path:
- **S.00**: _friend1_ reveals _hidingplace_ which contains _missingperson_ NPC
- **S.01**: _friend2_ refers to _friend4_ who had Smiley murder missing person
- **S.02**: _friend2_ knows missing person got arrested
- **S.03**: _friend1_ refers to _friend3_ who knows that missing person skipped town

As soon as _missingperson_ appears as a Where Is topic, path _S.00_ is ruined by mobile NPCs revealing _hidingplace_. Ideally, I think we either want _missingperson_ to never show up as a Where Is topic (which hopefully #2198  facilitates) or only after _friend1_ has revealed _hidingplace_ which we could do with an additional `add dialog for person` statement in the quest file (something I did not add in #2199 since I wanted to discuss first).

df-ferital:
[...]

PR #2194, opened 2021-07-29
Fix placed NPCs reverting to home location

mundifex:
Partial fix for #2190, specifically "NPCs who should appear in buildings where they don't."

The Tick method in Person.cs contains logic that attempts to hot-place an NPC at its home location when the player arrives there. While this behavior only occurs once, it did not contain a check to see whether the NPC had already been placed, e.g. by a PlaceNpc action. This had the effect of permitting quantum NPCs that could exist in two places at the same time: both at their placed-at location and at their home location. Once the player visits the NPC's home location, however, the NPC will no longer appear at the placed-at location. These anomalies account for the strange NPC availability noted in the issue.

Test case: Quest A0C00Y16
* _missingperson_ is placed at _hidingplace_ but still has a home location defined
* player visits _hidingplace_ first, _missingperson_ is present until... 
* player visits _missingperson_home_, _missingperson_ auto-placed at home and stops being present at _hidingplace_
* _missingperson_ will only exist at home location for remainder of quest, soft-locking quest path _S.00_

Adding a check that `lastAssignedPlaceSymbol` has not already been assigned prevents NPCs being erroneously hot-placed back to their home locations when placed elsewhere.

### DIFF
--- a/Assets/Scripts/Game/Questing/Person.cs
+++ b/Assets/Scripts/Game/Questing/Person.cs
@@ -378,10 +378,11 @@ namespace DaggerfallWorkshop.Game.Questing
             base.Tick(caller);
 
             // Auto-assign NPC to home Place if available and player enters
+            // (but only if not already placed, e.g. by PlaceNpc action)
             // This only happens for very specific NPC types
             // Equivalent to calling "place anNPC at aPlace" from script
             // Will not be called again as assignment is permanent for duration of quest
-            if (homePlaceSymbol != null && !assignedToHome)
+            if (lastAssignedPlaceSymbol == null && homePlaceSymbol != null && !assignedToHome)
             {
                 Place home = ParentQuest.GetPlace(homePlaceSymbol);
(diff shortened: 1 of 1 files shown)


Interkarma:
Thank you for this! I don't believe it will cause any regressions. I can't find any quest where NPC's automatic home is being used after "place npc" is deployed to move them somewhere else specific. This looks like a solid bug fix.

@JayH2971 can you think of any cases where a quest moves an NPC somewhere specific using place npc then relies on NPC being present in their automatic home (without first moving them back).

JayH2971:
> @JayH2971 can you think of any cases where a quest moves an NPC somewhere specific using place npc then relies on NPC being present in their automatic home (without first moving them back).

Nothing comes to mind... I think (hope) it's standard form among quest writers to ensure the NPC is manually placed every time it needs to be somewhere. Famous last words, perhaps!

Dezzj (commented):
Veiwed

PR #2195, opened 2021-07-29
Add console var "talk_npcBehavior" to override behavior in dialogs

mundifex:
Added new behavior "AlwaysFriendly" to force NPCs to respond helpfully even if they don't like you numerically. This is particularly helpful when debugging quests and you simply want answers from a mobile NPC instead of playing whack-a-mole to find one that likes you; `talk_npcBehavior 3` gets you omniscient, affable NPCs wherever you go.

Refactored "talk_npcsKnowEverything" and "talk_npcsKnowUsual" commands to drive a variant of new flags-based enum field for behavior states.

Finally, added a know-everything check to `GetAnswerWhereIs` method to fix an issue where extant mobile NPCs did not know answers to Where Is questions if they were not omniscient when spawned (i.e. developer changes behavior state to KnowEverything after NPC spawned).


KABoissonneault (approved):

Interkarma:
This looks good and will be great for testing. Thank you!

Thank you also for your help with review @KABoissonneault!

Issue #2191, opened 2021-07-26
Quest fix needed: K0C00Y06, A Mix-Up


JayH2971:
I'm aware I'm "the quest person," but there are times when a job is really, really beyond me. Anyone is free to pick this up and try to fix it. The sheer number of variables and paths through this quest makes it gigantic to debug and repair properly.

I'm not going to give a thorough explanation of what's broken in the quest (if I could, I'd fix it and be done with it), but a few of the problems are NPCs who should appear in buildings where they don't, quest topics that should be made available but don't, and paths that depend on certain actions to happen but don't. Whether people choose to collaborate or work on it separately is up to them.

This issue is for workflow's sake, so we don't forget about it by the time 1.0 comes around.

KABoissonneault:
Unrelated, but personally I'd like to restore the classic rep gains for that quest. I was thinking of instead adding a new check that cuts the quest early if your legal rep is 10 or above (ex: someone off-screen vouches for you). Our current workaround removes the primary option for restoring a bad legal rep.

I'll have a look on the code side and see if there's anything we don't handle well. Maybe there's some more stuff we don't handle properly, such as the semantics of the "Local" factions, as Ferital discovered

JayH2971:
> I was thinking of instead adding a new check that cuts the quest early if your legal rep is 10 or above

That isn't extant in the current quest conditions, so the first step would be designing that (assuming the idea is accepted; I'm neither for nor against it).

I think the exaggerated nature of the quest opens itself to criticism. It's kind of silly that a mass murderer could get accused of stealing a gem, and once it's returned, you suddenly become a local hero to law enforcement. It's mechanically useful, but a really wild concept when taken with fridge logic. Daggerfall really needs more methods to improve legal rep, IMO.

mundifex:
Along with the fixes introduced for #2190, the quest is now very playable. It's still the abstruse beast it always was in classic but the usual suspects are where they should be, the witness is where you're told, and I was able to both clear my name and return the stolen gem across several attempts.

JayH2971:
My gosh. I had no idea this much hecking help would come along for this thing. I can't call it finished until IK makes the decision, but I'm very grateful for what's been seen and fixed so far. No wonder trying to resolve this on my own felt like sinking into quicksand!

Interkarma:
I believe K0C00Y06 is working well enough to close this one now. Please let me know if some problems exist that still need attention and we can reopen or create a new issue.

[...]

PR #2197, opened 2021-07-30
Fix assigned quest places getting re-assigned when setting up places

mundifex:
IsBuildingAssigned method was comparing site's mapId against location's ExteriorData but everywhere else site mapIds are derived from location's MapTableData instead. Change comparison to compare against proper value.

This PR represents another partial fix for #2190 that addresses errors such as:
* incorrect building names causing confusion on automap and in dialog messages
* created NPCs sometimes existing at the same location
* NPCs placed at locations sometimes reusing their home location (e.g. _missingperson_'s home and _hidingplace_ being the same physical location but having different names)

With this fix in place, all locations chosen for quest A0C00Y16 were unique (within scope of quest itself) in 10 separate tests.

### DIFF
--- a/Assets/Scripts/Game/Questing/Place.cs
+++ b/Assets/Scripts/Game/Questing/Place.cs
@@ -1202,7 +1202,7 @@ namespace DaggerfallWorkshop.Game.Questing
                 {
                     Place place = (Place)resource;
                     if (place.siteDetails.siteType == SiteTypes.Building &&
-                        place.siteDetails.mapId == location.Exterior.ExteriorData.MapId &&
+                        place.siteDetails.mapId == location.MapTableData.MapId &&
                         place.siteDetails.buildingKey == buildingSummary.buildingKey)
                         return true;
                 }
@@ -1214,7 +1214,7 @@ namespace DaggerfallWorkshop.Game.Questing
                 foreach (SiteDetails site in activeQuestSites)
                 {
                     if (site.siteType == SiteTypes.Building &&
-                        site.mapId == location.Exterior.ExteriorData.MapId &&
+                        site.mapId == location.MapTableData.MapId &&
                         site.buildingKey == buildingSummary.buildingKey)
                         return true;
                 }


KABoissonneault (approved):

mundifex:
While investigating #2191, I noticed a possible issue with this change.

For context, my test bed is Gothway Garden and I use `startquest` to invoke a quest after speaking to the barmaid in The Pig and Cat tavern. Gothway Garden has only 3 taverns:
- The Toad and Goat
- The Dwarf and Huntsman
- The Pig and Cat

Quest K0C00Y06 requests several tavern locations during setup and with my changes in place, it exhausts the available taverns very quickly. The Pig and Cat automatically gets assigned to symbol qgiver_home since I'm invoking the quest with the barmaid as quest giver. The Toad and Goat has no valid quest markers and is thus discarded in `CollectQuestSitesOfBuildingType`. Finally, The Dwarf and Huntsman gets assigned to symbol witness_home since witness belongs to group Innkeeper. When we  get to a local tavern for symbol _inn_ we have exhausted all available unique taverns.

Without my changes, the quest is able to start successfully, even in Gothway Garden. I'm not sure it _should_ though, hence this comment. I think my fix is a valid one, as it fixes issues where quest resources overlap and cause naming mishaps (e.g. A0C00Y16), but it means that some quests that previously worked are going to fail to start when there aren't enough places/resources available. Is that an acceptable cost for better quest reliability across the board?

Related question: since I'm not as familiar with quest instantiation as others, is this what the "Forget it, I gave the quest to some other sellword" type message was designed to solve? (i.e. instances where the quest machine runs out of resources and scraps the quest altogether)

KABoissonneault:
For that last question, yes, it is intended that quests fail from being offered if it does not compile or if resources cannot be allocated properly, and this message is what is used.

I think not offering a quest is better than offering a buggy one, so I'm all for it. I doubt the classic behavior was any better.

Interkarma:
Thank you! This looks like a clear mistake on my end. I might have started one way then changed to another and missed a spot. I'll do a little backtracking first on my end first to be sure, but I'm reasonably confident this is a straight fix.

@KABoissonneault is correct, classic also uses this comment when a quest can't be setup for one reason or another. I think it's perfectly valid for K0C00Y06  not to compile if there aren't enough local resources. I'd consider that two fixes for the price of one. Great work. :)

Interkarma:
Very confident this is a mistake on my end and not properly tested. This change makes the code work as intended. Thank you! :)

PR #2198, opened 2021-07-30
Fix "dialog links" and "create npc" to behave as classic quest opcodes

df-ferital:
While working on A0C00Y16 to fix issue #2190, I noticed so-called "dialog links" were not working correctly. Furthermore, there were other issues, like the fact than all NPCs homes were available as topics under the "Where Is > Locations > General" list. That's not correct at all since these houses should only be available when mentioned by other NPCs, and none of the NPC houses are mentioned at start.

I also reverse engineered classic regarding the `dialog link` and `add dialog` opcodes and it appears that Tipton's documentation for these opcodes was inaccurate. In fact `dialog link` should have been named `remove dialog` as that's the only function it does. When several quest resources are mentioned in a `dialog link` command, this only means that all of these resources need to be hidden, nothing more. There is no "link" between any of the resources, which means than hearing about one does not necessarily reveal any of the other, except if the quest explicitly uses the `add dialog` command to do so.

I corrected all of this, and I now have a perfectly working version of A0C00Y16, which I will submit as another PR. Of course I tested other quests to be sure that my changes do not break any stuff and it appears to be working pretty well.

One other thing I corrected is the fact that quest journal entries should not reveal anything regarding quest resources, only speaking with other NPCs or reading letters and other written quest items should do so.

### DIFF
--- a/Assets/Scripts/Game/Questing/Actions/DialogLink.cs
+++ b/Assets/Scripts/Game/Questing/Actions/DialogLink.cs
@@ -1,4 +1,4 @@
-﻿// Project:         Daggerfall Tools For Unity
+// Project:         Daggerfall Tools For Unity
 // Copyright:       Copyright (C) 2009-2021 Daggerfall Workshop
 // Web Site:        http://www.dfworkshop.net
 // License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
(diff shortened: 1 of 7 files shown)


Assets/Scripts/Game/Questing/Actions/JournalNote.cs
```diff
@@ -49,7 +49,7 @@ public override void Update(Task caller)
         {
             Message message = ParentQuest.GetMessage(id);
 
-            GameManager.Instance.PlayerEntity.Notebook.AddNote(new List<TextFile.Token>(message.GetTextTokens()));
+            GameManager.Instance.PlayerEntity.Notebook.AddNote(new List<TextFile.Token>(message.GetTextTokens(-1, false)));
```
KABoissonneault, line 52: Maybe you could use named parameters for readabililty. `message.GetTextTokens(variant:-1, expandMacros:false)`

Assets/Scripts/Game/Questing/Actions/DialogLink.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Project:         Daggerfall Tools For Unity
+// Project:         Daggerfall Tools For Unity
```
KABoissonneault, line 1: Maybe we could add new patterns that reflect the use case better (ex: "remove dialogue for location (?<aSite>\w+) person (?<anNPC>\w+) item (?<anItem>\w+)").

Assets/Scripts/Game/Questing/Actions/JournalNote.cs
df-ferital: Good idea, I'll update the PR.

Assets/Scripts/Game/Questing/Actions/DialogLink.cs
df-ferital: Yes, but it should put in another PR I suppose. I should document in this one what "dialog links" really means though.

mundifex:
Thank you so much for these fixes! I had uncovered the same problems while investigating A0C00Y16 but did not feel comfortable enough with the codebase to make these changes. Regardless, my findings were the same (i.e. dialog link should be "remove dialog", we were "leaking" house locations, etc.) and I'm glad to see them fixed.

Nystul-the-Magician:
when originally implementing dialog link I extended intentionally vanilla behaviour whenever dialog link was used with 2 ressources, so that if you ask for say a location and the name of a person resource is mentioned (and only if that name is really told) this person resource gets unhidden, so you can ask for it. This was intended since it was more realistic in my opinion back then. I feel bad seeing this being patched out

df-ferital:
> when originally implementing dialog link I intentionally extended vanilla behaviour whenever dialog link was used with 2 ressources, so that if you ask for say a location and the name of a person resource is mentioned in the answer (and only if that person's name is really told by your dialog partner in the answer) this person resource gets unhidden, so you can ask for it. This was intended since it was more realistic in my opinion back then. I feel unhappy seeing this being patched out
> I agree though, that quest journal should not reveal anything. So this was a missing edge case I guess arising from the changed mechanism (I think we had a hotfix already for this earlier if i remember correctly but my memories might fool me here)

The problem with this approach is that it breaks things for some quests, e.g. when an NPC location is supposed to be hidden. It also reveals NPCs residences in the "General" section of the dialog panel even if the player never heard about the location. For example, if a quest tells you took look for Gondyrick Coppersmith but does not tell you where to find him, you're not supposed to know that he's in The Woodham Residence until you actually learn that he lives there.

Nystul-the-Magician:
this originally worked - not sure what broke this behavior :(

df-ferital:
The problem is more general and does not concern only quest A0C00Y16. As I said in this PR description, Tipton misunderstood this particular classic quest opcode and incorrectly interpreted it as `dialog link` whereas it was only supposed to do the opposite of `add dialog`, that is to hide resources, without any "link" between the quest resources themselves. That's why I'm removing dialog links in this PR: it's intentional as it was not intended by classic and, furthermore, it can break quests by revealing things which are not supposed to be revealed. Instead, what this PR does is to recreate the original purpose of classic opcode, which is to hide resources, nothing more.

However, while I agree with you regarding the fact that classic is sometimes broken, it is broken because quests themselves were often broken (I mean that the QBN file was broken) or, sometimes, the fix provided by Tipton or others was also broken, because of the original misinterpretation of classic quest behavior. But classic mechanics for this particular cases (the `add dialog` and the so-called `dialog link` opcodes) are not broken, I reversed engineered both functions and none of them contain any bug.

This PR is still in a draft state though, and I don't have any time to work on it for now, but I plan to do to so in a few weeks.

Nystul-the-Magician:
I am curious if original devs patched out some functionality regarding talk window/related quest op codes in later patches of vanilla like 2.13 - I remember playing the original shipped version (from cd) of df in 1996 giving me slightly different talk options than when using the patched 2.13 version some months later back then. I remember wondering that with 2.13 suddenly some quests felt much easier since dialog options were available from quest start (especially in "where is" - section) that were not available before - maybe devs decided change some behavior here? would be interesting to see icf cd-version was different than 2.13 in this regards... would it be possible to take a look?

df-ferital:
Yes, this would definitely be possible, however it would be a lot of work. My current reverse-engineered code is built on top of decompiled global structures, variables and enums which are tied to the 2.13 executable. Looking for the talk system functions within the old executables would require to identify the same structures and variables and this would be a little tedious, though not impossible.

Nystul-the-Magician:
so rather not feasible since it woud take a lot of work :/

df-ferital:
Hopefully, I expect to get some time in July to finally finish this PR, a year after I opened it...

Dezzj (commented):
Veiwed

KABoissonneault:
@df-ferital Are you still willing to complete this? The fix seems reasonable, but I'd rather not keep around old PRs forever.

df-ferital:
@KABoissonneault I still want to complete it, but I don't have much time to work on it these days and as it's already almost 3 years old, it's going to take me a while to get back into the code. I guess you can close it for now, as it can always be reopened if needed.

PR #2200, opened 2021-07-31
Fix questor NPC not hiding when using startquest via console

mundifex:
LastClickedNPC quest resource behavior fixup only occurred when starting quest via popup window. Moving it within StartQuest method of QuestMachine ensures it will take effect for any code path that starts a quest. AssignQuestResourceBehavior method already ignores non-questor NPCs and will only create a single behavior component per NPC so this should be a safe fix globally.

Quest K0C00Y06 hides its questor NPC shortly after quest start. Prior to this fix, this would only occur if starting the quest from a dialog popup. With fix in place, questor now also hides properly when invoking quest via `startquest` console command.

### DIFF
--- a/Assets/Scripts/Game/Questing/QuestMachine.cs
+++ b/Assets/Scripts/Game/Questing/QuestMachine.cs
@@ -702,6 +702,13 @@ namespace DaggerfallWorkshop.Game.Questing
             quests.Add(quest.UID, quest);            
 
             RaiseOnQuestStartedEvent(quest);
+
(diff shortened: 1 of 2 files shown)


mundifex:
Related to #2191 but not strictly necessary for 1.0 since this change only fixes a development tool.

KABoissonneault (approved):

rossipaolo (approved):

Interkarma:
This looks like a clear fix again, thank you. :)

I seem to recall this working properly at one point early in quest development. Probably things moved around in a refactor somewhere and we never caught the regression. Really great work @mundifex.

PR #2202, opened 2021-07-31
Destroy extant NPC object when placing it somewhere player is not

mundifex:
In K0C00Y06, the _witness_ NPC is placed at _withouse_ immediately after calling for the guards.

The current implementation of a hot-remove in Place.cs tried to hide the NPC in such an instance but the Tick method of QuestResourceBehavior would un-hide the NPC immediately after. Destroying the NPC's gameobject instead allows for hiding the NPC in a manner that QuestResourceBehavior cannot revert. The NPC will re-instantiate when the player visits its new location.

The only instance in which I can see this causing an issue is if an NPC is placed at an alternate location via `place npc` and then placed back at its original location while the player is still there. If such an instance exists, it should be changed to use `hide npc` and `restore` instead.

### DIFF
--- a/Assets/Scripts/Game/Questing/Place.cs
+++ b/Assets/Scripts/Game/Questing/Place.cs
@@ -495,10 +495,10 @@ namespace DaggerfallWorkshop.Game.Questing
                 }
             }
 
-            // Hot-remove resource is player already at this Place and resource was moved elsewhere
+            // Hot-remove resource if moved somewhere player is not
             if (!IsPlayerHere() && resource.QuestResourceBehaviour)
             {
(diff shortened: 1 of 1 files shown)


mundifex:
This fixes the bug in #2191 where the _witness_ NPC does not permanently hide after calling for the guards.

Interkarma:
The current code definitely isn't handling logic gap where an NPC is placed elsewhere and hot-removed, but technically not hidden so they appear again. It's messier for QuestResource to be checking where the NPC is before trying to show them again, I prefer how you've tweaked this. I haven't been able to produce any regressions with this change and can't think of any other than the one you've already covered.

Thanks again for all your patches. Great work on all of these. :)

PR #2203, opened 2021-07-31
Fix "Tell Me About" giving different person locations than "Where Is"

mundifex:
ExpandMacro method in Person relied on GetHomePlace which only ever took into account a person's home place symbol. Switched to using new method GetDialogPlace which will use assigned place first, falling back to home place if needed.

"Where Is" calls TalkManager::GetKeySubjectPersonHint which relies on GetPersonSiteDetails to check for an assigned place symbol, falling back to the person's home place symbol if no assignment exists. With this change, "Tell Me About" now employs similar logic, which works well for queries about questor and quest-related NPCs alike.

My test case was quest K0C00Y06 from #2191 wherein mobile NPCs reported two different locations for _witness_. They now report just one location, the correct one, which updates as the witness is moved from _inn_ to _withouse_.

### DIFF
--- a/Assets/Scripts/Game/Questing/Person.cs
+++ b/Assets/Scripts/Game/Questing/Person.cs
@@ -290,9 +290,12 @@ namespace DaggerfallWorkshop.Game.Questing
             // Store this person in quest as last Person encountered
             // This will be used for subsequent pronoun macros, etc.
             ParentQuest.LastResourceReferenced = this;
-            Place homePlace = GetHomePlace();
-            if (homePlace != null)
-                ParentQuest.LastPlaceReferenced = homePlace;
+
+            Place dialogPlace = GetDialogPlace();
+            if (dialogPlace != null)
+            {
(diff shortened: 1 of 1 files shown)


KABoissonneault (approved):

Assets/Scripts/Game/Questing/Person.cs
```diff
@@ -302,16 +305,16 @@ public override bool ExpandMacro(MacroTypes macro, out string textOut)
                     textOut = displayName;
                     break;
 
-                case MacroTypes.NameMacro2:             // Home building name
-                    textOut = GetHomeBuildingName();
+                case MacroTypes.NameMacro2:             // building name
+                    textOut = (dialogPlace != null ? dialogPlace.SiteDetails.buildingName : BLANK);
```
Interkarma, line 309: I always love a good fallback to BLANK in Daggerfall. 👍

Interkarma:
This makes sense to me. A good tweak, thank you!

PR #2209, opened 2021-08-11
Added PlayerSpeedChanger.ModifySpeedMod(), Includes Console Commands

l3lessed:
I've added ModifySpeedMod to PlayerSpeedChanger.cs. This object allows developers to modify an already injected walk or run speed modifier value. This helps resolve the issue of having to continually add and remove walk and run speed modifier values to update them by providing a method to update a current modifier value through its assigned UID value. Also, it will make lerping speed changes over time easier and less cycle expensive.

I also added commands to the set_walkspeed and set_runspeed commands for testing this out via console. To modify an already existing/inject speed modifier value merely call set_walkspeed or set_runspeed and after the wanted updated speed value put the UID you want to update. And you're done, value is updated.

### DIFF
--- a/Assets/Game/Addons/UnityConsole/Console/Scripts/DefaultCommands.cs
+++ b/Assets/Game/Addons/UnityConsole/Console/Scripts/DefaultCommands.cs
@@ -643,9 +643,10 @@ namespace Wenzil.Console
         private static class SetWalkSpeed
         {
             public static readonly string name = "set_walkspeed";
+            public static readonly string uid = "speedmod_uid";
             public static readonly string error = "Failed to set walk speed - invalid setting or PlayerMotor object not found";
-            public static readonly string description = "Set walk speed by multiplying current walk speed by inserted percentage. Set to -1 to disable or enable speed overrides. Set to -2 to clear out all speed modifiers";
-            public static readonly string usage = "set_walkspeed [#]";
+            public static readonly string description = "Set or update walk speed by multiplying current walk speed by inserted percentage. To update existing modifier, put in its UID after the updated speed value. Set to -1 to disable or enable speed overrides. Set to -2 to clear out all speed modifiers";
+            public static readonly string usage = "set_walkspeed [#] [uid]";
 
             public static string Execute(params string[] args)
             {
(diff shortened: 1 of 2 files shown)


Assets/Game/Addons/UnityConsole/Console/Scripts/DefaultCommands.cs
```diff
@@ -643,9 +643,10 @@ public static string Execute(params string[] args)
         private static class SetWalkSpeed
         {
             public static readonly string name = "set_walkspeed";
+            public static readonly string uid = "speedmod_uid";
```
KABoissonneault, line 646: What's the purpose of these lines? I can't see where it's referred from. Is this for mods?

l3lessed: That can be removed. It was an addition I forgot about. It is not needed.

Assets/Scripts/Game/Player/PlayerSpeedChanger.cs
```diff
@@ -216,6 +218,36 @@ public bool AddRunSpeedMod(out string UID, float speedModifier = 0, bool refresh
             return added;
         }
 
+        public bool ModifySpeedMod(string UUID, float updatedSpeed, bool modifyRunSpeed = false, bool refreshSpeed = true)
+        {
+            //setup false bool for manipulation.
+            bool modified = false;
+
+            //if there is no uuid put in, return false as error catching.
+            if (UUID == "" || UUID == null)
```
Interkarma, line 227: Please use string.IsNullOrEmpty() here. :)

Interkarma:
I'm not sure how practical it is to enter the long-format string UID in the DFU console here. We can copy-paste from Unity output console, but it's not so easy from builds with only DFU's console as it doesn't support copy-paste from output. I still find it easier to use "-2" to clear all modifiers and then re-enter the new walk/run modifier speed when testing the game. I use set_runspeed all the time when zipping around testing quests, etc.

Is it even necessary to use full-length string GUIDs here? They aren't serialised, so this could just as easily use an index system starting from 0 based on order speed mods are added. The mod itself will receive back an index like 0 or 1, so it still knows which modifier index it owns at runtime. And when entering from console, it's much easier for a human being to internalise and enter something like:

`set_runspeed 4 0`

vs.

`set_runspeed 4 79e84faa-4674-4dd0-ac40-891ea6669de0`

l3lessed:
When you explain it in that way, no we do not need a massive UID. I for some reason thought the indexing system would cause mods to use the same index position, but now that I read your comment and thought about how it is serialized through the speedchanger script, we do not need it. We can instead use the index position for sure.

Thanks for pointing this out. I'll update it and commit it by Friday.

Interkarma:
Thank you for taking a look, and no rush at all. Cheers! :)

Interkarma:
I'll close this one out for now. Don't hesitate to open up again later with the discussed changed if you still wish to work on this one. :)

PR #2210, opened 2021-08-11
Retrorendering in linear colorspace

petchema:
Do palette transformations in gamma colorspace


Interkarma:
Thank you for fixes! There's no way I would have been able to correct so quickly without your experience on this.

I did notice the retro shader still tends towards darker colours under linear than it did under gamma. We might still need to take a couple more passes over it. I'll merge this one for now as it's only a small change and we'll get back to it later. Cheers!

PR #2214, opened 2021-08-18
Check that dependencies of an enabled mod are enabled

petchema:
Minimum fix: adds a warning when a dependency of an enabled mod is not enabled

Forums: https://forums.dfworkshop.net/viewtopic.php?f=22&t=5020


Interkarma:
Thank you Pango!

PR #2215, opened 2021-08-19
Add divide by zero protection to ConditionPercentage property

ajrb:
Fixes an edge case for 2 vanilla items with 0 max condition defined (finger and dead body) - it was causing exceptions when the curse quest that gives the finger when used with a mod that repairs items as it checks all items in player inventory.


KABoissonneault (approved):

Interkarma:
Thank you Hazelnut! :)

PR #2218, opened 2021-08-19
A0C00Y16 minor fixes

JayH2971:


Assets/StreamingAssets/Quests/A0C00Y16.txt
```diff
@@ -170,36 +164,16 @@ Message:  1054
...
 QBN:
 Item _gold_ gold
 
 Person _questgiver_ group Questor
-Person _missingperson_ face 1 group Group_6.0 local
-Person _friend1_ group Local_3.1 local
-Person _friend2_ group Local_3.1 local
-Person _friend3_ group Local_3.1 local
-Person _friend4_ group Local_4.1 local
+Person _missingperson_ face 1 group Group_6.0
```
KABoissonneault, line 172: @df-ferital Do you think it's right to remove these flags? From what I understand, the classic quest has the "local" flag, and if it's removed, it's going to be 50% chance of either "local" or "remote", right?

Seems like folks are still confused as to what to do with these flags, especially since the VS Code plugin still doesn't recognize it

KABoissonneault: In case of confusion, I talked with JayH on Discord and he reverted those changes. I still hope you can clarify in case I got something wrong

df-ferital: I don't agree with this change. There is no need to set the location scope of the questgiver since he/she is always local, since we talk to him to get the quest. However, missingperson should remain local, else he/she can be placed in another town, which is not what we want there.

Assets/StreamingAssets/Quests/A0C00Y16.txt
```diff
@@ -217,11 +191,7 @@ Foe _F.00_ is 3 Barbarian
 	start timer _S.16_ 
 	log 1010 step 0 
 	pick one of _S.00_ _S.02_ _S.03_ _S.01_ 
-	create npc _friend4_ 
-	create npc _friend2_ 
-	create npc _friend3_ 
-	create npc _friend1_ 
-
```
df-ferital, line 224: I don't agree with this change either, if you remove that, NPCs won't be placed at all. Or maybe this is not needed in DFU, I need to check the code, but don't have much time to do that at the moment. At least in classic, I'm certain that this is required to place NPCs.

JayH2971: I'm not familiar with the new changes to the quest engine, and don't believe this PR will accomplish anything good at this point. I'd rather leave it alone rather than makes things worse.

I'm quite certain create NPC is unnecessary in DFU (it was required in classic but NPCs are automatically generated when the quest is started in DFU). Other than that, I'd likely cause problems with this. I'll withdraw it.

Interkarma:
Thanks for your comments everyone, very much appreciate everyone helping each other for best understanding and outcomes. :)

Issue #2184, opened 2021-07-24
The Mod Settings button in the pause menu should be hidden if the mod system is off.


SpicerXD:
Basically just remove the Mod Settings button in the pause menu when the mod system is disabled.

jefetienne:
Fair point – I'll look into adding some improvements.

qingpaul:
there isn't a "setting" button on the mods menu，how do get it out?

SpicerXD:
The button in question is in the in-game pause menu at the top left, not the mods menu in the launcher.

Interkarma:
> Fair point – I'll look into adding some improvements.

Maybe in addition to disabling if mod system off, also disable if no mods active. But either might be a problem if other features are added to that arrow later.

Another option is to change the texture to something a bit less eye-catching that the red arrow. Example being a grey upward chevron with transparent areas. Something like below scaled down into same space.

![image](https://user-images.githubusercontent.com/10426244/128652652-f4a9e940-e1a0-41ab-b9d2-5cbf2d32e4c3.png)

jefetienne:
I think what I will do is implement functionality to have certain ListBox items be disable-able and grayed out – that way people without editable mods can at least know where the option is.

I can definitely use a new texture. My first commit for the PR originally used [simple triangle textures](https://github.com/jefetienne/daggerfall-unity/blob/5a70647885ec609a944383f7c1852b95b19229f0/Assets/Resources/arrow_up.png) I made, but I think the chevron looks better.

PR #2219, opened 2021-08-20
Pause Options Dropdown and ListBox changes

jefetienne:
Addresses #2184 

- Added ability for ListItems to be manually disabled, which prevents selecting or using them
- Made Mod settings dropdown option disabled with this new ability if no applicable mods found
- Switch to using custom textures again (this time of the chevron icon - made by myself. If anyone wants to improve on them let me know)
- Bugfix: Make the ListItems property return a shallow copy to prevent bypassing adding and removing items
- Improvement: Return registered dropdown items in alphabetical order (except for the Mod Settings option) to keep a consistent order regardless of mod load order


Assets/Scripts/Game/UserInterface/ListBox.cs
```diff
@@ -367,17 +373,56 @@ private void DecideTextColor(TextLabel label, int i)
...
-                label.TextColor = listItems[i].textColor;
+                label.TextColor = listItems[i].Enabled ?
+                    listItems[i].textColor
+                    : listItems[i].disabledTextColor;
                 label.ShadowPosition = shadowPosition;
                 label.ShadowColor = listItems[i].shadowColor;
             }
         }
+
+        private int FindNextEnabled(int currentIndex, bool previous)
```
KABoissonneault, line 392: I feel like this could have been split into 2 functions, there's barely any common code between the two branches

KABoissonneault (approved):

Assets/Scripts/Game/UserInterface/ListBox.cs
jefetienne: Sure!

Assets/Scripts/Game/UserInterface/PauseOptionsDropdown.cs
```diff
@@ -152,7 +151,18 @@ public void SetDropdownExpand(bool expand)
 
         #region Private Methods
 
-        void SetBackground(BaseScreenComponent panel, Color color, string textureName)
+        private bool HasApplicableMods()
+        {
```
Interkarma, line 155: Please pop in a null check here for MidManager.Instance and bounce if null before loop. This can be null when starting from Game scene (e.g. directly loading a save or UI during testing) and will cause an exception if not started from Startup scene.

Interkarma:
Lovely, thank you @jefetienne! :)

PR #2220, opened 2021-08-20
Dehardcode sound effects

JaceySanders:
Moves the lists of sound effects found in AmbientEffectsPlayer.cs into text files in SteamingAssets/Tables, and modifies AmbientEffectsPlayer.cs to access them from there. This dehardcoding allows modders to change which sound effects are used (e.g. removing the door ambient sounds from dungeons.)

This is my first time contributing to a project on Github, so please be patient with me.


KABoissonneault:
Possible IK will want to put this on hold until 1.0. It's strictly a modding support feature, and we're in feature freeze.

Interkarma:
Hey there. :) As @KABoissonneault mentions, Daggerfall Unity is currently in feature freeze. Only bug fixes and delayed items previously scheduled on roadmap will be considered at this time. The freeze period will extend until after 1.0.

https://forums.dfworkshop.net/viewtopic.php?p=55617#p55617

I also don't think changes to core are necessary here. You can disable the gameobject Dungeon/AmbientEffects (or set its "Presets" to None) and use a custom effects player from your mod that plays these dungeon sounds in any way desired. If you start with a copy of AmbientEffectsPlayer.cs, you can trim out anything in custom behaviour unrelated to dungeon sounds. The exterior and weather ambient sounds are played elsewhere. Dungeon/AmbientEffects instance only plays dungeon effects. You can make the changes necessary from a .dfmod without altering core.

KABoissonneault:
There's a bit more history to this, I think the intent with this approach was to minimize mod conflicts. DFU has this tendency of having huge objects that only a single mod can replace at a time.

Granted, a mod could create its own Prefab that doesn't hardcode those values, having its own "modding tools" for other mods to contribute to sounds cooperatively.

Interkarma:
Would moving clip arrays to string tables help with this though? If two different mods wanted to replace the tables with two different sets of data, that's basically the same problem. And a custom effects player also has the benefit of modder being able to play entirely new sounds in addition to classic.

I understand that sometimes modders want to do different things to the same object (e.g. make two different changes to inventory management) and are constrained by this not being modable without replacement. But it's hard to imagine two different ambient effects mods playing together with different artistic goals and sound libraries without clashing anyway.

JaceySanders:
That makes sense. I have learned enough now that I think I can do that instead.

Interkarma:
@JaceySanders Good luck! Please don't hesitate to DM me on forums if you're stuck on something down that path and I'll be happy to try and help.

PR #2221, opened 2021-08-22
Don't level up if already dead

petchema:
Corner case seen in a stream, player tried to sleep over some poisoning, but was also close to leveling up.

He died while resting, but had to level up before death scene could start.


KABoissonneault (approved):

Interkarma:
Thanks Pango. :)

PR #2222, opened 2021-08-23
Allow Argonians to get random names in character creation

KABoissonneault:
The original code comment said:

>Disable random name button only for Argonians because their race id
>would give them Imperial names from namegen

but that's only true in games after Daggerfall. In both Arena and Daggerfall, Argonians get latin names, that's just how it works in those games. Imperials don't even exist. Even in Daggerfall Unity, an Argonian character will have named characters with latin names in their background History. 

There's no reason for this arbitrary exclusion in DFU. It's going to be a surprise to veterans of the series, but it's consistent with Daggerfall itself. Worst case, they can name their character Makes-Up-Names themselves.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/CreateCharNameSelect.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/CreateCharNameSelect.cs
@@ -104,9 +104,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
                 return;
             }
 
-            // Disable random name button only for Argonians because their race id
-            // would give them Imperial names from namegen
-            randomNameButton.Enabled = raceTemplate.ID != (int) Races.Argonian;
+            randomNameButton.Enabled = true;
 
             // Randomise DFRandom seed from System.Random
             // A bit of a hack but better than starting with a seed of 0 every time


Interkarma:
Imperials don't exist in Daggerall, but an Imperial namebank does in gamedata. This namebank is incorrectly indexed to Argonians.

After rolling a few Argonian characters in classic to look at their backstories, I've seen a mixture of Nord, Breton, and Elvish namebanks used here for random names. But they never selected from the Imperial name generator that I've seen. I'd actually consider this a bug in DFU and the Argonian backstories should _not_ be using the Imperial name bank for backstories either. I'm going to loop in and fix this shortly now that you've let me know.

I'm happy for Argonians to get a random name button, but I don't agree they should use the Imperial name bank. As you state, Imperials don't exist and this namebank isn't used anywhere that I'm aware of. That it shows up in backstories in DFU is an oversight that should be fixed.

But I would be OK with Argonians getting a random selection of names from the other name banks however, like their backstories do in classic. At least then the arbitrary exclusion is removed.

KABoissonneault:
Oh, really? Maybe it was just Arena then.

Interkarma:
I don't remember Arena very well. :) But just spun up 5 or 6 Argonian chars in classic and they were all a mixture of namebanks in backstories (but I never got the Roman-sounding Imperial namebank). Try it and let me know what you get.

Now I'm wondering if DFU should be using random namebanks everywhere for backstories, or if this is just an Argonian thing. More testing needed!

Look what you've done. :)

KABoissonneault:
Actually saw the Latin name on the Nightblade backstory.

image

The female name (%fn) is Nord, but the other one is a %bn macro, the only one in the game IIRC, and is Latin. DFU turns %bn to a normal male name.

Granted, UESP lists the Latin names as Argonian, and they're often wrong. Maybe it's a common misconception, like lots of Daggerfall urban legens.

Interkarma:
That's a good observation re. %bn. As an agent of the Emperor, it makes sense this name is Imperial. We should probably make %bn consistently expand to Imperial namebank then, like %imp. You're right it's not used anywhere else in game's TEXT.RSC data, so it's hard to call 100%. Maybe %bn is a standalone mistake and should be changed to %imp to be consistent?

The other observation when generating backstories is that DFU's BiogFileMCP should use a mixture of random namebanks as we see this in classic. For example, this Argonian Warrior backstory from classic is using a Wood Elf namebank for %fn. I've seen a mixture of races names in biogs here, but never Imperial or Khajiit after several attempts. Might just need more attempts, but it seems limited to namebanks 0-5 (Breton through WoodElf)

image

KABoissonneault:
So to summarize my current understanding:

1­. %mn and %fn should use a random non-beast race in BiogFileMCP
2. %bn should still match the player's race in BiogFileMCP (with still random gender)
3. Argonians use "Imperial" names for their race (as seen in %bn)

Does that match with your testing?

df-ferital:
From classic reverse-engineered functions:

1. %mn and %fn always indeed use a random non beast race, including the Imperial name bank from namegen.dat. However, we chose (i don't remember if it was me, or some othe dev before) to tie %mn and %fn to current province race _or_ to character in biography, as it made much more sense from the context.
2. %bn is always using the same race as the character, and a random gender. Therefore, since there is no Argonian name bank (or if, as you say, Argonians were supposed to use Imperial names), %bn generates an Imperial name for Argonians.
3. Argonians don't really use Imperial names in game since there is no Argonian NPC. Well, if we except the previous case of %bn.

In my opinion, we shouldn't change the current behavior as even if Argonian were supposed to bear Imperial names in Daggerfall, that was not apparent at all and most Elder Scrolls players would be confused to get a Cyrodiilic name when generating a random name.

KABoissonneault:
From the book [Jokes](https://en.uesp.net/wiki/Daggerfall:Jokes):
>"How is your wife," asked Zalither.
>"She's in bed with laryngitis," replied Harlyth.
>"Is that Argonian bastard back in town again?"

If you look it up, the original joke was "Greek" instead of Argonian.

Giving Argonians latin names is not a bug or an accident, that's just how it was in those days. I don't see why we should feel restrained by what later games do. Lots of content is incompatible with Morrowind and beyond, and we're not restricting content to avoid surprising those players.

jefetienne:
I skimmed through a little bit of code, what do you all think of this?:

We split the Resources/NameGen.txt into multiple files and move them into StreamingAssets. The "Random" button gets enabled if there exists a corresponding namegen_<race>.txt file, and modders can fill in the gap for Argonian names, and it benefits modders by introducing the ability for extending generated names for all races.

Interkarma:
Late to reply due to time zone differences. Thanks for your feedback @df-ferital. I'm happy with behaviour of item 1 if this was an intentional choice. My focus in replies above was mainly on this disparity because it was a variance that might need to be addressed. If this was an intentional tightening then I'm happy with current behaviour. Considering how minor this variance is in any case, it's really not worth me spending any more time discussing.

@KABoissonneault I don't feel restrained by what the later games have done. I actually don't think about the later games very much at all. The issue here is simply that an Argonian namebank doesn't exist. I don't agree the Imperial namebank belongs to Argonians - it's there for %imp, which is used in many more places throughout TEXT.RSC _specifically_ for Imperial names. The %bn usage by comparison is a single instance not used anywhere else. It looks just as accidental as Imperial namebank being next in sequence where Argonian namebank should be.

The main issue here is simply to give Argonians a random name button, so by all means let's activate this and give them a random name from all available namebanks (except Khajiit, I think). Then after 1.0 we can add new code to load custom namebanks from JSON as @jefetienne suggests and userbase can install random Argonian names to whatever preference they have.

Listen everyone. I'm really overworked and right now I feel too bloody tired and old to debate whether a Greek joke means that fantasy lizards should get exclusively Roman names because someone 25 years ago didn't have any better ideas. I don't go in for those kinds of discussions. But I'm more than happy for Argonians to get a random name button using the available name banks and call it good. That's the core issue being fixed by this PR and it really doesn't need much more time and energy than that to solve.

jefetienne:
I'm sorry to hear that, Interkarma. I hope everything clears up for you and feel better soon!

KABoissonneault:
>I don't agree the Imperial namebank belongs to Argonians - it's there for %imp, which is used in many more places throughout TEXT.RSC specifically for Imperial names.

Yet, we have a separate name list for %imp:
```
{ "Pelagius", "Cephorus", "Uriel", "Cassynder", "Voragiel", "Trabbatus" }
```

This is why I'm saying calling it Imperial is confusing. %imp is a separate list, and Argonians use the namebank they currently have assigned in both Arena and Daggerfall. I don't think there's any association to the Empire beyond sounding Roman. 

I don't think the fact that the namebank belongs to them is up for debate (anyone who thinks it is can reach out to me on another platform), but at the same time, the random name generation is a DFU improvement, not a classic issue, so I'm willing to let it go if it's too controversial, and wait for modding support later.

KABoissonneault:
(also, [Brief History of the Empire, Part 4](https://en.uesp.net/wiki/Daggerfall:Brief_History_of_the_Empire,_Part_4) has an Argonian character named Thoricles Romus)

Interkarma:
>Yet, we have a separate name list for %imp:

Then I have misremembered this macro expansion. Thank you for clearing that up.

I'm not much of a lore person, more of a systems kind of guy, but I am willing to see things from a different perspective. I'm going to try and see things from your point of view and why the Imperial names are arguably an intentional choice in the early games.

Daggerfall is set in the Third Era. Looking at the [Argonian history](https://elderscrolls.fandom.com/wiki/Argonians), they had by this stage been invaded, enslaved, and assimilated into the Empire. With this context in mind, it really does make sense for the majority of Argonians to have Imperial names. Other than a few holdouts, they had by this stage lost their own identity and become part of the empire. We've seen this happen to several cultures throughout history in the real world, and it makes sense for this to be an intentional case here.

So, sure. I'm doing a 180 and offering an apology. You're very likely right here and I'm wrong. I didn't think in terms of the lore and history within the game, and it really does make sense for Argonians at this time to have an Imperial namebank. By all means, re-open the PR and I'll merge as-is. Thank you. :)

PR #2226, opened 2021-08-24
Change protection to allow mod overriding for court UI window

ajrb:
This contains a slight refactoring of the main logic outside of the Update() method since this is called up the hierarchy. All other changes are protection only.


Interkarma:
Hey Hazelnut! OK, these are minor tweaks and don't run afoul of my freeze criteria. :)

https://forums.dfworkshop.net/viewtopic.php?p=55617#p55617

Once I get my other DFU workloads in that list out of the way, I'll probably be a bit more relaxed about the definition of minor tweaks too.

Part of this whole freeze process is clearing my backlog of work without drowning in PRs, and the other part is to clear my head so I can focus on the bits I still need to do that kept being pushed back until now because I couldn't find the headspace. So thank you for keeping the PRs minor and minimal.

ajrb:
No problem, I am completely on board with the freeze. That's why I kept this one separate so you could reject it if you wanted. :-)

PR #2231, opened 2021-08-29
Expose and virtualize members in Open and Lock magic effects

KABoissonneault:
Not really a fix, but it's small enough at least.

In my Unleveled Spells mod, I managed to remove level scaling from all the spells in the game, except for spells featuring the Open and Lock magic effects. My users would appreciate if I could fix this.

You can't replace those effects without inheriting from them and replacing `TriggerOpenEffect` / `TriggerLockEffect`, since the DFU code is hardcoded to search for them specifically, so a DFU code change was necessary.


Interkarma:
All good, cheers! I'm aiming to drop 0.12.3 builds this week while I finish preparing 0.13.0 for preview. Will roll this PR (and most of the others) into 0.12 builds this week.

Issue #2232, opened 2021-08-30
Touch spells don't stop interactions


KABoissonneault:
When the player has a readied spell, most interactions on objects are disabled until the spell is cast, except if the range is Touch.

This leads to battles using Touch spells looking like this.

![image](https://user-images.githubusercontent.com/5789925/131382499-fc5ccae1-1cac-4fc7-9228-b7f58a96421d.png)

This occurs for Steal, Grab, Info, and Dialogue interaction modes, for all sorts of interactible objects.

The line responsible for this is: https://github.com/Interkarma/daggerfall-unity/blob/master/Assets/Scripts/Game/PlayerActivate.cs#L244

The comment above has a reasonable explanation, that a player may want to ready a Touch range spell, open a door, then cast it at whatever was standing behind immediately.

Except, why are Touch spells an exception there? If a player wants to ready a spell before opening the door to danger, they could want use any sort of non-self spells. If the player is expected to be able to ready a ranged Wizard Fire *after* opening the door, then why can't they do it for their touch-range Shock spell too?

I think for consistency, we should also remove interactions while readying a Touch spell. 
1) This would remove lots of weird behavior in terms of interactions while a spell is readied. For example, looting a corpse may or may not cast your readied spell, depending on how recently you've readied the spell. This happens often to users which always have a spell readied, such as when using the Hotkey Bar mod
2) This removes the spam spellcasters are getting when using those touch spells. I think that classic didn't have the spam issue, because Interaction was on Space while casting the spell was on click
3) This doesn't break spells like Open or Lock, since you interact with the door after the spell is cast, not while it's readied

Suggested solutions:
1) remove the check for touch spells entirely
2) treat touch spells in a special way, and only allow door and "action" (ex: lever) interactions while they're readied

I'm curious for other people's opinions before sending a PR myself

Interkarma:
>Except, why are Touch spells an exception there?

Because that is the classic behaviour. My original implementation didn't have this check, and it caused problems with people believing their spells were missing because they weren't in range, or because the behaviour differed from classic.

Happy to refine this if it behaves differently to classic in some way.

Interkarma:
I found a few minutes to test against classic. Can definitely open doors/levers in classic with Touch equipped, and the other interactions also fall through when spamming touch spells. So looks like I was fairly thorough when I matched classic a few years back.

[...]

PR #2234, opened 2021-08-31
Disable interactions during touch-range spell

KABoissonneault:
UX improvement for issue #2232.

I considered disabling all interactions except doors, "action" (ex: levers), ladders, and custom. Then, I decided that for most other interactions, if you're clicking on them, you probably meant to interact, not cast a spell on them.

Maybe loot markers might be accidentally misclicked in combat, and some may prefer disabling interactions with them while casting spells.
I decided not to, since players might be readying a spell from an auto-recast mod (ex: Hotkey Bar) and deliberately clicking a loot container. 

So, in the end, I only disabled mobiles, since you're most likely intending to fight them. It was already dangerous to talk to villagers with a spell ready since it would go off even when you could successfully interact with them.


Interkarma:
I'm happy with that approach. Thanks so much Kab! :)

Issue #2230, opened 2021-08-26
MobileEnemy.CorpseTexture fails on texture archives outside of 0-511


KABoissonneault:
With custom enemies, you want custom corpse textures. Right now, if you try to use an archive beyond 511 (ex: `CorpseTexture = EnemyBasics.CorpseTexture(1000, 1)`), there's an error with the texture import down the line, and the corpse billboard won't appear.

I'm looking into it, but if someone more knowledgeable with the texture import system wants to have a look, feel free.

Interkarma:
Just spitballing since you've had no other replies. My feeling is this could happen at a low-level inside TextureFile.cs as archives above 511 don't exist in gamedata files. I'm not sure if texture replacement handles this case or not. Stepping into code in debugger should eventually show where it bounces as the call is passed down from mod through to DFU and eventually down into the basic reader classes.

Interkarma:
Hey KAB! I understand this one was resolved by #2242, so will close now.

PR #2242, opened 2021-09-14
Enemy corpse custom texture fix

KABoissonneault:
See issue #2230 

It seems lots of this code is older, back from 2019. For example, the Material code only checked loose files, not "all locations" (assets from the dfmod). 

Graphics isn't my area, and I don't know what are the users of these graphics settings. However, I tried to make my changes have no effects on people replacing texture 0 to 511 - there shouldn't be any users for textures outside of these in materials, as it didn't work before this PR.

I'm open to suggestions to make this cleaner.

Hopefully with this, Kamer can complete his mod.

EDIT: yes, this works for him :)
image

### DIFF
--- a/Assets/Scripts/MaterialReader.cs
+++ b/Assets/Scripts/MaterialReader.cs
@@ -416,7 +416,7 @@ namespace DaggerfallWorkshop
                 }
 
                 // Get texture
-                results = textureReader.GetTexture2D(settings, AlphaTextureFormat, TextureImport.LooseFiles);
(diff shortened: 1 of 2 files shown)


Interkarma:
This looks OK to me. It doesn't really alter the way core textures are handled, and I don't see any issues with regular asset injection. I tried with a few texture replacement mods (inc. DREAM and Animated People) and didn't spot any new issues. All the baseline stuff like emission, ghosts, etc. still all works OK.

I'll roll this one into future builds for community to play with. Cheers!

PR #2244, opened 2021-09-16
Color picker tweaks and fixes

rossipaolo:
Made a few changes to the color picker used for mod settings (for example by RealGrass mod):

* Improved scaling (text labels were too small).
* Color is picked (`OnColorPicked` event) only if window is closed with OK button, otherwise is discarded.
* Fixed an issue that caused a color to be selected if the mouse button is still pressed after popup window is opened by clicking a button positioned under the popup.


Interkarma:
Looks good to me, thank you @rossipaolo! :)

PR #2247, opened 2021-09-22
Fix out-of-range error in FormulaHelper.CalculateCastingCost 

KABoissonneault:
The issue was reported on my mod here: https://forums.dfworkshop.net/viewtopic.php?p=57172#p57172
Save with repro steps included in the mod, Hotkey Bar mod probably not actually necessary for repro.

Basically, my custom spells don't always have 3 spell effects in them, and it makes the magic item cost formula throw an exception. 
I just made it use the actual size of the array, instead of assuming all spells have 3 effects (most spells had invalid effects for slot 2 and 3, of course).

I took the opportunity to add a formula override.

### DIFF
--- a/Assets/Scripts/Game/Formulas/FormulaHelper.cs
+++ b/Assets/Scripts/Game/Formulas/FormulaHelper.cs
@@ -2394,6 +2394,10 @@ namespace DaggerfallWorkshop.Game.Formulas
         /// <param name="enchantingItem">True if the method is used from the magic item maker.</param>
         public static int CalculateCastingCost(SpellRecord.SpellRecordData spell, bool enchantingItem= true)
         {
+            Func<SpellRecord.SpellRecordData, bool, int> del;
+            if (TryGetOverride("CalculateCastingCost", out del))
+                return del(spell, enchantingItem);
+
             // Indices into effect settings array for each effect and its subtypes
             byte[] effectIndices = {    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, // Paralysis
                                         0x01, 0x02, 0x03, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, // Continuous Damage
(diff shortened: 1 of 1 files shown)


Interkarma:
I think spells should always have 3 effects and the unused effects are type -1 ( 255 in classic bytes?). It's been a while, so I don't have this off the top of my mind right now.

But I do think the fixed count of 3 is still required for total compatibility with classic. If the problem is only introduced by your custom spells, that seems like a problem with the custom spell data rather than something to be addressed in core.

KABoissonneault:
For the 3 last examples you've posted, my changes wouldn't affect classic, since that would still be a spell with an array of 3 effects (`spell.effects.Length == 3`), and all three effects will be traversed as before.

I think any software should be "liberal in what it accepts" (and conservative in what it emits). I think expecting mods to have a strict format or else you crash the game is not robust.
We've had no other issues with "magic 3s" for spells so far, so I think it's preferable to fix this case than have my mod update all existing saves, and expect all future moders to do follow this rule

Interkarma:
>I think expecting mods to have a strict format or else you crash the game is not robust.

Consider your spell templates aren't a mod, as such. Rather, you're using classic's template format and mixing in new data. This is technically data injection, not far removed from editing binary files directly. In these cases you shouldn't expect to add data and have it work without following the format data reader expects. You'd have just as much trouble adding new data to SPELLS.STD without following the expected format (which abstractly is what's happening here).

Another example of this style of mixing is custom locations. New RMB data has to follow the correct binary format or the game will crash or not behave as expected. It's really not that unusual to be expected to provide correctly formatted data when required. Otherwise, GIGO.

With all that said, your change will likely be OK. Spell cost formula is quite far down the production line and the data should already be sanitised into a neat list by this stage. But in cases like this, I need to raise this stuff and work through it. Changes like this should always consider classic compatibility first.

Interkarma:
How about a slightly modified approach that still uses 3-effect rule but also handles cases where the array doesn't count to 3? For example:

    for (int i = 0; i < 3; i++)
    {
        if (i < spell.effects.Length && spell.effects[i].type != -1)
        {

Look, I _know_ I'm being painful here. :)

Interkarma:
Thank you for change, I do appreciate it.

PR #2248, opened 2021-09-25
Fix missing sound and StaticNPC for imported RMB gameobjects

rossipaolo:


Assets/Scripts/Utility/RMBLayout.cs
```diff
@@ -349,39 +349,40 @@ public static GameObject CreateBaseGameObject(ref DFBlock blockData, int layoutX
                     -obj.YPos + blockFlatsOffsetY,
                     obj.ZPos + BlocksFile.RMBDimension) * MeshReader.GlobalScale;
 
-                // Import custom 3d gameobject instead of flat
-                if (MeshReplacement.ImportCustomFlatGameobject(obj.TextureArchive, obj.TextureRecord, billboardPosition, flatsParent) != null)
-                    continue;
+                bool isImported;
+                GameObject go = MeshReplacement.ImportCustomFlatGameobject(obj.TextureArchive, obj.TextureRecord, billboardPosition, flatsParent);
+                if (!(isImported = go))
```
Interkarma, line 354: Is this meant to be an assignment or a comparison? I'd prefer this is split out for clarity.

Assets/Scripts/Utility/RMBLayout.cs
```diff
@@ -349,39 +349,40 @@ public static GameObject CreateBaseGameObject(ref DFBlock blockData, int layoutX
...
+                }     
 
                 // Add animal sound
                 if (obj.TextureArchive == TextureReader.AnimalsTextureArchive)
-                    AddAnimalAudioSource(go);
+                    AddAnimalAudioSource(go, obj.TextureRecord);
 
                 // If flat record has a non-zero faction id, then it's an exterior NPC
-                if (obj.FactionID != 0)
+                if (!isImported && obj.FactionID != 0)
```
Interkarma, line 385: Wouldn't making this a conditional of `!isImported` result in successfully imported billboards not receiving their `StaticNPC` component? This component is still required for exterior NPCs even when texture is replaced (e.g. witch coven NPCs). Same for city lights if replaced and isImported == true, they won't receive the light prefab.

I might just be misunderstanding intent here though. :)

Assets/Scripts/Utility/RMBLayout.cs
rossipaolo: Done.

Assets/Scripts/Utility/RMBLayout.cs
rossipaolo: This is for custom prefabs, which are generally not billboards and don't have `DaggerfallBillboard` component. The intent is to ensure they have an audio source with the right sound if needed, while lights are tipically added as part of the prefab itself.  My aim was to fix issue with missing sound reported on [forums](https://forums.dfworkshop.net/viewtopic.php?f=22&t=5111&p=57298#p57160), tell me if you want these components to be handled differently.

Interkarma: They'll still need `StaticNPC` component though - this is involved in quest system. See comments in class.

https://github.com/Interkarma/daggerfall-unity/blob/master/Assets/Scripts/Game/StaticNPC.cs

`* Any NPC replacers (e.g. to 3D models) will need to assign this behaviour with original layout data.`

KABoissonneault: I think it's wrong to assume custom prefabs are not billboards. carademono's many mods (Villager Variety, Villain Variety, Animated People) all use custom prefabs and are all billboards.

My suggestion: just check `if(dfBillboard != null)` for the `dfBillboard.SetRMBPeopleData(obj.FactionID, obj.Flags, obj.Position);` line. Remove the `!isImported` check above

rossipaolo: Ok, thanks for pointing out the issue with npcs. This pr only included the minimal amount of changes to fix sound (StaticNPC component wasn't being added before), but I have now reworked it as suggested.

Do you think we should force the light prefab too?

Interkarma: If modder is replacing object entirely with their own prefab, probably best not to force another light prefab on them. They'll likely want to control the whole appearance themselves. But a basic texture swap should still receive internal light prefab as normal.

Interkarma:
Thank you for the updates @rossipaolo. :) I'll return after holidays to review this one again and merge into next release.

Assets/Scripts/Utility/RMBLayout.cs
```diff
@@ -348,44 +348,45 @@ public static GameObject CreateBaseGameObject(ref DFBlock blockData, int layoutX
...
+                    //        continue;
+                    //    }
+                    //}
+
+                    // Add to batch where available
+                    //if (obj.TextureArchive == TextureReader.AnimalsTextureArchive && animalsBillboardBatch != null)
+                    //{
+                    //    animalsBillboardBatch.AddItem(obj.TextureRecord, billboardPosition);
+                    //    continue;
+                    //}
```
ajrb, line 371: Do we need to keep this commented out code around anymore? Seems to have been commented out 4 years ago when some super atlas was disabled.

Interkarma: Yeah, the commented code can be edited out. I'll remove it after I merge.

PR #2249, opened 2021-10-02
added actions property to task class

Macadaynu:
Request to add a new "Actions" get property to the Task.cs class. This is to enable a mod to automatically add notes to the player's notebook with main story NPC quest dialogue, issue request [here](https://forums.dfworkshop.net/viewtopic.php?f=12&t=4942).
Also to help with the implementation of a hidden map location mod idea found [here](https://forums.dfworkshop.net/viewtopic.php?f=12&t=5117).

### DIFF
--- a/Assets/Scripts/Game/Questing/Task.cs
+++ b/Assets/Scripts/Game/Questing/Task.cs
@@ -95,7 +95,7 @@ namespace DaggerfallWorkshop.Game.Questing
             get { return globalVarLink; }
         }
 
-        public List<IQuestAction> Actions
+        public IEnumerable<IQuestAction> Actions
         {
             get { return actions; }
         }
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Game/Questing/Task.cs
```diff
@@ -95,6 +95,11 @@ public int GlobalVarLink
             get { return globalVarLink; }
         }
 
+        public List<IQuestAction> Actions
```
KABoissonneault, line 98: Do you really need access to the list? I'm concerned with mods trying to add and remove from it. If you just need to read the actions, consider returning `IEnumerable<IQuestAction>` (no other change needed otherwise)

Macadaynu: I only need read access to the list, have updated that to return an `IEnumerable`, I have given it a quick re-test my end and works fine as expected.

KABoissonneault (approved):

PR #2250, opened 2021-10-02
Improve choice of loose texture format

rossipaolo:
Checks width and height of png files directly from IHDR header and chooses appropriate texture format before data is loaded as Unity Texture2D. This avoids the need to reimport image if resolution is below threshold. @Interkarma let me know what you think of this approach.


Interkarma:
Thank you @rossipaolo! I was a bit concerned at first with the idea of grabbing width & height direct from binary data. But having seen your implementation and read a little on PNG format, it's such a straightforward thing to grab from first IHDR chunk. I'll give this another pass soon and merge as part of 0.13. Very nice work! :)

PR #2252, opened 2021-10-08
Avoid all quest clothing using blue dye

Ralzar81:
(Aw jeez, here I go attempting a pull request again)
Not sure if this is intentional design to copy vanilla or not, but it seems all clothing produced by the quest system defaults to blue clothing.
This change should make sure clothing gets a randomized clothing dye instead.


Interkarma:
Quests can potentially request a specific subtype of clothing. Setting the clothing dye after quest item is generated is the best way to do this, rather than override with a random item.

PR #2254, opened 2021-10-10
Allow cast when held to be overridden and modified by mods.

ajrb:
This is so travel options can prevent ticking of damage when time accelerated travelling, since fast travel doesn't damage condition.


KABoissonneault:
See:
https://github.com/ajrb/dfunity-mods/pull/28/files#diff-d29a0b10799964b00b1b1ce7fe7f214f891f2561096ca21f4de47e9754aa6035R61

if you want an idea of how Travel Options would make use of these changes

PR #2255, opened 2021-10-10
Fixed custom textures for NPC billboards not having a collider

KABoissonneault:
With #2242, mods can add static NPCs with new textures using world variants.

image
Pictured: Lady Flyte, in the Anticlere palace

However, there is an issue: they cannot be interacted with like a normal NPC.

`MaterialReader.GetFlatType` only handles textures between 0 and 511, so custom textures will not get detected as a `FlatTypes.NPC`. 

`DaggerfallBillboard.SetMaterial` uses this enum value to know whether to add a collider on it. Therefore, custom textures don't get that collider.

There were a few ways to go about fixing this, but I went, again, with something short that affects as few code as possible. Let me know if you think we could do with something cleaner.


Interkarma:
Happy with how you handled this one. Cheers. :)

PR #2271, opened 2021-12-17
Fix for switching back to original enemy Team

Macadaynu:
The EnemyBasics.Enemies array size is only 62, and the ids of the objects are not sequential in this array (It goes straight from Lamia id: 42 to Mage id: 128).

So just using the id as the array position would only work for enemies with an id of 42 or less. Better to check with Linq on the actual id of the object.

### DIFF
--- a/Assets/Scripts/Game/EnemyMotor.cs
+++ b/Assets/Scripts/Game/EnemyMotor.cs
@@ -14,6 +14,7 @@ using DaggerfallWorkshop.Game.Entity;
 using DaggerfallWorkshop.Game.MagicAndEffects;
 using System.Collections.Generic;
 using DaggerfallWorkshop.Utility;
+using System.Linq;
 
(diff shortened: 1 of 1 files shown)


KABoissonneault:
Do you have a scenario where you can reproduce that issue? I've considered fixing it when adding the modding support for custom enemies, but I thought at the time that only monsters could be a "Player Ally" (through the Skull of Corruption I think) and be attacked and turned hostile

Does this happen with your follower mod?

Macadaynu:
Yes I was playtesting the Language Skills Overhaul mod, and I had an enemy follower with an id of 133. 

I wanted to see what happened when attacking someone in my party, so attacked the follower and started getting spammed with errors, after investigation it turned out to be this code.

KABoissonneault (approved):

Interkarma:
Thank you for fix. I'll roll this into next release. :)

PR #2272, opened 2021-12-22
Update/world data editor object data

Cliffworms:
Added all interior building parts, exterior buildings, several new clutter and re-organized NPCs lists. There are around 1000 models or more added.

The other models and billboards left to add are linked to the main quest. This update at least allows someone to use every single interior parts for buildings.

Cheers!


Cliffworms:
It seems I made a mistake when first doing the PR. The WorldDataEditorObjectData.cs I had been working with didn't have the header found in the first 8 lines. You will thus see 2 commits. 

Commit #1 : A duplicate of my .cs file without the header
Commit #2 : Adds the header. 

Being not familiar at all with GitHub, I wonder if it may be better if I simply deleted this PR and did a new one with those 2 commits merged together?

KABoissonneault (approved):

PR #2273, opened 2021-12-30
Fix Loop for FactionsAndBuildings.  It was looping through 28 items i…

AsesinoBlade:
Fix loop for FactionsAndBuildings.  It was looping through 28 items instead of 29 so it was missing Gem Stores

See reference: https://forums.dfworkshop.net/viewtopic.php?p=59155#p59155 for bug report


KABoissonneault (approved):

Interkarma:
Thank you for fix! Will merge this ahead of next update.

PR #2276, opened 2022-01-08
Prevent shops being treated like FG

ajrb:
See https://forums.dfworkshop.net/viewtopic.php?t=5200

Alternative would be to change line 5956 to read "ggroup: -1" but that will only affect new games I think as faction data is saved.


Interkarma:
Thanks again for fix. I think I'll make the change in FACTION.TXT as well, but your PR is most suitable for current games using the bad data.

PR #2277, opened 2022-01-09
Add toggle noclip console command

extract:
This is the "simplest" method,
 but a bit hacky - hence experimental.

### DIFF
--- a/Assets/Game/Addons/UnityConsole/Console/Scripts/DefaultCommands.cs
+++ b/Assets/Game/Addons/UnityConsole/Console/Scripts/DefaultCommands.cs
@@ -36,6 +36,7 @@ namespace Wenzil.Console
             ConsoleCommandsDatabase.RegisterCommand(HelpCommand.name, HelpCommand.description, HelpCommand.usage, HelpCommand.Execute);
             ConsoleCommandsDatabase.RegisterCommand(LoadCommand.name, LoadCommand.description, LoadCommand.usage, LoadCommand.Execute);
             ConsoleCommandsDatabase.RegisterCommand(GodCommand.name, GodCommand.description, GodCommand.usage, GodCommand.Execute);
+            ConsoleCommandsDatabase.RegisterCommand(NoClipCommand.name, NoClipCommand.description, NoClipCommand.usage, NoClipCommand.Execute);
             ConsoleCommandsDatabase.RegisterCommand(NoTargetCommand.name, NoTargetCommand.description, NoTargetCommand.usage, NoTargetCommand.Execute);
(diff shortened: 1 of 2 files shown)


extract:
There's a few different ways that I can think of to add tcl, this solution is however the most generic that doesn't require too deep changes to for example CharacterController.

Interkarma:
Hey there and welcome. :)

There are a few changes in here unrelated to your PR (e.g. whitespace). Could you please clean these up and resubmit with just the required changes?

I'm not confident about using collider radius to create this effect. It should be adequate just to disable collider when noclip is enabled, and enable again when noclip disabled.

There will also be some conflicts if player has previously used "tgm" or "levitate on" commands. This implementation doesn't cache their previous state when noclip is enabled, so might improperly disable them when noclip is disabled. And what if player changes "tgm" or "levitate" state while noclip is enabled?

I'd have to play with this and see if there's a way to implement that doesn't trip up other console commands.

extract:
Hi and thank you. :)

You're right about the white space, this was more of a draft to get something in - I wanted your opinion.

Well, I am not confident either, but it seems you can not disable the collider in any real way without editing the character controller script. (Or temp disabling it and enabling another noclip controller... maybe that's a better way?).

Regarding, the console commands yes it's not considering tgm or levitate's prev/curr state - I thought it didn't matter to much though, and it should be fine to remove both of those changes but I was afraid the infinite radius may cause issues with health/insta-death if not god mode was enabled.

Anyway, I will keep looking into it as well and maybe play around with the noclip controller idea, I just didn't want to impact too much with my first commit here! haha :)

extract:
Hi :)

@Interkarma I had an idea that we could add a new Noclip layer, which then the controller could be put in. This would solve the issue - even quite neatly. However, I don't know how to add a layer to the unity project.

I also cleaned up what you requested earlier.

Ah and uh.. Ignore my failure to deal with git.

Interkarma:
No problems here with switching layers to ghost player collisions. I'm happy to add noclip layer into core as part of merging PR.

>Regarding, the console commands yes it's not considering tgm or levitate's prev/curr state - I thought it didn't matter to much though

You're right, probably not a big deal for console commands. I'll just need to play with it a bit to be sure and merge when I can. Cheers! :)

Interkarma:
I'm working towards another point release. I probably won't get this merged just yet, but I will loop back to this soon. Thanks again for tweaks. :)

extract:
That's fine, is there anyway I can assist you or contribute?

Interkarma:
After testing, I'm happy with this approach and using new "NoclipLayer" with physics matrix set as you describe.

There's also no problems setting levitate flag on/off, so this can be added back. My earlier concern was this might disrupt levitate effect as it uses same IsLevitating flag. But effect sets this continuously in ConstantEffect, so this won't be a problem at all.

Cheers! :)

extract:
Hi,

I have fixed these changes. :)

PR #2279, opened 2022-01-15
Make preset button in mod settings more obvious

ajrb:
I think a lot of players miss the presets function for mod settings and I think the button should be made clearer. Here is my proposal. I used the same background colour as the save button since I am terrible at colour selection and I think it looks fine. See picture below for comparison:

PresetButton


Interkarma:
Thank you. :)

PR #2282, opened 2022-01-23
World Data Editor - Fix truncation of water level in start markers

ajrb:


Assets/Game/Addons/WorldDataEditor/Editor/WorldDataEditor.cs
```diff
@@ -340,7 +340,7 @@ private void EditorWindow()
 
         private void FlatGroupSwitchers(ref int elementIndex, string flatId)
         {
-            Debug.LogFormat("flat group switch: {0}", flatId);
+            //Debug.LogFormat("flat group switch: {0}", flatId);
```
KABoissonneault, line 343: Is this an intended change? Might be worth removing the log if it's noisy

ajrb: It's helpful for debugging groups of flats if we add more. I had simply forgotten to comment out for the PR.

PR #2284, opened 2022-01-24
Extract abstract class from DaggerfallBillboard and move summary struct into MobileUnit base class

ajrb:
This is to allow mods contributing new MobileUnit or Billboard implementations to allow most DFU and other mod logic to continue to work. Billboard is based on the same base class structure as MobileUnit.

@Interkarma Would you like a different name for the base class, maybe AbstractBillboard or BillboardBase? I just followed the pattern already set. Apologies for the blast radius of this one. 

Pretty sure I did the refactoring correctly but could use some review in case I made a mistake somewhere. @KABoissonneault can you please review my work here?

### DIFF
--- a/Assets/Scripts/Internal/DaggerfallBillboard.cs
+++ b/Assets/Scripts/Internal/DaggerfallBillboard.cs
@@ -14,13 +14,8 @@ using UnityEditor;
 #endif
 using UnityEngine;
 using UnityEngine.Rendering;
-using System;
-using System.IO;
 using System.Collections;
-using System.Collections.Generic;
 using DaggerfallConnect;
-using DaggerfallConnect.Utility;
-using DaggerfallConnect.Arena2;
 using DaggerfallWorkshop.Utility.AssetInjection;
 using DaggerfallWorkshop.Game;
 using DaggerfallWorkshop.Utility;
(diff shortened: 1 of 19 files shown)


Assets/Scripts/Internal/DaggerfallInterior.cs
```diff
@@ -940,7 +940,7 @@ private void AddPeople(PlayerGPS.DiscoveredBuilding buildingData)
                     }
 
                     // Set position
-                    DaggerfallBillboard dfBillboard = go.GetComponent<DaggerfallBillboard>();
+                    Billboard dfBillboard = go.GetComponent<Billboard>();
```
KABoissonneault, line 943: I think this part of the code should probably be moved out of the "if".
 
For example, Vanilla Enhanced's animated people uses custom flat gameobjects (see line 929) to insert the custom billboard. After this PR, I'll change its prefabs to use the new Billboard interface, and it'd be best if `SetRMBPeopleData` was called on custom billboards too

Assets/Scripts/Internal/DaggerfallInterior.cs
```diff
@@ -620,7 +620,7 @@ private void AddFlats(PlayerGPS.DiscoveredBuilding buildingData)
                 GameObject go = GameObjectHelper.CreateDaggerfallBillboardGameObject(obj.TextureArchive, obj.TextureRecord, node.transform);
 
                 // Set position
-                DaggerfallBillboard dfBillboard = go.GetComponent<DaggerfallBillboard>();
+                Billboard dfBillboard = go.GetComponent<Billboard>();
```
KABoissonneault, line 623: For consistency, I think this part of the code should run on custom billboards too (currently, the loop stops at line 617 for them). 
he custom flat gameobjects were designed for 3d models in mind, but we've found use for it with custom billboard classes too. So I think as long as it has a `Billboard` component, it should have its elevation adjusted by DFU in all cases

Assets/Scripts/Utility/RDBLayout.cs
```diff
@@ -1211,7 +1211,7 @@ record = obj.Resources.FlatResource.TextureRecord;
                 // Setup standard billboard and assign RDB data
                 go = GameObjectHelper.CreateDaggerfallBillboardGameObject(archive, record, parent);
                 go.transform.position = targetPosition;
-                DaggerfallBillboard dfBillboard = go.GetComponent<DaggerfallBillboard>();
+                Billboard dfBillboard = go.GetComponent<Billboard>();
                 dfBillboard.SetRDBResourceData(obj.Resources.FlatResource);
```
KABoissonneault, line 1215: I think the `SetRDBResourceData` part should be moved out of the "if" (with an extra null check on the Billboard component of course), so mods have access to the RDB resource info.

KABoissonneault (changes requested):
I'd like `SetRDBResourceData` and `SetRMBPeopleData` to be called on custom flat gameobjects

Assets/Scripts/Internal/DaggerfallInterior.cs
ajrb: Looking at it, not sure about position. I've moved the set people data call. Will let Interkarma comment on the position adjustment because I am not sure either.

KABoissonneault (approved):

Assets/Scripts/Internal/DaggerfallInterior.cs
Interkarma: I note that `ImportCustomFlatGameobject` calls its own `AlignToBase` handler which does a position adjustment. If we also did that outside of the `if` block here then position would be adjusted twice for imported replacements.

I think it's fine the way it is.

Assets/Scripts/Internal/DaggerfallBillboard.cs
```diff
@@ -32,55 +27,38 @@ namespace DaggerfallWorkshop
...
+    public class DaggerfallBillboard : Billboard
     {
-        public int FramesPerSecond = 5;     // General FPS
-        public bool OneShot = false;        // Plays animation once then destroys GameObject
-        public bool FaceY = false;          // Billboard should also face camera up/down
-
-        [SerializeField]
-        BillboardSummary summary = new BillboardSummary();
-
-        public int customArchive = 210;
```
Interkarma, line 44: Please leave `customArchive` and `customRecord` intact. They're used by `DisplayAboutGUI` in DaggerfallBillboardEditor.cs. Removing these breaks the editor script and prevents changing billboard material in editor (billboard editor calls `SetMaterial` when user changes these fields).

The reason nothing appears to use these fields is because the editor abstracts scripts using `serializedObject.FindProperty(name);` which I call from a `Prop` helper.

ajrb: Sorry, that tricked me - I was really sure they were unused vestiges. Have restored them now.

ajrb:
The last question is whether I have named these classes correctly? 
I was thinking that Billboard & DaggerfallBillboard could become DaggerfallBilloard & DaggerfallUnityBillboard? 
Same for DaggerfallMobileUnit and DaggerfallUnityMobileUnit.

@Interkarma Please say if you would prefer this class naming scheme or favour the shorter names.

Interkarma:
I think this one is fine now. I'll merge and we'll see if any finds any problems in next build.
Thank you again @KABoissonneault for your help with reviews. I really appreciate your expert eyes. I have more confidence merging knowing you've also reviewed.

PR #2285, opened 2022-01-26
Add Pawn Shop to the Regional Topic

AsesinoBlade:
HI all,
I created this PR to add Pawn Shops as an option in Regional.   This code will only work after the addressable is rebuilt.   As per Internkarma's direction, I have not included that binary.

I understand from the Forums that Pawn Shop was also not available in Daggerfall Classic (although, I have not personally verified that), so if you would prefer not to include this code, I totally understand.   

Thank you for your consideration
a


Interkarma:
This works great, thank you. :)

PR #2289, opened 2022-01-28
Fixed PopulationManager's lag

John-Doom:
Implements coroutines to fix PopulationManager's lag spikes when spawning new people.


Interkarma:
Hi. :)

I'm seeing new classes files here (e.g. TextureReaderAtlas, TextureReplacementBillboard) that look to be re-implementing other parts of code from TextureReplacement or TextureReader) just to add coroutine versions of them. This would mean we now need to maintain both code paths if something changes affecting both. I'd prefer no such large change at this stage of development.

I'm also not aware of any lag when loading populations, certainly not in the base game. When it comes to modded materials, the materials are cached on first load, so any performance spikes should be fairly temporary and mitigated on subsequent drops of mobile NPCs.

Can you provide more context about what initiated this PR? What conditions/mods create the lag spikes and how bad are they? Does the material cache not mitigate this enough?

I try to remain flexible with very minor changes to core during hard feature lock, but this one is is unfortunately beyond that. I'll wait for your feedback, and possibly we can rework to be smaller. But at this stage, I'd say this one will need to wait until after 1.0 and I'm done with DFU.

John-Doom:
Base vanilla game. Profiling confirmed lag spikes on Populationmanager.Update putting fps around 15. They're caused by loading new textures, so it's probably going to be inevitable, nevermind with mods. New classes could recyle the original code by making and calling general-use versions of them.

Interkarma:
Textures are cached after load, shouldn't the only lag spike be when player starts game or enters a new region with new population sprites? Is the population manager a major cause of lag when loading textures for the first time only, or as player is moving around town and population is spawning/despawning?

John-Doom:
As population is spawning, moving around town.

Interkarma:
The thing is, I see no such lag when testing population manager and don't agree there's anything that needs fixing.

I tried to understand the problem and setup my own profiling. I tested in Daggerfall city so there's a max-size population and observed the following patterns:
* When loading textures first time, update takes around 25ms per load and is spread over multiple frames. A bit heavy but happens only once and materials are then cached. The highest time I saw for initial load was 75ms, but again a once-off.
* As population is spawned and despawned is typically 6-8ms. The highest I saw was 17ms before pooling was filled out.
* Population manager is only executed once every ten frames and uses object pooling alongside texture caching. Other than initial load time, ongoing maintenance cost is quite low.
* I see no frame drops (consistently 60fps) while moving around town and don't experience any missing frames or lag.

I think we need to rewind a bit and try to understand why this is happening for you in the first place. From everything I can see, population manager is quite performant and isn't doing much heavy work at all once textures are loaded. The class already has optimisations such as throttling, caching, and object pooling, so any further optimisation here is likely targeting the wrong problem.

If you're experiencing drops down to 15fps, that sounds very much like a local issue. All I can suggest for now is comparing your performance to a clean build (new DaggerfallGameFiles method > new folders > new settings.ini > no mods) and see if the problem continues. If you don't experience frame drops moving around town in clean build, the next step would be a fresh clone of the source project and try to better identify where the lag is coming from.

I do believe you that you're experiencing something here, but I don't observe same. We'd first have to identify we're even fixing the right problem to start with before changing any code.

I'll close this one for now. Please feel free to update comments and I'm happy to continue conversation. I may not be able to help much more, but I'll do my best. Good luck! :)

PR #2292, opened 2022-01-31
Only warn about non-optional missing dependencies

petchema:
Forums: https://forums.dfworkshop.net/viewtopic.php?p=56707#p56707


KABoissonneault:
Thank you, I keep noticing this when using my Unleveled Spells without Hotkey Bar and it bothered me, but never took the time to fix it :D

KABoissonneault (approved):

Interkarma:
Thank you Pango! :)

PR #2293, opened 2022-01-31
"Fixed" armor values showing up in the Character Sheet paperdoll­

KABoissonneault:
Request from a modder doing lots of paperdolls. The numbers are still in the Inventory window of course

EDIT: 
Images for comparison
image
image


Interkarma:
Thank you. :)

PR #2295, opened 2022-02-04
6 methods changed to virtual to allow modding

Osorkon21:
Apologies for the last pull request I immediately closed - still a novice with GitHub.

This changes 6 methods to virtual so I can mod them. Reasons are:

EnemyEntity: To unlevel enemies, nerf knockback stunlocks
CreateCharCustomClass: To tweak character creation
EnemyDeath: To display custom death messages
PlayerActivate: To display enemy level when enemy activated
DaggerfallDungeon: To mod dungeon enemies (I would've changed RDBLayout but it's static so I went upstream to non-static class)

### DIFF
--- a/Assets/Scripts/Game/Entities/EnemyEntity.cs
+++ b/Assets/Scripts/Game/Entities/EnemyEntity.cs
@@ -244,7 +244,7 @@ namespace DaggerfallWorkshop.Game.Entity
         /// <summary>
         /// Sets enemy career and prepares entity settings.
         /// </summary>
-        public void SetEnemyCareer(MobileEnemy mobileEnemy, EntityTypes entityType)
+        public virtual void SetEnemyCareer(MobileEnemy mobileEnemy, EntityTypes entityType)
         {
             // Try custom career first
             career = GetCustomCareerTemplate(mobileEnemy.ID);
@@ -609,7 +609,7 @@ namespace DaggerfallWorkshop.Game.Entity
             }
         }
 
-        public int GetWeightInClassicUnits()
+        public virtual int GetWeightInClassicUnits()
         {
             int itemWeightsClassic = (int)(Items.GetWeight() * 4);
             int baseWeight;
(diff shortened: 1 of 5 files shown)


Assets/Scripts/Game/Entities/EnemyEntity.cs
```diff
@@ -244,7 +244,7 @@ public override void ClearConstantEffects()
         /// <summary>
         /// Sets enemy career and prepares entity settings.
         /// </summary>
-        public void SetEnemyCareer(MobileEnemy mobileEnemy, EntityTypes entityType)
+        public virtual void SetEnemyCareer(MobileEnemy mobileEnemy, EntityTypes entityType)
         {
```
KABoissonneault, line 248: Note that this `virtual` isn't strictly necessary for mods. All the fields `SetEnemyCareer` modifies are public, so mods can register to `OnLootSpawned` and change back the fields from there.

But sure, it's cleaner if your custom entity type can just set them right to begin with

Assets/Scripts/Game/UserInterfaceWindows/CreateCharCustomClass.cs
```diff
@@ -439,7 +439,7 @@ void ExitButton_OnMouseClick(BaseScreenComponent sender, Vector2 position)
 
         #region Private methods
 
-        private void UpdateDifficulty()
+        protected virtual void UpdateDifficulty()
```
KABoissonneault, line 442: Curious, can you actually get DFU to use a derived `CreateCharCustomClass` from a mod? From what I'm seeing, `CreateCharCustomClass` is created by name from `StartNewGameWizard`, so I don't see how a mod would override this `virtual` function and have it get called

Assets/Scripts/Game/Entities/EnemyEntity.cs
Osorkon21: I use `SetEnemyCareer` to do a couple of things besides set field values. I could try to manually erase all generated items from every enemy and respawn them myself, for example, but like you said it'll be a lot cleaner if I can change this directly.

Assets/Scripts/Game/UserInterfaceWindows/CreateCharCustomClass.cs
Osorkon21: Good question, I haven't tried this myself yet and I don't know. 

This change is the least essential of the bunch - all I would here do is give player extra points on the difficulty dagger at character creation if player sets custom class HP lower than I recommend. If it works, great, if not, it's no big deal.

KABoissonneault (approved):

PR #2296, opened 2022-02-06
New races & Cloning images

John-Doom:
Implements:
- support to register new races;
- JsonClone: a new json template to create new images from existing data and new textures.

You can use this mod to test the new features: [OrcRace.zip](https://github.com/Interkarma/daggerfall-unity/files/8010161/OrcRace.zip)

### DIFF
--- a/Assets/Scripts/API/ItemsFile.cs
+++ b/Assets/Scripts/API/ItemsFile.cs
@@ -162,6 +162,11 @@ namespace DaggerfallConnect.FallExe
         public int id;
     }
 
+    public struct JsonClone
+    {
+        public string src;
+    }
+
     /// <summary>
     /// Reads item template data from FALL.EXE and structures for use.
     /// There is no "items file" as such for this data, just keeping
--- a/Assets/Scripts/Game/UserInterfaceWindows/CreateCharRaceSelect.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/CreateCharRaceSelect.cs
@@ -4,7 +4,7 @@
 // License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
 // Source Code:     https://github.com/Interkarma/daggerfall-unity
 // Original Author: Gavin Clayton (interkarma@dfworkshop.net)
-// Contributors:    
+// Contributors:    John Doom
 // 
 // Notes:
 //
@@ -32,6 +32,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
 
         Texture2D nativeTexture;
         TextLabel promptLabel;
+        TextLabel customLabel;
         DFBitmap racePickerBitmap;
         RaceTemplate selectedRace;
 
@@ -67,6 +68,16 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             promptLabel.ShadowColor = DaggerfallUI.DaggerfallDefaultShadowColor;
             promptLabel.ShadowPosition = DaggerfallUI.DaggerfallDefaultShadowPos;
 
+            // Add custom races prompt
+            if (RaceTemplate.CustomRaces.Count > 0)
+            {
+                customLabel = DaggerfallUI.AddTextLabel(DaggerfallUI.DefaultFont, new Vector2(0, 24), "Press R for more races.", NativePanel);
+                customLabel.HorizontalAlignment = promptLabel.HorizontalAlignment;
+                customLabel.TextColor = promptLabel.TextColor;
+                customLabel.ShadowColor = promptLabel.ShadowColor;
+                customLabel.ShadowPosition = promptLabel.ShadowPosition;
+            }
+
             // Handle clicks
             NativePanel.OnMouseClick += ClickHandler;
         }
@@ -80,6 +91,10 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
                 selectedRace = null;
                 CancelWindow();
             }
+
+            //test
+            if (Input.GetKeyDown(KeyCode.R))
+                ShowList();
         }
 
         public void Reset()
@@ -87,6 +102,8 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             selectedRace = null;
             if (promptLabel != null)
                 promptLabel.Enabled = true;
+            if (customLabel != null)
+                customLabel.Enabled = true;
         }
 
         void ClickHandler(BaseScreenComponent sender, Vector2 position)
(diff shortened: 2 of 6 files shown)


Assets/Scripts/Utility/ImageReader.cs
```diff
@@ -371,6 +371,17 @@ public static ImageData GetImageData(string filename, int record = 0, int frame
 
                     break;
 
+                case ImageTypes.JSON:
+                    string text;
+                    AssetInjection.TextAssetReader.TryRead(filename, out text);
+                    DaggerfallConnect.FallExe.JsonClone jsonClone =
+                        (DaggerfallConnect.FallExe.JsonClone)Game.Serialization.SaveLoadManager.Deserialize(typeof(DaggerfallConnect.FallExe.JsonClone), text);
+                    string sourceName = jsonClone.src;
+                    string actualName = filename.ToLower().Replace(".json", "") + "_" + record + "-" + frame;
```
KABoissonneault, line 380: `OrcRace_PaperDollHeadsMale.json` in your mod has a `dest` field, so it seems you intended to put "actual name" in there at first.
Any reason you chose to replace `dest` with `filename.ToLower().Replace(".json", "")`? I guess maybe it's best if the JSON has the same name as the replacement assets. Wondering if you had issues with the `dest` approach

Assets/Scripts/Utility/ImageReader.cs
```diff
@@ -371,6 +371,17 @@ public static ImageData GetImageData(string filename, int record = 0, int frame
 
                     break;
 
+                case ImageTypes.JSON:
+                    string text;
+                    AssetInjection.TextAssetReader.TryRead(filename, out text);
```
KABoissonneault, line 376: Just a general question about the general need for the new archive format.
Do I understand that this format allows you to take the image data of another archive while using new textures? Seems kind of orthogonal to the feature, but I suppose it's useful

Assets/Scripts/Utility/ImageReader.cs
John-Doom: No issues, just tried to make life easier for modders😁

Assets/Scripts/API/ItemsFile.cs
```diff
@@ -162,6 +162,11 @@ public struct Ingredient
         public int id;
     }
 
+    public struct JsonClone
```
KABoissonneault, line 165: Can you add move context to describe what this is used for? `JsonClone` doesn't say much in itself. Maybe `JsonArchiveClone` would at least be more clear to someone familiar with texture archives.

Assets/Scripts/Utility/ImageReader.cs
John-Doom: That's exactly the case. I was trying to find a way to make new image files with all the data required (offsets, sizes etc). After a few experiments, i found this solution to be the easiest for modding

Assets/Scripts/API/ItemsFile.cs
John-Doom: Would it make sensefor other file formats as well? JsonClone's struct could be potentially used for them as well

KABoissonneault: Just handle the code you're adding, no need to touch other code

John-Doom: What do you mean? I was wondering if the name "JsonArchoveClone" would fit other ends as well in the future, besides cloning image files

ajrb:
Could you explain why you've taken this approach rather than the existing asset replacement mechanisms?

John-Doom:
> Could you explain why you've taken this approach rather than the existing asset replacement mechanisms?

To add new body textures for new races, not replace existing ones

Assets/Scripts/API/ItemsFile.cs
John-Doom: > Just handle the code you're adding, no need to touch other code

What do you mean? I was wondering if the name "JsonArchiveClone" would fit other ends as well in the future, besides cloning image files

ajrb:
The asset replacement systems have been used for new images in other cases. I think it would be helpful to explain the technical reasons that was not viable, or the trade offs that you considered before taking this approach to make it easier for reviewers.

John-Doom:
> The asset replacement systems have been used for new images in other cases. I think it would be helpful to explain the technical reasons that was not viable, or the trade offs that you considered before taking this approach to make it easier for reviewers.

Unless I'm mistaken, there's no way to import new image files, because the game checks them in the arena2 folder first; using this new system let you overcome this problem and clone data needed from the original files.

KABoissonneault:
> there's no way to import new image files, because the game checks them in the arena2 folder first
> Some mods managed to add new sprites for NPCs, for example, but they had to make new unity prefabs to do that

While it has been true by default in DFU's development, it's not true nowadays. Many mods do import new textures without providing their own prefab. Examples of this include custom inventory icons, Cliff's new sprites for Famous People of the Bay, and Kamer's work on custom enemies. 
What happens in this case is that DFU does not use any image data to provide properties for the textures, and relies instead on defaults, best guesses computed from other data, and properties set by mod-provided XML files. Those XML files, however, are notoriously tedious to do, because you need to essentially copy-paste them for every frame.
What your new solution provides is the ability to associate a bunch of frames with an existing archive, reusing its image data.

So it does provide something new, hopefully this clears it up for Hazel. But it means this PR is two new orthogonal features being sent during the feature freeze, though not that big and pretty low impact on existing code.

John-Doom:
I've forgot to upload my patch to the choose race screen. In case we're going to remove JsonClone, I've just tested a way to make new sprites using xmls and a new function to retrieve imageData.
I'm willing to rework the code as many times as you need, IK.

Assets/Scripts/Game/UserInterfaceWindows/CreateCharRaceSelect.cs
```diff
@@ -80,13 +91,19 @@ public override void Update()
                 selectedRace = null;
                 CancelWindow();
             }
+
+            //test
+            if (Input.GetKeyDown(KeyCode.R))
```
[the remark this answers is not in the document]
John-Doom: it is, but could be updated later. i'd like to make a button, but i dunno how it would look like with mods like dream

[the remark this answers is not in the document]
John-Doom: I will look into it, if this patch will move forward o.O

Interkarma:
If it's OK with everyone, I'll leave this one in the hands of @ajrb and @KABoissonneault to approve. If they approve, I'm happy to merge and roll into upcoming preview builds (0.14).

KABoissonneault:
In my opinion, this is too rushed for a stable release. Once we release this into a major release, mods are going to depend on this new modding API, and in theory that means we won't be able to make breaking modifications on it. Custom races are a big enough feature that we're definitely going to need extra work on it (ex: racial bonuses, modifying the continent map, lots of dialogue stuff, etc).
If we're going to integrate it, at least be sure to make clear that this is an experimental API, so only modders ready to live with the consequences start implementing it.

I do like the JSON utility, though. I'm confident it will be useful.

Really wish we could change the hardcoded key to a visible button though.

John-Doom:
Since I was able to remake this into a mod, you can probably close this pr. The only thing that still deserves attention is the json/xml way to implement new image files; that can't be properly modded and may be useful for all modders. If you're interested, I can make a new pr just for that one.

Interkarma:
Hi @John-Doom thanks for update. I'll close this one for now. I think the new JSON method looks interesting, and the others also seemed positive about it. If you're up to it, go ahead and submit that as its own PR.

PR #2297, opened 2022-02-10
Fix bug where ally/enemy propagation is not performed for Knightly Orders

ajrb:
Unintended bug in my code that i spotted.


Interkarma:
Thank you Hazelnut! :)

PR #2298, opened 2022-02-11
Fix 2 usages of Clamp() that were not assigning the result

ajrb:


Interkarma:
Thank you again Hazelnut. :)

PR #2300, opened 2022-02-17
PR #2284 mod compilation issue workaround

KABoissonneault:
As discussed with IK, the change from `DaggerfallBillboard.BillboardSummary` to `Billboard.BillboardSummary` can cause mods relying on `BillboardSummary` to fail to compile. To mitigate this, I moved the struct to the DaggerfallWorkshop namespace instead.


Interkarma:
Thank you Kab! I'll merge this and proceed to next builds later this week when I have some downtime again.

PR #2301, opened 2022-03-08
Changed the way rotations are applied on models / doors in RDBLayout

BadLuckBurt:
Changed the way rotation is applied to the models to avoid breaking orientation. All rotations are now set at once instead of being applied one after another.

Me and Ralzar both ran into the issue of certain rotation combinations not working, it will look fine in the editor but not when the RDB data is loaded in-game.

I linked to Imgur but I've added the images here now so that they won't go missing at some point. 

I compared the way rotation is handled in RMBLayout, there the model rotations are applied in one pass instead of applying each axis rotation separately. I have changed the code in RDBLayout to do the same and this solves the rotation issues. Most classic models only rotate around 1 axis so they won't exhibit this issue.

In the Unity Editor:
UnityEditor1

In the live build:
LiveBuild1

In my custom build:
MyBuild1

In the Unity Editor:
UnityEditor2

In the live build:
LiveBuild2

In my custom build:
MyBuild2
)

I've also supplied Ralzar's RDB, go to the Bish Mines in Dragontail Mountains, when you enter you will be right beside the two beams that have the wrong orientation in the live build.

[N0000052.RDB.zip](https://github.com/Interkarma/daggerfall-unity/files/8204594/N0000052.RDB.zip

### DIFF
--- a/Assets/Scripts/Utility/RDBLayout.cs
+++ b/Assets/Scripts/Utility/RDBLayout.cs
@@ -687,23 +687,14 @@ namespace DaggerfallWorkshop.Utility
         /// </summary>
         private static Matrix4x4 GetModelMatrix(DFBlock.RdbObject obj)
         {
-            // Get rotation angle for each axis
-            float degreesX = -obj.Resources.ModelResource.XRotation / BlocksFile.RotationDivisor;
-            float degreesY = -obj.Resources.ModelResource.YRotation / BlocksFile.RotationDivisor;
-            float degreesZ = -obj.Resources.ModelResource.ZRotation / BlocksFile.RotationDivisor;
-
-            // Calcuate transform
+            // Calculate transform
             Vector3 position = new Vector3(obj.XPos, -obj.YPos, obj.ZPos) * MeshReader.GlobalScale;
 
-            // Calculate matrix
-            Vector3 rx = new Vector3(degreesX, 0, 0);
-            Vector3 ry = new Vector3(0, degreesY, 0);
-            Vector3 rz = new Vector3(0, 0, degreesZ);
+            // Calculate rotation
+            Vector3 rotation = new Vector3(-obj.Resources.ModelResource.XRotation / BlocksFile.RotationDivisor, -obj.Resources.ModelResource.YRotation / BlocksFile.RotationDivisor, -obj.Resources.ModelResource.ZRotation / BlocksFile.RotationDivisor);
+
             Matrix4x4 modelMatrix = Matrix4x4.identity;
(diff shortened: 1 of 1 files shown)


KABoissonneault:
From Unity's doc (https://docs.unity3d.com/ScriptReference/Transform.Rotate.html):

> Applies a rotation of eulerAngles.z degrees around the z-axis, eulerAngles.x degrees around the x-axis, and eulerAngles.y degrees around the y-axis (in that order).

So seems like `GetModelMatrix` had the order consistent with `Transform.Rotate`, but not `AddActionDoor`.
Your fix is good though. Just wanted to add this in case you were wondering why a "one pass" rotation gave different results from the previous code.

Interkarma:
Sadly this approach breaks some objects in game. My original attempts back in 2009 looked a lot like this, and it unfortunately doesn't work reliably. The reason I know this and where the points of failure are is because I still have PTSD-like flashbacks just looking at that block of code.

One example are the coffins in Scourg Barrow, which will end up rotated the wrong way.

image

Another are levers that end up not attached to walls properly or rotating in wrong direction.

image

The story I told on Discord wasn't just for humour's sake - making this work as well as it does literally gave me nightmares.

I'm sorry, I'm not willing to change this so close to release for the sake of mods. I know it's frustrating trying to work in with Daggerfall's data jank, but that's the hand we've been dealt.

BadLuckBurt:
No problem and I am sorry, I did not mean to give you flashbacks. I've dealt with these rotations in Blender too when I worked on the importing stuff so I completely understand the pain.

I went and checked several dungeons with this PR active but clearly not enough, I even went to Scourg Barrow but missed what you show here. My 'solution' basically reverses the problem and we need levers more than well positioned beams so I wholeheartedly agree with the closure of this PR. 

Rotations are just an immense PITA. I will see if there is any other approach I can take to get past this problem but otherwise it is what it is.

BadLuckBurt:
I have thought of another solution using a mod with custom activations to exclude the models from being combined and then running a separate script on the interior transition to apply the rotations for those models in the way that works for them. There aren't that many and if I just keep that mod updated with new models as we find them, it will work fine.

Again, I'm sorry. I know how horrible this stuff is and this solution will work without interfering with anything so nothing lost, everything gained.

BadLuckBurt:
No worries, it's nightmarish stuff. I will deal with the mod first and probably set up an RDB with every single model to test problematic rotations. It's quite easy to generate that with the data I have, a lot of models can be excluded too because we're dealing with props. I'll see what I find and take it from there. I don't believe that custom models suffer from this problem, at least I haven't witnessed it with mine yet but I'll have to investigate.

PR #2302, opened 2022-03-09
Don't override faction id if the replacement id is zero

ajrb:
This was added to allow named buildings to be altered without affecting the pool matching. Unfortunately I missed this one when I made that change.


BadLuckBurt:
Yes, this is the same change I tested in my own build when Cliffworms mentioned the issue on Discord, this should be all that's needed to make it fall in line with the named buildings behaviour.

ajrb:
Yes I am pretty confident on this change, and usually would just merge it myself but a release is imminent so leaving it to Interkarma whether he includes this fix. It's not urgent at all since Cliff has a work around that causes no issues.

Interkarma:
I'm still testing builds. Happy to include this and re-run them. When I tag release on git, I'll tag this merge.

PR #2316, opened 2022-04-24
4 packages to unlock some future performance gains (for either vanilla release or for modders)

andrew-raphael-lukasik:
Hi,

This is a small and basic PR. It adds:
- `com.unity.mathematics`
- `com.unity.collections`
- `com.unity.jobs`
- `com.unity.burst`
Hope you approve it.

### Reasons behind it

Adding these 4 packages won't change anything by virtue of just adding them but those can enable some worthwhile performance gains in few areas where the workload can be considerably high. There is probably not a lot of cases like that, but those few still impact the performance & gameplay experience and that is noticeable with older CPUs (my i3-4170 among them).

To give an example, a Burst compilation could make all existing `IJob`s about >=10x faster after adding `BurstCompile` attribute and few changes (removing managed objects and static code references). This won't make a noticeable difference in vanilla release but I saw that a popular terrain mod can stall the main thread for up to few dozen ms for like >1.5s - and this PR could help fix that in the future releases.
To give another example, not a big deal but I also saw a cpu-based billboards somewhere - this one can be flattened with super performant Burst-compiled `IJobParallelForTransform`. More small cases like that will appear in the future I'm sure.

Screenshot 2022-04-24 164021

### DIFF
--- a/Packages/manifest.json
+++ b/Packages/manifest.json
@@ -1,11 +1,15 @@
 {
   "dependencies": {
     "com.unity.addressables": "1.19.11",
+	"com.unity.burst": "1.7.1",
     "com.unity.collab-proxy": "1.2.16",
+	"com.unity.collections": "0.9.0-preview.6",
     "com.unity.ide.rider": "1.2.1",
     "com.unity.ide.visualstudio": "2.0.8",
     "com.unity.ide.vscode": "1.2.3",
+    "com.unity.jobs": "0.2.10-preview.13",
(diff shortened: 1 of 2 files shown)


Interkarma:
I can't see any downsides to including those packages. It shouldn't break anything and doesn't increase build size.

As this one triggers a reimport, I'll merge later once I've cleared the majority of PRs in my queue and I'm done switching around.

PR #2317, opened 2022-04-25
Fix ground seams on some hardware/drivers

petchema:
Attempt to fix ground seams to appear on some hardware/drivers (AMD, integrated Intel graphics), based on
https://forum.unity.com/threads/tiling-textures-within-an-atlas-by-wrapping-uvs-within-frag-shader-getting-artifacts.535793/

Again, I no longer have access to some hardware to test this issue, so I need feedback from developers with such hardware!

Forums: https://forums.dfworkshop.net/viewtopic.php?t=1464

### DIFF
--- a/Assets/Shaders/DaggerfallTilemap.shader
+++ b/Assets/Shaders/DaggerfallTilemap.shader
@@ -40,9 +40,9 @@ Shader "Daggerfall/Tilemap" {
 		sampler2D _TileAtlasTex;
 		sampler2D _TilemapTex;
 		sampler2D _BumpMap;
-		int _TilesetDim;
-		int _TilemapDim;
-		int _MaxIndex;
+		uint _TilesetDim;
+		uint _TilemapDim;
+		uint _MaxIndex;
 		float _AtlasSize;
 		float _GutterSize;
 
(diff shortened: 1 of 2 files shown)


KABoissonneault:
Worth considering making a custom DFU build with the changes and uploading it on the thread, see if you can get any user to test it

magicono43:
I just tested your changes on my system through changing the lines in that file for my Unity Editor version of DFU 0.13.5. So this test was not in the compiled version just the Unity Editor environment.

System Specs shown from Player.log
Apparent_System_Specs

Example of seams before putting your changes.
Unchanged File

Example of seams after putting your changes, also exited and entered a nearby building to make sure there was not anything sticking from loading the save or whatever.
Changed File

So at least on my side through the editor by changing the shader file directly the seams appear to still be there.

petchema:
From live experimentation with Magicono on Discord, modifying Padding shifts the textures, but not the seams.
So it doesn't fix that issue :(

Interkarma:
I bookmarked this a while back.

https://bgolus.medium.com/distinctive-derivative-differences-cce38d36797b

Feels like it could have some good info to digest, but I haven't had time to unpack it yet. Sorry if you've come across this one already, Pango.

petchema:
I managed to reproduce the problem on some laptop, forcing the Unity Editor to use the integrated Intel chip.
And I think I've found a good fix, commit message:

> The issue does not seem related to coordinates wrapping, but the fact we
rely on 2 separate computations (texture sampling for tile index on one
side, frac() of uv on the other for tile coordinates) that do not
necessarily match perfectly.
>
> To fix that in a portable and reliable way, use floor()/frac() for both.
Cost is an extra division by _TilemapDim, that's not huge. Maybe
somebody will find a way to get rid of it.

magicono43:
Here is my results after testing with your new commit. Top is without, below that is with.

Interkarma:
Congrats Pango, looks like you've slain that dragon. :)

ajrb:
Oh wow!! Brilliant, I'd almost given up hope we were going to be able to fix this.

@petchema Awesome work mate!

petchema:
@ajrb Thank you!

I totally forgot about DaggerfallTilemap.shader, what it is used for already? I tried to follow references, but it's hard for me to tell what's deprecated

Interkarma:
DaggerfallTimemap.shader is still used when system doesn't support texture arrays or user sets EnableTextureArrays=false.

Provider is selected in TerrainMaterialProvider.Default property, and each provider uses the other shader.

https://github.com/Interkarma/daggerfall-unity/blob/982a071f3fb42e20dbb300a19e2235a842da0dd2/Assets/Scripts/Terrain/TerrainMaterialProvider.cs#L90

Disabling EnableTextureArrays in settings.ini should be enough to trigger a test.

Interkarma:
Again, fantastic work. Thank you Pango!

PR #2319, opened 2022-04-29
QoL: Replace ClickToAttack with WeaponSwingMode (Vanilla, Click, Hold)

Nattfarinn:
`ClickToAttack` option is awesome Quality of Life addition to the main game, but constant clicking may (and from my experience, is) be tiresome for player and for other family members. Also deadly to the mouse. The quality of life improvement proposed here is to allow "Hold to Attack" option without disabling ones previously available.

Unfortunately this change is not achievable from mod and has to touch the Core (`WeaponManager`), but in a surprisingly non-invasive manner.

The options are:
 - `Vanilla`: plain-old "hold and swing",
 - `Click`: DFU standard,
 - `Hold`: keeping the keybind pressed triggers new attack after another.

Startup settings:
Screenshot 2022-04-29 at 02 10 36
In-game Advanced settings:
Screenshot 2022-04-29 at 02 31 06


Assets/Scripts/Game/WeaponManager.cs
```diff
@@ -202,6 +202,13 @@ void Start()
             SetMelee(ScreenWeapon);
         }
 
+        bool IsHoldAttackMode()
+        {
+            return DaggerfallUnity.Settings.AttackMode == 2 && InputManager.Instance.GetKey(
```
KABoissonneault, line 207: Nitpick, the function could be "IsHoldingAttackMode" I think. The current name made me think it was checking for attack mode, but it's also checking for input being held

KABoissonneault (approved):

Assets/Scripts/Game/WeaponManager.cs
Nattfarinn: You're right. What method actually does is `IsHoldingHoldAttackMode`. I've moved the button held check to InputManager.

Assets/Scripts/Game/WeaponManager.cs
```diff
@@ -311,8 +318,9 @@ void Update()
                 }
                 else
                 {
-                    // Player must click to attack
-                    if (InputManager.Instance.ActionStarted(InputManager.Actions.SwingWeapon))
+                    // Player must click or hold to attack, depending on attack mode
+                    if (InputManager.Instance.ActionStarted(InputManager.Actions.SwingWeapon)
+                        || (DaggerfallUnity.Settings.AttackMode == 2 && InputManager.Instance.IsSwingWeaponHeld()))
```
jefetienne, line 323: You can replace this 'if' statement line with

```cs
 if (InputManager.Instance.ActionStarted(InputManager.Actions.SwingWeapon)
     || (InputManager.Instance.HasAction(InputManager.Actions.SwingWeapon)
        && DaggerfallUnity.Settings.AttackMode == 2))
```

And remove all changes from InputManager. What will happen is if the ActionStarted, then it will do ClickToAttack normally. When it hits the next frame, ActionStarted will be false, but it will still notice that SwingWeapon is still being held, and it will verify that the AttackMode is 2 (held) in order to proceed the swinging.

Assets/Scripts/Game/WeaponManager.cs
```diff
@@ -220,6 +220,13 @@ void Update()
                 isAttacking = false;
                 isDamageFinished = false;
                 isBowSoundFinished = false;
+
+                if (InputManager.Instance.HasAction(InputManager.Actions.SwingWeapon)
+                    && DaggerfallUnity.Settings.AttackMode == 2
+                    && InputManager.Instance.IsSwingWeaponHeld())
+                {
+                    InputManager.Instance.StopAction(InputManager.Actions.SwingWeapon);
```
jefetienne, line 228: This block can be removed as well based on the if-statement comment

Assets/Resources/defaults.ini.txt
```diff
@@ -134,7 +134,7 @@ WeaponAttackThreshold=0.03
 WeaponSensitivity=1.0
 MovementAcceleration=False
 ToggleSneak=False
-ClickToAttack = False
+AttackMode = 0
```
jefetienne, line 137: Not a huge thing, but while I think "Click to Attack" wasn't a great setting name, I think "AttackMode" might sound a bit more vague. Maybe call it "WeaponSwingMode"?

Assets/Scripts/Game/InputManager.cs
```diff
@@ -79,6 +79,7 @@ public class InputManager : MonoBehaviour
         KeyCode[] heldKeys = new KeyCode[totalHeldKeys];
         KeyCode[] previousKeys = new KeyCode[totalHeldKeys];
         KeyCode heldModifier;
+        KeyCode swingWeaponBinding = KeyCode.Mouse1;
```
jefetienne, line 82: All changes in InputManager can be removed from the if-statement comment suggestion.

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallUnityMouseControlsWindow.cs
```diff
@@ -123,15 +123,15 @@ protected override void Setup()
...
+            mouseSensitivitySlider = CreateSlider("Mouse Look Sensitivity", 15, 100, 0.1f, 8.0f, DaggerfallUnity.Settings.MouseLookSensitivity);
+            invertMouseVerticalCheckbox = AddOption(20, 120, "Invert Look-Y", DaggerfallUnity.Settings.InvertMouseVertical);
+            mouseSmoothingCheckbox = AddOption(20, 130, "Mouse Smoothing", DaggerfallUnity.Settings.MouseLookSmoothing);
+            moveSpeedCheckbox = AddOption(20, 140, "Movement Acceleration", DaggerfallUnity.Settings.MovementAcceleration);
 
             //weaponSensitivitySlider = CreateSlider("Mouse Weapon Sensitivity", 115, 80, 0.1f, 10.0f, DaggerfallUnity.Settings.WeaponSensitivity);
-            clickToAttackCheckbox = AddOption(115, 100, "Click to Attack", DaggerfallUnity.Settings.ClickToAttack);
-            bowDrawbackCheckbox = AddOption(115, 110, "Bows - draw and release", DaggerfallUnity.Settings.BowDrawback);
-            toggleSneakCheckbox = AddOption(115, 120, "Toggle Sneak", DaggerfallUnity.Settings.ToggleSneak);
+            attackModeSlider = CreateSlider("Attack Mode", 115, 100, DaggerfallUnity.Settings.AttackMode, "Vanilla", "Click", "Hold");
```
jefetienne, line 132: From a UI perspective I found the placement of this slider to be a bit awkward. Could you put the two sliders next to each other above, and then the mouse weapon threshold below them, so it looks like this? (Image was done with Paint):

https://imgur.com/a/irC2ncp

jefetienne (changes requested):
Hey there, I like this idea!

I pulled your branch to test it out and I found a much easier way to deal with held swinging with a one-liner in one of your if-statements. With that, you can remove all of the InputManager changes and remove the code block that has the StopAction. I also think the Advanced controls window might need a bit of a tidier adjustment so I made an image with Paint for you to see what I think would look a little nicer.

Nattfarinn:
@jefetienne wow, thank you very much for the code review! I am in the middle of nowhere for a weekend but I promise to address everything once I'm back.

Assets/Resources/defaults.ini.txt
Nattfarinn: That makes sense code-wise. But should I also change the UI labels?

jefetienne: I would say so. The name would make it more clear to the player of what kind of setting it is.

Nattfarinn:
@jefetienne can't re-request review for some reason, so I ping you directly. :) Everything fixed.

jefetienne:
It looks good to me, thanks! :)

Nattfarinn:
@jefetienne Can you approve the PR? :) (not sure what are procedures here, but ypur the code review remains in Request Changes state :) )

jefetienne (approved):

Interkarma:
Hey there! I really like this concept. This is also a good change from an accessibility point of view - much easier to hold a button than to keep mashing it or swinging a mouse around.

I'll definitely merge this one for 0.14. Changes look good on a read, and than you @jefetienne for your help with review. I'll also give it a pass when I can and merge.

Interkarma:
This works great! No problems found here. Thanks for the cool feature. :)

petchema:
In retrospect I find 2 issues with how this feature is presented:

- people can miss that they're 3 choices ("who uses a slider in an UI to pick among 3 choices?");
- "click-to-attack", a term that many already know and that is rather obvious, has disappeared; So again some people can miss the fact this feature exists, and it's also somewhat harder to explain where to find it/how to enable it

Nattfarinn:
> - people can miss that they're 3 choices ("who uses a slider in an UI to pick among 3 choices?");
I agree, but I have checked if there is reusable dropdown UI element available but couldn't find one. There are few places already where slider is used instead, like ie. dungeon texure setting, helm and shield material display setting or all filter display modes.

> - "click-to-attack", a term that many already know and that is rather obvious, has disappeared; So again some people can miss the fact this feature exists, and it's also somewhat harder to explain where to find it/how to enable it
Not sure how severe that issue is. Setting is in the same place as previous one, is called "Attack Mode" and also has a tooltip that explains what it does. And has been released Jun 26, 2022 and as far as I know there were no questions what happened to "click-to-attack" on Discord or Forum since then.

petchema:
For the latter, that's something I suspect from a tweet about not being able to hit the first rat and rage quitting, but since it seems default in click-to-attack, that may have been my interpretation of the issue.
And yes, I know not all usual widget types are available; I'm not sure the situation can be improved without implementing say a combo box; Radio buttons would use more screen space

PR #2322, opened 2022-05-01
Make controller cursor and controller-look frame-independent

jefetienne:
Using the Steam Deck and capping the framerate, I noticed that if I set the game to 30 fps, the cursor and controller-look moved half as fast, 15 fps a quarter speed, and 60 the desired speed.

This commit fixes the issue by adding a multiplier with an inverse relationship where if the frames go down, the speed goes up and vice versa.

You can test the different speeds by adding an Awake function and comparing:

```cs
void Awake()
{
    Application.targetFrameRate = 30;
}
```

### DIFF
--- a/Assets/Scripts/Game/InputManager.cs
+++ b/Assets/Scripts/Game/InputManager.cs
@@ -47,6 +47,8 @@ namespace DaggerfallWorkshop.Game
         const float deadZone = 0.05f;
         const float inputWaitTotal = 0.0833f;
         const float moveAccelerationConst = 9.8f;
+        const float controllerCursorHorizontalSpeed = 900.0f;
+        const float controllerCursorVerticalSpeed = 900.0f;
 
         KeyCode[] keyCodeList;
         KeyCode[] reservedKeys = new KeyCode[] { };
@@ -113,11 +115,10 @@ namespace DaggerfallWorkshop.Game
         float joystickUIMouseSensitivity = 1.0f;
         bool cursorVisible = true;
         bool usingControllerCursor;
-        Vector2 controllerCursorPosition = new Vector2(0,0);
+        Vector2 controllerCursorPosition = new Vector2(0, 0);
         int controllerCursorWidth = 32;
         int controllerCursorHeight = 32;
-        float controllerCursorHorizontalSpeed = 300.0F;
-        float controllerCursorVerticalSpeed = 300.0F;
+        Rect controllerCursorRect;
 
         bool pauseController = false;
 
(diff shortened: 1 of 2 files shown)


KABoissonneault:
From https://docs.unity3d.com/ScriptReference/Time-deltaTime.html:

> deltaTime inside [MonoBehaviour.OnGUI](https://docs.unity3d.com/ScriptReference/MonoBehaviour.OnGUI.html) is not reliable, because Unity might call it multiple times per frame.

I'm wondering if we couldn't just make it so:
- `OnGui` only calls `GUI.DrawTexture`
- The rest is moved to `Update`
This way, that code could rely on `Time.deltaTime` (I guess unscaled so it works during Pause?) and be processed only once per frame, while the texture would still draw any time the GUI has to draw

jefetienne:
> From https://docs.unity3d.com/ScriptReference/Time-deltaTime.html:
> 
> > deltaTime inside [MonoBehaviour.OnGUI](https://docs.unity3d.com/ScriptReference/MonoBehaviour.OnGUI.html) is not reliable, because Unity might call it multiple times per frame.
> 
> I'm wondering if we couldn't just make it so:
> 
> * `OnGui` only calls `GUI.DrawTexture`
> * The rest is moved to `Update`
> 
> This way, that code could rely on `Time.deltaTime` (I guess unscaled so it works during Pause?) and be processed only once per frame, while the texture would still draw any time the GUI has to draw

Good idea, updated.

Assets/Scripts/Game/InputManager.cs
```diff
@@ -47,6 +47,8 @@ public class InputManager : MonoBehaviour
         const float deadZone = 0.05f;
         const float inputWaitTotal = 0.0833f;
         const float moveAccelerationConst = 9.8f;
+        const float controllerCursorHorizontalSpeed = 900.0f;
```
KABoissonneault, line 50: Any reason these were changed from 300.0f to 900.0f?

jefetienne: When switching over to the Update function, the cursor speed became slower. I think it's because I read that OnGUI runs more times in a frame than Update, which may be why switching to Update decreased the speed.

KABoissonneault: Yeah okay, that makes sense

KABoissonneault (approved):

Interkarma:
Thank you @jefetienne and @KABoissonneault for help with review.

PR #2323, opened 2022-05-02
Prevent setting null values for setting text in TextBox and TextLabel with empty string

jefetienne:
Setting these values to null would create exceptions elsewhere in their respective classes, including in the `set` action for TextBox.Text and TextLabel.Text.


Interkarma:
Thank you!

PR #2324, opened 2022-05-04
Add "Done" button to GameEffectsConfigWindow

jefetienne:
- Menu could only be escaped by the "Esc" button which a controller is unable to do
- Added a "buttonPanel" to occupy the bottom part, so if settings pages were to scroll, other UI elements would cut off and not interfere with the button

### DIFF
--- a/Assets/Localization/StringTables/Internal_Strings Shared Data.asset
+++ b/Assets/Localization/StringTables/Internal_Strings Shared Data.asset
@@ -3259,6 +3259,10 @@ MonoBehaviour:
     m_Key: settingsNotApplied
     m_Metadata:
       m_Items: []
+  - m_Id: 814
+    m_Key: doneButton
+    m_Metadata:
+      m_Items: []
   - m_Id: 29516644352
     m_Key: temple
     m_Metadata:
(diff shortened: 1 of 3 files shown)


KABoissonneault (approved):

Interkarma:
This should be OK as-is, but can you please check you have an up-to-date version of Localization in your Package Manager? It should be on 1.0.5 at the moment.

I just note the new text string added to the legacy incremental pool at "m_Id: 814", when the updated localization strings should use a larger unique key like "m_Id: 9244149909045248"

Otherwise PR works great and I don't need anything changed. The m_Id is fine for that new text key. Please just check Localization is on the same version. Cheers!

jefetienne:
> This should be OK as-is, but can you please check you have an up-to-date version of Localization in your Package Manager? It should be on 1.0.5 at the moment.
> 
> I just note the new text string added to the legacy incremental pool at "m_Id: 814", when the updated localization strings should use a larger unique key like "m_Id: 9244149909045248"
> 
> Otherwise PR works great and I don't need anything changed. The m_Id is fine for that new text key. Please just check Localization is on the same version. Cheers!

I just checked it and I am on 1.0.5. Up until now, I have only been adding localization strings via Text Editor, but now I know how to do it in the editor. Will do in the future!

Interkarma:
Ah, that would explain it. Cheers! :)

PR #2325, opened 2022-05-07
Quest can target Individual Static NPCs outside of dungeons

KABoissonneault:
For example, mods can add NPCs with unused faction IDs (ex: Lady Doryanna Flyte) to normal city Palaces. Unlike Daggerfall, Sentinel, and Wayrest Palace, the other palaces aren't dungeons, and previously no QuestResourceBehaviour would be allocated to Individual NPCs in RMBs (RDBLayout.AddFlat is what's handling that for vanilla Main Quest NPCs). This changelist adds the QuestResourceBehaviour component for Individuals NPCs in DaggerfallInteriors and RMB exteriors.

Impact:
- Classic NPCs with garbage faction ids corresponding to Individual factions (ex: Lord Kilbar) will now have an extra QuestResourceBehaviour component

Tested with:
- Famous Faces of the Iliac Bay (mod)
- Test quest here: [testface.txt](https://github.com/Interkarma/daggerfall-unity/files/8645740/testface.txt)
- Go to Anticlere Palace and `startquest testface`. Click on Lord Flyte (he's on the third floor) to start the quest.

Without the changes, `clicked npc _npc_` won't recognize Lord Flyte.

I also fixed an issue where non-God factions would check their child factions for guilds, which would break on Knight Order logic. For example, Lady Flyte has the Knights of the Flame as child faction, but 400 does not represent the Orders enum. This would cause GuildManager.GetGuild to throw an exception. This made quests involving custom Individual NPCs spam errors in the log, though DFU handles it gracefully, so this change is not necessary.

Impact:
- Hopefully this code path was really only used by deities...

### DIFF
--- a/Assets/Scripts/Game/Guilds/GuildManager.cs
+++ b/Assets/Scripts/Game/Guilds/GuildManager.cs
@@ -266,7 +266,8 @@ namespace DaggerfallWorkshop.Game.Guilds
                 guildGroup = (FactionFile.GuildGroups)factionData.ggroup;
 
                 // Handle temples nested under deity
-                if (factionData.children != null && (guildGroup == FactionFile.GuildGroups.None && factionData.children.Count > 0))
+                if (factionData.type == (int)FactionFile.FactionTypes.God
+                    && factionData.children != null && (guildGroup == FactionFile.GuildGroups.None && factionData.children.Count > 0))
                 {
                     FactionFile.FactionData firstChild;
                     if (persistentFactionData.GetFactionData(factionData.children[0], out firstChild))
--- a/Assets/Scripts/Game/Questing/QuestMachine.cs
+++ b/Assets/Scripts/Game/Questing/QuestMachine.cs
@@ -1293,6 +1293,44 @@ namespace DaggerfallWorkshop.Game.Questing
         }
 
         /// <summary>
+        /// Checks if NPC is a special individual NPC, then sets it up with components required for quests.
+        /// If the NPC has an Individual faction, and the NPC is currently placed somewhere else in a quest,
+        /// the GameObject will be deactivated
+        /// </summary>
+        /// <param name="go">GameObject representing the static NPC</param>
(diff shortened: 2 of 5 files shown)


Interkarma:
Thank you. This is a good change, and I can't see any issues with it. I'll work it over a bit more and merge soon.

Assets/Scripts/Game/Guilds/GuildManager.cs
```diff
@@ -266,7 +266,8 @@ public FactionFile.GuildGroups GetGuildGroup(int factionId)
                 guildGroup = (FactionFile.GuildGroups)factionData.ggroup;
 
                 // Handle temples nested under deity
-                if (factionData.children != null && (guildGroup == FactionFile.GuildGroups.None && factionData.children.Count > 0))
+                if (factionData.type == (int)FactionFile.FactionTypes.God
```
Interkarma, line 269: I wanted to look through the history of this one. Other than a comment change, this part of code has been relatively untouched since 2018.

image

Looking at the methods which call this one, I can't see any immediate problems with limiting this to just `FactionTypes.God`. That seems to be the intent of this conditional from the start.

I'll quickly tag @ajrb for feedback in case we're making an error here. Hazelnut wrote this and might recall if there are issues limiting to just `FactionTypes.God` child factions.

If nobody raises any issues, I'll go ahead and merge.

ajrb: Pretty sure this will be okay. It was only temples that would have children and no guild on the faction id so I never thought to limit it just to god factions. As long as temple guilds still work then this change is fine I think.

Interkarma: Thank you Hazel! I'll run through some tests with temple guilds and merge if no problems found.

Interkarma:
No problems found with temple quests or named NPCs linked to quests. Merging for 0.14.0 release.

PR #2328, opened 2022-05-11
CreateCharClassQuestion: Allow scrolling with controller right stick

muttleyxd:
I tried playing this game on Steam Deck, but I was unable to scroll through questionnaire during character creation (it doesn't react to any input, apart from clicking first visible answer).

I have quickly hacked a solution for this - simulating mouse scroll down event when right stick is being used (though possibly actual scrolling code could have been moved to separate function [so we won't call function with misleading name when using controller input], but I decided to leave it as it is).

Also I made it react to both axes - I have tested this on controller emulator first (because I didn't have any to plug to my dev PC) and it apparently doesn't output proper vertical values. So this version works fine both on Deck and with controller emulator.

### DIFF
--- a/Assets/Scripts/Game/InputManager.cs
+++ b/Assets/Scripts/Game/InputManager.cs
@@ -1199,6 +1199,41 @@ namespace DaggerfallWorkshop.Game
             }
         }
 
+        /// <summary>
+        /// Calculates UI scroll movement value based on analog axes input.
+        /// Return value:
+        /// Positive values (> 0.1f) - scrolling downwards
+        /// Negative values (< -0.1f) - scrolling upwards
+        /// </summary>
+        public float GetUIScrollMovement()
+        {
+            float horizontal = Input.GetAxis(cameraAxisBindingCache[0]);
+            float vertical = Input.GetAxis(cameraAxisBindingCache[1]);
+
+            if (GetAxisActionInversion(AxisActions.CameraHorizontal))
+                horizontal *= -1;
+
+            if (GetAxisActionInversion(AxisActions.CameraVertical))
+                vertical *= -1;
+
+            float calculatedDeadzone = Mathf.Sqrt(horizontal * horizontal + vertical * vertical);
+            if (calculatedDeadzone <= JoystickDeadzone)
+                return 0.0f;
+
+            /*
+             * With most controllers, positive values on vertical axis usually mean that stick is being tilted upwards.
+             * Similarly, negative values mean that stick is tilted downwards.
+             *
+             * Most common behavior when scrolling is to scroll in the same direction, where the stick is being tilted.
+             * So to achieve this we need to invert vertical axis value.
+             * This issue does not occur with horizontal axis - right tilt is being interpreted as scrolling downwards.
+             */
+            vertical *= -1;
+
+            // Use both analog axes for scrolling movement - for now it doesn't matter which axis we use for scrolling
+            return horizontal + vertical;
+        }
+
         #endregion
 
         #region Public Static Methods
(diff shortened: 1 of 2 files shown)


Interkarma:
Hey there, thank you for PR. :)

Just checking something - on Steam Deck, are you not able to click on the top and bottom parts of scroll to scroll up and down?

muttleyxd:
> Just checking something - on Steam Deck, are you not able to click on the top and bottom parts of scroll to scroll up and down?

This works, I would have never thought of that (this is my first attempt of playing Daggerfall in any form).

I don't know if we want to merge this (in this or different form), but from my perspective original behavior is not intuitive (and I assume I'm not doing any harm by adding additional method to scroll this)

jefetienne:
I haven't tested yet, but I think this would be good to make a general function out of in the InputManager and used in multiple scrollable UI objects.

This PR should utilize `InputManager.Instance.JoystickDeadzone` and `InputManager.Instance.GetAxisActionInversion(..)` to handle the player's personal joystick settings. The methods `FindInputAxisActions` or `UpdateControllerCursorPosition` are good examples to see how they get used in practice.

muttleyxd:
@jefetienne I'll look into it during next week (I'm new to this code so everything takes probably 10x longer than it should :D)

jefetienne:
> @jefetienne I'll look into it during next week (I'm new to this code so everything takes probably 10x longer than it should :D)

No worries! To clarify, I think it's fine that this PR targets at least this one window for now. I think the logic for calculating that `movement` variable should be moved over to a new method in the InputManager so you can call it like

```
float movement = InputManager.Instance.GetUIScrollMovement();
```

muttleyxd:
@jefetienne here's my second attempt at this

Assets/Scripts/Game/InputManager.cs
```diff
@@ -1199,6 +1199,28 @@ public KeyCode ParseKeyCodeString(String s)
...
+
+            if (GetAxisActionInversion(AxisActions.CameraVertical))
+                vertical *= -1;
+
+            float calculated_deadzone = Mathf.Sqrt(horizontal * horizontal + vertical * vertical);
+            if (calculated_deadzone <= JoystickDeadzone)
+                return 0.0f;
+
+            // Use both analog axes for scrolling movement - for now it doesn't matter which axis we use for scrolling
+            return horizontal + vertical;
```
jefetienne, line 1221: Not sure if this was your preference, but with my controller (Xbox on Windows/Linux, aimed for as the standard), I get inverse movement where Up moves it down and Down moves it up, but I think it's more standard for joysticks that they move in the same direction. Can you do `return -(horizontal + vertical);`?

jefetienne (commented):
Just tested it now, looks mostly good! I added a few comments.

Assets/Scripts/Game/InputManager.cs
```diff
@@ -1199,6 +1199,28 @@ public KeyCode ParseKeyCodeString(String s)
...
+            float horizontal = Input.GetAxis(cameraAxisBindingCache[0]);
+            float vertical = Input.GetAxis(cameraAxisBindingCache[1]);
+
+            if (GetAxisActionInversion(AxisActions.CameraHorizontal))
+                horizontal *= -1;
+
+            if (GetAxisActionInversion(AxisActions.CameraVertical))
+                vertical *= -1;
+
+            float calculated_deadzone = Mathf.Sqrt(horizontal * horizontal + vertical * vertical);
```
jefetienne, line 1216: I'd rename this to camelCase format (`calculatedDeadzone`)

muttleyxd: Done

Assets/Scripts/Game/InputManager.cs
muttleyxd: You're right, I was so excited to get it working, that I didn't care which direction it scrolled.

I inverted only the vertical axis - I tested it now on Deck (I assume it should mimic Xbox controller, as it's configured to do so in the options) and now it scrolls in the following directions:
```
stick up - scrolls up
stick down - scrolls down

stick left - scrolls up
stick right - scroll down
```

Do we want to invert horizontal axis too? Current behavior looks most intuitive to me.

Interkarma:
Thanks for the refinement. Thank you @jefetienne for the help with review - are you happy with this one for me to merge?

jefetienne:
I am!

PR #2329, opened 2022-05-12
Fix accidentally closing multiple windows at once

jefetienne:
When multiple calls to `UserInterfaceWindow.CloseWindow()` are made in one frame, it will pop the current window from the UIManager each time and accidentally close other windows below it.

How to reproduce:
1. Bind "Cast Spell" to "Escape"
2. Open Character Sheet Window
3. Click the "Spellbook" button within the Character Sheet Window
4. Press Escape
5. Observe that both windows closed instead of just the spellbook window

It was called twice because the Escape key to exit is already observed in the base class's `DaggerfallPopupWindow.Update()`, and the binding for `Escape` is also observed as the `toggleClosedBinding` in the Spellbook window.

This PR fixes the issue by peeking for an existing message to close the window, which should not appear if it was only called once.

While it would be unusual to bind the Escape key to something else, this fix is important for adding the escape key as a joystick keybind, where it could be bound as the "B" button which may be shared as a keybind for opening up a different window.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallAutomapWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallAutomapWindow.cs
@@ -200,6 +200,8 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
         public DaggerfallAutomapWindow(IUserInterfaceManager uiManager)
             : base(uiManager)
         {
+            // Prevent duplicate close calls with base class's exitKey (Escape)
+            AllowCancel = false;
         }
 
         /// <summary>
@@ -698,11 +700,11 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
 
             HotkeySequence.KeyModifiers keyModifiers = HotkeySequence.GetKeyboardKeyModifiers();
 
-            if (Input.GetKeyDown(KeyCode.Escape) ||
+            if (Input.GetKeyDown(exitKey) ||
                 // Toggle window closed with same hotkey used to open it
                 InputManager.Instance.GetKeyDown(automapBinding))
                 isCloseWindowDeferred = true;
-            else if ((Input.GetKeyUp(KeyCode.Escape) ||
+            else if ((Input.GetKeyUp(exitKey) ||
                 // Toggle window closed with same hotkey used to open it
                 InputManager.Instance.GetKeyUp(automapBinding)) && isCloseWindowDeferred)
             {
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallCharacterSheetWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallCharacterSheetWindow.cs
@@ -85,6 +85,8 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
         public DaggerfallCharacterSheetWindow(IUserInterfaceManager uiManager)
             : base(uiManager)
         {
+            // Prevent duplicate close calls with base class's exitKey (Escape)
+            AllowCancel = false;
         }
 
         #endregion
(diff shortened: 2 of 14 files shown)


petchema:
The method spuriously also observing the closing KeyUp event may have taken other actions than just calling CloseWindow(), so doing nothing in CloseWindow is not guaranteed to leave the interface in a consistent state. I'd prefer a mechanism that either "consumes" the KeyUp after its handling, or at least that hides it after its handling, to be sure that no 2nd handling is even attempted

jefetienne:
> I'd prefer a mechanism that either "consumes" the KeyUp after its handling, or at least that hides it after its handling, to be sure that no 2nd handling is even attempted

You raise a good point on not leaving the interface in a consistent state, but I think we would narrow a feature like this just for the window closing. I could see in certain instances it being desirable to be still be triggered in parallel with other KeyUp/Down functions, especially if a mod is trying to listen in on input.

I might look into simply adding a state like "isClosed" to the UserInterfaceWindow and harden all of the inherited windows' code to check if the closed signal has happened.

jefetienne:
It's not perfect, but I added a committed that reverted the change and instead checks for every window Update() whether or not the window was set to be canceled, since this issue is resulting from DaggerfallPopupWindow's Update method calling CancelWindow() when the "exitKey" is key'd up.

Interkarma:
@jefetienne Proposing an alternate approach for your feedback.

The core issue here that `DaggerfallPopupWindow` has a hardcoded cancel button (Escape) _and_ this behaviour is enabled by default. Things come unstuck higher up in UI windows that have their own ability to exit via toggle-close and might prefer to manage this behaviour themselves.

So rather than check `if (!Cancelled)` in each window, we first disable cancel behaviour by adding the below to constructor.

`AllowCancel = false;`

Then handle either the toggle-close or Escape at same time in the UI window.

    // Toggle window closed with same hotkey used to open it or using exit key
    if (InputManager.Instance.GetKeyUp(toggleClosedBinding) || Input.GetKeyUp(exitKey))
        CloseWindow();

Now the higher-level UI window is in control of its own lifetime and can exit using logic specific to that window. The above could also be offloaded to a common helper that accepts toggleClosedBinding as a param - checks toggle, Escape, and even a controller exit button at the same time. See #2367 for some work along those lines.

I tested the above with your repro steps and `DaggerfallSpellBookWindow` and it also stopped the double handling without needing to eat the `Cancelled` flag (because it never gets raised).

I prefer this approach as it's explicitly stopping the bad behaviour rather than just eating it higher up. Then the window can handle itself, and that handling can be extended.

Let me know your thoughts. :)

jefetienne:
> @jefetienne Proposing an alternate approach for your feedback.
> 
> The core issue here that `DaggerfallPopupWindow` has a hardcoded cancel button (Escape) _and_ this behaviour is enabled by default. Things come unstuck higher up in UI windows that have their own ability to exit via toggle-close and might prefer to manage this behaviour themselves.
> 
> So rather than check `if (!Cancelled)` in each window, we first disable cancel behaviour by adding the below to constructor.
> 
> `AllowCancel = false;`
> 
> Then handle either the toggle-close or Escape at same time in the UI window.
> 
> ```
> // Toggle window closed with same hotkey used to open it or using exit key
> if (InputManager.Instance.GetKeyUp(toggleClosedBinding) || Input.GetKeyUp(exitKey))
>     CloseWindow();
> ```
> 
> Now the higher-level UI window is in control of its own lifetime and can exit using logic specific to that window. The above could also be offloaded to a common helper that accepts toggleClosedBinding as a param - checks toggle, Escape, and even a controller exit button at the same time. See #2367 for some work along those lines.
> 
> I tested the above with your repro steps and `DaggerfallSpellBookWindow` and it also stopped the double handling without needing to eat the `Cancelled` flag (because it never gets raised).
> 
> I prefer this approach as it's explicitly stopping the bad behaviour rather than just eating it higher up. Then the window can handle itself, and that handling can be extended.
> 
> Let me know your thoughts. :)

This sounds like a better idea! I may be busy the next few days but I'll definitely update the code to reflect what you said.

jefetienne:
Sorry for the late update - I was on vacation and did not have enough signal to get the right Unity version installed.

I have pushed the fix for this as well as updated instances where "KeyCode.Escape" is used with "exitKey" instead -- I know @argonius2 your PR at #2367 deletes this variable, but I think it will be good since you'll need to fix merge conflicts anyway and the compiler errors for deleting it can help you find all the places where you'll need to make changes.

Interkarma:
Thanks so much for taking the time on the multiple iterations here @jefetienne. I really like the final form this is taking now.

Looks good on a read. I'll give it a proper review soon and merge. Cheers!

PR #2330, opened 2022-05-12
Quest fixes May 2022

JayH2971:
I tested The Possessed Child fix in live game, to be certain of it.


Interkarma:
Thank you for fixes Jay! :)

PR #2333, opened 2022-05-13
Fixed multiple calls to Billboard.SetMaterial

KABoissonneault:
It was observed while using the World Data Editor that multiple calls to `SetMaterial` on NPC Billboards causes many Box Colliders to be added.

image

This probably doesn't occur in gameplay, but I don't think my fix hurts gameplay either, so might as well be more robust here.

### DIFF
--- a/Assets/Scripts/Internal/DaggerfallBillboard.cs
+++ b/Assets/Scripts/Internal/DaggerfallBillboard.cs
@@ -338,7 +338,9 @@ namespace DaggerfallWorkshop
             // Add NPC trigger collider
             if (summary.FlatType == FlatTypes.NPC)
             {
-                Collider col = gameObject.AddComponent<BoxCollider>();
+                Collider col = gameObject.GetComponent<BoxCollider>();
+                if(col == null)
+                    col = gameObject.AddComponent<BoxCollider>();
                 col.isTrigger = true;
             }
 


jonoser:
When I was working on this last night, I used the following:
```
BoxCollider col = gameObject.GetComponent<BoxCollider>();
if(col)
{
    col.size = summary.Size;
}
else
{
    col = gameObject.AddComponent<BoxCollider>();
}
```
that way it updates the collider to have the size of the new sprite.

KABoissonneault:
It sounds right, but at the same time, I didn't want to change the observable behavior of the NPC triggers in this changelist. If you have cases where the box collider doesn't match the sprite, you can make another PR

jonoser:
That's okay, I just figured I'd mention it. I don't want to overstep or anything so I'll just leave it at that.

Interkarma:
While SetMaterial() was originally intended to be called only once at time of creation, I can see needs have evolved and agree a fix is needed. Thank you for proposed fix @KABoissonneault.

@jonoser is also right about collider too. Any time billboard texture is changed, it should be updated for new texture dimension.

Let me think on it for a bit. I think it might be better to just split out setting material and collider into separate calls, rather than keep them together in SetMaterial like this.

Interkarma:
After working through this, I think Kab's solution is the simplest. During normal play, the collider should only be added once. It's really just the WorldDataEditor where things end up in the situation Kab is addressing above.

In regards to new collider dimensions, this can be done with a new button in `DaggerfallBillboardEditor` to strip the collider and  add it again so it picks up the new billboard quad dims.

I'll merge this one as is and we can refine.

PR #2336, opened 2022-05-14
Do not attempt to list drives on unix

### LINKED ISSUE #2309: 0.13.5: Daggerfall directory impossible to select on Linux
**Describe the bug**
The root directory is not displayed in the drive list on Linux when selecting the Daggerfall game files during setup.

**To Reproduce**
1. Launch Daggerfall Unity

**Expected behavior**
The root directory `/` should always be an option on Unix-like operating systems (macOS, Linux). Due to the unified directory tree this will be sufficient to locate any path, unlike Windows which has a concept of drives.

It seems that Unity only detects my boot partition and some virtual drives on my system. As amusing as it would be to install Daggerfall on my boot partition, that seems impractical.

**Screenshots**
Screenshot_2022-04-18_23-21-15

 - OS: Manjaro Linux
 - Version: 2022-04-18

**Additional context**

ahouts:
resolves #2309

### DIFF
--- a/Assets/Scripts/Game/UserInterface/FolderBrowser.cs
+++ b/Assets/Scripts/Game/UserInterface/FolderBrowser.cs
@@ -4,8 +4,8 @@
 // License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
 // Source Code:     https://github.com/Interkarma/daggerfall-unity
 // Original Author: Gavin Clayton (interkarma@dfworkshop.net)
-// Contributors:    
-// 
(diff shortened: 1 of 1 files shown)


KABoissonneault (approved):

d10sfan:
This does seem to fix the issue I'm seeing when attempting to run on Linux.

Interkarma:
Thank you @ahouts for submitting a PR, and thank you @KABoissonneault for review. :)

PR #2340, opened 2022-05-17
Fix to allow animals loaded in RDBs to make sounds

jeniferirwin:


Assets/Scripts/Utility/RMBLayout.cs
```diff
@@ -1001,7 +1001,7 @@ private static bool IsBulletinBoard(uint modelID)
             return modelID == BulletinBoardModelID;
         }
 
-        private static void AddAnimalAudioSource(GameObject go, int record)
+        public static void AddAnimalAudioSource(GameObject go, int record)
```
df-ferital, line 1004: This method should be moved outside RMBLayout since it is now also used for RDB layout.

KABoissonneault: @df-ferital Do you have any suggestions? I tried looking a bit, and couldn't find a good place. Maybe `GameObjectHelper`?

df-ferital: Yes, I suppose `GameObjectHelper` would be the best place.

jeniferirwin: I am new to the DFU scene. This change you're discussing - is it a change that I should be making, or is it something that more experienced DFU programmers will make after the pull request is merged? If I'm the one that should be making this change, should I cancel this pull request and re-do it? Thanks in advance for your time!

KABoissonneault: You should do it, but you don't have to cancel. You can make a new commit, push the commit to the same branch (jeniferirwin:master), and the PR will update accordingly

jeniferirwin: I've made the change and added the commit. Thanks for the explanation!

KABoissonneault (approved):

df-ferital (approved):

Interkarma:
Thank you for the refinement @jeniferirwin. And thank you @KABoissonneault and @df-ferital for help with review. :)

PR #2343, opened 2022-05-21
Start difficulty dagger trail in correct place

numidium:
When the custom class creator window appears, the difficulty dagger trail starts at the top left of the screen and fades out. This moves it to the correct position at the start.


KABoissonneault (approved):

Interkarma:
Thank you @numidium! :)

PR #2346, opened 2022-05-29
Allow overriding standard guild services

ajrb:
his allows standard service handling to be replaced by mods.

Examples:

```
// Override cure service without changing label or shortcut key
Services.RegisterGuildService(GuildServices.CureDisease, CustomCureService);

// Override cure service, changing label
Services.RegisterGuildService((int)GuildServices.CureDisease, CustomCureService, "Custom Cure");
```

### DIFF
--- a/Assets/Scripts/Game/Guilds/Services.cs
+++ b/Assets/Scripts/Game/Guilds/Services.cs
@@ -342,6 +342,11 @@ namespace DaggerfallWorkshop.Game.Guilds
 
         public static string GetServiceLabelText(GuildServices service)
         {
+            // Check custom names
+            string serviceName;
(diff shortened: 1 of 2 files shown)


ajrb:
Only downside I can see is the first 19 faction npcs can't have custom service handlers without conflicts but they are the Daedra so I don't see it as an issue. (with the exception of "The Master of Initiates" in Theves Guild but Mods can always create a new faction entry to deal with it if desired)

KABoissonneault (approved):

magicono43:
Hey just wanted to make a comment about some confusion I'm having with this feature, I'll just attach a picture of what I wrote, thanks.

Capture

PR #2348, opened 2022-06-05
Per-enemy name for custom enemies

KABoissonneault:
Added ability for mods to provider per enemy name through a new TextProvider function. Previously, custom enemies could only be named after their career name.

This removes a TODO I left last year :)

Usage example: 
https://github.com/SquidKamer/DaggerfallBestiaryProject/blob/feat/enemy-name/Scripts/BestiaryTextProvider.cs#L16

For custom enemies, it makes sense to me that mods should be able to reuse the same DFCareer for multiple enemy ids. For example, a mod might have a Goblin Soldier, a Goblin Archer, and a Goblin Skirmisher, all using the same Goblin career. Before this change, however, info-clicking the goblin would always say "You see a Goblin" - you'd need a unique career for each to get their unique name.

This change allows the career reuse.

### DIFF
--- a/Assets/Scripts/Game/TextManager.cs
+++ b/Assets/Scripts/Game/TextManager.cs
@@ -360,7 +360,13 @@ namespace DaggerfallWorkshop.Game
             // Handle custom enemies
             else
             {
-                // TODO: maybe go through TextProvider so mods can try to offer localization?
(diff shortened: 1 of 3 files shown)


Interkarma:
Thank you Kab.

Issue #2349, opened 2022-06-08
[0.13.5] Arrows Having Materials In Loot Is Still A Problem


magicono43:
So yeah, did some testing and found currently arrows spawned in loot are still being given materials that they should not be. The effect is more noticeable when your character's level is fairly high such as 15+.

Here are some screenshots that will hopefully explain better

Daedric Arrows found on a knight, level 18 character.
![2022_06_08_13_59_16](https://user-images.githubusercontent.com/39843717/172687711-47510986-b2ea-477e-818c-4169e3117309.jpg)

Iron? Arrows found on a knight, level 9 character.
![2022_06_08_13_32_06](https://user-images.githubusercontent.com/39843717/172687859-d317a89f-448a-47c3-b14e-19dc273d80a4.jpg)

Silver or Elven? Arrows found on a knight, level 9 character.
![2022_06_08_13_38_19](https://user-images.githubusercontent.com/39843717/172687965-552da4c6-f62a-42bb-8a28-143dc2d3abd9.jpg)

Found these from using "killall" on the guards in Castle Daggerfall, also when using "cm 145" to spawn knights and killing them in the same way. For some reason when trying to test this with archers and rangers they were basically never dropping arrows for me, I'm not sure why. Even on the knights the arrow drop chances were very low.

Potential Solution: Not sure on this one, have not read through the loot code too recently, so not sure why this might still be happening, but apparently it is. Also just a note this is a save that did have mods, but in this testing no mods were enabled.

magicono43:
Here is some more info after Hazelnut asked for what it says in the save-data.

Arrows from my original stack that apparently were "normal"
![Capture1](https://user-images.githubusercontent.com/39843717/172691034-3de7e52e-98ca-4492-8f00-8f52e41addc5.PNG)

Ebony Arrows
![2022_06_08_14_29_10](https://user-images.githubusercontent.com/39843717/172691139-fd8b0198-e9db-4f4f-8a75-54d17198f2f1.jpg)

What the save-data entry looks like for those new ebony arrows.
![Capture](https://user-images.githubusercontent.com/39843717/172691202-81751aab-5ad8-4c29-bf2f-1542c0d333c1.PNG)

ajrb:
@Interkarma has already fixed part of the issue and I have submitted a PR for the remaining edge case I believe.

PR #2350, opened 2022-06-08
Fix edge case for arrow generation which would still set the material properties

ajrb:
Fixes remainder of issue #2349


Interkarma:
Good catch, thanks for the extra set of eyes on this one.

PR #2357, opened 2022-06-24
Fixed "PayMoney" quest action infinite money drain

KABoissonneault:
Issue reported here:
https://old.reddit.com/r/Daggerfall/comments/vjsdkn/disappearing_gold_glitch_unity_a_cohort_of_the/

Quest in question:
https://github.com/JayH2971/dfunity-questpacks/blob/master/QP1/QuestPacks/QP1/Temples/JHTP111.txt

The custom DFU quest action "PayMoney" was missing a call to `SetComplete()`. Therefore, the task would infinitely try to pay the specified amount, rather than only once as expected.

### DIFF
--- a/Assets/Scripts/Game/Questing/Actions/PayMoney.cs
+++ b/Assets/Scripts/Game/Questing/Actions/PayMoney.cs
@@ -89,6 +89,8 @@ namespace DaggerfallWorkshop.Game.Questing.Actions
                     }
                 }
             }
+
+            SetComplete();
         }
 
         #region Serialization


Interkarma:
Very nice catch, thank you!

ajrb:
It's supposed to work that way I think. This will probably break my Mountain Rumors quest which was designed to allow paying the lord off and ensures this action only ever triggers once. I would have appreciated being consulted before this 'fix' went in.

Not sure what to do, except add a different version of this in my mod I guess.

Interkarma:
Hey Andy! Sorry mate, I try but can't win them all.

Have a think about it and let me know what solution/alternative you prefer. I'll be happy to work with you on next round.

ajrb:
@Interkarma No worries mate, I need to test how this behaves now before deciding what to do - I looked in history and I removed the SetComplete() back when you removed the per task click rearming, which was restored a few weeks later. I never went back and looked at whether I should restore the SetComplete() and also, except for answering a query, never really made it clear that this action worked differently to others.. not even a comment in the code. Lazyness in the past on my part there. :-) It's definitely caught Jay out so I should have been more communicative when I changed it.

I will look into it and see where we are, and it's no big deal if I have to have my own custom action. I missed the whole discussion until I saw this closed PR this morning just before I had to leave to meet my brother so it was a surprise. Will update you once I know what is what.

ajrb:
Looks like it all works just fine with the change KAB made. I think I probably should have added SetComplete() back when you restored the per click rearming code, so as you said a good catch.

Interkarma:
Thanks for confirming, one less thing to worry about. :)

PR #2359, opened 2022-06-25
Make MobilePersonMotor.MobilePersonAsset public

AncientGrief:
Maybe I am wrong, but as of now I have to do this in my custom implementation to fallback to BillboardAssets without breaking some behaviour like NPCs stopping and looking at you when close up:

```cs
public class MyPersonAsset : MobilePersonAsset
{
	//...
	
    public override void SetPerson(Races race, Genders gender, int personVariant, bool isGuard, int personFaceVariant, int personFaceRecordId)
    {
        // ...
		
        //fallback to billboard
        if (!useThisPersonAsset)
        {
            Transform parent = this.gameObject.transform.parent; //MobilePersonNPC 
			
            //Get the MobilePersonBillboard
            GameObject goBillboard = parent.Find("MobilePersonBillboard").gameObject;
            MobilePersonBillboard billboard = goBillboard.GetComponent<MobilePersonBillboard>();
            
            goBillboard.SetActive(true);
            this.gameObject.SetActive(false); 
            
            //Assign the billboard to the motor
            parent.gameObject.GetComponent<MobilePersonMotor>().MobileAsset = billboard;
            //setup the Billboard with correct Person data
            billboard.SetPerson(race, gender, personVariant, isGuard, personFaceVariant, personFaceRecordId);
            
            //Get rid of this Asset, since it's not needed
            Destroy(this.gameObject);
            
            return;
        }
	
	}
}
```

### DIFF
--- a/Assets/Scripts/Game/MobilePersonMotor.cs
+++ b/Assets/Scripts/Game/MobilePersonMotor.cs
@@ -54,7 +54,7 @@ namespace DaggerfallWorkshop.Game
 
         // References
         DaggerfallEntityBehaviour entityBehaviour;
-        MobilePersonAsset mobileAsset;
+        public MobilePersonAsset MobileAsset { get; set; }
 
         #endregion
 
@@ -162,7 +162,7 @@ namespace DaggerfallWorkshop.Game
         {
             // Cache references
             entityBehaviour = GetComponent<DaggerfallEntityBehaviour>();
-            mobileAsset = FindMobilePersonAsset();
+            MobileAsset = FindMobilePersonAsset();
 
             // Need to repath if floating origin ticks while in range
             FloatingOrigin.OnPositionUpdate += FloatingOrigin_OnPositionUpdate;
@@ -183,7 +183,7 @@ namespace DaggerfallWorkshop.Game
             // Do nothing if paralyzed
             if (entityBehaviour.Entity.IsParalyzed)
             {
-                mobileAsset.IsIdle = false;
+                MobileAsset.IsIdle = false;
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Game/MobilePersonMotor.cs
```diff
@@ -54,7 +54,7 @@ public class MobilePersonMotor : MonoBehaviour
 
         // References
         DaggerfallEntityBehaviour entityBehaviour;
-        MobilePersonAsset mobileAsset;
```
KABoissonneault, line 57: You could have made this a single line change by changing this to:
```
MobilePersonAsset mobileAsset;
public MobilePersonAsset MobileAsset { get { return mobileAsset; } set { mobileAsset = value; } }
```
The rest of the code can still use the private version

AncientGrief: True but with this approach you have two members doing the same thing. I can change it to your approach if the overall request is okay to you?

KABoissonneault:
I'm not sure I understand what you're trying to achieve.

So, DFU spawns the DaggerfallMobilePerson prefab through the PopulationManager. This prefab has a `MobilePersonMotor` component, and contains a default child GameObject called "MobilePersonBillboard". On `Awake`, `MobilePersonMotor` tries to see if any mod provides a "MobilePersonAsset" prefab, and if so, disables its own MobilePersonBillboard and uses the mod MobilePersonAsset instead.

So, based on some condition, you want to change back some of the randomly spawned mobiles to the DFU default? 
My first question would be, why can't your "MyPersonAsset" conditionally reproduce the behavior of "MobilePersonBillboard"? Why does the motor *have* to revert back to its default person asset?

AncientGrief:
Correct, I am trying to implement an optional 3D Replacement for 2D Billboards and I don't want to copy the Billboard code, this would duplicate a lot of behaviour. This way I can switch back to the 2D Billboard if no 3D model is present and my implementation can focus around 3D stuff instead of 2D stuff.

Assets/Scripts/Game/MobilePersonMotor.cs
KABoissonneault: They're not doing the same thing: one is private access, one is public. It's idiomatic use of C# properties

AncientGrief: IF `mobileAsset` is made public because something else needs access to it from the outside and there are no other reasons to keep it private, it makes no sense to have it this way. Unless private mobileAsset is doing "behind the scene" stuff imho.

With you approach I can use `mobileAsset` and `MobileAsset` to achieve the same things in the class.
Another approach would be to write `public MobileAsset {private get; set}`, this way you don't have two variables doing the same thing and allowing access from the outside. Maybe this is a compromise? :D

PR #2361, opened 2022-06-26
WorldDataEditor bugfix and QoL features (editor only)

jonoser:
Fix for a bug where some billboards would be repositioned after saving due to their Summary.Size not being updated when cycling through the records.

Added additional quality of life features to the editor both of which function on all selected objects in the scene:
 - Added buttons that move an object (3d model, flat) in a direction based on that object's size
image
 - Added a button that exposes the helper function AlignBillboardToGround() to the editor.
image

Though this isn't potentially the cleanest way to do it, I allow AlignBillboardToGround to be used for 3d objects within the editor since it contains no code specific to billboards.

### DIFF
--- a/Assets/Game/Addons/WorldDataEditor/Editor/WorldDataEditor.cs
+++ b/Assets/Game/Addons/WorldDataEditor/Editor/WorldDataEditor.cs
@@ -260,6 +260,7 @@ namespace DaggerfallWorkshop.Game.Utility.WorldDataEditor
                     if (data.type == (int)WorldDataEditorObject.DataType.Object3D)
                     {
                         GUIElementPosition(ref data, ref elementIndex);
+                        GUIElementShift(ref data, ref elementIndex);//Shift by size
                         GUIElementRotation(ref data, ref elementIndex);
                         if (dataMode == DataMode.Building)
                             GUIElementScale(ref data, ref elementIndex);
@@ -270,6 +271,7 @@ namespace DaggerfallWorkshop.Game.Utility.WorldDataEditor
                             GUIElementModelActionRecord(ref data, ref elementIndex);
                             GUIElementNextObject(ref data, ref elementIndex);
                             GUIElementPrevObject(ref data, ref elementIndex);
+                            GUIElementGroundButton(ref elementIndex);//move to ground
                         }
                         else if (dataMode == DataMode.Building)
                         {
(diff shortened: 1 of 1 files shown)


ajrb:
Thanks, I'll review this when I can.

Assets/Game/Addons/WorldDataEditor/Editor/WorldDataEditor.cs
```diff
@@ -670,24 +790,17 @@ private void GUIElementFlatSwitchType(ref WorldDataEditorObject data, ref int el
...
+                    Billboard billboard = Selection.activeGameObject.GetComponent<Billboard>();
+
+                    data.gameObject.name = DaggerfallWorkshop.Utility.GameObjectHelper.GetGoFlatName(archive, record);
+                    
+                    //Dungeon mode needs us to update the rdbObj and building mode ignores it.
+                    data.rdbObj.Resources.FlatResource.TextureArchive = archive;
+                    data.rdbObj.Resources.FlatResource.TextureRecord = record;
+                    
+                    DestroyImmediate(Selection.activeGameObject.GetComponent<Collider>());
+                    billboard.SetMaterial(archive, record);
```
ajrb, line 803: Not sure about this change to the flat switching for buildings. That is code from UVs original building editor that I basically left alone. This change seems to mean that the id in the attached WorldDataEditorObject won't be getting updated with the new archive and texture which are subsequently used to save to the JSON.

This is only by reviewing code visually, I have not had a chance to try it out yet. Could you comment?

jonoser: I tested to make sure that the output from saving was not affected by this change for both RDBs and RMBs.
The original "building mode" code saves the new archive ID data with the line "data.id = tempArray[pos];" just above here.
Lines 799 and 800 are from the "dungeon mode" code that you added should update the necessary data saved to RDBs

Lines 673 - 682 in the original code only updates the image displayed on the billboard by creating a new temporary billboard and copying its renderer.  

However, doing it this way didn't update the billboard's Summary.Size.  This is the part that caused objects to reposition when saving.  If an archive started with a large image, and you switched it to a smaller one, it would retain the size data of the first.  When you load, the billboards would be recreated from the json and load with the correct size data which would reposition them.

I decided to replace that with a call to SetMaterial since that will correctly update the image and size data.  Though, since it's intended to be called once on a billboard, it duplicates the collider.  I get around that by just removing the collider before SetMaterial since the editor doesn't really care about it.

ajrb: Yes I tested also and it works. I think there is some redundant code after I made the dungeon editor which had to work differently from UncannyValley's building editor (that was the genesis of the WDE) - I should have tidied up better, but it was a lot more work than I had expected.

ajrb:
I've done some testing and it all looks good.

Great work, thank you!

PR #2362, opened 2022-06-26
Allow overriding temple and knightly orders

ajrb:
These guilds require a init parameter to set the variant


KABoissonneault (approved):

KABoissonneault:
Does this allow adding new knight orders, somehow? or just changing existing orders

ajrb:
Yes it should in theory allow that, although this fix is just an error in my instantiation code where I didn't take into account overriding Temples or Orders which need the init parameter. Since in C# enums are just ints the type wont restrict adding new Deity or Order elements in custom implementations.

Still only one mod can override each guild implementation at a time so that has to be borne in mind.

Interkarma:
Thank you @ajrb!

PR #2363, opened 2022-06-27
Custom enemy support in RDB

KABoissonneault:
It's something @ajrb and I discussed a while ago.

We have a conflicting situation when it comes to the `FactionOrMobileId` field of the `FlatResource` of a Monster marker (archive 199, record 16). 
- classic DF data needs to be masked with `& 0xFF` to get the correct mobile Id
- custom enemies need to support ids bigger than 255

My solution here is to add a bit flag to annotate a marker as "custom" (ie: not from DF data).

Flags usage is a bit hard to find in DFU code, but this is the assumptions I've relied on:
- In DF data, "Monster" marker uses Flags=0 for unspecified gender, Flags=1 for Female, Flags=2 for Male (source: `RDBLayout.AddEnemy`, with `useGenderFlag==true` only for FixedRDB Enemy)
- "Random Monster" marker uses Flags as a seed between 0 and 255 for indexing into the monster spawn table (source: `RDBLayout.AddRandomRDBEnemyClassic`)
- Static NPC flats use `.Flags & 32 == 32` to identify a Female character

To make sure "Random Monster" marker doesn't get affected, I've limited the use of this flag to only the fixed "Monster" marker. I've kept 1 as the Female flag, 2 as the Male flag. If neither are set, the gender is left Unspecified as before.

With this, I can make use of custom monster ids in the WorldData Editor. For example, I've added our mod's id 261 "Medusa" at the start of Privateer's Hold.

image

and once I load in-game, I get one-shot by Kamer's nightmare fuel (might be a bit too dark lol)

image

### DIFF
--- a/Assets/Game/Addons/WorldDataEditor/Editor/WorldDataEditor.cs
+++ b/Assets/Game/Addons/WorldDataEditor/Editor/WorldDataEditor.cs
@@ -934,13 +934,7 @@ namespace DaggerfallWorkshop.Game.Utility.WorldDataEditor
                 obj.Type = DFBlock.RdbResourceTypes.Flat;
                 obj.Resources.FlatResource.TextureArchive = int.Parse(selectedObjectID.Split('.')[0]);
                 obj.Resources.FlatResource.TextureRecord = int.Parse(selectedObjectID.Split('.')[1]);
-
-                // Only add "Custom Marker" flag to this marker for now
-                // Some markers (ex: Random Monster) may use Flags differently
-                if (selectedObjectID == "199.16") // Marker: Monster
-                {
-                    obj.Resources.FlatResource.Flags |= RDBLayout.CustomMarkerFlag;
-                }
+                obj.Resources.FlatResource.IsCustomData = true;
 
                 go = WorldDataEditorDungeonHelper.AddFlatObject(obj);
                 if (go != null)
--- a/Assets/Game/Addons/WorldDataEditor/Editor/WorldDataEditor.cs
+++ b/Assets/Game/Addons/WorldDataEditor/Editor/WorldDataEditor.cs
@@ -13,6 +13,7 @@ using UnityEngine;
 using DaggerfallConnect;
 using DaggerfallConnect.Arena2;
 using DaggerfallWorkshop.Utility.AssetInjection;
+using DaggerfallWorkshop.Utility;
 
 namespace DaggerfallWorkshop.Game.Utility.WorldDataEditor
 {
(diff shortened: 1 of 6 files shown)


Assets/Game/Addons/WorldDataEditor/Editor/WorldDataEditor.cs
```diff
@@ -933,6 +934,14 @@ private GameObject AddNewDungeonObject()
                 obj.Type = DFBlock.RdbResourceTypes.Flat;
                 obj.Resources.FlatResource.TextureArchive = int.Parse(selectedObjectID.Split('.')[0]);
                 obj.Resources.FlatResource.TextureRecord = int.Parse(selectedObjectID.Split('.')[1]);
+
+                // Only add "Custom Marker" flag to this marker for now
+                // Some markers (ex: Random Monster) may use Flags differently
+                if(selectedObjectID == "199.16") // Marker: Monster
```
ajrb, line 940: Missing a space before (

KABoissonneault:
> Why does that first screenshot have flags == 0? shouldn't it be 128 or 129 if you want to set female too?
> 
> Other than that confusion of mine, this looks good to me.

`WorldDataEditor.GUIElementFlatActionRecord` (which renders the text you're referring to) uses `rdbObj.Resources.ModelResource.ActionResource` for flags, not `rdbObj.Resources.FlatResource`. The latter is where the "128" is. This seemed wrong to me, but I didn't know how to fix it (there's no ActionResource for Flats), so I didn't touch it. Maybe you'd know what the intent was

ajrb:
It is confusing, but I think you've found a bug in the editor code. Working through it and I will submit a PR. I assume you tested by setting flags field in world data JSON directly?

KABoissonneault:
> I assume you tested by setting flags field in world data JSON directly?

No, because the flag is set by WorldData Editor itself when creating any new Monster marker. I think having to set the flag manually every time would be very error prone.
Ultimately, the only data that wouldn't have the flag is the one coming from arena2, which is the data that needs masking

ajrb:
Okay this looks good to me now I've gone through it. The only comment I have is the terminology used "CustomMarker" is quite general and not specific to fixed monster markers. I wonder if using more specific terminology such as ExtendedMobMarker or something would be clearer, especially if other flag values get added in future. 

e.g. CustomMarkerFlag -> ExtendedMobMarkerFlag

Thoughts?

KABoissonneault:
I've considered it. I came up with the name when I thought we could have the flag on all resources placed by DFU editors, but I realized that I couldn't do this everywhere (ex: it affects the Random Monster marker seed). If we don't see a future use for such a flag, I can make it more specific then

Interkarma:
Would it simplify things if we added custom fields in `RdbFlatResource` to store your data? Then you wouldn't need to pack the bits into some other field, and the layout code can just check if this should be custom and run with the data provided.

These enums do map to binary data being read from classic, but it's still possible to add new fields for DFU use only.

Would that help or just complicate things for WDE instead?

KABoissonneault:
I don't mind taking that approach. Hazel had made me think that we couldn't change the binary layout of these structures, but if they can be extended, then I could do that. But then again, is it worth reimplementing and retesting it...

Interkarma:
They can be extended (though I still prefer a light touch and of course don't change the fields mapping to native data).

A good example here is the `TextureBitfield` from binary data, which is then split into `TextureArchive` and `TextureRecord` for DFU's convenience and so we don't have to manipulate the bitfield every time. `Position` is also something recorded for DFU's own use.

So if that helps, go for it. :)

ajrb (approved):

KABoissonneault:
In addition to changing the flag to a boolean field, I have also fixed a "crash" that occurs if a RDB contains a marker with an invalid FactionOrMobileID. This can happen if, say, a user removes a mod that adds monsters, but forgets to remove WorldData loose files. The game softlocks in the loading process. With this change, the enemy simply does not appear.

Interkarma:
Looks good, thanks for the rework and fix. I'll get this merged ahead of next builds.

Thanks also for your help with review @ajrb. As this crosses over into WDE, I appreciate the extra set of eyes.

PR #2365, opened 2022-06-27
Update Jetbrains Rider Editor package to latest version.

XJDHDR:
The Jetbrains Rider Editor package that DFU currently uses is locked to v1.2.1. This package is not compatible with the current version of Rider, which causes a number of issues. Some of the ones I've noticed are:

- The External Script Editor option is labelled as "Rider.exe" rather than the program's name and version.
- Project files can't be regenerated from the Unity editor.
- Rider complains about missing Unity classes.
- Autocomplete doesn't suggest anything from the Unity API.

I haven't noticed any issues caused by telling the Editor to use the latest version of this package instead of the one recommended for this version of Unity.

### DIFF
--- a/Packages/manifest.json
+++ b/Packages/manifest.json
@@ -1,10 +1,10 @@
 {
   "dependencies": {
     "com.unity.addressables": "1.19.11",
-	"com.unity.burst": "1.7.1",
+    "com.unity.burst": "1.7.1",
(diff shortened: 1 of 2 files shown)


Interkarma:
Thank you. I don't believe this will cause any issues, no problems updating that package.

PR #2367, opened 2022-06-28
Allowing players to exit out of UI screens on a controller

argonius2:
Adding support so players can exit out of UI screens with a button on the controller, just like how the escape key is used to back out of UI screens with the keyboard. The key binding can be set in the joystick controller configuration screen.

image

This also allows users to cancel rest with the controller by pressing the same UI back button, since now in code where it was previously hardcoded to check for the keyboard escape key are now checking both for the escape key and the UI exit controller button.

Also noticed that the pause binding was named Escape in code, which could be confused with this new UI exit binding, so that has been renamed to Pause.

### DIFF
--- a/Assets/Scripts/Game/InputManager.cs
+++ b/Assets/Scripts/Game/InputManager.cs
@@ -1041,12 +1041,12 @@ namespace DaggerfallWorkshop.Game
             return Input.GetMouseButton(button) || (EnableController && GetKey(joystickUICache[button], false));
         }
 
-        public bool GetBackKeyDown()
+        public bool GetBackButtonDown()
         {
             return GetKeyDown(KeyCode.Escape) || (EnableController && GetKeyDown(joystickUICache[3], false));
         }
 
-        public bool GetBackKeyUp()
+        public bool GetBackButtonUp()
         {
             return GetKeyUp(KeyCode.Escape) || (EnableController && GetKeyUp(joystickUICache[3], false));
         }
--- a/Assets/Scripts/Game/InputManager.cs
+++ b/Assets/Scripts/Game/InputManager.cs
@@ -296,12 +296,13 @@ namespace DaggerfallWorkshop.Game
         {
             LeftClick,
             RightClick,
-            MiddleClick
+            MiddleClick,
+            UiExit
         }
 
         public enum Actions
         {
-            Escape,
+            Pause,
             ToggleConsole,
 
             MoveForwards,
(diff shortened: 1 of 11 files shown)


Interkarma:
Thank you for the PR. I appreciate the problem being solved and do see this could be one path. However, there's already steps being taken towards this in #2329 with @jefetienne, who implemented controller support and is currently maintaining this feature set. We first working on solving some unwanted window behaviours, then will start moving towards better support for cancelling out of UIs with controller.

I tagged jefetienne above so they're aware of this PR and can offer any feedback. We'll loop back to this PR once the other set of problems has been resolved. Just a heads up, this most likely require this PR to be closed or heavily reworked by the time we reach that point.

Cheers.

jefetienne:
I like the changes here, and some of it was better than what I had initially planned! What Interkarma said is true since some of these escape changes will probably be need to reworked.

One thing I would say is that changing the 'Actions.Escape' enum name to 'Pause' had failed auto-rebinding for me when the 'Escape' KeyCode was already bound to something else. This is probably very unusual for someone to have done that but it would technically break backwards compatibility for them and they would have to modify their Keybinds.txt file to get it working.

Assets/Scripts/Game/InputManager.cs
```diff
@@ -1081,6 +1083,16 @@ public bool AnyKeyUpIgnoreAxisBinds
             get => GetAnyKeyUpIgnoreAxisBinds() != KeyCode.None;
         }
 
+        public bool GetUiExitKeyDown()
```
jefetienne, line 1086: These should be moved underneath where the "GetMouseButton" functions are. These methods should also be use the word "Button" instead of "Key".

Assets/Scripts/Game/InputManager.cs
```diff
@@ -1081,6 +1083,16 @@ public bool AnyKeyUpIgnoreAxisBinds
...
+        {
+            return GetKeyDown(KeyCode.Escape) || GetKeyDown(GetJoystickUIBinding(JoystickUIActions.UiExit));
```
jefetienne, line 1088: You should extend the `joystickUICache` array to be size 4 and include the UiExit button along with it in all the places that variable is used.

Then you can use this line instead, where it uses this cache to exclude and apply the same logic to the rest:
`return Input.GetKeyDown(KeyCode.Escape) || (EnableController && GetKeyDown(joystickUICache[3], false));`

jefetienne (changes requested):
I also would personally opt for calling this new button the "Back" button instead of "UI Exit" so it's less technical-sounding.

Interkarma:
I'll temp close this one for now as it needs to be refactored a bit with #2329 merged. Feel free to reopen once changes made.

Having reviewed this one in first pass, I would prefer for existing keybind format not to be changed - i.e. please don't rename Escape to Pause. As @jefetienne uncovered this can break backwards compatibility.

Thank you for your patience on this one while the previous PR was resolved.

jefetienne:
@argonius2 Are you still interested in creating another PR for this?

PR #2370, opened 2022-06-30
Null reference exception when mod enemies cast

KABoissonneault:
In mods that spawn enemies with a custom flow (ex: World of Daggerfall), a custom serialization component might be used (ex: for extra, mod-specific data).

Previously, `EntityEffectManager.GetCasterLoadID` explicitly checked for the component `SerializableEnemy`, when it really only needs a component that implements the `ISerializableGameObject`, for the LoadID.

I've loosened the `GetComponent` to work with custom `ISerializableGameObject` types.

I've also added a null check, to avoid the null reference exception in the future.


Interkarma:
This makes sense. Thank you.

PR #2374, opened 2022-07-01
Dungeon Micro-Map QoL Option, 7/1/2022

magicono43:
Dungeon Micro-Map QoL Option, 7/1/2022.

Originally in relation to this forum thread: https://forums.dfworkshop.net/viewtopic.php?t=5242

This post is mostly related to the above forum thread that I made a good few months back, and finally decided to make a PR for it. I think the difference is fairly clear from the below attached screenshots. I set the option to be disabled by default. Also the purple background is not what occurs in practice, that is just a weird quirk when you try to take a screenshot with "F8" while in the map screen. 

Please do any edits you feel would make this better. I was considering for small possible performance reasons to put the settings check into another for-loop on it's own, but did not want to make the PR look to ugly by doing that, but up to whoever if they feel there is a better way to do this etc. Thanks as usual, I'm still quite the noob when it comes to using Git and pushing commits and such.

Default
Dungeon_Mini-map_Example

Option On Example 1

Option On Example 2

### DIFF
--- a/Assets/Resources/defaults.ini.txt
+++ b/Assets/Resources/defaults.ini.txt
@@ -174,6 +174,7 @@ AssetInjection=True
 CompressModdedTextures=True
 NearDeathWarning=True
 AdvancedClimbing=False
+MoreReadableDungeonMicroMap=False
 AlternateRandomEnemySelection=False
 DungeonAmbientLightScale=1.0
 NightAmbientLightScale=1.0
--- a/Assets/Scripts/Game/Automap.cs
+++ b/Assets/Scripts/Game/Automap.cs
@@ -1787,7 +1787,17 @@ namespace DaggerfallWorkshop.Game
                 {
                     for (int x = 0; x < microMapBlockSizeInPixels; x++)
                     {
-                        textureMicroMap.SetPixel(xBlockPos * microMapBlockSizeInPixels + x, yBlockPos * microMapBlockSizeInPixels + y, Color.yellow);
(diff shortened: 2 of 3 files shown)


Ralzar81:
Just a suggestion: Could not this and Smaller Dungeons be combined in a slider setting?

So it is Off, Edges Visible, Smaller Dungeons

Or something like that.

magicono43:
> Just a suggestion: Could not this and Smaller Dungeons be combined in a slider setting?
> 
> So it is Off, Edges Visible, Smaller Dungeons
> 
> Or something like that.

That would be up to the others honestly, I wanted to make this initially a "hidden" feature as I know that options menu creep has always been an issue, lol. Also I think Smaller Dungeons just recently got added to be visible in the launcher instead of being "hidden" in 0.14.0?

Interkarma:
BTW, not opposed to adding an option to launcher UI. This might be a candidate for Accessibility tab? I know things are getting crowded on the other tabs.

magicono43:
> BTW, not opposed to adding an option to launcher UI. This might be a candidate for Accessibility tab? I know things are getting crowded on the other tabs.

I'd be all up for it, just did not want to get ahead of myself and initially add the settings with the rest of the launcher stuff, lol. (Also was not 100% sure how to do the settings for launcher, so just decided to push the PR this way.) But yeah, I'm totally for doing that if others find it appropriate.

Interkarma:
I can help if needed. I just follow the examples of the other options anyway. Let me know if any help needed.

Feedback also welcome from the others about this.

PR #2378, opened 2022-07-04
Fix bright UI rendering on Mac

Interkarma:
Implemented DaggerfallUI.DrawTexture to handle separate UI rendering pathways for Mac and PC/Linux.
The PC/Linux pathway is still using GUI.DrawTexture as before. Mac pathway uses Graphics.DrawTexture and font shader customisations to yield correct colour space output.
This should be a low impact change for mods as only internal UI DrawTexture handling involved. Any mods using GUI.DrawTexture or GUI.DrawTextureWithTexCoords should use the equivalent method in DaggerfallUI to render textures in correct colour space on Mac.
Tested with and without DREAM in editor and builds across Mac/PC and working OK, so texture mods should not be impacted by this.
Tagging @ajrb and @KABoissonneault - please quickly test your mods build with the changes in this PR and it doesn't cause any problems you can see. I'd like to include this in 0.14.1 if no issues found.

### DIFF
--- a/Assets/Scripts/Game/DaggerfallUI.cs
+++ b/Assets/Scripts/Game/DaggerfallUI.cs
@@ -126,6 +126,8 @@ namespace DaggerfallWorkshop.Game
 
         Material pixelFontMaterial;
         Material sdfFontMaterial;
+        Material uiBlendMaterial;
+        Material uiBlitMaterial;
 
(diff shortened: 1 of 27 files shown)


Interkarma:
I'm going to merge this one so it's in play for general dev testing and next preview release. Feedback welcome if any troubles found after merge.

PR #2380, opened 2022-07-04
Enable enhanced item lists by default

ajrb:
When I originally implemented enhanced item lists (16 items instead of 4) DFU was more aiming at recreating the classic experience out of the box, so we made this switched off by default. I think things have evolved to the point where this is backwards. The majority of players want a modern Daggerfall experience, and any that are looking for the classic already need to change settings to get that. They will know the original game just had 4 visible items per list and can then switch it off, whereas new players have no idea they can get 16 unless they happen to see someone else's inventory window.

This is just a proposal, @Interkarma you can merge or close this as you prefer.


Interkarma:
Hey Hazel. I'm happy to merge this one.

To explain from my side, I see DFU as "Classic Daggerfall Plus". I wanted DFU to be stick fairly close to classic while still enabling some nice QoL features by default (e.g. scroll bars, tooltips, etc.). I agree this change falls into that range, and users can still disable if they prefer the default 1x4 layout.

We can always reconsider if there's a lot of pushback, but I don't think there will be. It's a nice QoL feature and I think most people appreciate it once they find it.

PR #2381, opened 2022-07-04
Allow building greeting message boxes to be disabled by mods

ajrb:
Simple switch for the building greetings so mods can turn them off, and optionally replace using enter exit events

### DIFF
--- a/Assets/Scripts/Game/PlayerActivate.cs
+++ b/Assets/Scripts/Game/PlayerActivate.cs
@@ -105,7 +105,7 @@ namespace DaggerfallWorkshop.Game
         }
 
         // Allows the building greeting message boxes to be disabled by mods
-        public static bool buildingGreetingEnabled = true;
+        public static bool buildingGreetingsEnabled = true;
 
         #region custom mod activation
         private struct CustomModActivation
@@ -561,7 +561,7 @@ namespace DaggerfallWorkshop.Game
 
                     // If entering a shop let player know the quality level
                     // If entering an open home, show greeting
-                    if (hitBuilding && buildingGreetingEnabled)
+                    if (hitBuilding && buildingGreetingsEnabled)
                     {
                         const int houseGreetingsTextId = 256;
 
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Game/PlayerActivate.cs
```diff
@@ -558,7 +561,7 @@ void Update()
 
                     // If entering a shop let player know the quality level
                     // If entering an open home, show greeting
-                    if (hitBuilding)
+                    if (hitBuilding && buildingGreetingsEnabled)
```
KABoissonneault, line 564: Doesn't this also prevent the transition to interior (line 594)? Why not only wrap line 573-584 with that check instead?

Also, instead of this approach, we could have a sort of "building greeting message box override" event, which lets mod override `mb` to either return null, or a new message box, after DFU has chosen whatever message box should go there. Between line 584 and line 586

Interkarma: The transition on line 594 is outside conditional block starting 561 and ending 591. That block is only used to display greeting messages, so should be fine to filter here based on `buildingGreetingsEnabled`.

Not opposed to making it more generally useful as Kab suggests. Up to Hazel if he wants to implement that. I'm happy to take this one now though, it's a low impact change.

KABoissonneault: Oh god, this branch runs all the way back from line 502, I got confused lol. Fair enough

Interkarma: You should have seen it before I cleaned it up a bit haha.

ajrb: > Doesn't this also prevent the transition to interior (line 594)? Why not only wrap line 573-584 with that check instead?
> 
> Also, instead of this approach, we could have a sort of "building greeting message box override" event, which lets mod override `mb` to either return null, or a new message box, after DFU has chosen whatever message box should go there. Between line 584 and line 586

I thought about it, but went for simple here. Feel free to replace it with something more refined if you want to. This took 10 mins so I don't care if it's immediately replaced. :-)

PR #2383, opened 2022-07-05
Revert "Dungeon Micro-Map QoL Option, 7/1/2022"

magicono43:
Reverts Interkarma/daggerfall-unity#2374

I'm not sure if this is the proper way to do this. But I figure revert this PR so I can replace it with a new one with improvements/revisions?


magicono43:
If this is not the correct way to do this, please feel free to close this one and I'll just do the PR for the revision.

PR #2386, opened 2022-07-11
Player companion

dythervin:
Added:

HealthChanged handler for the entity,
Player companion: 
- spawns on the game start/loaded, 
- not serialized,
- spawns a health potion when player health falls below a specific level
- follows the player, teleports when too far


Interkarma:
Hi,  I'm sorry but this is something that should be created as a mod. This is not the type of change I'll accept to core.

If help is needed creating a mod, please engage with the wider community for help with mod development.

https://forums.dfworkshop.net/viewforum.php?f=22

KABoissonneault:
Feel free to submit the `DaggerfallEntity.OnHealthChanged` event as a separate PR though. I can see this being useful

Interkarma:
I'd prefer not to have the event as implemented though. It's also easy enough just to track this anywhere using `GameManager.Instance.PlayerEntity.CurrentHealth` and do work when last check != current health. I'm not sure an event is really needed here.

KABoissonneault:
Eh. Such polling approaches don't make the cleanest code, and can add redundant computations every frame. Plus, you can run into "ABA problems" (https://en.wikipedia.org/wiki/ABA_problem) where the value changed multiple times in one frame, and you accidentally end up not detecting the change.

Interkarma:
Cheers Kab, I see your point. If this is something you think will be useful to the mod community, happy to review that event if it's submitted in its own PR.

PR #2388, opened 2022-07-14
K0C0XY01 fix

JayH2971:
K0C0XY01 is a harpy-slaying quest with several dummied-out items that still appear as dialogue choices. Not only that, the text is strange, like a description rather than dialogue. It makes more sense to remove them from the game, since they contribute nothing at all. The quest is still completable by delivering the leader's harpy feather to the questgiver, like vanilla.


Interkarma:
Thanks Jay!

PR #2389, opened 2022-07-15
optimization: `DaggerfallBillboardSystem` increases performance of `DaggerfallBillboard` with `IJobParallelForTransform`-based multithreading

andrew-raphael-lukasik:
Hi guys,

This is my first attempt here at implementing something that #2316 enabled.
I picked billboards not because those were causing any mayor performance issues, **they weren't**, but because I knew how to optimize them already and it is pretty much an isolated system from many complexities of this project.

<br>

These are **very** rough (editor & debug) numbers how billboards performed
> **Before**: `0.6 ms` (main thread)

Screenshot 2022-07-12 181808

> **After**:  `0.03 ms` (main thread) + `0.06 ms` (different thread)

Screenshot 2022-07-13 120343

Screenshot 2022-07-13 120516
> ☝ different thread

We can see that raw computation cost decreased about 10x, but that is not even that important. The thing that matters here is that this `0.5 ms` disappeared from the main thread, so it will no longer affect the main gameplay/rendering loop from now on.

<br>

If you point me any cases which I should test and make sure it works, that would be great. It looks fine on my machine, but I want to make sure my changes won't break something that I didn't even consider.

### DIFF
--- a/Assets/Scripts/Internal/DaggerfallBillboard.cs
+++ b/Assets/Scripts/Internal/DaggerfallBillboard.cs
@@ -80,10 +80,19 @@ namespace DaggerfallWorkshop
                     meshRenderer.enabled = false;
                 }
 
-                yield return null;// without this delay many billboards were broken (not animating or visible when shouldn't)
+                // wait a single frame so all objects initialize
+                yield return null;
 
-                if (faceY) DaggerfallBillboardSystem.AddPointBillboard(transform);
-                else DaggerfallBillboardSystem.AddAxialBillboard(transform);
+                // Rotate to face camera in game
+                // Do not rotate if MeshRenderer disabled. The player can't see it anyway and this could be a hidden editor marker with child objects.
+                // In the case of hidden editor markers with child treasure objects, we don't want a 3D replacement spinning around like a billboard.
+                // Treasure objects are parented to editor marker in this way as the moving action data for treasure is actually on editor marker parent.
+                // Visible child of treasure objects have their own MeshRenderer and DaggerfallBillboard to apply rotations.
+                if (meshRenderer.enabled)
+                {
+                    if (faceY) DaggerfallBillboardSystem.AddPointBillboard(transform);
+                    else DaggerfallBillboardSystem.AddAxialBillboard(transform);
+                }
 
                 // Restart animation coroutine if not running
                 if (summary.AnimatedMaterial)
--- a/Assets/Scripts/Internal/DaggerfallBillboard.cs
+++ b/Assets/Scripts/Internal/DaggerfallBillboard.cs
@@ -98,7 +98,6 @@ namespace DaggerfallWorkshop
                 if (summary.AnimatedMaterial)
                     StartCoroutine(AnimateBillboard());
             }
-            else yield return null;
         }
 
         void OnDestroy()
--- a/Assets/Scripts/Internal/DaggerfallBillboard.cs
+++ b/Assets/Scripts/Internal/DaggerfallBillboard.cs
@@ -39,6 +39,7 @@ namespace DaggerfallWorkshop
         const int lightFps = 12;
 
         MeshFilter meshFilter = null;
+        bool restartAnims = true;
         MeshRenderer meshRenderer;
 
         int framesPerSecond = 5;     // General FPS
@@ -63,7 +64,7 @@ namespace DaggerfallWorkshop
             set { faceY = value; }
         }
 
-        IEnumerator Start()
+        void Start()
         {
             if (Application.isPlaying)
             {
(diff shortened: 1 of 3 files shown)


Assets/Scripts/Internal/DaggerfallBillboard.cs
```diff
@@ -65,12 +63,11 @@ public override bool FaceY
             set { faceY = value; }
         }
 
-        void Start()
```
KABoissonneault, line 68: I actually had no idea you could turn most of Unity's events into coroutines... good to know

andrew-raphael-lukasik: yes `IEnumerator Start()` is handy sometimes, just a shorthand but useful nevertheless

Assets/Scripts/Internal/DaggerfallBillboard.cs
```diff
@@ -82,34 +79,31 @@ void Start()
...
-        }
 
-        void OnDisable()
-        {
-            restartAnims = true;
-        }
+                // wait a single frame so all objects initialize
+                yield return null;
 
-        void Update()
```
andrew-raphael-lukasik, line 93: I got rid of `Update` and moved this code to `Start` but I worry a bit this could have introduced some kind of unforeseen consequence. If you can point me any cases which I should test and make sure it works, that would be great. It looks fine on my machine, but I want to make sure my changes won't break something that I didn't even consider.

Assets/Scripts/Internal/DaggerfallBillboard.cs
```diff
@@ -82,34 +79,31 @@ void Start()
...
-            if (mainCamera && Application.isPlaying && meshRenderer.enabled)
-            {
-                float y = (FaceY) ? mainCamera.transform.forward.y : 0;
-                Vector3 viewDirection = -new Vector3(mainCamera.transform.forward.x, y, mainCamera.transform.forward.z);
-                transform.LookAt(transform.position + viewDirection);
+                // Restart animation coroutine if not running
+                if (summary.AnimatedMaterial)
+                    StartCoroutine(AnimateBillboard());
             }
+            else yield return null;
```
KABoissonneault, line 101: Why wait the extra frame there?

andrew-raphael-lukasik: [49ae337](https://github.com/Interkarma/daggerfall-unity/pull/2389/commits/49ae33762506451b4d9f93d88513da3608822a9c) fixed

Assets/Scripts/Internal/DaggerfallBillboardSystem.cs
```diff
@@ -0,0 +1,173 @@
...
+            void IJobParallelForTransform.Execute(int index, TransformAccess transform)
+            {
+                float3 position = (float3)transform.position;
+                transform.rotation = quaternion.LookRotationSafe(
+                    new float3 { x = Point.x, y = position.y, z = Point.z } - position,
+                    new float3 { y = 1 }
+                );
+            }
+        }
+
```
andrew-raphael-lukasik, line 167: **Sidenote**: `AxialLookAtPointJob` introduces an alternative axial alignment method. Just something to consider or just delete if deemed not worthy.

To test it locally just comment/uncomment these 2 lines: https://github.com/Interkarma/daggerfall-unity/pull/2389/files#diff-61902a9d59619dfbe36dfe1b492f7eeb4192bce7eb5f336d20d61f682fa4f806R81-R82

> `vanilla` left,  `AxialLookAtPointJob` right
<p float="center">
    
    
</p>

> `vanilla` left,  `AxialLookAtPointJob` right
<p float="center">
    
    
</p>

I am not sure which one is better, seems rather arbitrary, just something to consider maybe.

Assets/Scripts/Internal/DaggerfallBillboard.cs
```diff
@@ -82,36 +79,32 @@ void Start()
                     // Example is the treasury in Daggerfall castle, some action records flow through the quest item marker
                     meshRenderer.enabled = false;
                 }
-            }
-        }
 
-        void OnDisable()
```
Interkarma, line 88: Why was this removed? This pattern is here to restart animations after gameobject is disabled. For example, going inside and outside of a building will disable then enable billboard objects. With this removed in your PR, animating billboards are now broken when going in and out of a building.

Assets/Scripts/Internal/DaggerfallBillboardSystem.cs
Interkarma: Daggerfall's billboards should align to camera facing as they do in classic and in DFU. They both have issues depending on camera alignment and where player is standing, but I find camera-facing to be less obnoxious than billboards spinning around just by walking closer to them. It also matches classic, and I don't wish to change the current behaviour.

Interkarma:
Thank you for PR. It's interesting to see this as a test example, though I'm not convinced a 0.5ms saving per frame is worth the additional complexity. The PR also changes code for no reason, such as removing pattern to restart anims after disable and introducing alternate alignment methods. I'd prefer to see a much lighter touch here focused on optimisation effort only.

Please also keep in mind that I'm trying to reach 1.0 milestone, so any refactors that don't serve bug fix or localisation aren't on my radar right now. I'm still open to small changes, particularly for mod support, but they have to be very light touch.

It would be interesting to compare the savings in axial alignment using a shader (similar to what I did with billboard batch). Though I'm not sure it would be worth the additional complexity at this point either.

Assets/Scripts/Internal/DaggerfallBillboardSystem.cs
andrew-raphael-lukasik: I understand, [8a50def](https://github.com/Interkarma/daggerfall-unity/pull/2389/commits/8a50deff0992dfa5cc24e24a0061a34fc6502686) reverted this already

Assets/Scripts/Internal/DaggerfallBillboard.cs
andrew-raphael-lukasik: a8a36c44f5b61423dc50e18460d4dbbeff243db5 reverts these ill changes

Assets/Scripts/Internal/DaggerfallBillboard.cs
andrew-raphael-lukasik: a8a36c44f5b61423dc50e18460d4dbbeff243db5 reverts this mistake

Assets/Scripts/Internal/DaggerfallBillboardSystem.cs
```diff
@@ -0,0 +1,150 @@
+﻿// Project:         Daggerfall Unity
+// Copyright:       Copyright (C) 2009-2022 Daggerfall Workshop
+// Web Site:        http://www.dfworkshop.net
+// License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
+// Source Code:     https://github.com/Interkarma/daggerfall-unity
+// Original Author: Gavin Clayton (interkarma@dfworkshop.net)
```
Interkarma, line 6: Please don't forget to set yourself as the original author of this file. :)

Interkarma:
Thank you for refinement. The animation stall is now resolved and easier overall for me to review the optimisation changes.

This does make a quite a difference to the billboard alignment times. I profiled across locations of all sizes and found the 0.5ms per frame to be a very consistent saving. I can see how this setup could be useful in other places.

I'm a bit hesitant to merge right now. I'm committed to a stable 0.14.3 release by next week and don't want to introduce any last-second changes. But the process appears good and now doesn't introduce any problems I'm able to find. I'll work on merging this one sometime in next few releases.

andrew-raphael-lukasik:
I have doubts about this PR - is it really a good enough of a solution. Mainly because that occlusion-related overhead since [a8a36c4](https://github.com/Interkarma/daggerfall-unity/pull/2389/commits/a8a36c44f5b61423dc50e18460d4dbbeff243db5).

PR #2390, opened 2022-07-16
optimization: lets enable incremental garbage collector

andrew-raphael-lukasik:
# what:

This tiny PR enables an incremental garbage collector via `ProjectSetting.asset` file (a single line changed), nothing else.

# why:

As we know managed allocations pile-up leading to `GC.Collect` being called sooner or later. In 2018/2019 Unity introduced an incremental GC which is no longer an experimental feature for some time now. Here is an official article announcing it's introduction: https://blog.unity.com/technology/feature-preview-incremental-garbage-collection

`gcIncremental: 0` leads to sudden, very noticeable CPU spikes, freezing the game rendering for few frames. This is how it work now:

Screenshot 2022-07-16 232340

`gcIncremental: 1`, spike is spread across few frames, framerate is maintained:

Screenshot 2022-07-16 232929

### DIFF
--- a/ProjectSettings/ProjectSettings.asset
+++ b/ProjectSettings/ProjectSettings.asset
@@ -683,7 +683,7 @@ PlayerSettings:
   allowUnsafeCode: 0
   additionalIl2CppArgs: 
   scriptingRuntimeVersion: 1
-  gcIncremental: 0
+  gcIncremental: 1
   assemblyVersionValidation: 1
   gcWBarrierValidation: 0
   apiCompatibilityLevelPerPlatform:


KABoissonneault (approved):

KABoissonneault:
It will probably be really good for World of Daggerfall :P we tend to allocate a lot a new objects on the fly without pooling them. This leaves large though somewhat infrequent garbage, which is probably much better smoothed across multiple frames.

Interkarma:
Cheers! I remember reading about this ages ago and thinking "must check this out when it's ready". Then promptly lost track of it. :)

I think it's a valuable and low impact change. I'll work on including this one sometime in next few releases. It's probably fine, but I don't want to introduce any last minute changes before 0.14.3.

petchema:
Is all the logic to prevent calling Resources.UnloadUnusedAssets() too often (#1787) still relevant? Can it be simplified?

Interkarma:
Good memory @petchema. Yeah, that could possibly be simplified or removed now.

Let's see how things go with incremental GC enabled then we can look at reworking this.

PR #2391, opened 2022-07-19
Weapon HUD Animations Mod Support 1, 7/19/2022

magicono43:
Very basic mod support for making it possible through a mod with texture replacements to add unique HUD animation frames for the various different vanilla weapons.

So instead of only being able to replace the entire group of the generic "long-blade" animations that most bladed weapons use, the modder can specify through texture naming if they want to replace the animations for say the Claymore and Katana so they look different from one another on the HUD. Or even adding unique HUD animations for the vanilla weapon artifacts and enchanted magic weapons, etc, 7/19/2022.

### DIFF
--- a/Assets/Scripts/Game/FPSWeapon.cs
+++ b/Assets/Scripts/Game/FPSWeapon.cs
@@ -549,7 +549,7 @@ namespace DaggerfallWorkshop.Game
                     {
                         filename = WeaponBasics.GetModdedWeaponFilename(SpecificWeapon);
 
-                        if (filename == "")
+                        if (string.IsNullOrEmpty(filename))
                             filename = WeaponBasics.GetWeaponFilename(WeaponType); // Possibly make support for custom weapon types for the HUD in the future.
                     }
 
--- a/Assets/Scripts/Game/FPSWeapon.cs
+++ b/Assets/Scripts/Game/FPSWeapon.cs
@@ -4,7 +4,7 @@
 // License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
 // Source Code:     https://github.com/Interkarma/daggerfall-unity
 // Original Author: Gavin Clayton (interkarma@dfworkshop.net)
-// Contributors:    Hazelnut
+// Contributors:    Hazelnut, Kirk.O
 // 
 // Notes:
 //
@@ -22,6 +22,7 @@ using DaggerfallWorkshop.Utility.AssetInjection;
 using DaggerfallWorkshop.Game.Entity;
 using DaggerfallWorkshop.Game.Serialization;
 using DaggerfallWorkshop.Game.Formulas;
+using DaggerfallWorkshop.Game.Items;
 
 namespace DaggerfallWorkshop.Game
 {
@@ -38,6 +39,7 @@ namespace DaggerfallWorkshop.Game
         public WeaponTypes WeaponType = WeaponTypes.None;
         public MetalTypes MetalType = MetalTypes.None;
         public ItemHands WeaponHands = ItemHands.None;
+        public DaggerfallUnityItem SpecificWeapon = null;
         public float Reach = 2.5f;
         public float AttackSpeedScale = 1.0f;
         public float Cooldown = 0.0f;
@@ -74,6 +76,9 @@ namespace DaggerfallWorkshop.Game
 
         readonly Dictionary<int, Texture2D> customTextures = new Dictionary<int, Texture2D>();
         Texture2D curCustomTexture;
+
+        // Allows a mod to specify if DFU should use custom HUD animations for weapons
+        public static bool moddedWeaponHUDAnimsEnabled = false;
 
         float lastScreenWidth, lastScreenHeight;
         bool lastLargeHUDSetting, lastLargeHUDDockSetting;
@@ -539,6 +544,14 @@ namespace DaggerfallWorkshop.Game
                 for (int frame = 0; frame < frames; frame++)
                 {
                     textures.Add(GetWeaponTexture2D(filename, record, frame, metalType, out rect, border, dilate));
+
+                    if (moddedWeaponHUDAnimsEnabled && SpecificWeapon != null)
+                    {
+                        filename = WeaponBasics.GetModdedWeaponFilename(SpecificWeapon);
+
+                        if (filename == "")
+                            filename = WeaponBasics.GetWeaponFilename(WeaponType); // Possibly make support for custom weapon types for the HUD in the future.
+                    }
 
                     Texture2D tex;
                     if (TextureReplacement.TryImportCifRci(filename, record, frame, metalType, true, out tex))
--- a/Assets/Scripts/Utility/WeaponBasics.cs
+++ b/Assets/Scripts/Utility/WeaponBasics.cs
@@ -214,10 +214,17 @@ namespace DaggerfallWorkshop.Utility
                         case (int)Weapons.Dagger:
                             return "MEHRUNESRAZOR.CIF";
                         case (int)Weapons.Staff:
-                            if (weapon.ItemName == "The Wabbajack")
-                                return "WABBAJACK.CIF";
-                            else
-                                return "STAFFOFMAGNUS.CIF";
+                            foreach (DaggerfallConnect.FallExe.DaggerfallEnchantment enchantment in weapon.LegacyEnchantments)
+                            {
+                                if (enchantment.type == DaggerfallConnect.FallExe.EnchantmentTypes.SpecialArtifactEffect)
+                                {
+                                    if (enchantment.param == (int)ArtifactsSubTypes.Wabbajack)
+                                        return "WABBAJACK.CIF";
+                                }
+                                else
+                                    return "STAFFOFMAGNUS.CIF";
+                            }
+                            return "";
                         case (int)Weapons.Katana:
                             return "EBONYBLADE.CIF";
                         case (int)Weapons.Claymore:
--- a/Assets/Scripts/Utility/WeaponBasics.cs
+++ b/Assets/Scripts/Utility/WeaponBasics.cs
@@ -214,7 +214,7 @@ namespace DaggerfallWorkshop.Utility
                         case (int)Weapons.Dagger:
                             return "MEHRUNESRAZOR.CIF";
                         case (int)Weapons.Staff:
-                            if (weapon.ItemName == "Wabbajack")
+                            if (weapon.ItemName == "The Wabbajack")
                                 return "WABBAJACK.CIF";
                             else
                                 return "STAFFOFMAGNUS.CIF";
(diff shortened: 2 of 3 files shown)


magicono43:
So this was some work I decided to do so another mod author could more fully accomplish the goal of their mod, this mod being "Diverse Weapons" [Nexus Page](https://www.nexusmods.com/daggerfallunity/mods/242)

The mod aiming to give all the vanilla weapons a more unique identity visually, both on the Paper-doll inventory sprites, as well as the HUD attack animations. Unfortunately they were previously only able to replace some of the HUD animations, but this PR aims to allow them to do this for ALL of the vanilla weapon types, with hopefully little impact or conflict for the main code-base, since atm the feature is only active if a mod specifically sets a new bool value from "false" to "true" through the mod.

Here are some simple examples on how the naming works currently to replace specific weapon types:

In this case we are replacing the "Katana" of "Steel" material type.
HowToNameNormalWeapon

Here we are replacing the "Dagger" of "Ebony" material type, that also happens to be enchanted in some way.
HowToNameGenericMagicWeapon

Finally here we replace the "Volendrung" daedric artifact. Annoyingly we still have to include the material it is made out of, in this case that being "Dwarven." This will hopefully not be too confusing just as long as the mod creator of said artifact texture replacements is informed of what DFU sets these various artifacts to for their "base" material type.
HowToNameArtifactWeapon

magicono43:
Just for any mod creator reference here are the current as the making of this PR ways to name the weapons you would want to replace the HUD animations for:

Daedric Artifact Weapons (Material types not yet included)
ArtifactNamingConv

Magically Enchanted Weapons
MagicGenericNamingConv

Normal Weapons
GenericWeaponNamingConv

magicono43:
Finally on the info part, here are some useful resources for somebody wanting to test, as well as an example mod to use to show the stuff in action and a bare-bones mod script to show how to use it in a mod:

Unfortunately can't link compressed archive files on Github comments, so I made a forum post with the attached useful files:
[Forum Thread With Useful Links](https://forums.dfworkshop.net/viewtopic.php?t=5759)

magicono43:
Here is a small list with some examples of observed potential bugs/quirks/issues that came about while doing testing with this feature as it currently is:

Known Observed Bug With My PR 2:
So the artifact thing actually works thankfully, but it's a bit weird. If you equip another weapon that has the same vanilla sprites, at least the idle one right after equipping the other similar item, the texture will sort of "linger" from the last weapon. 
But if you equip a weapon from the inventory, then exit the inventory screen
Open it up again and equip the similar thing, if you don't exit the inventory after equipping (or something) the previous one will linger basically, it's consistent but hard to describe, likely something with how the screen refreshes/rechecks some values, will do more testing.

Observed Bug 1:
Weapons that normally share the same vanilla HUD animations, examples being say a "Enchanted Dwarven Warhammer" and "Volendrung." If you change these weapons in the inventory screen. Say if you currently have Volendrung equipped, but then you switch the a Enchanted Dwarven Warhammer, when you exit the inventory and unsheath your weapon, it is likely that the Idle sprite at least for Volendrung will used instead of the Enchanted Dwarven Warhammer. But if you instead unequip Volendrung, exit the inventory, the reenter and equip the Enchanted Dwarven Warhammer, you will then properly see the EDW's idle sprite when you unsheath.

I did not do testing for the normal weapons in this case, say a steel katana and a steel saber (which normally share the same HUD animations in vanilla), this behavior likely will happen in that case as well. It's kind of minor, but just something to keep in mind as a known issue. I believe it has something to do with how the FPSWeapon logic "refreshes" some values only when certain things happen such as closing the inventory screen with a new weapon equipped. Not certain how this could be fixed, but I'm sure it could be if I understood the running logic a bit better.

Observed Quirk 2:
Weapons that have a specific "WeaponType" as defined in the code for their vanilla behavior will keep the screen dimensions that are given even if replaced by another texture. So if you say design a short-sword to work with the size of the dagger or the long-sword in terms of the width and height they are normally defined in vanilla on the HUD. If you put those same replaced sprites for say a flail or warhammer, the replacement sprites will be warped/stretch to fit the screen based on the attributes that the flail and warhammer would normally use.

This is likely something that always has to be worked with when doing any texture replacements, but I figured I would mention it as something that might be undesired, I'm also not certain how to fix this, possibly after "ri" is defined in the logic you would instead put your own values after the texture replacement process or something, but that' just my guess. Here are some examples that hopefully explain better what I mean:

Katana:
Katana1

Dagger:
Dagger1

Flail:
Flail1

Magic Warhammer:
Magic_Warhammer1

magicono43:
RealAKP sent me some more of the HUD animations they are working on, tested them in-game as well and they worked fine. Here is some images samples for those interested:

Steel Katana Idle:
Steel Katana Example

Ebony Katana Idle:
Ebony Katana Example

Assets/Scripts/Game/FPSWeapon.cs
```diff
@@ -540,6 +545,14 @@ private void LoadWeaponAtlas()
                 {
                     textures.Add(GetWeaponTexture2D(filename, record, frame, metalType, out rect, border, dilate));
 
+                    if (moddedWeaponHUDAnimsEnabled && SpecificWeapon != null)
+                    {
+                        filename = WeaponBasics.GetModdedWeaponFilename(SpecificWeapon);
+
+                        if (filename == "")
```
Interkarma, line 552: Please use string.IsNullOrEmpty for these checks.

Assets/Scripts/Utility/WeaponBasics.cs
```diff
@@ -202,6 +203,130 @@ public static string GetMagicAnimFilename(ElementTypes elementType)
...
+            if (weapon.IsEnchanted)
+            {
+                if (weapon.IsArtifact)
+                {
+                    switch (weapon.TemplateIndex)
+                    {
+                        case (int)Weapons.Dagger:
+                            return "MEHRUNESRAZOR.CIF";
+                        case (int)Weapons.Staff:
+                            if (weapon.ItemName == "The Wabbajack")
```
Interkarma, line 217: Best to avoid a string literal check here, as someone might translate the name in future, e.g. "Der Wabbajack" or "
Le Wabbajack" or even "ワバジャック" :)

There must be some other unique way of identifying Wabbajack item here.

Assets/Scripts/Utility/WeaponBasics.cs
```diff
@@ -202,6 +203,130 @@ public static string GetMagicAnimFilename(ElementTypes elementType)
             }
         }
 
+        public static string GetModdedWeaponFilename(DaggerfallUnityItem weapon)
```
Interkarma, line 206: I was trying to find a way to use the Table system to break this into external data for you, but it's a bit of a pain to manage that way. The switch statement is easier and for now fits in with how DFU already does this work.

I'm OK with this handling in WeaponBasics. :)

magicono43: Great, I know it's not the cleanest way of doing stuff, but it seems like an alright option especially for the vanilla weapon types at least. Sucks that it takes up a bunch of lines, but there are worse examples of this already I suppose, lol.

Assets/Scripts/Utility/WeaponBasics.cs
magicono43: Yeah, I did not consider that fact, I'll try and use one of the other ways that it is done in the code. Which I think is mainly through checking unique spell effects for some of them. The reason I mainly went with the string check being I was not really 100% understanding how those unique effects checks were working. Also a note that there is at least a few examples in "FormulaHelper.cs" that is checking artifacts based on their literal item-name.

magicono43: I was looking at some possible alternative in the code, but I'm sort of concerned that the only good potential alternative I found might be slower? That being checking the "SpecialArtifactEffects", but at least in this example it seems to need to loop through the entire item.LegacyEnchantments to get these values? I might just be misunderstanding it though.

Capture

This is also where I got the idea of using the literal item-name as a way of identification, since it's used at the bottom there for Chrysamere and Staff of Magnus, which apparently don't have any "unique" effect like some of the other artifacts.

Assets/Scripts/Utility/WeaponBasics.cs
Interkarma: Yeah, it sort of plays on my mind how I'd like to do this better. But we're kind of stuck with interfacing to classic data and there's only so much work we can reasonably do as unpaid devs. It's fine. :)

Assets/Scripts/Game/FPSWeapon.cs
magicono43: Changed the if-statement check to instead use, string.IsNullOrEmpty(filename)

Assets/Scripts/Utility/WeaponBasics.cs
Interkarma: > Also a note that there is at least a few examples in "FormulaHelper.cs" that is checking artifacts based on their literal item-name

Right you are. :) For my part, I'm gradually (slowly) working back through all the use of string literals and trying to migrate these over to support localization in some way. This has been a real agony for me and I often feel like I'm barely making progress. What I'm trying to do (if possible) is not introduce any more cases of this in new code. That way I only have to fix the old code and not keep chasing new instances being added by new code. The gap here of course is that handling for this hasn't been added yet so there's no good solution for you to move to either.

I'm sorry if that's adding complexity for you. I can definitely make an exception in this case if we can't find another way to identify cleanly. What's one more. :)

Assets/Scripts/Utility/WeaponBasics.cs
Interkarma: > Also a note that there is at least a few examples in "FormulaHelper.cs" that is checking artifacts based on their literal item-name

Right you are. :) For my part, I'm gradually (slowly) working back through all the use of string literals and trying to migrate these over to support localization in some way. This has been a real agony for me and I often feel like I'm barely making progress. What I'm trying to do (if possible) is not introduce any more cases of this in new code. That way I only have to fix the old code and not keep chasing new instances being added by new code. The gap here of course is that handling for this hasn't been added yet so there's no good solution for you to move to either.

I'm sorry if that's adding complexity for you. I can definitely make an exception in this case if we can't find another way to identify cleanly. What's one more. :)

magicono43: Yeah, it does not help obviously that artifacts are such a "unique" case. I'm sure this sort of thing happens in other code-bases within the Elder Scrolls series, just sort of hard-coded values and names. I'll see if I can do something with the unique effect checking like is shown in that screenshot, but I don't have too much confidence if it will be much better from me, lol.

Interkarma: It's fine, let's not stress this one for now. :) I'll review this one better in another pass when I have a bit more time and will raise any better ideas if one comes up.

magicono43: I just made another commit that will hopefully be a less string literal solution, using the same principle as the attached screenshot, just slightly modified.

PR #2392, opened 2022-07-19
Two Artifacts HUD Color Inaccuracies Correction, 7/19/2022

magicono43:
Realized while doing some testing on my recent PR that some of the artifact weapons have unfitting or inaccurate material values set that make their HUD weapon colors not very fitting to what their actual Paper-doll sprite would suggest. This might be mostly up to opinion, but I figure I bring it up.

So it seems that Chrysamere was set to "Elven" which was making it look pretty stark white on the HUD, which I don't think looks right for it at all myself. The unofficial wiki says it is considered Adamantium, but I personally don't think that colors fits well either since it's basically as dark as ebony. So I decided that really the closest color imo was mithril, so I decided to change it to that. It's not perfect since it's a pretty dark-blue, rather than the sort of greenish teal that the paper-doll sprite looks like, here are just some comparison images:

In the code what it was previously:
Seems Like Not Fitting Materials

Paper-doll color:
Capture3

Before Change:
Chrysamere Before Elven

After Change:
Chrysamere After Mithril

### DIFF
--- a/Assets/Scripts/Game/FPSWeapon.cs
+++ b/Assets/Scripts/Game/FPSWeapon.cs
@@ -4,7 +4,7 @@
 // License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
 // Source Code:     https://github.com/Interkarma/daggerfall-unity
 // Original Author: Gavin Clayton (interkarma@dfworkshop.net)
-// Contributors:    Hazelnut, Kirk.O
+// Contributors:    Hazelnut
 // 
 // Notes:
 //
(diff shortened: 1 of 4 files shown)


magicono43:
And now for the stuff about "The Wabbajack" so the reason I thought this deserved a change as well is just that steel looks nothing like the actual color of the "metal caps" on the edges of the staff for the paper-doll. If you look at the paper-doll sprite it's actually just a copy/paste from the dwarven staff for the metal cap parts:

Wabbajack Paper-doll:
Capture2

Dwarven Staff Paper-doll:
Capture

So I sort of felt it was more fitting to have it as Dwarven rather than steel, once again up to interpretation, but I think it looks better anyway for an item of it's infamy, here are the comparisons:

Wabbajack Before:
Wabbajack Before Steel

Wabbajack After:
Wabbajack After Dwarven

Interkarma:
This PR is fine, no problems with change. But due to commit sequence this one conflicts with #2391 changes once merged. I'll just close this one and quickly make those two changes on my end.

PR #2394, opened 2022-07-24
Remove stutter when switching between hands

numidium:
I get a stutter on my mid-tier PC whenever I switch between my right and left hand weapons in DFU. Unity profiler says it's ~200ms each time. After some analysis I found that the culprit, GetWeaponTextureAtlas, re-draws the entire weapon atlas each time the player switches weapons.

I opted to use a queue to implement this since I didn't want to clog the memory with atlases for every permutation of weapon type and metal dye. I settled on a cache size of 4 to accommodate lower-end systems while also allowing players to juggle a number of weapons without hitching.

Note that this currently only works for vanilla weapon atlases. I wanted to know if you think this optimization warrants a merge before continuing on.

EDIT: I refined the caching scheme a little further to only perform a texture load when equipping a weapon. This is normally done from the inventory menu unless you're using a mod so it should be less noticeable than a hitch occurring outside of a menu in my opinion. With this change applied, the player will never experience hitching when swapping hands.

The only drawback is the memory footprint. My cache holds 4 atlases at a time which comes out to about 64 MB of texture data. It would be much smaller if the textures could use a 256 color palette in memory but I don't know if Unity has any support for such a thing.


Interkarma:
Thanks Numidium. :) I'll need to think this one through a bit, I might want to try and solve a different way. I'd also prefer any solution to support both classic and modded textures while having the smallest footprint possible for either.

For some background - I purposefully didn't cache weapons using the default MaterialReader dict so that weapon anims didn't pile up over time. You've at least addressed that by limiting history in a custom cache setup.

The obvious reason not to cache anything as a 256 colour palette is that weapon images can be replaced like anything else. The  intended pipeline in DFU is that it can read/convert palettized images from classic data, while internally using 32-bit textures so that classic and modded assets can mix freely.

numidium:
I modified my changes a bit to support modded textures. The vanilla and modded textures have their own separate cache queues. I reduced the size of both to 2 since you said you wanted the smallest possible memory footprint (1 for each hand). I'm fine if we ultimately don't end up solving the issue this way, just wanted to make a proof of concept.

Interkarma:
Thank you @numidium. The state of this PR is very good, and I don't think I could do any better. I'll do some more testing and aim to merge for next release.

Interkarma:
Hey @numidium! There looks to be an issue with cache and modded weapon textures. Report raised by KoW below on forums.

https://forums.dfworkshop.net/viewtopic.php?t=5954

numidium:
Oh dear, that's a really bad oversight on my part. I'll get on it ASAP.

Interkarma:
Thanks for taking a look. :) I'm sorry I didn't find it in review. I tried to find any problems, and DREAM was one of the mods I tested. I just didn't hit the right sequence to reproduce this.

PR #2396, opened 2022-07-27
Mods can provide a female corpse texture for classic or custom enemies

KABoissonneault:
Kamer would really like if we could have unique corpses for each enemy, but human enemies have uh, sexual dimorphism that's not reflected on their gender neutral corpse.

With this simple change, ladies can now turn to dust on death
image


Interkarma:
Simple enough change, thanks Kab.

PR #2397, opened 2022-07-31
Disable doppler effect

numidium:
The Doppler effect implementation in Unity causes odd-sounding, and at times extreme, pitch shifts to occur during play. This is most noticeable to me when I run past an enemy as it plays its bark. It's a matter of preference but I believe it may be best to simply disable it since it was not something present in classic.


Interkarma:
This is fine. I don't have strong feelings about the default doppler either way. I'll merge this one in next round of builds.

PR #2399, opened 2022-08-01
latest-savegame-name-automated

andreiserbanrazvan:
Name of latest save game gets auto populated in the saveNameTextBox inside Save Game Window

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallUnitySaveGameWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallUnitySaveGameWindow.cs
@@ -87,6 +87,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
         {
             this.mode = mode;
             this.displayMostRecentChar = displayMostRecentChar;
+            LoadLatestSaveGameIntoSaveNameTextBox();
         }
 
         #endregion
@@ -460,6 +461,18 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
 #pragma warning restore 0618
         }
 
+        #endregion
+
+        #region Private Methods
+        private void LoadLatestSaveGameIntoSaveNameTextBox()
+        {
+            int mostRecentSave = GameManager.Instance.SaveLoadManager.FindMostRecentSave();
+            SaveInfo_v1 latestSaveInfo = GameManager.Instance.SaveLoadManager.GetSaveInfo(mostRecentSave);
+            currentPlayerName = latestSaveInfo.characterName;
+            saveNameTextBox.Text = latestSaveInfo.saveName;
+            screenshotPanel.BackgroundTexture = new Texture2D(0, 0);
+            UpdateSelectedSaveInfo();
+        }
         #endregion
 
         #region Event Handlers
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallUnitySaveGameWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallUnitySaveGameWindow.cs
@@ -87,10 +87,6 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
         {
             this.mode = mode;
             this.displayMostRecentChar = displayMostRecentChar;
-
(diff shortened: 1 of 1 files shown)


Interkarma:
Hey there. :)

What's the use case of this change? I'm not sure that I like the name of most recent save being auto-populated here. Player could just as likely be going for some other save than most recent. I'm not sure what the benefit is of filling out the text string like this.

KABoissonneault:
I see it as a compromise on the Continue button found in other games, which we can't add because of the classic DF menu.
I think it's a slight upside with no downside.
It saves a mouse click for a common use case (left-handed person with a right-handed mouse here, I like saving a right-hand action). You can immediately press the hotkey (Enter?) to load it.
Users who want an older save can still click on it in one click, same as before.

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallUnitySaveGameWindow.cs
```diff
@@ -462,6 +470,18 @@ protected virtual void SetBackground(Button button, Color color, string textureN
...
         #endregion
 
+        #region Private Methods
+        private void LoadLatestSaveGameIntoSaveNameTextBox()
+        {
+            int mostRecentSave = GameManager.Instance.SaveLoadManager.FindMostRecentSave();
+            SaveInfo_v1 latestSaveInfo = GameManager.Instance.SaveLoadManager.GetSaveInfo(mostRecentSave);
+            currentPlayerName = latestSaveInfo.characterName;
+            saveNameTextBox.Text = latestSaveInfo.saveName;
+            screenshotPanel.BackgroundTexture = new Texture2D(0, 0);
```
KABoissonneault, line 480: Couldn't we fill this with the save's picture, or do we want to avoid loading a file without waiting for user action?

Interkarma:
> I think it's a slight upside with no downside.

Thanks for the input Kab. I still not convinced of the change though. The most recent save for character is already at top of list, so a simple double-click at top of list or quicksave/quickload in-game already has this covered. I personally don't like having the most recent save queued here.

In any case, this PR could be implemented a bit better - e.g. simply select first item in list. This is already sorted in reverse chrono order and a select should populate screenshot and other data normally.

Interkarma:
To accomplish this a bit better, all you need is the following code in OnPush() at line 293.

    // Autoselect save at top of list
    if (mode == Modes.LoadGame && savesList.Count > 0)
        savesList.SelectIndex(0);

This will auto-select the most recent save and ensure screenshot and other information is properly populated.

image

Interkarma:
I'm reflecting on why I personally find this so distracting. I think it's because I use saves mainly for test setups and I rarely play sequentially. So when I load a save from one test setup then open UI again, I experience some discontinuity because the save selected is not the save I just opened. I find it easier to see a blank slate then move to the next save I need. That's just how my own mental map works.

With that in mind, I understand most people would in fact play sequentially, so auto-selecting the most recent save would have a small upside as Kab points out.

If PR can please be updated to use the method I point out above, I'd be happy to accept this and credit @andreiserbanrazvan for change.

andreiserbanrazvan:
Hello :), I updated the PR according to your suggestions @Interkarma. I can't believe I wrote all that for something that was a 1 liner.

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallUnitySaveGameWindow.cs
andreiserbanrazvan: Hey, I wrote that (line 480) only to have the flux work as usual without breaking.

andreiserbanrazvan:
@Interkarma, @KABoissonneault Yes indeed. I played Skyrim mainly. Maybe that's why it felt more natural to me. But my primary motivation was that I wanted to start learning about the codebase, so I needed to pick something up and do it. I was still in the tutorial dungeon, loading a lot, and noticed I needed 2 clicks to reload that save. It seemed like a good place to start for me to gain some experience with the game and codebase. I tried looking at existing issues first, but they required deeper knowledge of Daggerfall. If you know of anything I could pick up, please let me know.

Interkarma:
Thanks for updating. :) I'll get this one merged in next round of releases after 0.14.3.

PR #2401, opened 2022-08-05
Fixed "MuteHumanSounds" and "CharmEffect" for custom class enemies

KABoissonneault:
Replace "id >= 128" logic in DFU with a new helper, `DaggerfallEntity.IsClassEnemyId`, for better consistency in custom enemy support. Remember that for custom enemies, for every multiple of 256, the first 128 ids are monsters, and the last 128 ids are class enemies (this allows infinite extensibility while allowing mods to reserve an id space for themselves).

I made this change at first thinking MuteHumanSounds was an actual user-facing feature that players could turn off, and I wanted custom class enemies to properly support it. But I eventually realized only code can disable it. So that part of the change is not really necessary, I can just remove the sounds from my custom enemies. 

But the CharmEffect fix is useful, so I kept it all. It's not that big

### DIFF
--- a/Assets/Scripts/Game/EnemySounds.cs
+++ b/Assets/Scripts/Game/EnemySounds.cs
@@ -12,6 +12,7 @@
 using UnityEngine;
 using System.Collections;
 using DaggerfallWorkshop.Utility;
+using DaggerfallWorkshop.Game.Entity;
 
 namespace DaggerfallWorkshop.Game
 {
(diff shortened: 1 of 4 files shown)


petchema:
There's a regression with City Watches, they no longer shout "Halt"
Forums: https://forums.dfworkshop.net/viewtopic.php?t=5847

PR #2403, opened 2022-08-06
Classic player death sound option

petchema:
Ties classic player death sound to combats vocalization; No combats vocalization <=> classic player death sound.
Rationale:
- players bothered by "non classic" death gasp are more than likely also bothered by combats vocalization;
- on the other hand I see a possible case for players that don't want combats vocalization but like race and gender aware death sounds. Seems pretty niche case though.


Interkarma:
"Nyuck!" makes a comeback. :) Your reasoning is fine here and I'm happy with the approach. Preferable to making yet another launcher option for such a minor thing. Thank you @petchema!

PR #2405, opened 2022-08-10
Keep levitating as long as NoClip is active

petchema:
Levitation could stop under 2 conditions while NoClip status was still active:
- some levitation spell (on top of levitation caused by NoClip) expired;
- a game was (re)loaded; NoClip status was persistent, but not its associated flying effect.

It is still possible to disable it with "levitation off" in console, in case one absolutely wants NoClip with no levitation.

Another implementation could be to have LevitationMotor::IsLevitating() return playerLevitating || GameManager.Instance.PlayerEntity.NoClipMode if its entity is the player; It could be a little cleaner than spreading PlayerEntity.NoClipMode checks in several places, but a little more expensive at runtime.


Interkarma:
This should be fine. Thank you. :)

PR #2406, opened 2022-08-10
Always consume teleportation anchor when recalling

petchema:
When recalling, anchor was not consumed in all code paths

Regression since 239ad207661804c33d02a86ed9aa5e05db9a484e

Forums: https://forums.dfworkshop.net/viewtopic.php?t=5788


Interkarma:
Thank you @petchema!

PR #2408, opened 2022-08-18
Click to attack: swing from all directions

petchema:
UpLeft and Up direction could not happen with click-to-attack

Reddit: https://www.reddit.com/r/daggerfallunity/comments/wqgt4d/comment/iko3rg9/?utm_source=reddit&utm_medium=web2x&context=3

### DIFF
--- a/Assets/Scripts/Game/WeaponManager.cs
+++ b/Assets/Scripts/Game/WeaponManager.cs
@@ -338,7 +338,7 @@ namespace DaggerfallWorkshop.Game
                 }
                 else if (isClickAttack)
                 {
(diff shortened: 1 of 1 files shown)


petchema:
It seems there's more logic around click-to-attack than I thought
https://github.com/Interkarma/daggerfall-unity/commit/c1b3b1b613f9239c8da0b2cfcad7531260dc2ea7
Will need to test it more

petchema:
Okay, so since there's no UpLeft and UpRight animations they're conflated with Up; and click-to-attack only generated UpRight to avoid picking Up directions 3 times more often than others. I assume this is correct.

PR #2412, opened 2022-08-29
Fixed regression where City watch lost its human sounds (HALT)

KABoissonneault:
Fix for [https://forums.dfworkshop.net/viewtopic.php?t=5847](https://forums.dfworkshop.net/viewtopic.php?t=5847)

I also added an extra custom enemy id fix for Foe quest resources. Once we get quests targeting custom enemies, it would be nice if custom class enemies didn't have monster names.


Interkarma:
Thank you Kab!

PR #2414, opened 2022-09-03
typo quest exposition J0B00Y00

petchema:
I may a job -> I may have a jpb


crowbardoctor:
Isn't "jpb" also misspelled?

PR #2415, opened 2022-09-04
Changes to allow randomized billboard scaling in billboard batches.

BadLuckBurt:
I've added three new properties and one method to facilitate randomized billboard scaling when applying a billboard batch. This allows batched flats to vary in size, making for a more natural looking canopy / skyline in case of the nature flats but it can be applied to any billboard batch.

I exaggerated the scaling values in the screenshot, min scale was 0.75, max scale was 2.75 but it does help to illustrate the effect. I've seen performance drops when I set max scale to 4.0, probably due to overlapping sprites, at some point they just become too big. 

natureScaleExample

### DIFF
--- a/Assets/Scripts/Internal/DaggerfallBillboardBatch.cs
+++ b/Assets/Scripts/Internal/DaggerfallBillboardBatch.cs
@@ -61,6 +61,9 @@ namespace DaggerfallWorkshop
         [Range(1, 127)]
         public int RandomDepth = 16;
         public float RandomSpacing = BlocksFile.TileDimension * MeshReader.GlobalScale;
+        public float MinRandomScale = 1.0f;
+        public float MaxRandomScale = 1.0f;
+        public int RandomSeed = -1;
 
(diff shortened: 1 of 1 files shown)


andrew-raphael-lukasik:
Oh dear, the unfortunate timing. My https://github.com/Interkarma/daggerfall-unity/pull/2416 changes this class **a lot** and touches some of these lines as well. Merge can be resolved if needed ofc.

BadLuckBurt:
I had a look at your version, while it certainly changes a lot, it doesn't seem like my changes would be hard to integrate if this PR even makes it into the core.

BadLuckBurt:
While this PR is still relevant for random scaling of vanilla billboards that are loaded from Daggerfall's data, my nature layout mod has fully controllable scaling already via use of a custom texture atlas. I'd definitely prefer Andrew's newer PR for the updated DaggerfallBillboardBatch class to be merged because of the significant performance increase, resubmitting this PR for the updated class is trivial.

Issue #2417, opened 2022-09-05
player falls through reality if levitate spell expires while no clipping


breakfastitem:
**To Reproduce**

1. Go to anywhere
2. cast levitate spell
3. tcl in console and fly into the air
4.  wait for levitate to expire
5. player falls

**Expected behavior**

no clipping should be maintained even if levitate expires
 ____________________________________
tcl command must be retoggled to noclip

breakfastitem:
this also appears to be a duplicate of #2405

PR #2418, opened 2022-09-05
add not noclipping stipulation before canceling levitate

breakfastitem:
I'm still investigating no clip behaviors.


breakfastitem:
addresses #2417

PR #2419, opened 2022-09-05
Added a "generator" customization point for PopulationManager

KABoissonneault:
One of our mod plans is to expand the selection of mobile NPCs generated in towns.

Currently, the PopulationManager chooses the race of the mobile using the climate settings, then passes it on to `MobilePersonNPC.RandomiseNPC` which takes care of the rest. Other NPC settings like gender, outfit, face, guard, and name are determined internally.

We've previously worked around some of these limitations with `PopulationManager.OnMobileNPCEnable`, but all we can really do is call `RandomiseNPC` a second time with another race (we've put Redguards in Swamp and Subtropical climates, for example). This approach only works for Bretons, Redguards, and Nords, since other races aren't supported.
In addition to the limitations of this approach, this also had the issue of causing two calls to `MobilePersonAsset.SetPerson`, which cause two sets of textures to be loaded per NPC.

This PR thus does two things:
- give mods access to setting `MobilePersonNPC` properties from the outside
- Give `PopulationManager` a hook for mods to do their own thing instead of calling `.RandomiseNPC` (avoids the double call to `MobilePersonAsset.SetPerson` problem)

With this, we can generate NPCs of any race, with any outfit and face we want. Custom portrait records are already supported by `DaggerfallTalkWindow.SetNPCPortrait`.

Results (ignore sprite scale lol):
image
image

### DIFF
--- a/Assets/Scripts/Game/Utility/PopulationManager.cs
+++ b/Assets/Scripts/Game/Utility/PopulationManager.cs
@@ -192,7 +192,16 @@ namespace DaggerfallWorkshop.Game.Utility
                 {
                     poolItem.npc.Motor.gameObject.SetActive(true);
                     poolItem.scheduleEnable = false;
-                    poolItem.npc.RandomiseNPC(GetEntityRace());
+
+                    if (MobileNPCGenerator != null)
+                    {
+                        MobileNPCGenerator(poolItem);
+                    }
+                    else
+                    {
+                        poolItem.npc.RandomiseNPC(GetEntityRace());
+                    }
+
                     poolItem.npc.Motor.InitMotor();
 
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Game/Utility/PopulationManager.cs
```diff
@@ -279,6 +288,7 @@ int CreateNewPoolItem()
             PoolItem poolItem = new PoolItem();
             poolItem.npc = npc;
             poolItem.npc.Motor = motor;
+            poolItem.npc.Asset = motor.MobileAsset;
```
Interkarma, line 291: Good idea. This is a better place to set `Asset` rather than use `GetComponentInChildren`. :)

PR #2420, opened 2022-09-05
performance and productivity utilities

andrew-raphael-lukasik:
Hi! This PR is part of #2416 saga.

## `UniversalJobs.sc`
contains jobs that can come useful anywhere:
- `DeallocateArrayJob`
- `FillJob`
- `CopyToJob`
- `ReleaseGCObjectJob` - more on what this does below

## `NativeArrayExtensionMethods.cs`
contains methods such as
- `CopyTo(this NativeArray<T> src, T[,] dst)`
- `CopyTo(this NativeArray<T> src, NativeArray<T> dst, JobHandle dependency)` - delayed copy
- `Fill(this NativeArray<T> Array, T value)`
- `AssertPtrScope(this NativeSlice<ARRAY_ITEM> array, POINTER* ptr)` - pointer safety assertion (mandatory before dereferencing any pointer)

## `NativeArrayExtensionMethodsTests.cs`
unit tests

## `ArrayExtensionMethods.cs`
contains:
- `AsNativeArray(this T[] array, out ulong gcHandle)` variants
Yes you red that right - this is not a pedestrian `ToNativeArray` (copy) but `AsNativeArray` and creates a `NativeArray` struct that represents any managed array. Im super exited for this as I finally figured how to do this and figured out the proper safe workflow (as seen in #2416)

## `JobHelpers.cs`

Hosts `JobUtility` as I couldn't find a more suitable place for this.

It allows to call `JobUtility.ReleaseGCObject(ulong gcHandle, JobHandle dependency)` which is a convenient way of releasing managed arrays (it schedules `ReleaseGCObjectJob`).

##

### DIFF
--- a/Assets/Scripts/Terrain/JobHelpers.cs
+++ b/Assets/Scripts/Terrain/JobHelpers.cs
@@ -4,12 +4,14 @@
 // License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
 // Source Code:     https://github.com/Interkarma/daggerfall-unity
 // Original Author: Hazelnut
-// Contributors:    
+// Contributors:    Andrzej Łukasik (andrew.r.lukasik)
 // 
 // Notes:
 //
 
 using System;
+using Unity.Collections;
+using Unity.Jobs;
 
 namespace DaggerfallWorkshop
 {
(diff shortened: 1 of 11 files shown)


Interkarma:
Thank you @andrew-raphael-lukasik! This is a great starting point. I'm very happy to see tools in a separate file, and we can start reviewing functional changes to the loading process in smaller stages.

Just so you know where I'm headed in the next few months - I'm aiming to reach a highly stable release of 0.14.x with some additional fixes and complete localization to a point where translation mods can be distributed. I also have some external factors driving my desire to make as few changes to 0.14 as possible and not break any mods.

The short version is that I'm about to roll out of Beta into 0.15 Release Candidate. Daggerfall Unity is going to have a lot of visibility during this time and a lot of first impressions will be made for new players. So I'm going to be _very_ conservative with any changes that might break mods or introduce any compatibility issues with saves.

Once a few of these things settle in, I'll be able to start rolling in those changes.

PR #2421, opened 2022-09-06
[BurstCompile] for all existing jobs

andrew-raphael-lukasik:
Hi! This PR is part of #2416 saga.

## What

Burst compilation enabled for all existing jobs. That's it, nothing else.

## Why

Burst is not a magical wand necessarily. It is powerful when there is a lot of math calculations but it won't fix a memory access-bound code (L2 cache misses).

## Results

It sped up these existing jobs by around a factor of 10x and nearly `*halves*` my loading time for exteriors (20-40% reduction)

- before: `5 seconds` total of work (left), after `[BurstCompile]`: `0.4 second` total of work (right)
<p float="center">
    
    
</p>

> **I mean... optimization is *never* this easy, so merge while PR is still hot :T**

### DIFF
--- a/Assets/Scripts/Terrain/TerrainTexturing.cs
+++ b/Assets/Scripts/Terrain/TerrainTexturing.cs
@@ -14,6 +14,7 @@ using System;
 using DaggerfallConnect.Arena2;
 using Unity.Jobs;
 using Unity.Collections;
+using Unity.Mathematics;
 
 namespace DaggerfallWorkshop
 {
@@ -103,6 +104,7 @@ namespace DaggerfallWorkshop
         // Very basic marching squares for water > dirt > grass > stone transitions.
         // Cannot handle water > grass or water > stone, etc.
         // Will improve this at later date to use a wider range of transitions.
+        [Unity.Burst.BurstCompile]
         protected struct AssignTilesJob : IJobParallelFor
         {
             [ReadOnly]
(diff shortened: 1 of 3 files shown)


Interkarma:
Nice one! This one should be fine to just merge. I'll aim to get this into next release for general testing. Thank you. :)

Assets/Scripts/Terrain/TerrainTexturing.cs
```diff
@@ -219,7 +222,7 @@ public void Execute(int index)
                 }
                 // Beach texture
                 // Adds a little +/- randomness to threshold so beach line isn't too regular
-                if (height <= beachElevation + (JobRand.Next(-15000000, 15000000) / 10000000f))
+                if (height <= beachElevation + Unity.Mathematics.Random.CreateFromIndex((uint)index).NextFloat(-1.5f, 1.5f))
```
ajrb, line 225: Hi, could you explain more about this change please? The JobRand helper class providing basic thread safe access to system random was the best solution I could find when migrating the original code into Jobs.

Why are you changing it here? Is the helper class not compatible with burst?

andrew-raphael-lukasik: > Why are you changing it here? Is the helper class not compatible with burst?
Yes exactly, `System.Random`, being a managed type, is not compatible with Burst

andrew-raphael-lukasik: Before ending with this particular solution I dig into this and run some tests to figure out how to generate random numbers without memory allocation with `Unity.Mathematics.Random`. You can look through these numerical results in my post [here](https://gist.github.com/andrew-raphael-lukasik/b767c82969eb67ae76b6351bd31bb558?permalink_comment_id=4233622#gistcomment-4233622)

PR #2431, opened 2022-09-18
Fix flying enemies fall damage

petchema:
Flying enemies hitting ground (recoil, under paralysis,...) were taking fall damage relative to their spawn location, because their lastGroundedY was never updated.

This can be fixed several ways, in this patch I extend lastGroundedY to flying enemies by updating it as long as flyingFalls is false; It could be renamed say lastAltitudeControlY, the last value of Y when the enemy had control over its altitude.

Forums: https://forums.dfworkshop.net/viewtopic.php?t=5874


Interkarma:
Thank you for fix @petchema!

PR #2433, opened 2022-09-23
Quest L0B30Y09 typo: mars -> marks

deepfighter:
A small typo correction from classic.


breakfastitem:
Seems like mar might be the intended word.

mar
/mär/
verb
3rd person present: mars
impair the appearance of; disfigure.
"no wrinkles marred her face"

deepfighter:
Never heard of this term, but therefore we have native speaker in our community. :) The word didn't sound very common and as the wording of this quests aren't very high-level (like the main quest), I thought this might be an typo. I will close this PR then. Thanks for checking!

PR #2427, opened 2022-09-12
DaggerfallBillboardBatch: mesh data creation process multi-threaded

andrew-raphael-lukasik:
Hi! This PR is part of https://github.com/Interkarma/daggerfall-unity/pull/2416 saga.

# What

I refactored #2416 changes to the `DaggerfallBillboardBatch.cs` to make them more conservative, safe and limited to a single file. It delivers less benefits than original, but will serve as a good foundation for future improvements which has been identified so far.

I know you guys have other priorities now so no worries - take your time. I am posting this now while my mind is still familiar with this code.

# Results

- `CreateMesh` takes less time from the main thread. Before: about `266 ms` (left), after: about `34 [s]` (right):
<p float="center">
    
    
</p>

- `AddItemsAsync` were added to open the door for future decrease in the overhead from the thousands of separate  `AddItem` calls (about an `100 ms`, sometimes more)

- `ProfilerMarker` introduced, to make profiling much more granular and informative. My controversial proposal here is in it's naming convention ie. `____PascalCase` (for methods) and `____camelCase` (for generic markers). I tried different schemes and this was my conclusion as most consistent and readable, but you might have different conclusions. Prefix `___` especially helps with readability as it makes these markers * immediately * apparent and distinguishable from the rest of the code mid-all these busy methods.

PR #2436, opened 2022-10-02
DaggerfallBillboardBatch: mesh data creation process multi-threaded (take 2)

andrew-raphael-lukasik:
#2427 re-uploaded for further consideration

> Hi! This PR is part of https://github.com/Interkarma/daggerfall-unity/pull/2416 saga.
> 
> # What
> 
> I refactored #2416 changes to the `DaggerfallBillboardBatch.cs` to make them more conservative, safe and limited to a single file. It delivers less benefits than original, but will serve as a good foundation for future improvements which has been identified so far.
> 
> I know you guys have other priorities now so no worries - take your time. I am posting this now while my mind is still familiar with this code.
> 
> # Results
> 
> - `CreateMesh` takes less time from the main thread. Before: about `266 ms` (left), after: about `34 [s]` (right):
> <p float="center">
>     
>     
> </p>
> 
> - `AddItemsAsync` were added to open the door for future decrease in the overhead from the thousands of separate  `AddItem` calls (about an `100 ms`, sometimes more)
> 
> - `ProfilerMarker` introduced, to make profiling much more granular and informative. My controversial proposal here is in it's naming convention ie. `____PascalCase` (for methods) and `____camelCase` (for generic markers). I tried different schemes and this was my conclusion as most consistent and readable, but you might have different conclusions. Prefix `___` especially helps with readability as it makes these markers * immediately * apparent and distinguishable from the rest of the code mid-all these busy methods.
>

### DIFF
--- a/Assets/Scripts/Internal/DaggerfallBillboardBatch.cs
+++ b/Assets/Scripts/Internal/DaggerfallBillboardBatch.cs
@@ -4,7 +4,7 @@
 // License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
 // Source Code:     https://github.com/Interkarma/daggerfall-unity
 // Original Author: Gavin Clayton (interkarma@dfworkshop.net)
-// Contributors:    
+// Contributors:    Andrzej Łukasik (andrew.r.lukasik)
 // 
 // Notes:
 //
@@ -12,12 +12,11 @@
 using UnityEngine;
 using UnityEngine.Rendering;
 using System.Collections;
-using System.Collections.Generic;
-using System;
-using System.IO;
-using DaggerfallConnect;
-using DaggerfallConnect.Utility;
 using DaggerfallConnect.Arena2;
+using Unity.Profiling;
+using Unity.Collections;
+using Unity.Jobs;
+using Unity.Mathematics;
 
 namespace DaggerfallWorkshop
 {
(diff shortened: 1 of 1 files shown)


andrew-raphael-lukasik:
This recreates the #2427 and incorporates the 2e3f2f7 and 4723886 bug fixes

BadLuckBurt:
I'll update the version I have in my mod accordingly. I haven't run into these issues as I only deal with the static billboards in my nature mod and not the mobile NPCs or animals but still, better safe than sorry :)

andrew-raphael-lukasik:
I haven't seen these either and I tested it for >hour but that is not an sensible excuse. I think I will look into gameplay test automation maybe. Hitting a button and letting your player visit dozens of locations could help catch these kind of errors before pr merge and prevent unnecessary drama.

Interkarma:
Thanks for reworking and resubmitting with fixes @andrew-raphael-lukasik. My apologies for not giving you much time to fix in master before reverting.

I'll merge this one again now so it's in next release for general testing by player base.

andrew-raphael-lukasik:
I hope it works

<p float="center">
    
</p>

Because I haven't unit tested it yet...

Interkarma:
Don't stress it. :) I spend a lot of time testing on my end, including working with other PRs and my own fixes, and we have a robust support community to help catch and communicate edge cases when found by player base.

Usually we catch any problems before release. Otherwise a player will let us know. The main thing for me is we don't release with any big obvious problems that break games (especially saves). Everything else can be hotfixed quickly if needed.

BadLuckBurt:
I've not been ready to release my mod yet hence the silence but as far as unit testing / stress testing goes:

I'm generating 3 billboard batches at the same time for each map pixel and ignoring the default billboard batch that is usually used for the terrain. 

The 3 batches I generate are added as children to the terrain and are further tracked by StreamingWorld's TrackLooseGameObjects functionality, that seems to be the only way to reliably get rid of them as the terrains get shuffled around. 

If my code finds existing billboard batches as children of the terrain, it just Destroys those and creates new ones. I've also had another setup where I just cleared existing ones I found. 

Both approaches seem to work fine without any hits to performance, there's always the inevitable lag but it's really minor compared to what I had happen before. 

I'd be surprised if there are suddenly major issues with your code once it goes live.

Interkarma:
I haven't confirmed yet, but DunnyOfPenwick on forums has raised another warning about native collection not being disposed. This one happen when game closes from inside a building directly after load, so should only be a minor problem and not impact gameplay.

https://forums.dfworkshop.net/viewtopic.php?t=5937

Issue #2428, opened 2022-09-14
Cinematics Closing/Skipping When Running 0.14.4 Via Steam


PhluXx1:
**Describe the issue**
Please be advised I've discovered an issue with Daggerfall Unity when adding it as a non-Steam game and playing it. Cinematics will close at random as if being skipped with a key press. I have replicated this issue and confirmed it does not occur if Daggerfall Unity is launched normally. Initially I thought it was an installation or mod issue but have ruled this out, after troubleshooting the issue I've confirmed it only occurs via Steam. I've also verified that this is not a problem with the Vanilla version of Daggerfall available through the Steam store.

**To Reproduce**
1. Launch Steam.
2. Click Library
3. Bottom left click + icon Add Game
4. Choose Add a Non-Steam Game
5. Browse to DaggerfallUnity.exe and add
6. Once successfully added to steam launch Daggerfall Unity
7. Play game and let cinematic splash screen play it should automatically close at random
8. Create new character and watch first in-game cinematic which should also close at some point.

**Desktop (please complete the following information):**
 - OS: Windows 10 Pro 64bit 21H2
 - Version 0.14.4 BETA

Interkarma:
Hi, please use the forums for troubleshooting personal support problems. Adding DFU as a non-Steam game is a user-side configuration unrelated to the game itself.

The recommended method for setup is using the DaggerfallGameFiles method below.

https://forums.dfworkshop.net/viewtopic.php?p=27538#p27538

PhluXx1:
I'm not requesting support I'm reporting a interoperability bug with your software and the biggest game launcher in the world (your software is still in beta AKA the phase to report such bugs so that you may improve it). Again it's a bug report I don't need support also raising this issue in the forums will yield no results as it's a bug. Whether you want to address the bug or not is entirely your own prerogative however. Thank you for your reply.

petchema:
Currently mouse clicks any keyboard event (even pressing modifier keys like Alt(*)) will abort cinematics; If this problem happens without any interaction with the keyboard, then I assume the Steam overlay generates phantom events, or otherwise messes with events handling.
Filtering what keys can interrupt cinematics could help, assuming that's the only problem overlay causes.

(*). I know this because sometimes Twitch streamers want to adjust volume or otherwise switch to another app while cinematics are playing, but Alt-Tab immediately interrupt playback

jefetienne:
[...]

PR #2437, opened 2022-10-04
Exclude modifier keys from being used in AnyKey methods

jefetienne:
From the discussion of #2428.

Originally I was going to go with a whitelist of keys to be accepted, but since there isn't any formalized controller default buttons yet, I decided instead to create a list of unaccepted keys.

### DIFF
--- a/Assets/Scripts/Game/InputManager.cs
+++ b/Assets/Scripts/Game/InputManager.cs
@@ -51,6 +51,24 @@ namespace DaggerfallWorkshop.Game
         const float controllerCursorHorizontalSpeed = 900.0f;
         const float controllerCursorVerticalSpeed = 900.0f;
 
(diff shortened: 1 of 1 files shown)


Interkarma:
My preference would be to set `bool endOnAnyKey = false;` by default in `DaggerfallVidPlayerWindow`. We already do this for really important videos like special infections. I'd prefer this over adding more complexity to what should be a simple change. Players can use their bound back-key / escape to cancel videos.

I know this changes habits for people used to clicking through videos with mouse. But they already can't click through several videos, and this would just standardise the behaviour.

jefetienne:
That sounds okay with me. Would adding at least the spacebar seem okay to you?

Interkarma:
Yeah, I'd be fine with spacebar also skipping videos.

PR #2438, opened 2022-10-08
Prevent ending on any key by default for video playback

jefetienne:
#2437

Set endOnAnyKey to false by default in DaggerfallVidPlayerWindow, but allow spacebar and Escape/back buttons.


Interkarma:
Thanks for update @jefetienne. :) Let's roll this into live and see what the feedback is like.

PR #2441, opened 2022-10-09
Improved Mouse Look Camera Smoothing

AverniteDF:
This submission contains a reworking of the camera smoothing code located in PlayerMouseLook.cs. The purpose of this reworking is primarily to increase behavioral parity between the "smoothing" and "no smoothing" modes. Moving the mouse an inch in a certain direction should give identical yaw/pitch deltas (ultimately) for both modes, it's just that one mode is progressive and the other is immediate.

Note: To increase or decrease the camera smoothing effect you may edit the following value (located near top of class):

float smoothing = 0.5f;

For example, a value of 0.3f will give a very subtle smoothing effect and a value of 0.8f will give a very pronounced smoothing effect (at the cost of less responsiveness/higher latency). I have chosen a value of 0.5f because I feel it is suitable for production. It is possible that the "Mouse Smoothing" checkbox could be replaced with a slider ranged 0.0 to 0.9, in future releases.


Interkarma:
Something I like about this change is the mouselook sensitivity feels equivalent with smoothing on or off. In the current implementation, mouselook sensitivity must be set differently based on smoothing.

This also raises a minor issue. Anyone with current smoothing on will experience a slower-feeling mouselook sensitivity once change goes live. Multiplying `sensitivityX` and `sensitivityY` by 2.5 - 3.0 lands about the same "feel" as the current mouselook sensitivity 1.0. But not sure adding magic numbers is the best approach. Probably best just to increase default mouselook sensitivity and try our best to communicate change so existing users retune mouselook sensitivity.

Interkarma:
>One idea would be to inform players with an in-game popup message.

No need for anything like that. I'll communicate change via patch notes and socials, then support anyone who doesn't get the memo.

Please don't make  style-only changes to existing code such as `PitchMin` to `PITCH_MIN`. I prefer contributors to match the style of original author and only make changes with a clear functional purpose.

Stripping this PR down to bare essentials of improved smoothing calculation, what's the minimum number of lines changed could this be accomplished in?

AverniteDF:
Understood, Gavin. I'll look at putting together a minimum change version
later today.

AverniteDF:
Minimal changes version is now up.

AverniteDF:
Code changes for mouse smoothing slider are now up.

The enable mouse smoothing checkbox has been eliminated and the
MouseLookSmoothing field in settings.ini is now a float.

DFU 0.14.5+ should probably examine that field to see if it has a boolean
value. If so, then False would import as 0.0 and True as 0.5, or some other
value you think is appropriate.

It could have been left for backwards compatibility and a different field
name chosen for the slider such as MouseLookSmoothingFactor. What do you
think?

Interkarma:
SettingsManager will sync with player's live settings by removing old values and adding new values based on defaults. The best approach is simply to add the new setting with a different name to defaults.ini.txt and remove the old setting. There's no need to retain old values for backwards compatibility, this will be automatic if player runs an older version of the game.

Something like `MouseLookSmoothingStrength` or similar will do the job for new value.

Thanks for your PR. I'll review when I can. It won't make it into 0.14.5, as that version is already in testing. I'll aim to get it into 0.14.6.

Assets/Scripts/Game/PlayerMouseLook.cs
```diff
@@ -4,9 +4,9 @@
 // License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
 // Source Code:     https://github.com/Interkarma/daggerfall-unity
 // Original Author: Gavin Clayton (interkarma@dfworkshop.net)
-// Contributors:    
+// Contributors:    Avernite (avernite@gmail.com)
 // 
-// Notes:
+// Notes:           Look-smoothing code changed in 0.14.5 (Avernite)
```
Interkarma, line 9: No need to add this note here (that's what git is for). Also, this change won't be in 0.14.5. :)

Assets/Scripts/Game/PlayerMouseLook.cs
```diff
@@ -86,6 +97,76 @@ public float PitchMinLimit
...
+        void ApplyLook()
+        {
+            // Scale sensitivity
+            float sensitivityX = 1.0f;
+            float sensitivityY = 1.0f;
+
+            if (InputManager.Instance.UsingController)
+            {
+                // Make sure it keeps consistent speed regardless of framerate
+                // Speed = speed * 60 frames / (1 / unscaledDeltaTime) or speed * 60 * unscaledDeltaTime
```
Interkarma, line 118: I was wondering what happens if user has a high refresh monitor (e.g. 120fps or higher)? Does the new smoothing method still scale to very high refresh rates?

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallUnitySetupGameWizard.cs
```diff
@@ -423,13 +423,30 @@ HorizontalSlider AddSlider(float x, string key, int selected, params string[] ch
...
+            int min = Mathf.RoundToInt(minValue * 10);
+            int max = Mathf.RoundToInt(maxValue * 10);
+
+            int numValues = max - min + 1;
+
+            string[] choices = new string[numValues];
+
+            --min;
+
+            for (int i = 0; i < numValues; ++i)
```
Interkarma, line 444: Can you explain this to me? There should be no need to add multiple choices here, the slider control can do 0.0 to 0.9 in 0.1 increments if that's what you're aiming for?

Assets/Scripts/SettingsManager.cs
```diff
@@ -443,7 +443,7 @@ public void LoadSettings()
             EnableSpellShadows = GetBool(sectionSpells, "EnableSpellShadows");
 
             InvertMouseVertical = GetBool(sectionControls, "InvertMouseVertical");
-            MouseLookSmoothing = GetBool(sectionControls, "MouseLookSmoothing");
+            MouseLookSmoothing = GetFloat(sectionControls, "MouseLookSmoothing");
```
Interkarma, line 446: Don't forget to enforce valid range when reading a setting like this.

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallUnitySetupGameWizard.cs
AverniteDF: Just a bit of shoe-horning. Has been reworked to instantiate proper float slider.

Assets/Scripts/Game/PlayerMouseLook.cs
AverniteDF: Note removed.

Assets/Scripts/SettingsManager.cs
AverniteDF: Got it, thanks.

Assets/Scripts/Game/PlayerMouseLook.cs
AverniteDF: Yes. But you don't have to take my word for it!

I've added a block of test code so you can try it out yourself (if you wish).

Pressing LEFT CTRL key activates the test.

Here's what the test code does:
1. Sets smoothing to a high value (0.99f) to give an exaggerated, more drawn-out effect
2. Levels pitch, then sets lookCurrent yaw to 0° and lookTarget yaw to 90°
3. Monitors lookCurrent and finishes test when lookCurrent yaw reaches 60° (two-thirds of the way to lookTarget)
4. Displays a popup showing how much time elapsed and what the average FPS were over that period

Once you've run the test you've got a reference point.

There are three ways of getting additional reference points for comparison:

1. Change your display's physical refresh rate or FPS limit and repeat the test (time should be the same)
2. Uncomment the "hang time" code at the bottom of the test function to lower FPS (time should be the same)
3. *Simulate* a different frame rate by adding a multiplier to Time.unscaledDeltaTime in the PlayerMouseLook.GetFrameRateScaledFractionOfProgression method

For the third option:

a) If you doubled delta time (multiplied by 2) then the test should take half as long to complete (simulating half frame rate so do twice as much per update)
b) If you halved delta time (multiplied by 0.5) then the test should take twice as long to complete (because you're simulating 120 FPS but are actually only running at 60 FPS)

So:

`float frames = Time.unscaledDeltaTime * 60f;`

Becomes something like:

`float frames = Time.unscaledDeltaTime * 2 * 60f;`

If the times come out as expected then the scaling is correct.

Results using option #2:

test_time_at_55_fps
test_time_at_47_fps
test_time_at_34_fps
test_time_at_22_fps

You asked specifically about higher FPS (120+) and the results check out using option #3.

AverniteDF: Simulating 120 FPS at 60 FPS (time should be doubled):

test_time_simulated_120_fps

Simulating 300 FPS at 60 FPS (time should be five times longer):

test_time_simulated_300_fps

That last result is slightly higher than the expected 9.25 sec but still quite acceptable.

Interkarma:
This is looking good so far, thank you for all the work on this one. :)

I'll aim to complete review and have it merged ahead of next release. I'll let you know if any issues are found.

Assets/Game/Addons/UnityConsole/Console/Scripts/DefaultCommands.cs
```diff
@@ -864,7 +864,7 @@ public static string Execute(params string[] args)
         }
 
 
-        private static class ToggleMouseSmoothing
+        /*private static class ToggleMouseSmoothing
```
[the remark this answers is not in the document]
Interkarma: Fair point. I'm rather guilty of this myself. :)

PR #2442, opened 2022-10-11
Test around; find out

andrew-raphael-lukasik:
# What

A foundational setup to execute automated gameplay (play mode) tests. It includes few test examples to help everyone figure out how to write their own test cases.

Test will fail if an exception is thrown anywhere. So it won't catch all the types of errors by itself. For the exception-less cases place `Assert.IsTrue()` to make sure the code throws when some specific data is outside of valid range.

One important requirement of this setup is an [assembly definition](https://github.com/andrew-raphael-lukasik/daggerfall-unity/blob/CASE17_test-around-find-out/Assets/DFU.asmdef) that includes the game code. This means that the game code wont go into `GameAssembly.dll` (or something like that) but into `DFU.dll` now.

# Why

Merging a branch and then realizing it broke something that worked perfectly until now is not a productive way to merge branches (case of #2427 being an inspiration here). But there is a way to prevent that without sinking hours on testing everything manually every time - enter *Unit Tests*.

# Results

With a single click and bit of patience (<5min leaving Unity do it's thing) you can, for example, visit 100 locations now:

Screenshot 2022-10-11 143526

PR #2445, opened 2022-10-13
added: `StreamingWorld` unregisters events on `OnDestroy()`

andrew-raphael-lukasik:
# What

`OnDestroy()` added so `StreamingWorld` can unregister it's events. That's all.

# Why

`StreamingWorld` registers events (`+=`) on `Awake` but was never unregistering them back (`-=`). This was probably never an issue until #2442 where `DaggerfallUnityGame` scene can be loaded and unloaded back and forth for testing purposes.

# Results

`StreamingWorld` cleans up after itself.


Interkarma:
This change looks to be duplicated in #2447.

Dezzj (commented):
Veiwed

PR #2446, opened 2022-10-13
added: `ModManager.OnDestroy()` to unload loaded asset bundles

andrew-raphael-lukasik:
# What

 `OnDestroy()` added to unload asset bundles from `List<Mod> mods;`. That's all.

# Why
#2442, for better or worse, requires that all scripts clean up after themselves

# Results

Destroying a `ModManager` will unload asset bundles associated with loaded mods.

### DIFF
--- a/Assets/Game/Addons/ModSupport/ModManager.cs
+++ b/Assets/Game/Addons/ModSupport/ModManager.cs
@@ -158,6 +158,12 @@ namespace DaggerfallWorkshop.Game.Utility.ModSupport
             }
         }
 
+        void OnDestroy()
+        {
+            foreach (Mod mod in mods)
+                mod.UnloadAssetBundle(unloadAllObjects: true);
+        }
+
         #endregion
 
         #region Public Methods


Assets/Game/Addons/ModSupport/ModManager.cs
```diff
@@ -158,6 +158,12 @@ void Start()
             }
         }
 
+        void OnDestroy()
+        {
+            foreach (Mod mod in mods)
+                mod.UnloadAssetBundle(unloadAllObjects: true);
+        }
```
andrew-raphael-lukasik, line 165: This line needs to be added here in `OnDestroy()`. It makes sense to add this line from #2447 but that could cause a merge conflict between these 2 PR
```
void OnDestroy()
{
    StateManager.OnStateChange -= StateManager_OnStateChange;
    /* code this PR adds goes here */
}
```
I am writing this down here to not forget about it

Interkarma:
I don't expect this to create any issues as it should only happen when game is shutting down. I don't believe ModManager will trigger this when resetting scene (e.g. after player death).

@KABoissonneault @ajrb Can you think of any problems this simple PR might cause? Thank you. :)

ajrb:
I can't see any issues, but I know next to nothing about the intricacies of the mod support asset bundle usage so it's not a very informed opinion. :) 

Would be nice to have TheLacus's opinion on this one, but not seen anything from them for a long time..

Dezzj (commented):
Veiwed

KABoissonneault:
Gonna close this, to keep the repo clean. I'm not against someone reviving the testing / performance work that was being done here, but without author activity, I don't see myself merging this. We can always reopen later

PR #2447, opened 2022-10-13
RAII compliance for all static events

andrew-raphael-lukasik:
# What

I scanned the entire codebase for classes that register `static event`s `+=` but never bother to unregister them `-=` once class is destroyed etc. and made sure they do that.

Screenshot 2022-10-13 233742

# Why

1. I need this for #2442
2. I am 100% positive this caused errors in the past and will cause them in the future, if not fixed

At first I needed this for test setup only. But after digging into the codebase and realizing how prevalent this issue is I convinced myself that this is serious.

I am definitely no fan of programming dogma but RAII makes sense even if we call it differently or none at all. Making sure that resource is allocated (here: event registered) on object construction and then deallocated (here: event unregistered) makes sense on a very practical level - there are always errors somewhere eventually if we don't fancy doing that.

# Results

This PR spans across many files over different domains unfortunately so it will require some additional attention to make sure this wont break anything when merged. If you read the code changes you can see that these are not invasive changes, thankfully. All are either `OnDestroy` or destructor additions - so are vastly safer that most code changes as they won't break existing code execution loops (these calls are managed by Unity engine and CLR).

### DIFF
--- a/Assets/Game/Addons/UnityConsole/Console/Scripts/ConsoleController.cs
+++ b/Assets/Game/Addons/UnityConsole/Console/Scripts/ConsoleController.cs
@@ -34,6 +34,11 @@ namespace Wenzil.Console
             DaggerfallWorkshop.Game.InputManager.OnSavedKeyBinds += GetConsoleKeyBind;
         }
 
+        void OnDestroy()
+        {
+            DaggerfallWorkshop.Game.InputManager.OnSavedKeyBinds -= GetConsoleKeyBind;
+        }
+
         void OnEnable()
         {
             DaggerfallWorkshop.Game.InputManager.OnLoadedKeyBinds += GetConsoleKeyBind;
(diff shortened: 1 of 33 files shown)


Interkarma:
While the majority of these are singleton controller objects that persist through entire runtime, I agree it's a good idea to unregister events when object is destroyed. You also found a few cases of non-singleton objects (e.g. entities) where this was missing. I appreciate the catch, thank you. :)

I don't believe this will cause any issues, but I'll do my best to uncover any before merging. Also, I recall some of these objects are tagged as "do not destroy" and are intended to persist forever. I'm not sure if OnDestroy will even be called in those cases. I'll unpack it all later as part of review.

Dezzj (commented):
Veiwed

KABoissonneault:
Gonna close this, to keep the repo clean. I'm not against someone reviving the testing / performance work that was being done here, but without author activity, I don't see myself merging this. We can always reopen later

Honestly I do appreciate what this PR does on a code cleanliness principle, I just can't bring myself to test it for unknown gains

PR #2448, opened 2022-10-14
Prevent missile collision from moving entity

numidium:
In some cases spell or arrow missiles could cause their targets to bump over them as mentioned in the code comment. Setting the collider to trigger and the rigid body to kinematic at start seemed to fix this without causing any other issues in my tests.

### DIFF
--- a/Assets/Scripts/Game/DaggerfallMissile.cs
+++ b/Assets/Scripts/Game/DaggerfallMissile.cs
@@ -184,10 +184,12 @@ namespace DaggerfallWorkshop.Game
             // Setup collider
             myCollider = GetComponent<SphereCollider>();
             myCollider.radius = ColliderRadius;
+            myCollider.isTrigger = true;
 
             // Setup rigidbody
             myRigidbody = GetComponent<Rigidbody>();
             myRigidbody.useGravity = false;
+            myRigidbody.isKinematic = true;
 
             // Use payload when available
             if (payload != null)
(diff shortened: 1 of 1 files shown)


Interkarma:
There is unfortunately another issue with triggers and static geometry. The way it's handled now is a compromise between two different issues.

https://github.com/Interkarma/daggerfall-unity/commit/a953155f81c7809c398672d90f20ce667742c60f

Admittedly, this was a few Unity versions ago. The current version of Unity might not exhibit the original problem prompting original change from a trigger collider. Also the original setup was to configure as a trigger collider in prefab (not in code) so that might make a difference also.

I'll work back through my memory and history, and try to ensure I'm not re-introducing and old problem with this change.

Thanks for the suggestion. :)

numidium:
I've done some more testing and now I think I see what you mean. If I hug a wall and shoot either an arrow or spell missile then it sometimes goes right through depending on the angle I'm facing. I got the arrows to work correctly by spawning them directly on the player and ignoring trigger collisions with the caster but the problem persists with spells. Still looking into it.

Interkarma:
Yeah, it caused me a lot of grief a few years back. I recall working through a lot of different solutions that just had other problems. I'm happy with the current setup as it has had no serious issues for years now.

When it comes to bumping enemies, there's two things that could be happening. Physics is one, and obstacle avoidance in AI is another. Might be worth disabling Advanced AI to check what you're trying to fix is actually physics and not steering.

numidium:
> When it comes to bumping enemies, there's two things that could be happening. Physics is one, and obstacle avoidance in AI is another. Might be worth disabling Advanced AI to check what you're trying to fix is actually physics and not steering.

I've had the same thing happen to the player entity when enemies cast missiles at me so I don't think it's AI. I believe the solution is to somehow have all mobile entities simply ignore missiles so they can't "climb" over them. The missiles can handle the collision on their end. There's probably a setting in Unity for this but I'm still not great with the engine despite working on this project for a couple years. Your code has mostly kept me insulated from its quirks (which I appreciate).

Interkarma:
From memory - the issue with tiggers for missiles was related to the missile being consumed when encountering other trigger volumes. I'll need to go back through the history to see what the original issue was being correct and regression test against this change to ensure we're not re-introducing another bug that was fixed years ago.

Interkarma:
Unfortunately this change does reintroduce the collision problem outlined above. It won't be an appropriate fix, I'm sorry.

PR #2449, opened 2022-10-17
Fix exceptions when bashing certain doors

numidium:
Leverages ActivateStaticDoor when bashing static doors so all door types are handled. Currently, only exterior doors in settlements are handled. This mimics classic behavior with the exception that dungeon exterior doors can be bashed too.

### DIFF
--- a/Assets/Scripts/Game/PlayerActivate.cs
+++ b/Assets/Scripts/Game/PlayerActivate.cs
@@ -490,7 +490,8 @@ namespace DaggerfallWorkshop.Game
             DFLocation.BuildingTypes buildingType,
             bool buildingUnlocked,
             int buildingLockValue,
-            Transform doorOwner)
+            Transform doorOwner,
+            bool isBash = false)
         {
             StaticDoor door;
             if (CustomDoor.HasHit(hit, out door) || (doors && doors.HasHit(hit.point, out door)))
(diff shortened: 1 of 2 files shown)


Interkarma:
This looks like a rather large refactor just to fix an exception. I'd prefer to see fewer lines touched.

What's the actual issue being fixed here? If you can give me a repro save, I might like to handle this a different way.

numidium:
Sorry, I could have explained better. Right now you will get exceptions thrown when either bashing a building interior door or dungeon exterior door. In addition, it doesn't transition the player through the door.

I'm letting ActivateStaticDoor handle the static door transitions for bashing since most of the logic is already there. I added a little extra logic to handle bashing but github's diffing went a little wild. If you look at what was "removed" vs what was "added" it is largely the same with a few exceptions.

I zipped a couple of saves with a character standing next to doors that will yield exceptions when you bash them.
[Saves.zip](https://github.com/Interkarma/daggerfall-unity/files/9797018/Saves.zip)

Interkarma:
Thank you for save. I'll look into the issue when I can. :)

Dezzj (commented):
Veiwed

KABoissonneault:
This error is still being hit by users. After fixing the public API changes, I'd be willing to merge this

numidium:
There was one public method that I renamed. Its name is now reverted to what it was originally per Kab's request.

Assets/Scripts/Game/WeaponManager.cs
```diff
@@ -467,8 +467,8 @@ public bool WeaponEnvDamage(DaggerfallUnityItem strikingWeapon, RaycastHit hit)
                 return true;
             }
 
-            // Check if player hit a static exterior door
-            if (GameManager.Instance.PlayerActivate.AttemptExteriorDoorBash(hit))
+            // Check if player hit a static door
+            if (GameManager.Instance.PlayerActivate.AttemptStaticDoorBash(hit))
```
KABoissonneault, line 471: I think you forgot to revert this change too?

numidium: Hmm, looks like something went wrong when I was merging. It's fixed now.

Assets/Scripts/Game/PlayerActivate.cs
```diff
@@ -502,70 +503,92 @@ void Update()
...
+                        dfAudioSource.PlayOneShot(SoundClips.PlayerDoorBash);
+
                 if (door.doorType == DoorTypes.Building && !playerEnterExit.IsPlayerInside)
                 {
                     // Discover building
                     GameManager.Instance.PlayerGPS.DiscoverBuilding(building.buildingKey);
 
                     // Handle clicking exterior door with Open spell active
-                    if (HandleOpenEffectOnExteriorDoor(buildingLockValue))
+                    if (HandleOpenEffectOnExteriorDoor(buildingLockValue) && !isBash)
```
KABoissonneault, line 517: I think you put the check in the wrong order. Currently, a bash will consume an incumbent Open effect, but will not activate it. If you rewrite it as `!isBash && HandleOpenEffectOnExteriorDoor(buildingLockValue)`, then a bash will neither consume nor activate the Open effect.

Assets/Scripts/Game/PlayerActivate.cs
```diff
@@ -502,70 +503,92 @@ void Update()
                     return;
                 }
 
+                // Play sound when bashing.
+                if (isBash && door.doorType != DoorTypes.DungeonExit)
+                    if (TryGetComponent<DaggerfallAudioSource>(out var dfAudioSource))
+                        dfAudioSource.PlayOneShot(SoundClips.PlayerDoorBash);
+
                 if (door.doorType == DoorTypes.Building && !playerEnterExit.IsPlayerInside)
```
KABoissonneault, line 511: Might be cleaner if we just skipped this whole if on a bash.
`door.doorType == DoorTypes.Building && !playerEnterExit.IsPlayerInside && !isBash`
The Discovery part does not seem necessary, you seem to have found how to get the lock level without it

Assets/Scripts/Game/PlayerActivate.cs
KABoissonneault: I've also noticed another DFU issue here: by making the door "unlocked", you enable the "greeting" message, despite having bypassed the lock. This is inconsistent with lockpicking, which moves forward down the function without unlocking the door.

The check should probably be rewritten as `if(!buildingUnlocked || !HandleOpenEffectOnExterior(buildingLockValue)` (assuming you still have the `!isBash` check above), so that `buildingUnlocked` remains false, and the greeting message will be avoided like in the lockpicking case.

KABoissonneault:
After having run tests, the changelist mostly works as advertised, outside the Open issue I've mentioned. 

I've run tests on classic DF, and here's what I've concluded from bashing doors:
- Trying to bash the exterior door from the inside causes you to warp inside as if you'd just broken in from the outside, with possibility of guards. I'm fine with the approach this PR takes instead
- Trying to bash internal doors always succeeds if they are unlocked, and will not trigger guards, not even in a dungeon palace like Wayrest. The sound is not bash, but just a normal open
- Dungeon doors are not bashable, not even a sound

I'm fine with DFU diverging on point 1 and 3. I know this changelist is about external doors, but while we're on it, I think we should update `DaggerfallActionDoor.AttemptBash` to 1) always open door if the lock value is 0 2) not make enemies hostile if the lock value is 0. Note that closing the door currently behaves properly, save for the alerting enemies part.

Summary of my recommandations:

Must:
- Fix so that Open is not consumed when bashing

Nice to have:
- Bit of cleanup around `isBash` checks
- Remove the Greeting for Open
- Bashing both external and interior doors should always work if the door has no lock
- Bashing unlocked interior doors should not make Palace guards hostile (see Wayrest). Bashing locked interior doors should still always alert (see Daggerfall)

KABoissonneault (dismissed):

numidium:
OK, should be ready for merging now.

KABoissonneault (approved):

PR #2450, opened 2022-10-19
added: `DaggerfallStaticDoors.OnDrawGizmosSelected()` to debug door trigger placements

andrew-raphael-lukasik:
# What

> This is small and unimportant PR. You can close it rn if you see it of no worth to the project.

Moves commented out "Debug trigger placement at start" code block from `Start()` and puts it into `OnDrawGizmosSelected()`

# Why

I needed to see where the door triggers were... the code was ready written already, so.

# Results

Door gizmos shows up when `DaggerfallStaticDoors` instance is selected only, so it wont pollute the screen nor slow down the editor everywhere. `#if UNITY_EDITOR` makes sure this won't ever end up in a build.

Screenshot 2022-10-18 232154

### DIFF
--- a/Assets/Scripts/Internal/DaggerfallStaticDoors.cs
+++ b/Assets/Scripts/Internal/DaggerfallStaticDoors.cs
@@ -32,26 +32,6 @@ namespace DaggerfallWorkshop
 
         void Start()
         {
-            //// Debug trigger placement at start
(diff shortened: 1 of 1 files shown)


Interkarma:
I think this is fine. I only used that block of code when developing door system and left it behind in case I needed it later, but I don't see any harm in showing these info as gizmos in editor.

PR #2452, opened 2022-10-24
Allow access to tagged spells even without a spellbook

petchema:
Fix would be relatively simple if DaggerfallSpellBookWindow.spellListBox and PlayerEntity.spellbook didn't use the same indices.
To be able to only show a subset of the known spells, some index mapping (spellBookIndices list) has been added.

Forums: https://forums.dfworkshop.net/viewtopic.php?t=5924

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallSpellBookWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallSpellBookWindow.cs
@@ -240,6 +240,12 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
                 {
                     spellBookIndices = new List<int>();
                     PopulateSpellsList(spellbook.ToList(), GameManager.Instance.PlayerEntity.CurrentMagicka);
+                    // in case a mod overrode PopulateSpellsList() with a method that doesn't update spellBookIndices
+                    if (spellBookIndices.Count == 0)
+                    {
+                        for (int i = 0; i < spellsListBox.Count; i++)
+                            spellBookIndices.Add(i);
+                    }
                 }
             }
 
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallSpellBookWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallSpellBookWindow.cs
@@ -108,6 +108,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
         protected EffectBundleSettings renamedSpellSettings;
         protected int deleteSpellIndex = -1;
         protected KeyCode toggleClosedBinding;
+        protected List<int> spellBookIndices = null;
         protected List<EffectBundleSettings> offeredSpells = new List<EffectBundleSettings>();
         protected PlayerGPS.DiscoveredBuilding buildingDiscoveryData;
         protected int presentedCost;
@@ -222,6 +223,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
 
             // Clear existing list
             spellsListBox.ClearItems();
+            spellBookIndices = null;
 
             // Add spells based on mode
             if (buyMode)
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Game/UserInterfaceWindows/DaggerfallSpellBookWindow.cs
```diff
@@ -253,32 +255,43 @@ protected virtual void RefreshSpellsList(bool preservePosition)
             }
         }
 
-        protected virtual void PopulateSpellsList(List<EffectBundleSettings> spells, int? availableSpellPoints = null)
+        protected virtual List<int> PopulateSpellsList(List<EffectBundleSettings> spells, int? availableSpellPoints = null)
```
KABoissonneault, line 258: I'm not too keen of changing the interface of `protected` access methods. It will break mods who override this function, which is not something DFU is meant to do at this stage.
I can't think of another way to do this without that, except by letting `PopulateSpellsList` set `spellBookIndices` from within, and have all the other methods check `spellBookIndices` for null, in case a mod replaced this function without updating

petchema: Does such mod exist? Because the implicit invariant that generated spell listbox must use the same indices as the spellbook seriously limited the usefulness of overriding that method.

I did not modify spellbookIndices directly because that method is also used in buymode == true, when populating spellbookIndices would make no sense. But I'm not particularly fond of the solution I found either.

petchema: Should be better now

KABoissonneault (approved):

Interkarma:
With this change, player drops a spellbook and happens to have tagged spells, they see tagged spells but can't access all of them. They probably won't understand the problem or what they have to do to fix. At least with the current setup, they see the message "you don't have a spellbook".

My original setup was to disallow player from dropping their spellbook. The community steered me around to making this work like classic, which it now does.

I'm not sure I prefer this solution over not dropping the spellbook or simply doing away with it altogether. It's just an item that allows access to a UI, it's not like each spellbooks has different spells or properties. The whole book could be removed and it wouldn't impact the gameplay, just the fantasy element of holding an item.

Interkarma:
I disagree with this being tagged as a bug. This is how classic operates and is working as intended in DFU.

It should be considered an enhancement or QoL feature that varies from classic. As such, the best solution may be something totally different.

petchema:
What was a bug in my mind was preventing lycanthropes and vampires from accessing their hard owned capabilities because they didn't have a spellbook at hand

petchema:
Okay, so in classic too, if you drop your spellbook you cannot use your racial capabilities.
Damn, so weird. When I saw this case in a stream that really felt like a bug. But at least it's iso with classic., my bad

Interkarma:
For my part, I fully agree this shouldn't happen. Let me think about it a bit and we'll find a solution. :)

petchema:
Just a remark,I think it would better to have another flag ("intrinsic"?) on spells rather than (ab)using Tag.
Tag is just used to be able to track subsets of spells, say remove them when you're cured, but doesn't in itself say anything about the need for a spellbook to be cast. Even if today we decide to make all existing tagged spells intrinsic.

Interkarma:
I'm really not abusing the Tag field, I put it there specifically for this purpose. :) Tag can also communicate if intrinsic: `if (tag == vampireSpellTag || tag == lycanthropySpellTag)` then it's intrinsic.

I agree it would be nice to solve the problem you've raised as a QoL change. Something I'd consider is adding vampire/lycanthropy spells to the U key popup along with magic items. My reasoning:

* This is a spellcasting UI that's always available regardless of spellbook status
* UI could be easily refactored to allow casting spells from both magic items and innate abilities. When it comes to innate abilities, the player's own body is essentially a magic item.
* The QoL behaviour could be disabled from settings.ini for purists.

What are your thoughts on this alternate approach to solving core problem?

petchema:
Yeah, alternate approach to disconnect such capabilities from spellbook is to bind them to another key, a la Skyrim shouts, etc. I didn't think about gathering it with U / magic items, but why not. Not exactly the same thing, but it has the advantage of not requiring a new key binding.
Main criticism of all those alternate approaches is that they're departures from classic; Old-timers won't find lycanthropy where they expect it.

Interkarma:
Yeah, I mean both in spellbook and (optionally) available in U list as a default QoL feature. Purists can disable option if they don't want native powers to be visible in U list.

Happy you find that a good approach. :)

PR #2454, opened 2022-11-01
Immediate Ambient Light Updating To Prevent Flicker On Transition To …

AverniteDF:
…Exterior

The purpose of this PR is to eliminate the glitchy flicker you sometimes get when exiting a building.

What the world outside looks like for a few frames before lighting gets updated:
transition_to_exterior_first_few_frames

And after:
transition_to_exterior_after

Immediate calls to PlayerAmbientLight.UpdateAmbientLight solves this.

The proposed code calls that method twice: once in the current frame and again on the next. The combination of the two calls completely eliminates all flickering as far as I can tell (I disabled HUD fade in so I could see every frame).

### DIFF
--- a/Assets/Scripts/Game/PlayerAmbientLight.cs
+++ b/Assets/Scripts/Game/PlayerAmbientLight.cs
@@ -48,6 +48,7 @@ namespace DaggerfallWorkshop.Game
         
         private void OnTransitionToExterior(PlayerEnterExit.TransitionEventArgs args)
         {
+            sunlightManager.Update(); // Ensure that SunlightManager is in a ready state
             UpdateAmbientLight();
         }
 
@@ -73,7 +74,7 @@ namespace DaggerfallWorkshop.Game
             }
             else if (playerEnterExit.IsPlayerInside && !playerEnterExit.IsPlayerInsideDungeon)
             {
-                if (DaggerfallUnity.Instance.WorldTime.Now.IsNight && !playerEnterExit.IsPlayerInsideTavern) // Interiors of taverns are always welcomingly bright
+                if (DaggerfallUnity.Instance.WorldTime.Now.IsNight)
                     targetAmbientLight = (DaggerfallUnity.Settings.AmbientLitInteriors) ? InteriorNightAmbientLight_AmbientOnly : InteriorNightAmbientLight;
                 else
                     targetAmbientLight = (DaggerfallUnity.Settings.AmbientLitInteriors) ? InteriorAmbientLight_AmbientOnly : InteriorAmbientLight;
--- a/Assets/Scripts/Game/PlayerAmbientLight.cs
+++ b/Assets/Scripts/Game/PlayerAmbientLight.cs
@@ -39,14 +39,16 @@ namespace DaggerfallWorkshop.Game
         Color targetAmbientLight;
         bool fadeRunning;
 
-        public static PlayerAmbientLight Instance = null;
-
         void Start()
         {
-            Instance = this;
-
+            PlayerEnterExit.OnTransitionExterior += OnTransitionToExterior;
             playerEnterExit = GetComponent<PlayerEnterExit>();
             sunlightManager = GameManager.Instance.SunlightManager;
+        }
+        
+        private void OnTransitionToExterior(PlayerEnterExit.TransitionEventArgs args)
+        {
+            UpdateAmbientLight();
         }
 
         void Update()
--- a/Assets/Scripts/Game/PlayerAmbientLight.cs
+++ b/Assets/Scripts/Game/PlayerAmbientLight.cs
@@ -38,7 +38,6 @@ namespace DaggerfallWorkshop.Game
         SunlightManager sunlightManager;
         Color targetAmbientLight;
         bool fadeRunning;
-        float timeOfLastAmbientLightUpdate;
 
         public static PlayerAmbientLight Instance = null;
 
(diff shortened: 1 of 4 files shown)


Assets/Scripts/Game/Serialization/SaveLoadManager.cs
```diff
@@ -587,6 +587,8 @@ private IEnumerator RestoreCachedSceneNextFrame(string sceneName)
             yield return new WaitForEndOfFrame();
             // Restore the scene from cache
             stateManager.RestoreCachedScene(sceneName);
+
+            PlayerAmbientLight.Instance.UpdateAmbientLight(true); // Immediately update ambient lighting
```
petchema, line 591: I think Event/Delegate would be better style here (less coupling).

petchema:
I did not notice the 3Hz polling here, that explains some delayed updates...
I wonder if it would be possible to use other mechanisms (events/degates?) to get rid of polling (and its delays vs performance tradeoffs) without introducing coupling, are there best practices in Unity/C# in this domain?

Assets/Scripts/Game/PlayerAmbientLight.cs
```diff
@@ -52,34 +56,49 @@ void Update()
                 StartCoroutine(ChangeAmbientLight());
         }
 
+        bool mutex;
```
petchema, line 59: To ensure that "mutex" changes are immediately seen by all threads, I think it should at very least be declared volatile.
Busy looping is also terrible. If you don't want to go the full Mutex class way, use at least a lock() statement

AverniteDF:
The code given is primitive, yes. I'm not super familiar with C# yet or
multithreading.

It doesn't look like there's anything too expensive in the ambient light
update logic. Could it simply be called every game tick instead of being
run as a coroutine?

AverniteDF:
I've put up a cleaner implementation (no mutex stuff and ManageAmbientLight coroutine has been removed). In current form it still requires the two calls to PlayerAmbientLight.UpdateAmbientLight from other classes, otherwise you get two frames of incorrect lighting after interior->exterior transition.

It is helpful to set FadeBehaviour.allowFade to false when testing this so you can more clearly see what's going on during transition.

AverniteDF:
I think it would be best to call UpdateAmbientLight() in the main thread, as late as possible and immediately before the drawing of each frame.

Are there any suggestions as to where that might be? I tried doing it in LateUpdate() but that didn't seem late enough.

The other option as I see it is to call UpdateAmbientLight() on demand as needed, in which case there'd be no need to call it every frame or periodically.

With the second option, UpdateAmbientLight() would be called after, and whenever, player location or game time is changed.

petchema:
> The other option as I see it is to call UpdateAmbientLight() on demand as needed, in which case there'd be no need to call it every frame or periodically.
> 
> With the second option, UpdateAmbientLight() would be called after, and whenever, player location or game time is changed.

Yes, that's what events/delegates could be used for. It's a publication/subscription mechanism, so code notifying about changes (location, weather,...) would unaware of the PlayerAmbientLight object (and possibly other objects) listening for those notifications; This decreases coupling vs plain method calls.

AverniteDF:
There is an existing event that gets fired at the end of the transition to
exterior logic called RaiseOnTransitionExteriorEvent() that may be of note.

I've further simplified the PR code. Calling UpdateAmbientLight() every
tick in LateUpdate() eliminates the need to call it from
SaveLoadManager.RestoreCachedSceneNextFrame(), so that class no longer
needs to be modified, and now there is just the one call from
PlayerEnterExit.BuildingTransitionExteriorLogic() which could probably be
moved into the event like you suggested.

AverniteDF:
Ah, that *does* work better. The event handler got rid of the need to modify PlayerEnterExit and also the Instance member I added to PlayerAmbientLight.

AverniteDF:
There's a wrinkle in`PlayerEnterExit.BuildingTransitionExteriorLogic()`that I'd like to point out.

It has to do with the last two lines of code in that method:

```
// Update serializable state from scene cache for interior->exterior transition
SaveLoadManager.RestoreCachedScene(world.SceneName);

// Fire event
RaiseOnTransitionExteriorEvent();
```

The call to`RaiseOnTransitionExteriorEvent()`is not quite accurate here because the previous line,`SaveLoadManager.RestoreCachedScene()`, attempts to restore the cached scene on a separate thread.

So other classes are being notified that the transition has completed when in fact it is still pending.

I believe this is the reason why I kept getting a pesky ambient lighting flicker when exiting buildings, and I originally needed a second call to`UpdateAmbientLighting()`at the bottom of`RestoreCachedSceneNextFrame()`.

The`RaiseOnTransitionExteriorEvent()`call could be passed to`RestoreCachedScene()`as a post-Action and this would fire the event upon actual transition completion. However, downstream code in the calling thread may be relying on the event being fired as currently implemented.

It would be nice to eliminate the`RestoreCachedSceneNextFrame()`method, if possible. Find out what needs to be registered and do the scene restoration immediately instead of on a separate thread.

AverniteDF:
Okay I found what was causing the pesky flicker. It wasn't the delayed cached scene restoration. It was happening because `SunlightManager` wasn't in a ready state when `CalcDaytimeAmbientLight()` was being called.

It seems that `SunlightManager.Upate()` isn't called per tick when the player is indoors, or at least not reliably. I know that if you load a character who is inside a building the first time you exit the data members of `SunlightManager` aren't correct until its `Update()` method gets called.

It's working quite well now.

Interkarma:
Thank you for the PR. I know the flicker you're looking at fixing, and appreciate the suggested fix. I also appreciate the refinement after @petchema's review. Cheers both. :)

Dezzj (commented):
Veiwed

KABoissonneault:
Is this still relevant? If the original contributor is still around, I'd like for them to revert the changes PlayerEnterExit.cs and SaveLoadManager.cs, since it's just whitespace. Otherwise, I can give this a run and see if I notice a difference

petchema:
The original issue should still be around, easiest way to spot it is probably to exit a tavern at night, and notice the exterior is bright for a split second before light abruptly adjusts to dark then ramps up to final correct exterior lighting. Effect varies a bit between tries, depending on when the 3hz light adjustment polling takes place vs. your building exit, but it's there.
With this patch when you exit that same tavern the light is immediately dark and ramps up to final correct exterior lighting, no initial flash.
(Whether it's the best way to solve that issue is another story, cleanly handling transient states is hard.)

KABoissonneault (approved):
Finally went around to testing this. I did frame-by-frame pausing, and it does indeed fix the issue reliably. Other light transitions (ex: sleeping in an inn) are still handled.

Issue #2455, opened 2022-11-02
A `MonoBehaviour` is constructed but never enabled so `OnDestroy` is never called


andrew-raphael-lukasik:
This thread will attempt to document a case of: "a `MonoBehaviour` is constructed but end up never being enabled so it's `OnDestroy` is never called". Strap in.

This wouldn't be much of a problem if not for that there is this [DaggerfallBillboardBatch](https://github.com/Interkarma/daggerfall-unity/blob/bb67c52/Assets/Scripts/Internal/DaggerfallBillboardBatch.cs) class that allocate unmanaged (`NativeArray<T>`) memory **AND** you can create a situation (in the Editor) where it's `GameObject` end up never being enabled before exiting the play mode - which in turn causes Unity to never call `OnDestroy` for some unclear reasons (backward compatibility being my guess).

### Repro steps:
- run the current DFU game (master branch @ d855b812c4d0f5702e3949932125c3d661fdf6eb) from an Unity 2019.4.x editor
- start a game in a building (tavern etc) by loading a save
- exit play mode
Result: Console window is flooded with hundreds of `unmanaged memory exception` noises.

### Original report:

> I haven't confirmed yet, but DunnyOfPenwick on forums has raised another warning about native collection not being disposed. This one happen when game closes from inside a building directly after load, so should only be a minor problem and not impact gameplay.
> 
> https://forums.dfworkshop.net/viewtopic.php?t=5937
_Originally posted by @Interkarma in https://github.com/Interkarma/daggerfall-unity/issues/2436#issuecomment-1297843900_
      
This report led me to conclude that this issue arised after [this commit](https://github.com/Interkarma/daggerfall-unity/commit/2e3f2f7e8f3df21647d4e951dbb4a7ad07bcd389) which was put forward by me to address https://github.com/Interkarma/daggerfall-unity/pull/2427#issuecomment-1264545004 bug report. My first intuition was to revert the commit. **BUT** after a while I realized that this would achieve nothing as both cases require unmanaged memory being allocated before object being enabled. This is because other classes in the project (`TerrainNature`, `RMBLayout`) push billboard data on it's own schedule (before `DaggerfallBillboardBatch.Awake()` is called).
This was not a problem when all the memory was managed, but became one once we moved into unmanaged memory domain.

#

### Things I tried that does not work:

```csharp
void OnDestroy()
{
    /* not called */
}
```
```csharp
void OnDisable()
{
    /* not called either */
}
```
```csharp
[...]

PR #2456, opened 2022-11-02
`DaggerfallBillboardBatch` fix for an issue #2455

andrew-raphael-lukasik:
# What:
Calls `OnDestroy()` on `PlayModeStateChange.ExitingPlayMode` (editor only).

# Why:

> I haven't confirmed yet, but DunnyOfPenwick on forums has raised another warning about native collection not being disposed. This one happen when game closes from inside a building directly after load, so should only be a minor problem and not impact gameplay.
> 
> https://forums.dfworkshop.net/viewtopic.php?t=5937
_Originally posted by @Interkarma in https://github.com/Interkarma/daggerfall-unity/issues/2436#issuecomment-1297843900_

Also seen as https://github.com/Interkarma/daggerfall-unity/issues/2455

# Results:
Exceptions are gone. Builds unaffected.

### DIFF
--- a/Assets/Scripts/Internal/DaggerfallBillboardBatch.cs
+++ b/Assets/Scripts/Internal/DaggerfallBillboardBatch.cs
@@ -9,14 +9,15 @@
 // Notes:
 //
 
+using System.Collections;
 using UnityEngine;
 using UnityEngine.Rendering;
-using System.Collections;
-using DaggerfallConnect.Arena2;
+using UnityEngine.Assertions;
 using Unity.Profiling;
 using Unity.Collections;
 using Unity.Jobs;
 using Unity.Mathematics;
+using DaggerfallConnect.Arena2;
 
 namespace DaggerfallWorkshop
 {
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Internal/DaggerfallBillboardBatch.cs
```diff
@@ -214,6 +228,44 @@ IEnumerator AnimateBillboards()
             }
         }
 
+        void AllocateUnmanagedMemory()
+        {
+            ___AllocateUnmanagedMemory.Begin();
+
+            Assert.AreEqual(expected: (byte)0, actual: isUnmanagedMemoryAllocated);
```
andrew-raphael-lukasik, line 235: `Assert`ions are only part of `DEBUG` builds

Assets/Scripts/Internal/DaggerfallBillboardBatch.cs
```diff
@@ -967,6 +1044,10 @@ private void UpdateName()
 
         private bool ReadyCheck()
         {
+            // Get references
+            if (meshRenderer == null)
+                meshRenderer = GetComponent<MeshRenderer>();
+
```
andrew-raphael-lukasik, line 1050: This is totally opportunistic change, hope you don't mind. Logic behind it: `GetComponent` is more costly that null check and `ReadyCheck()` is called repeatedly.

Interkarma: That's no problems, cheers. :)

Interkarma:
Thank you for fix!

PR #2457, opened 2022-11-03
Add missing fireplace audio (burning loop)

AverniteDF:
The code adding burning sound to torches is easily adapted for use with fireplaces too.


Interkarma:
This isn't missing, fireplaces don't make any sound in Daggerfall. :) This is something that can be added via mod system.

PR #2458, opened 2022-11-03
Fix 'unaccepted keys' not being allowed for bindings in control settings

jefetienne:
From Pango's [comment](https://github.com/Interkarma/daggerfall-unity/issues/2428#issuecomment-1301489156).

My apologies for not checking this earlier! Since it does break binding things as simple as "Shift" back to sprint for example, I'm not sure if this warrants a hotfix release. I do know as a workaround, you can simply rebind in the Keybinds.txt file.


Interkarma:
Thank you for the fix!

PR #2459, opened 2022-11-03
Taverns too dim at night when using Simplified Interior Lighting option

AverniteDF:
Taverns are lively places that should be well lit and welcoming all hours of the day.

tavern_dim
Inside of tavern at night when Simplified Interior Lighting is enabled

tavern_lit
How it should look (in my opinion) so players feel like taverns are a safe haven from the night


Interkarma:
Taverns also get darker at night in Daggerfall. When compared in classic:

**Daytime**
image

**Evening**
image

Simplified lighting is currently a pretty good match for classic. I don't want to modify this behaviour as in this PR. People might interpret this as a bug as it no longer matches classic behaviour.

PR #2460, opened 2022-11-04
SwingWeapon can be converted to ActivateCenterObject when weapon is sheathed

AverniteDF:
Either mouse button will then work for opening doors, etc. when weapon is sheathed.

### DIFF
--- a/Assets/Scripts/Game/GameManager.cs
+++ b/Assets/Scripts/Game/GameManager.cs
@@ -835,6 +835,12 @@ namespace DaggerfallWorkshop.Game
                 }
             }
         }
+        
+        // Check if player object exists without invoking GameManager instance creation
+        public static bool PlayerObjectExists()
+        {
+            return instance != null && instance.playerObject != null;
+        }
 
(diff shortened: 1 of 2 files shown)


jefetienne:
I don't have final say, but I think the mechanic might cause some confusion or accidental clicks. If someone is using click to attack, and are used to interacting with the world with the same attack button, then when they forget they have their weapon unsheathed that they may accidentally attack a neutral NPC or accidentally bash a residential door.

I think this should be more of a mod ultimately.

Interkarma:
Daggerfall uses the "activate centre object" key binding to interact with objects such as doors. I think that making the weapon button sometimes work, depending on weapon being sheathed or not, is only going to confuse players.

Please understand, DFU is functionally complete at this stage. I don't want to change appearance or behaviours. Once DFU is past the 1.0 milestone and I see where all this takes me, I'm planning to step away from DFU to work on a new project of my own. Whoever ends up curating DFU moving forwards might be more open to these kinds of changes.

PR #2461, opened 2022-11-05
Scale sky panels according to current fov relative to start-up fov

AverniteDF:
The `DaggerfallSky.skyScale` member needs to be accessible for mods that would like to alter zoom level.

For example, if a modder wanted to apply zoom while bow is drawn and held the skywall backdrop needs to be scaled in addition to modifying the fov for proper zoom effect.

A more compete solution for this would be the addition of an FPSZoom class to provide an obvious API for zoom behavior.

I would be interested in implementing such a class just let me know if there's an interest.

**UPDATE:**

Actually, exposing that field probably isn't necessary.

DaggerfallSky can scale its panels according to fov in a way that doesn't change the way the backdrop looks when no fov modifiers are applied.

This can be done by scaling the panels by the following factor:

```
float relativeZoom = Mathf.Tan((DaggerfallUnity.Settings.FieldOfView * 0.50f) * Mathf.Deg2Rad) / Mathf.Tan((mainCamera.fieldOfView * 0.50f) * Mathf.Deg2Rad);

float width = (int)(Screen.width * skyScale * relativeZoom); // Panel width
float height = Screen.height * skyScale * relativeZoom; // Panel height

```
This works because when main camera fov is the same as game settings (initial) fov you get a `relativeZoom` of 1.0f (no change). The only limitation is on zooming out because the panels get smaller, but zooming in from initial pov is unlimited and accurate.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallAdvancedSettingsWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallAdvancedSettingsWindow.cs
@@ -526,7 +526,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             DaggerfallUnity.Settings.GUIFilterMode = guiFilterMode.ScrollIndex;
             DaggerfallUnity.Settings.VideoFilterMode = videoFilterMode.ScrollIndex;
 
-            DaggerfallUnity.Settings.FieldOfView = fovSlider.Value;
+            GameManager.Instance.MainCamera.fieldOfView = DaggerfallUnity.Settings.FieldOfView = fovSlider.Value;
             DaggerfallUnity.Settings.TerrainDistance = terrainDistance.Value;
             DaggerfallUnity.Settings.ShadowResolutionMode = shadowResolutionMode.ScrollIndex;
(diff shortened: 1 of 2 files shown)


Interkarma:
Hi - there's quite a bit more here than just making a field public for mods. I'm happy to accept just that change, but nothing else.

I'm not interested in camera zooming of any kind, thank you.

AverniteDF:
Hi Gavin,

I apologize if there was any miscommunication. My original PR was for the
field accessibility only, then I changed the PR title and context. Perhaps
I should have made a separate PR for the latter.

/Rob

Interkarma:
Hi Rob, happy to make the field public so this can be accessed by a mod. It's also possible to replace the sky entirely through mod system so it behaves however needed. However, from the base game perspective, the classic sky is already completed. I don't want to risk any regressions or create new support problems.

From my end, I only have limited time to work on DFU now. Unfortunately, I'm stuck in a loop where my limited time is always spent on code review or fixing other people's problems, and not actually working on my milestones for 1.0. I need to be very selective about which changes are accepted to the game until I'm clear of project, otherwise I'll never get it out of my life. Thank you for understanding.

PR #2462, opened 2022-11-08
re-work caching, de-couple gfx change from caching

numidium:
Had to work through multiple issues.
1. SpecificWeapon needed to be set right before the caching routine so proper cache keys can be referenced.
2. Used a pigeon-hole cache instead of a queue for the right/left hands. The queue couldn't cache for both hands in every case.
3. Modified GetWeaponTextureAtlas to not alter the screen weapon's current custom textures. The caching routine uses this method so I opted to put the newly loaded textures in an output parameter where it can be assigned to customTextures as needed.

I went through the steps to reproduce the bug outlined in https://forums.dfworkshop.net/viewtopic.php?t=5954. I do not have DREAM installed but I have another texture replacer (Diverse Weapons). I was seeing similar issues before while using it but they are now resolved with these changes.

### DIFF
--- a/Assets/Scripts/Game/FPSWeapon.cs
+++ b/Assets/Scripts/Game/FPSWeapon.cs
@@ -74,8 +74,8 @@ namespace DaggerfallWorkshop.Game
         DaggerfallUnity dfUnity;
         CifRciFile cifFile;
         WeaponAtlas weaponAtlas;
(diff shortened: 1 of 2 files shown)


Interkarma:
Thanks for looking at this so quickly @numidium :)

PR #2465, opened 2022-11-20
Allow commas in localization table values

AverniteDF:
If there's a way to do this already please let me know because it seems there ought to be.

Two options presented:

1) Allow alternative value separator (pipe) to be defined in schema
schema_pipes

2) Parse escaped character values
schema_escaped_values


Interkarma:
This is already possible. :) Using $ on field name in schema enabled quotes around that field in data.

I might revise the legacy text tables now that CSV parsing is being added for in-game localisation. The data tables were originally only meant for data in quest system, then were later co-opted for menu text.

PR #2466, opened 2022-11-22
Quest fix Nov 2022

JayH2971:
r0c11y28, Mad Slayer: this one didn't have "clearclick" mechanisms in place. Also removing every element of the "dead body," which appears for no reason and has no function. Am cutting it out as a presumed mistake.


JorisVanEijden:
The dead body is a prop.
The researcher is doing research and autopsies on victims of the ripper.
You are supposed to find a skeleton at their place in the beginning, representing the autopsied remains of an earlier victim, and later a dead body is added, representing a new victim, being autopsied.
Both are not essential to the story or quest, just scene-setting background props.

JayH2971:
Interesting! That gives some useful background information. As you said in the other post, placing the quest item and person at the same place makes this a bit clunky. But there's the answer, at least.

Interkarma:
Thank you Jay!

PR #2469, opened 2022-11-28
Rectify lycanthrope guard spawning behavior 

nights192:
Relatively simple change that prevents guards from despawning with player crime state when the player is a transformed lycanthrope.


Interkarma:
Thank you for fix. This solves the core despawning problem in a very direct manner. I've worked through a few different solutions and they all have different problem. I prefer this approach. Cheers. :)

PR #2471, opened 2022-12-04
Adjust arrow position relative to camera rotation

numidium:
Prevents arrows spawning off-center from player perspective.

**Problem:**
When player aims the bow far enough up or down, arrows spawn noticeably off-center from the camera's perspective.

**Solution:**
Adjust the arrow origin "down" from the camera's perspective instead of moving it down the Y-axis.


Interkarma:
Thank you, this looks better. It's a minor misalignment at very steep angles in any case, but this change shouldn't hurt anything.

I tested change as thoroughly as I could, and arrow still seem to go where the crosshair is aiming. That's good enough for me. :)

PR #2476, opened 2022-12-10
Increase Maximum Mouse Sensitivity to 20

mr-sihc:
This will allow people like me to look around better, since i like my sensitivities REAL high!

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallAdvancedSettingsWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallAdvancedSettingsWindow.cs
@@ -251,7 +251,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
 
             // Controls
             AddSectionTitle(leftPanel, "controls");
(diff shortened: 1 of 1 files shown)


mr-sihc:
And going back to the issue, no not a bug, i just like the sensitivity that high like 11, i played half life like it, wasnt a bug

mr-sihc:
I do beg you not to close this, as this is important to me, for my game

Interkarma:
This is not the way to go about this. If you open a discussion in the appropriate place - the forums - and your request has merit and support from others, then you'll have a much better chance of a change actually being made. The way you're approaching this now is not productive.

You PR does not address the change correctly in any case. You're only expanding range on slider in advanced settings UI. Your setting will be lost when closing and re-opening game as you have not also updated range enforcement in settings manager. You also missed the slider control in launcher UI where same setting can be adjusted at startup.

Interkarma:
As this is a generally low impact setting, I've made the required changes to double maximum mouselook to 16.

https://github.com/Interkarma/daggerfall-unity/commit/c22bf050c40801014c6106dc5f9687a5eb4b4c0a

mr-sihc:
heyyyyy, thank you!

PR #2477, opened 2022-12-12
Add prohibited action event for DaggerfallEntity and Callbacks in ItemEquip and ReadySpell

GoldenKaiju:
Mod extension for prohibited items and and spell silence/disabled check event

This creates a callback in DaggerfallEntity for "prohibited actions", and changes the bool checks within a few functions to be events with a bool return type. When a function is made to complete early due to a check failing, an OnProhibitedAction event is raised.

The main behavioral changes in the commit:
1. Any player equip OnDisableEquip events that fail now produce the prohibited message. Before this change, only the EquipItem method within DaggerfallInventoryWindow would display the prohibited message.
2. Check events which fail produce another callback, potentially creating 2 callbacks total if they fail. I believe the additional complexity and small computational cost for the callbacks are worth it in exchange for mod accessibility.

Minor differences are:
- UnequipItem now has a disabled check event. However, the current result of event always unequips the item like before.
- UnequipItem with 2 params slot EquipSlot and list, now returns a bool, which was previously void. Since this was only a helper function used within ItemsEquipTable, it seemed fine to do and it was necessary for prohibited unequips to work properly.

The reason for this change is exposing these prohibited checks to modders for custom behavior and a new event modders can use when prohibited action has occurred. I intend to use some of these myself in upcoming mods I am creating.

This commit opens:
- Prohibited weapons and armor equips and OnProhibitedAction callback for enemies and non-player classes inheriting DaggerfallEntity
- Prohibited spells based on any arbitrary condition
- More fine detail prohibited rules for item equips
- Prohibited unequips for weapons, cursed items
- Callback events when any prohibited action occurs

While there were more changes than I originally expected, a good chunk of these changes are cut and paste reorganization. The second commit with OnDisableReadySpell shows how few lines are needed when replacing an if statement and how simple the changes can be.

I think the prohibited action event for daggerfall entities is a useful extension. It can be expanded to include other types of behavior/systems in the future. I wouldn't know of another way to expose this functionality for modders without being mod incompatible. One thing I decided to not modify was the special weapon handling for arrows. Since arrows in EquipItem always return null, you could argue it's part of the OnDisabledEquip check. It wasn't necessary to modify as part these commits, although it could be added in the future.

### DIFF
--- a/Assets/Scripts/Game/Entities/DaggerfallEntity.cs
+++ b/Assets/Scripts/Game/Entities/DaggerfallEntity.cs
@@ -288,6 +288,7 @@ namespace DaggerfallWorkshop.Game.Entity
             // Allow for resetting specific player state on new game or when game starts loading
             SaveLoadManager.OnStartLoad += SaveLoadManager_OnStartLoad;
             StartGameBehaviour.OnNewGame += StartGameBehaviour_OnNewGame;
+            ItemEquipTable.OnDisableEquipitem += DaggerfallEntity_OnDisableEquipEvent;
         }
 
         #endregion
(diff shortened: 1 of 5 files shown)


Interkarma:
Hi, please be aware that Daggerfall Unity is about to enter Release Candidate. To maintain stability and mod compatibility leading up to 1.0, I am no longer accepting new new Pull Requests other than direct bug fixes. All changes that do not fix an accepted issue will not be merged.

This ensures I have time to complete my personal commitments to core project, rather than only servicing requests from mod authors. Once Daggerfall Unity reaches 1.0, I will be stepping away from this project to work on something new. All Pull Requests not accepted at that time will be closed.

Thank you for your contribution and understanding.

-Interkarma

PR #2478, opened 2022-12-15
2430 fix %god macro for fighters guild

JorisVanEijden:
This is a fix for using a %god macro in a quest when inside a Fighters Guild hall in a region where this is considered a temple.
It will pick a random deity name, rather than crashing the quest.

I also fixed the existing random deity name selection, which could pick non-existing factionIds 23, 30, 31, 32 and 34.

### DIFF
--- a/Assets/Scripts/Game/Questing/QuestMCP.cs
+++ b/Assets/Scripts/Game/Questing/QuestMCP.cs
@@ -217,10 +217,8 @@ namespace DaggerfallWorkshop.Game.Questing
                 {
                     // Classic returns "BLANK" if no temple is found, here we return a random deity name.
                     // We do the same for Fighters Guild halls, which are are considered temples in some areas.
-                    const int minGodID = 21;
-                    const int maxGodID = 35;
+                    var god = GetRandomDivine();
 
-                    FactionFile.FactionIDs god = (FactionFile.FactionIDs)UnityEngine.Random.Range(minGodID, maxGodID + 1);
                     return god.ToString();
                 }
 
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Game/Questing/QuestMCP.cs
```diff
@@ -258,6 +257,22 @@ public override string Direction()
...
             }
+
+            private static FactionFile.FactionIDs GetRandomDivine()
+            {
+                switch (UnityEngine.Random.Range(0, 9))
+                {
+                    case 0: return FactionFile.FactionIDs.Arkay;
+                    case 1: return FactionFile.FactionIDs.Zen;
+                    case 2: return FactionFile.FactionIDs.Mara;
+                    case 3: return FactionFile.FactionIDs.Ebonarm;
```
KABoissonneault, line 268: Not sure about Ebonarm being in the random selection.
We could go full "Daggerfall urban legend" and make Ebonarm the "FG with Temple building type" deity, instead of a random selection. Having a random deity there might get them confused with religious knight orders

JorisVanEijden: I'm not familiar enough with the lore to be able to say which divines would be inappropriate to mention in a Fighters Guild hall.

I went with the original code comment "Better just to provide something than let let game loop crash".

Interkarma:
Everything is a step up from just returning BLANK like classic. Not crashing is definitely an improvement over that. :)

@KABoissonneault - I'm not much of a lore guy, but I see your point here. I'm happy to exclude Ebonarm after merge if you like?

PR #2482, opened 2022-12-18
Bug when giving the totem to Gothryd

JorisVanEijden:
Removes the requirement to be outside the castle (S.07) before Gothryd will "thank" and reward you for giving him the Totem.

Fix for https://forums.dfworkshop.net/viewtopic.php?t=4845


Interkarma:
Thank you!

PR #2483, opened 2022-12-18
[C0B10Y15] The Relic does not give extra reputation

JorisVanEijden:
Move the ending of the quest to the end of the quest. This should give S.03 a chance to trigger without being stopped at S.01.

Fixes https://forums.dfworkshop.net/viewtopic.php?t=5679


Interkarma:
Nice and simple. Thank you for fix @JorisVanEijden.

Issue #2491, opened 2023-02-04
Linux 0.14.5: paintings do not display in inventory


JayH2971:
ISSUE:

A Windows user told me this does not happen for them, so this may be Linux-related. No mods, 0.14.5.

Clicking on any painting provides no full-size image of it in the inventory. The paintings were spawned by Biography questions in chargen and by the console.

![image](https://user-images.githubusercontent.com/37541932/216775756-a0e7f62b-2147-4f3b-adf1-8bf246719944.png)

The player log generated the following error after clicking:

> UnityException: Texture '' is not configured correctly to allow GetPixels
>   at (wrapper managed-to-native) UnityEngine.Texture2D.GetPixels32(UnityEngine.Texture2D,int)
>   at UnityEngine.Texture2D.GetPixels32 () [0x00001] in <0012ea2f37974907a308009fd693c396>:0 
>   at DaggerfallWorkshop.Utility.ImageReader.GetColors (DaggerfallWorkshop.ImageData imageData) [0x00015] in <89ba9f00299043cca178a23f1cb68cb2>:0 
>   at DaggerfallWorkshop.Utility.ImageReader.GetImageData (System.String filename, System.Int32 record, System.Int32 frame, System.Boolean hasAlpha, System.Boolean createTexture, System.Boolean createAllFrameTextures, System.Int32 alphaIndex) [0x0044d] in <89ba9f00299043cca178a23f1cb68cb2>:0 
>   at DaggerfallWorkshop.Game.UserInterfaceWindows.DaggerfallInventoryWindow.ShowInfoPopup (DaggerfallWorkshop.Game.Items.DaggerfallUnityItem item) [0x000d8] in <89ba9f00299043cca178a23f1cb68cb2>:0 
>   at DaggerfallWorkshop.Game.UserInterfaceWindows.DaggerfallInventoryWindow.LocalItemListScroller_OnItemClick (DaggerfallWorkshop.Game.Items.DaggerfallUnityItem item, DaggerfallWorkshop.Game.UserInterfaceWindows.DaggerfallInventoryWindow+ActionModes actionMode) [0x000c7] in <89ba9f00299043cca178a23f1cb68cb2>:0 
[...]

PR #2492, opened 2023-02-05
Use uppercase ?PAINT.CIF filenames

petchema:
Matters for case-sensitive operating systems, like Linux or OS X

Fix for #2491 

I wonder if there could be other similar bugs hiding, because that one was almost silent (unless you keep an eye on the logs)


KABoissonneault (approved):

Interkarma:
Thank you Pango!

PR #2493, opened 2023-02-07
Moved EnemyEntity methods to FormulaHelper for mod override

KABoissonneault:
Added FormulaHelpers for EnemyEntity:

- GetEnemyGroup
- GetLanguageSkill
- GetWeightInClassicUnits

The first is necessary for the functioning of Pacify and Dispel spells effects to work on custom enemies. The second is for the language pacification feature.
image

The third one, I don't plan on using just yet, but I felt I might as well add it for completion.


Interkarma:
I think this one is OK despite the feature lock. The original methods are still in place with their original signatures, I don't _think_ this will break any mods calling those public methods. Is that right, Kab?

PR #2495, opened 2023-02-08
Extended max "terrain distance" to 6

KABoissonneault:
I've seen players use a custom DFU build to run terrain distance at 6. Seems viable on some PCs. 
For a while, people have been running Distant Terrain, which adds an extra 4 low-quality tiles around the terrain, which kind of looks like a distance of 8.
However, users have reported memory and other issues with DT in recent versions. Running DFU with a higher terrain distance seems easier for them.

So, I don't think allowing this is harmful. Users can experiment and see what they prefer.

I could consider letting the max go to 8, not sure if that's something anyone would run.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallAdvancedSettingsWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallAdvancedSettingsWindow.cs
@@ -373,7 +373,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             // Advanced settings
             AddSectionTitle(rightPanel, "advanced");
             fovSlider = AddSlider(rightPanel, "fovSlider", 60, 120, DaggerfallUnity.Settings.FieldOfView);
-            terrainDistance = AddSlider(rightPanel, "terrainDistance", 1, 4, DaggerfallUnity.Settings.TerrainDistance);
+            terrainDistance = AddSlider(rightPanel, "terrainDistance", 1, 6, DaggerfallUnity.Settings.TerrainDistance);
             shadowResolutionMode = AddSlider(rightPanel, "shadowResolutionMode",
                 DaggerfallUnity.Settings.ShadowResolutionMode, "Low", "Medium", "High", "Very High");
             dungeonLightShadows = AddCheckbox(rightPanel, "dungeonLightShadows", DaggerfallUnity.Settings.DungeonLightShadows);
--- a/Assets/Scripts/SettingsManager.cs
+++ b/Assets/Scripts/SettingsManager.cs
@@ -513,7 +513,7 @@ namespace DaggerfallWorkshop
             StartCellY = GetInt(sectionStartup, "StartCellY", 2, 497);
             StartInDungeon = GetBool(sectionStartup, "StartInDungeon");
 
-            TerrainDistance = GetInt(sectionExperimental, "TerrainDistance", 1, 4);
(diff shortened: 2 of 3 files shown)


petchema:
My main concern is people setting it to the max blindly and then getting huge performance drop on their subpar hardware.
I did like the more resources conscious distant terrain approach, I hope it'll get some love too...

KABoissonneault:
I can always leave the launcher slider to 4 and only allow the higher setting in the INI... meh

petchema:
Agreed, not a good idea; Beside arbitrary inconvenience, the setting could be reset inadvertently, maybe just by opening the settings page.

Interkarma:
I'm also concerned users who don't understand the settings will blow out their performance and just assume "Daggerfall Unity Bad" without realising the changes they're making can seriously impact performance. I generally feel this way about a max terrain distance of 4 now, let alone much higher.

Distant Terrain has the right idea by setting up a ring of low-detail terrain around the main world grid. It's just unfortunate that mod isn't actively maintained. Maybe a better initiative would be for someone to revive and update that mod.

chappjo:
Hi, I was considering creating an issue regarding increasing terrain distance, and came across this.

I know people have been asking for increased render distance for many years at this point, but Interkarma has always said no due to performance concerns. While I think it is a valid concern, I think it's a bit silly at this point. On my laptop with **integrated graphics** and terrain distance of 4 my framerate sits around 90 FPS (100FPS without AO) in Daggerfall City, and 130FPS (145FPS without AO) at some random Cemetery in the middle of the Wrothgarian Mountains.

I have a couple of suggestions for options which might address Interkarma's concerns:

- An "unlock higher graphics settings" toggle which would unlock terrain distances above 4, while warning the user performance may suffer.

- An unlocked terrain distance slider, which would warn users with a yes/no warning popup if they set a terrain distance above 4.

- As KABoissonneault suggested, maybe just let users change it in the ini file, without having it be forcibly changed back when the game loads.


If it was up to me I would unlock terrain distances up to 8, or maybe even higher, however even 6 would be a huge improvement.

On my desktop I've played around a terrain distance of 8 through the Unity Editor. With my RX570, a 6 year old mid-range graphics card, the game is running at around 80FPS in the Wrothgarian Mountains wilderness, and 35FPS in Daggerfall City. This would be easily playable on any modern graphics card.

Interkarma:
Hey there. :) The problem of adding more terrain tiles is the increase to loading time and memory, especially when more locations are loaded in the expanded terrain radius. Framerate isn't too much of an issue.

To help explain it better, the current maximum is 4 because that's what I'm _willing to support_. Remember, people come to my forums asking for help when things go wrong. I'm often the one left with the burden of supporting these changes after the person who requested them moves on. I filter PRs not only through a lens of "is this reasonable" but also "will this increase my support burden". That's just not something people tend to consider when asking for things, because it's my life and not theirs.

I am considering merging Kab's increase to 6, but any higher isn't going to happen, I'm sorry.

Dezzj (commented):
Veiwed

AssassinLV:
You could block the max distance for non-discrete video cards and for those who don't overcome certain VRAM requirements, while warning those who passes the said requirements about possible performance impact and to revert back the max distance to 4 or bellow if freezes or low FPS is encountered. :)

numidium (approved):

KABoissonneault:
Closing this. I'm not willing to merge this in the near future, and the change is easy to redo if we ever need it in the future.

PR #2501, opened 2023-03-16
Use custom billboard gameobject for quest NPCs

KABoissonneault:
Unlike other static NPCs, quest NPCs didn't take custom flat gameobjects into account. So, no custom animation for Vanilla Enhanced, no 3d model.

With the fix:
image


Interkarma:
Hey Kab! I haven't spent a lot of time testing this one, but it should be OK. I'll merge and keep playing with it, and it will be in 0.15.3 if no problems found.

PR #2507, opened 2023-04-08
Force clickAnyWhereToClose behavior on message boxes with no buttons

DFIronman:
Currently, if you create a message box with `clickAnywhereToClose` set to `false` (the default), there's no way to get rid of the message box if it doesn't have a button that closes it. Unless there's a reason why this behavior is needed, here's a very small change that will allow the player to close buttonless message boxes that neglected to set `clickAnywhereToClose` to `true`.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallMessageBox.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallMessageBox.cs
@@ -657,7 +657,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             {
                 if (nextMessageBox != null)
                     nextMessageBox.Show();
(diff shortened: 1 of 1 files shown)


KABoissonneault:
Is there a way to close this window from code? I'd think being able to have a window that can only be closed from a code trigger (ex: delay) would be useful. The issue you're describing hopefully should never occur outside of debugging

DFIronman:
> Is there a way to close this window from code? I'd think being able to have a window that can only be closed from a code trigger (ex: delay) would be useful. The issue you're describing hopefully should never occur outside of debugging

You should be able to close it through code if you have a reference to it - should be somewhere on the stack in UIManager, if not already the TopWindow.

Your scenario is one I hadn't thought of. I think I'll just close this one.

PR #2513, opened 2023-04-27
Rewrite of CSV regex parsing to avoid exponential behavior

petchema:
Problem was discovered after DFU froze (busy-looping) while parsing a CSV with a faulty

    1554,"%it, Le Protecteur

(missing closing quotes) line in Text/Internal_MagicItems.csv

Nested star operators - like here, `(a*b*)*` - can take exponential time under the right conditions.
Minimal fix is to replace this pattern with `(a+|b+)*`, assuming a and b share no common prefix.

I did not stop there and removed the intermediate `matches` array by directly parsing key and value columns together in the regex.

Another potential bug fix, `.Trim("\"")` may remove more than the quotes surrounding quoted strings, ex.

    """Approach"" said the King" -> Approach" said the King

so I replaced it with a more explicit removal of surrounding quotes.

### DIFF
--- a/Assets/Scripts/Game/StringTableCSVParser.cs
+++ b/Assets/Scripts/Game/StringTableCSVParser.cs
@@ -102,10 +102,12 @@ public class StringTableCSVParser
         // First row will be accepted if any other key/value pair is present instead
         if (rows.Count > 0 && rows[0].Key == keyString && rows[0].Value == valueString)
             rows.RemoveAt(0);
+
         return rows.ToArray();
     }
 
-    static string UnescapeCSVvalue(string value) {
+    static string UnescapeCSVvalue(string value)
+    {
         if (value.Length > 0 && value[0] == '"')
         {
             return value.Substring(1, value.Length - 2)
--- a/Assets/Scripts/Game/StringTableCSVParser.cs
+++ b/Assets/Scripts/Game/StringTableCSVParser.cs
@@ -36,7 +36,7 @@ public class StringTableCSVParser
     /// </summary>
     /// <param name="filename">Filename of StringTable CSV file.</param>
     /// <returns>KeyValuePair for each row if successful, otherwise null.</returns>
-    public static KeyValuePair<string, string>[] Load(string filename)
+    public static List<KeyValuePair<string, string>> Load(string filename)
     {
         string csvText = null;
 
@@ -64,7 +64,7 @@ public class StringTableCSVParser
             return null;
 
         // Parse into CSV rows
-        KeyValuePair<string, string>[] rows = null;
+        List<KeyValuePair<string, string>> rows;
         try
         {
             rows = ParseCSVRows(csvText);
@@ -84,7 +84,7 @@ public class StringTableCSVParser
     /// </summary>
     /// <param name="csvText">Source CSV data.</param>
     /// <returns>KeyValuePair for each row.</returns>
-    static KeyValuePair<string, string>[] ParseCSVRows(string csvText)
+    static List<KeyValuePair<string, string>> ParseCSVRows(string csvText)
     {
         // Regex pattern inspired by https://gist.github.com/awwsmm/886ac0ce0cef517ad7092915f708175f
         // but without the exponential behavior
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Game/StringTableCSVParser.cs
```diff
@@ -84,37 +84,36 @@ public class StringTableCSVParser
...
         if (rows.Count > 0 && rows[0].Key == keyString && rows[0].Value == valueString)
             rows.RemoveAt(0);
-
         return rows.ToArray();
     }
 
+    static string UnescapeCSVvalue(string value) {
+        if (value.Length > 0 && value[0] == '"')
+        {
+            return value.Substring(1, value.Length - 2)
```
KABoissonneault, line 111: Maybe we should check and report an error if the last character is not also a quote.

Assets/Scripts/Game/StringTableCSVParser.cs
```diff
@@ -84,37 +84,36 @@ public class StringTableCSVParser
...
         // Remove first row if it contains "Key" as key and "Value" as value
         // This is the expected header row but doesn't need to be present
         // First row will be accepted if any other key/value pair is present instead
         if (rows.Count > 0 && rows[0].Key == keyString && rows[0].Value == valueString)
             rows.RemoveAt(0);
-
         return rows.ToArray();
     }
 
+    static string UnescapeCSVvalue(string value) {
```
KABoissonneault, line 108: Minor, DFU puts the opening brace on a new line

Assets/Scripts/Game/StringTableCSVParser.cs
petchema: No, because that's already guaranteed by the regex that if it starts with a quote, you're in the case it ends with a quote too

Interkarma:
Thanks for suggested fix Pango, I appreciate it. :)

I'll just need to regression test change against master CSV files. These are exported direct from string tables using Unity Editor ([here](https://docs.unity3d.com/Packages/com.unity.localization@0.9/manual/CSV.html)), so that's always the baseline for compatibility and expected format of CSV input.

Assets/Scripts/Game/StringTableCSVParser.cs
petchema: Same goes for unescaping quotes only in this case by the way; If the string does not start with a quote, then it cannot contain any quote.

Assets/Scripts/Game/StringTableCSVParser.cs
```diff
@@ -84,28 +84,19 @@ public class StringTableCSVParser
...
-                break;
-            }
-            string key = matches[pos++].Trim(trimChars);
-            string value = matches[pos++].Trim(trimChars);
-            value = value.Replace("\"\"", "\""); // Replace escaped quotes in value with single quote marks
-            KeyValuePair<string, string> kvp = new KeyValuePair<string, string>(key, value);
-            rows.Add(kvp);
-        }
-
+        char[] trimChars = { '\r', '\n' };
```
jefetienne, line 92: This could be moved outside the method so it doesn't have to allocate a new array each time

Assets/Scripts/Game/StringTableCSVParser.cs
```diff
@@ -84,28 +84,19 @@ public class StringTableCSVParser
...
+        List<KeyValuePair<string, string>> rows = (from Match m in
+            Regex.Matches(csvText, linePattern)
+            select new KeyValuePair<string, string>(
+                m.Groups[1].Value.Trim(trimChars),
+                UnescapeCSVvalue(m.Groups[2].Value).Trim(trimChars)
+                )
+            ).ToList();
```
jefetienne, line 99: You could change it to `.ToArray()` and do `return rows;`

petchema: How to replace the `rows.RemoveAt(0)` though?

petchema: Ended up doing just that. See also commit message for why I chose to return a List

Interkarma:
This is excellent! It handles all master CSVs perfectly, plus some edge cases I look out for. It stands up to bad input in value field much better than the previous. Thank you Pango!

PR #2515, opened 2023-04-28
Allow custom sound indices in SoundReader.GetAudioClip. 

KABoissonneault:
This can be used for custom enemies to pass new sounds in EnemySounds (ex: by providing `1000.mp3` and using `1000` as the attack sound in its Mobile structure). The value is forwarded to `SoundReplacement`, which handles non-standard values just fine.

I added some extra sanitization for negative numbers in BsaFile, because `SoundClips.None` is `-1`, which is often used as a default.


Interkarma:
This is good, thank you Kab. :)

PR #2516, opened 2023-05-06
Remove ! prefix from spell bundles in Active Spells HUD tooltips

KABoissonneault:
I've recently learned that spell bundle names are user-facing

image

There's an existing practice in DF data that spells that shouldn't be sold at the Mages Guild start with ! (ex: `!Lycanthropy`). Having one of these as an active effect is not really possible in classic. With mods, it's desirable to give enemies custom spell bundles without filling up the vendor.
So, this change just removes the ! prefix of spells in the active spells HUD, making the name look less like a debug placeholder


Interkarma:
Thank you Kab!

PR #2518, opened 2023-05-15
[RFC] Use hamburger button instead of chevron for options

petchema:
Use a more recognizable (by contemporary players) hamburger button for options
DFU hamburger button options

Discord: https://discord.com/channels/559842137374457876/650507773389307935/1107786016007409714

(I actually learned this is an old symbol: https://en.wikipedia.org/wiki/Hamburger_button )


jefetienne:
I think it would look better if the bars were a little wider and spread out more evenly from the top left-corner like this

https://imgur.com/a/8ZaJ1ob

petchema:
Changed, hopefully got it right from looking at the screenshot

Interkarma:
This looks good, hopefully it better helps communicate the purpose. Thank you! :)

PR #2520, opened 2023-05-20
Remove delay between asset unloads

numidium:
The addition of incremental GC makes asset unloads less jarring in my experience. Waiting less time between asset unloads should give the GC less work to do when it is called. Incremental GC was enabled here: #2390

### DIFF
--- a/Assets/Scripts/Game/UserInterface/UserInterfaceWindow.cs
+++ b/Assets/Scripts/Game/UserInterface/UserInterfaceWindow.cs
@@ -128,7 +128,7 @@ namespace DaggerfallWorkshop.Game.UserInterface
         {
             uiManager.PopWindow();
             RaiseOnCloseHandler();
-            DaggerfallGC.ThrottledUnloadUnusedAssets();
+            DaggerfallGC.ForcedUnloadUnusedAssets();
         }
 
         public void PopWindow()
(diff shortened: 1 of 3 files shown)


petchema:
Memory graph traversal required for full GC depends mostly on the number of remaining live objects, not just the number of objects to collect, so I'm not convinced that incremental GC significantly improved system behavior when full GC is called; Which for many people is very annoying sound skipping, and the reason why I tried not to call full GC systematically when UI windows are closed.
So to me the only way to improve upon existing would be to find a way not to call UnloadUnusedAssets (which triggers a full GC), not to call it more often.

numidium:
You're right. I should gather metrics to justify this PR. If it turns out that the GC doesn't happen any faster when done more frequently then I'll close.

I find it a little odd that we need to force GC in the first place. I thought that was an implementation detail that is typically abstracted away when coding at the application level in C#. Does DFU eat up all the system memory on people's systems if we don't do this?

petchema:
> I find it a little odd that we need to force GC in the first place.
We don't. It seems that we need to periodically force Unity to release unused assets, which is a similar, but distinct layer on top of the first one.
Problem was that UnloadUnusedAssets was triggering a full GC _after_ it released unused assets, and that increased its latency significantly. I wonder if that changed when incremental GC is enabled.

numidium:
I looked into this a bit more. It seems the issue is that when we allocate a Unity asset without calling Destroy on that asset it just stays in the asset pool occupying memory. UnloadUnusedAssets identifies assets that aren't in use anymore but if we want to avoid using that then I think we'll have to go through the code and proactively destroy unused assets (textures, meshes, etc.) manually.

petchema:
Yes, and require mod authors to do the same with the assets they create

PR #2521, opened 2023-05-21
Add an option to hide PersistentDataPath in launcher

petchema:
Some streamers would prefer not to show PersistentDataPath in the launcher because it reveals their login name.

This PR adds a (not published) setting to hide that path. I called the setting "HideLoginName" to match its true intent, it case it becomes necessary in other places.

(I kept this PR minimal given feature freeze).


Interkarma:
This is fine, and I understand the purpose. Thank you Pango. :)

PR #2522, opened 2023-05-21
Allow clicking on save thumbnail to open save directory on Linux

petchema:
Tested that it works on Linux too, so lift the restriction for Windows and OSX only


Interkarma:
Very happy this works! When I first added this some years back the docs specifically said this method of opening a folder was Windows-only. I should have realised when it worked on Mac that it should be available for Linux as well. Thank you. :)

PR #2523, opened 2023-05-22
Localize "USE"d maps

petchema:
Translate the location written on maps used from the inventory (usually maps found in loot).
Tested to translate correctly the dialog box, the notes, and the regional map.

Screenshot before PR:
Capture d’écran du 2023-05-22 01-18-01


Interkarma:
Always finding a new spot where pre-localization text is slipped in somewhere. I catch most of them, but not all. Appreciate the help. :)

PR #2524, opened 2023-05-25
"IsShield" Changed To Virtual So Custom Shields Can Be Added, 5/24/2023

magicono43:
"IsShield" Changed To Virtual So Custom Shields Can Be Added, 5/24/2023.

So with this property being private there is no way afaik (without hacky stuff of course) to make a custom item be considered a shield just as the vanilla ones are.

So hopefully this is a small enough change to not be an issue, thanks.


magicono43:
Messed up the pull-request, my bad.

PR #2525, opened 2023-05-29
Fix biography spurious questionmarks on empty lines

petchema:
Only happens when:
- not using SDF
- character has been refreshly created (not saved/reloaded)

Forums: https://forums.dfworkshop.net/viewtopic.php?t=6298


Interkarma:
Thank you Pango!

PR #2527, opened 2023-06-21
Spawn fix for Former Student part 2

JayH2971:
Players would suffer enemy spawns in Direnni Tower after completing Former Student 2, this will prevent that.


Interkarma:
Thank you Jay. I saw you supporting this on Discord earlier. Thanks for the assist! :)

PR #2528, opened 2023-06-23
Proactive asset cleaning

numidium:
The purpose of this PR is to prevent any allocated Unity assets from becoming orphaned so that we can be free of asset leaks in the DFU code base.

This is currently a work-in-progress. As of the creation of this PR I've eliminated most (if not all) asset leaks in the windowing system. I will make another pass at it after I work on the world streaming and anything else that might leak memory.

This will _not_ prevent mods from leaking memory. Therefore, we will probably still need to periodically call a global asset cleanup. However, I've noticed in my tests that the memory footprint is greatly decreased between cleanup calls. Whereas the editor would show memory usage exceeding 3 gb after opening and closing shelves in a shop in under a minute, it now never exceeds 2.45 gb as long as I don't travel around the world map and stream in new areas. I haven't tested this with high-definition texture mods but I can imagine the effect is even more dramatic.


Assets/Scripts/Game/UserInterface/ItemListScroller.cs
```diff
@@ -508,18 +509,21 @@ void LoadTextures(bool enhanced)
                 Texture2D redArrowsTexture = ImageReader.GetTexture(redArrowsTextureName);
                 redUpArrow = ImageReader.GetSubTexture(redArrowsTexture, upArrowRect, arrowsFullSize);
                 redDownArrow = ImageReader.GetSubTexture(redArrowsTexture, downArrowRect, arrowsFullSize);
+                GameObject.Destroy(redArrowsTexture);
```
ajrb, line 512: I'd like to understand why this destroy call is needed, or what benefits it brings. I suspect this may be an aspect of Unity game objects I am unaware of, but when I look at this code above it's getting a texture from image reader and using a locally scoped variable to hold a reference to it. Once the code block is exited this variable is out of scope and assuming there are no other references to this texture then it should simply by cleaned by the GC.

Is this destroy to expedite that? Or does it remove hidden unity references that are still held that will be preventing GC? Pretty sure the class references of subtextures wont hold a reference since the subdata is copied.

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
```diff
@@ -1042,6 +1046,38 @@ protected virtual void LoadTextures()
 
         }
 
+        protected virtual void UnloadTextures()
```
ajrb, line 1049: I've not been active on the project for a while, but if I remember correctly the Inventory window (plus a few others) essentially use singletons that are initialised each time they're shown, which I don't think re-creates the texture data being destroyed here. If you destroy the textures then they will need recreating every time the window is show. (a lot in most playing I think) I also can't see what's calling this method, but that's probably due to not diving deep enough into the PR.

ajrb (commented):
Couple of comments more from curiosity than criticism... I'm far to out of the loop for that these days but I'd like to understand why I didn't see these issues at the time. Thanks.

Assets/Scripts/Game/UserInterface/ItemListScroller.cs
numidium: What I realized about Unity assets (and what prompted me to start this PR) is that they don't respect the usual GC. ImageReader.GetTexture calls a Texture2D constructor which allocates a Unity Texture asset. If I don't explicitly Destroy that asset at some point then it will simply linger in memory. I've verified this by watching the texture number increase over time when opening a window that isn't a singleton. This is mitigated by a periodic call to Resources.UnloadUnusedAssets() in DaggerfallGC.cs. I'm hoping by proactively cleaning assets we can call that less often since 1. it's an expensive function and 2. the leaked memory can really pile up in between calls depending on what the player does.

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
numidium: That's correct but this is only used by the trade window which inherits from the inventory window. Unlike its base class the trade window isn't persistent so it has to be cleaned up when it's popped.

Assets/Scripts/Game/UserInterface/ItemListScroller.cs
ajrb: Oh wow, that's huge.. either Unity don't make that clear enough or I just somehow never realised that. Thanks for the explanation

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallInventoryWindow.cs
ajrb: Ah, very good. Makes sense now.

numidium:
This seems to be in a good state now. I'm not noticing any textures left behind by the UI or meshes by the world streaming. I was having issues getting dungeons to clean up all their meshes when the player leaves them so I opted to force a full asset clean when that occurs. Your mileage may vary on whether that's too aggressive but I suppose it depends on the performance hit. I think that, since there's a bit of a hitch when leaving a dungeon regardless, players won't notice it.

I want to do a bit more testing with the periodic collection turned off. Still need to try starting a new character with D.R.E.A.M installed so I can be sure custom assets don't throw a wrench in anything. I think that we should eventually turn the periodic collection back on (albeit at a slower pace) since we can't guarantee mods will collect their orphaned assets.

numidium:
I've found that loading assets from asset bundles mitigates memory leak issues. Assets loaded in this way are cleaned from memory without having to call Destroy. Since most assets in DFU can be swapped out with mod asset bundles I wrapped the Destroy method into my own method that ensures the asset is not in any bundles before actually calling it.

Not noticing any problems currently with the periodic memory sweep disabled. There _is_ a full memory sweep when leaving dungeons but nowhere else. I'm open to reinstating the periodic sweep but we shouldn't have to do it as often now.

Dezzj (commented):
Veiwed

KABoissonneault:
Now that we're in post-1.0, we can consider this change. It's a bit big, but I think the only obstacles now are some of the public API changes. If we can remove or workaround these, I'd like to give this a try. Unlike some of the other PRs, I don't think we'll be able to extensively test it by just one maintainer, we'll need to deploy this to playtesters. 
If we get a stable 1.0.1 build, we can also deploy a "1.0.1+" build later with this and see if people notice new issues with their mod setup

ajrb:
Sounds like a good plan Kab. I would advise only making the experimental build available to people on discord/forums so we actually hear feedback.

ajrb (dismissed):
Looks okay to me from reviewing code changes. 

I wonder if there's any noticable performance impact for players who have a lot of mods that all the asset cleanup calls have to iterate though all of their asset bundles? 

(I still can't believe Unity requires devs to manually handle this kind of thing when they're using a garbage collected language)

numidium:
> I wonder if there's any noticable performance impact for players who have a lot of mods that all the asset cleanup calls have to iterate though all of their asset bundles?

This is definitely something worth considering. I haven't yet looked into ways to search the library without enumerating the entire thing but I'll see if that's possible.

ajrb:
> This is definitely something worth considering. I haven't yet looked into ways to search the library without enumerating the entire thing but I'll see if that's possible.

Probably isn't, but worth a look. it's something that we should get someone running 100+ mods to test out.

Assets/Scripts/Game/Automap.cs
```diff
@@ -1500,9 +1503,10 @@ private void DestroyUserMarkerNotes()
         {
             if (gameObjectUserNoteMarkers != null)
             {
+                var meshFilters = gameObjectUserNoteMarkers.GetComponentsInChildren<MeshFilter>();
```
KABoissonneault, line 1506: Why do you get `meshFilters` here? It's not used in this scope.
Did you mean to `Destroy` those too?

Same for line 1700

Assets/Scripts/Game/UserInterface/UserInterfaceWindow.cs
```diff
@@ -128,7 +128,7 @@ public void CloseWindow()
         {
             uiManager.PopWindow();
             RaiseOnCloseHandler();
-            DaggerfallGC.ThrottledUnloadUnusedAssets();
+            //DaggerfallGC.ThrottledUnloadUnusedAssets();
```
KABoissonneault, line 131: Did you mean to keep this comment? Should we remove the line instead?

numidium: Line is now removed.

Assets/Scripts/Game/Automap.cs
numidium: Those were leftovers from some debugging I was doing. They're removed now.

ajrb (approved):

KABoissonneault:
Oh yeah, I should probably run a test build of this, while we have no new changes compared to the latest release

KABoissonneault:
Test build can be found here: https://github.com/Interkarma/daggerfall-unity/releases/tag/v1.1.1-pac. I've posted it on the Lysandus' Tomb Discord test channel

KABoissonneault:
I can't rule out setup issues, but it seems like multiple people get missing doors inside. Maybe this is while using Transparent Windows?

image
image

numidium:
I'm going to take a different approach using finalizers. That should hopefully require a lot less code and be less error-prone.

PR #2529, opened 2023-07-01
Reset seed to ensure magic item shop theft randomization

chloelcdev:
DaggerfallGuildServicePopupWindow.cs line 231 (in the GetMerchantMagicItems function) sets a consistent seed when opening the magic items window. This means there is no randomization when you go to steal an item from the magic shop. This change re-randomizes the seed when you're about to steal so that quickload-stealing is possible from the magic shop as I have to assume it was in vanilla.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTradeWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTradeWindow.cs
@@ -913,6 +913,10 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
                 int chanceBeingDetected = FormulaHelper.CalculateShopliftingChance(PlayerEntity, buildingDiscoveryData.quality, weightAndNumItems);
                 PlayerEntity.TallySkill(DFCareer.Skills.Pickpocket, 1);
 
+                // Here we ensure that stealing will always be random
+                // DaggerfallGuildServicePopupWindow.cs:231 sets a seed which will keep you from being able to steal from magic item shops, which I don't think is purposeful
+                UnityEngine.Random.InitState(System.DateTime.Now.Millisecond);
+
                 if (Dice100.FailedRoll(chanceBeingDetected))
                 {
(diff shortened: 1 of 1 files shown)


Interkarma:
Hey there. :) How it functions now is the intended behaviour. The seed is not fixed, it rotates so there's different stock on a daily basis.

chloelcdev:
I understand the behaviour is working as intended for filling the shop inventory. The issue I was attempting to solve is not the generation of the items, it's the chances of stealing those items. Essentially when stealing something from that shop, your chances are set in stone for that menu open. You cannot save and load to try stealing again, as it will generate the same numbers each time you load, since the seed is set each time you open the menu.

I would also assume that if the same numbers are going to be generated each time you open the menu, that if you are able to steal an item, I would think you'd be able to repeatedly reopen the menu and resteal the same object, since it will always generate the same chance value.

Interkarma:
Ah, I seem to have misunderstood the intent of your change then. My apologies. I'll reopen and revisit this one again with fresh eyes. Thank you. :)

ajrb:
Hi, original author of that trade window code here. I can say for sure that this was a circumstance I never considered when Interkarma added the 24h deterministic random generation to merchant magic items the following year, even though I reveiwed it. Unless someone confirms as classic behaviour, by pure chance, I think this is an edge case we missed 4 years ago. (wow doesn't time fly?) It certainly wasn't intended as far as I can recall.

Good spot in that case. Thanks for the PR. 

I don't think there's any ill effect from setting the seed to random in trade window, but I wonder if a better approach would be to change GetMerchantMagicItems() to save before, and then restore the seed after, seeding from time to keep it self contained. What do you think @Interkarma? I'm a bit rusty with DFU code so take this as a suggestion only.

Interkarma:
Hey thank you for spotting this one! I'll merge this change so you're in the contributor list and will add you to credits. I'll probably revert to #2534 as it fixes the problem at source (where I stuffed things up a little while back), but it's still your fix in principle.

PR #2531, opened 2023-07-16
Fix painting description for paintingRecord #70

petchema:
PAINT.DAT's #70 paintingRecord alternatives list for %pp2 is empty, raising an exception.

Runtime patch a list for this macro, compatible with painting gfx. In case this paintingRecord is hardcoded to be used for Wayrest Painting, it will also be compatible with the "fog over Cryngaine Fields" event.

It is also possible that classic does not generate text (%pp2 expands to the empty string) instead.

Forums: https://forums.dfworkshop.net/viewtopic.php?t=6336


Interkarma:
Thank you Pango, I appreciate it. :)

PR #2532, opened 2023-07-17
Dim backgound for DaggerfallEffectSettingsEditorWindow

DunnyOfPenwick:
The background panel of the DaggerfallEffectSettingsEditorWindow should be dimmed like in the original DF.


Interkarma:
Thank you! :)

PR #2533, opened 2023-07-26
Fix regression caused by #1861

petchema:
PR #1861 was added to prevent caster enemies from nuking themselves with spells at range in the presence of low obstacles.

A side effect of this PR however, is that daedra lords in the "crossbow room" near the end of Mantellan Crux stopped shooting at the player altogether, making this room much less threatening.

This new PR checks for obstacles a bit in front of the enemies, so they can still cast from behind thin obstacles (rationale: spells start from arm's reach). It's a bit crude, for example it doesn't change from where the spells will be really cast. While some spells get thru, daedra lords sometimes nuke themselves.

### DIFF
--- a/Assets/Scripts/Game/DaggerfallMissile.cs
+++ b/Assets/Scripts/Game/DaggerfallMissile.cs
@@ -37,6 +37,7 @@ namespace DaggerfallWorkshop.Game
 
         public float MovementSpeed = 25.0f;                     // Speed missile moves through world
         public float ColliderRadius = 0.45f;                    // Radius of missile contact sphere
+        public static float ArmLength = 0.9f;                   // Distance of cast origin, >= ColliderRadius
         public float ExplosionRadius = 4.0f;                    // Radius of area of effect explosion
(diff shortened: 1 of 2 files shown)


petchema:
Casters from the main quest, for tests
[Orsinium Shaman.zip](https://github.com/Interkarma/daggerfall-unity/files/12170462/Orsinium.Shaman.zip)
[Karolys.zip](https://github.com/Interkarma/daggerfall-unity/files/12170463/Karolys.zip)
[Daedra Lords.zip](https://github.com/Interkarma/daggerfall-unity/files/12170464/Daedra.Lords.zip)

petchema:
Update: casting origin is now consistent between EnemyMotor and DaggerfallMissile, enemies should cast safely again.
As an improvement, casting from a higher position (instead of only ahead) would help daedra lords to cast from a wider variety of angles; But that may require adding enemy-type specific logic.

Interkarma:
You're a treasure, thank you Pango! I appreciate the saves for testing. :)

PR #2534, opened 2023-07-26
Ensure GetMerchantMagicItems preserves seed (alternative to #2529)

ajrb:
GetMerchantMagicItems() sets the random seed based on the current day. Any activity reliant on random sequence which occurs after opening the mages item seller will always have the same set of outcomes. Preserve the seed to avoid this.

This is an alternative fix to the one contributed in PR #2529 to limit the effect the change may have on the rest of the codebase. I am pretty sure there would be no ill effects from the other approach, but it's possible and given how close to release of v1.0 I am very cautious.

All credit to @chloelcdev for finding the issue, and contributing a perfectly fine fix.


Interkarma:
Thank you Hazel!

PR #2539, opened 2023-09-24
Allow extension to travel calculator for mods

ajrb:
Required for new settings for Travel Options mod in DFU release 1.0

### DIFF
--- a/Assets/Scripts/Game/Utility/TravelTimeCalculator.cs
+++ b/Assets/Scripts/Game/Utility/TravelTimeCalculator.cs
@@ -61,7 +61,7 @@ namespace DaggerfallWorkshop.Game.Utility
         /// returns minutes taken to travel.
         /// </summary>
         /// <param name="endPos">Endpoint in map pixel coordinates.</param>
-        public int CalculateTravelTime(DFPosition endPos,
+        public virtual int CalculateTravelTime(DFPosition endPos,
             bool speedCautious = false,
             bool sleepModeInn = false,
             bool travelShip = false,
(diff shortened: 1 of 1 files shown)


KABoissonneault:
I assume the intent of this change is so `DaggerfallTravelPopup` can be overriden to use a custom `TravelTimeCalculator`. But the calculator is also used by the Questing `Clock`, in `GetTravelTimeInSeconds`

```
            // Create a path to location
            // Use the most cautious time possible allowing for player to camp out or stop at inns along the way
            TravelTimeCalculator travelTimeCalculator = new TravelTimeCalculator();
            int travelTimeMinutes = travelTimeCalculator.CalculateTravelTime(endPos, true, false, false, false, true);
```

I feel like we should also provide a way for mods to override the quest travel time calculator for consistency.

I'd suggest adding a `public static Func<TravelTimeCalculator> TravelTimeCalculatorFactory = () => new TravelTimeCalculator();` in `Clock`, and replacing the above snippet with
```
            // Create a path to location
            // Use the most cautious time possible allowing for player to camp out or stop at inns along the way
            TravelTimeCalculator travelTimeCalculator = TravelTimeCalculatorFactory();
            int travelTimeMinutes = travelTimeCalculator.CalculateTravelTime(endPos, true, false, false, false, true);
```

ajrb:
I simply want to be able to manipulate the travel costs after they're calculated, and actually only need the virtual on  CalculateTripCost() but I thought it was worth making the others virtual as well.

Keeping the changes as simple as possible to get them in to 1.0 was my focus, but by all means create a follow on PR. I won't require that myself though.

PR #2540, opened 2023-09-24
Synchronize char creation wizard screens

numidium:
Addresses a bug described here: https://forums.dfworkshop.net/viewtopic.php?p=66200

The skill bonus screen de-synchronizes with the summary screen when the user goes back and forth in the character creation wizard. My fix was to ensure either window is synced to the character document whenever they are pushed. I also made it so that changes made on the summary page are propagated backward in the wizard as far as makes sense. Skills bonuses will still be wiped if the user creates a new biography since that alters them.


Dezzj (commented):
Veiwed

Assets/Scripts/Game/UserInterface/SkillsRollout.cs
```diff
@@ -146,7 +146,7 @@ public void Reroll()
             UpdateSkillValueLabels();
         }
 
-        public void SetClassSkills(DFCareer dfClass)
+        public void SetClassSkills(DFCareer dfClass, bool doReroll = true)
```
KABoissonneault, line 149: After 1.0, we have to be careful with modifying the signature of public functions. This breaks precompiled mods

https://learn.microsoft.com/en-us/dotnet/standard/library-guidance/breaking-changes#binary-breaking-change
> A binary breaking change happens when you change the public API of your library, so assemblies compiled against older versions of your library are no longer able to call the API. For example, **changing a method's signature by adding a new parameter** will cause assemblies compiled against the older version of the library to throw a [MissingMethodException](https://learn.microsoft.com/en-us/dotnet/api/system.missingmethodexception).

I think in this case, we can avoid it by simply overloading the function instead. Keep the old signature, add a new version with the bool parameter, and the old function simply calls `SetClassSkills(dfClass, true);`

numidium: Ok, added the old method back in.

KABoissonneault:
I've tested it and the fix works. Now, when going back to the biography, the "add bonus" window is reset, rather than preserve the state of the last time the screen was entered, which makes sense given the biography changes skill values.
I do wish we could have preserved the bonus assignment while modifying the base skills, given how other screens preserve such assignments. But it's probably not worth the extra complication

KABoissonneault (approved):

ajrb (approved):

PR #2543, opened 2023-10-29
Add null check to inventory window

ajrb:
Spotted this and figured it should be corrected before release.


Interkarma:
Thank you Hazel. I'll get this merged before 1.0 builds.

PR #2545, opened 2023-11-18
Fix for RuntimeMaterials to pick the right ground texture in Woodland Hills

KABoissonneault:
A user was having a mismatch in RuntimeMaterials when replacing ground textures according to climate/season/weather.

In this example:
image

The bright green of the hill clashes with the dark green of the rest of the area. This is because the hill has been made to use archive 302-304, and Woodland Hills uses archive 102-104 for ground.

A concrete climate's ground archive is defined in `MapsFile.GetWorldClimateSettings`, along with its base climate setting. Without the concrete climate, we can get a default ground archive matching the base climate as defined in `ClimateSwaps` logic (either in `ApplyClimate` or `GetGroundArchive`).
In the latter case, we have for the ground:
- Desert = archive 2
- Mountain = archive 102
- Temperate = archive 302
- Swamp = archive 402
With winter variants at x03 and rain variants at x04.

For the most part, the climates in `MapsFile.GetWorldClimateSettings` have a ground archive matching the base climate's in `ClimateSwaps.GetGroundArchive`. We only have one exception: `Climates.MountainWoods` (aka "Woodland Hills") has a Temperate base climate (normally ground archive 302), but with a specific ground archive of 102 (the mountain one).
102 is the archive used by the terrain generator. Therefore, any ground generated in this climate needs to match 102, not 302.

This fix adds an explicit check in RuntimeMaterials for whether we're replacing a ground archive, and if so, use the climate settings' ground archive, not the naive climate-based swap.

And the result is a more natural hill
image

Similarly, `DaggerfallMesh.SetClimate` uses `MaterialReader.ChangeClimate` which also relies on `ClimateSwaps` logic to customize by weather. However, I'm not sure whether a `DaggerfallMesh` ever uses the ground archive, so I did not insert this fix in this PR.

### DIFF
--- a/Assets/Scripts/Utility/AssetInjection/Components/RuntimeMaterials.cs
+++ b/Assets/Scripts/Utility/AssetInjection/Components/RuntimeMaterials.cs
@@ -198,10 +198,28 @@ namespace DaggerfallWorkshop.Utility.AssetInjection
             int archive = runtimeMaterial.Archive;
             int record = runtimeMaterial.Record;
 
+            ClimateTextureInfo ci = ClimateSwaps.GetClimateTextureInfo(archive);
             if (dungeonTextureTable != null)
+            {
                 archive = DungeonTextureTables.ApplyTextureTable(archive, dungeonTextureTable, climateBaseType);
+            }
             else if (runtimeMaterial.ApplyClimate)
-                archive = ClimateSwaps.ApplyClimate(archive, record, climate, season);
+            {
+                if (ci.textureSet == DFLocation.ClimateTextureSet.Exterior_Terrain)
+                {
+                    // For exterior terrain, always apply the ground based on climate settings, not climate swap
+                    // Most climates match, except Woodland Hills, which is Temperate with Mountain ground
+                    archive = GameManager.Instance.PlayerGPS.ClimateSettings.GroundArchive;
+                    if (season == ClimateSeason.Winter && ci.supportsWinter)
+                        archive += (int)DFLocation.ClimateWeather.Winter;
+                    else if (season == ClimateSeason.Rain && ci.supportsRain)
+                        archive += (int)DFLocation.ClimateWeather.Rain;
+                }
+                else
+                {
+                    archive = ClimateSwaps.ApplyClimate(archive, record, climate, season);
+                }
+            }
 
             return DaggerfallUnity.Instance.MaterialReader.GetMaterial(archive, record);
         }


KABoissonneault:
We can confirm that the issue also occurs for `DaggerfallMesh`. 
In this example below,
image

The mound uses the grass from archive 302 (Temperate), despite the surrounding ground being archive 102 (Mountain). 

I'm not fixing this yet, because `DaggerfallMesh` only has info on the current "base climate", not the actual climate. Too big of a change for the scope of this. But it's worth noting for the future

drcarademono:
There's another bug with RuntimeMaterials or a related script, which occurs when users put nature flats in an RMB.json override, like this:

```
        "MiscFlatObjectRecords": [
            {
                "Position": 0,
                "XPos": 1511,
                "YPos": -1,
                "ZPos": -2546,
                "TextureArchive": 504,
                "TextureRecord": 16,
                "FactionID": 0,
                "Flags": 0
            }]
```

Nature flats in RMB.jsons are placed correctly in their original climate (in this case Temperate), but in other climates are floating or buried the ground:

Correctly placed in Temperate:


Incorrectly placed in Mountain:


It's possible that the texture is being swapped _after_ a BillboardMesh is created based on the original texture, but that is only my conjecture. Nature flats placed the normal way, through the GroundScenery section in RMB.jsons, work fine.

I'm attaching a modified FILLAA00.RMB.json and two save folders (in Horbury, Dwynnen and Crossgate, Wrothgarian Mountains that can be used to reproduce the issue:
[Nature Flats Placement Bug.zip](https://github.com/Interkarma/daggerfall-unity/files/13422270/Nature.Flats.Placement.Bug.zip)

drcarademono:
In case it helps, this is the section in DaggerfallLocation.cs that's responsible for swapping out nature flats in RMBs. It only handles loose flats, NOT nature flats defined by the GroundScenery section in RMBs. Line 257 (                    db.SetMaterial(natureArchive, db.Summary.Record);) handles the swap.

```            // Determine correct nature archive
            int natureArchive = GetNatureArchive();

            // Process all DaggerfallBillboard child components
            Billboard[] billboardArray = GetComponentsInChildren<Billboard>();
            foreach (var db in billboardArray)
            {
                if (db.Summary.FlatType == FlatTypes.Nature)
                {
                    // Apply recalculated nature archive
                    db.SetMaterial(natureArchive, db.Summary.Record);
                }
                else
                {
                    // All other flats are just reapplied to handle any other changes
                    db.SetMaterial(db.Summary.Archive, db.Summary.Record);
                }
            }

            // Process nature billboard batch
            if (summary.NatureBillboardBatch != null)
            {
                summary.NatureBillboardBatch.SetMaterial(natureArchive, true);
                summary.NatureBillboardBatch.Apply();
            }
        }```

numidium (approved):

ajrb (approved):

PR #2547, opened 2023-11-23
Add setter for StaticNPC.Data

numidium:
Allows modders to edit static NPC data.

### DIFF
--- a/Assets/Scripts/Game/StaticNPC.cs
+++ b/Assets/Scripts/Game/StaticNPC.cs
@@ -69,6 +69,14 @@ namespace DaggerfallWorkshop.Game
             get { return IsChildNPCData(Data); }
         }
 
+        /// <summary>
+        /// Sets the NPC name bank for name generation.
+        /// </summary>
+        public NameHelper.BankTypes NameBank
+        {
+            set { npcData.nameBank = value; }
+        }
+
         #endregion
 
         #region Structs & Enums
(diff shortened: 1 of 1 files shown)


KABoissonneault:
Can you explain the shortcomings of using the existing `SetLayoutData` methods? What sort of modification did you wanna make that couldn't be done using the existing APIs?
As far as I see, all of properties can be set from there. However, there are some restrictions, as a few of them are derived from your input, rather than allowing for an exact value to be assigned. 

If only a few specific fields needs to be overwritten, we could add more public methods for those fields only.

I'm concerned we're relaxing invariants and letting people create code that breaks in subtle ways.

In spite of all this, I'm not a fan of the property setter on the struct. As far as I remember, you can't just do `staticNpc.Data.nameSeed = ...`, you gotta assign the whole struct for each modification. Might as well just make the private `npcData` public if you wanna throw all the invariants off the window anyway

numidium:
You are correct. I added another SetLayoutData that only alters what I wanted access to. The issue was that the name bank was inaccessible. Also, it was not possible to alter other npc layout data without altering the hash unless the original location data is fetched from the RDB record. I added a method that allows mutating data without touching the hash.

KABoissonneault:
The `nameBank` is documented in `struct NPCData` as a "runtime" property. When the `StaticNPC` enters play, the `Start` event is called, which calls `SetRuntimeData()`, which sets the nameBank. I feel like your function mixes up two responsibility, serialization data and runtime data. Honestly not sure in which order these are called (Runtime before Layout, or Layout before Runtime), but it seems fragile to me if the Runtime can override the namebank after you thought you assigned it in your new `SetLayoutData`.

`SetLayoutData` should only set the serialization data, and runtime data should be somewhere else.
IMO keep `SetLayoutData` as it was, and add a new setter for the namebank, which is expected to be called after `Start` has run.

numidium:
Okay, I moved the code that sets the name bank to its own setter. I want to keep a version of SetLayoutData that doesn't rely on block x,y,z values to pass in a hash though. This allows the existing npcData to be altered without retrieving its previous x, y, and z values from the block record.

KABoissonneault (approved):

Interkarma (approved):

PR #2550, opened 2023-12-14
God's name localized in the "Donation" dialog

Jagget:
God's name localized in the "Donation" dialog


Assets/Scripts/Game/Guilds/Temple.cs
```diff
@@ -320,6 +320,15 @@ public static Divines GetDivine(int factionId)
             throw new ArgumentOutOfRangeException("There is no Divine that matches the factionId: " + factionId);
         }
 
+        public static string GetDivineLocalized(int factionId)
+        {
+            string god = Temple.GetDivine(factionId).ToString();
+            if (!string.IsNullOrEmpty(god))
+                return TextManager.Instance.GetLocalizedText(god);
```
KABoissonneault, line 327: Should this use the factionId as key and the Factions collection? I'm not sure about how this works, but I don't think there are entries named after the gods

Jagget: Technically I could call GetLocalizedText in DaggerfallGuildServiceDonation.cs

Will it be a better solution?

KABoissonneault: Gonna let @Interkarma judge on this

Interkarma: My mental map is a bit fuzzy here, but I can quickly confirm the following keys are present in the string tables.

```
Akatosh
Arkay
Dibella
Julianos
Kynareth
Mara
Stendarr
Zenithar
```
If the faction ID to string is guaranteed to always return one of those keys (case sensitive) in this process, then it's at least on the right track. I'd have to step through this one to be sure of the outcome however. Daggerfall being Daggerfall, there always seems to be some edge case I don't expect. :)

Jagget: Well, in this exact dialog we have a Macro Context, that overrides the %god macro, and the text is not translated.

KABoissonneault (approved):

Assets/Scripts/Game/UserInterfaceWindows/DaggerfallGuildServiceDonation.cs
```diff
@@ -105,7 +105,7 @@ public GuildServiceDonationMacroDataSource(DaggerfallGuildServiceDonation guildS
 
             public override string God()
             {
-                return Temple.GetDivine(parent.BuildingFactionId).ToString();
+                return Temple.GetDivineLocalized(parent.BuildingFactionId).ToString();
```
KABoissonneault, line 108: The ".ToString()" can be removed, `GetDivineLocalized` already returns a string

Jagget: Fixed

KABoissonneault (approved):

Interkarma (approved):

PR #2552, opened 2023-12-14
Added new terminator for macros that could be used for postfixes

Jagget:
Added new terminator for macros that could be used for postfixes

### DIFF
--- a/Assets/Scripts/Utility/MacroHelper.cs
+++ b/Assets/Scripts/Utility/MacroHelper.cs
@@ -395,7 +395,7 @@ namespace DaggerfallWorkshop.Utility
         #region Macro Expansion Code
 
         // Any non-alpha characters that can be on the end of a macro symbol need adding here.
-        static readonly char[] MACRO_TERMINATORS = { ' ', '%', '.', ',', '\'', '?', '!', '/', '(', ')', '{', '}', '[', ']', '\"', ';', ':' };
+        static readonly char[] MACRO_TERMINATORS = { ' ', '%', '.', ',', '\'', '?', '!', '/', '(', ')', '{', '}', '[', ']', '\"', ';', ':', '|' };
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Utility/MacroHelper.cs
```diff
@@ -453,6 +453,11 @@ public static void ExpandMacros(ref TextFile.Token[] tokens, IMacroContextProvid
 
                             // Iterate from macro end
                             currentPos = endPos;
+
+                            // "Eating" the | terminator (for postfixes purpose)
+                            // Example: '%di|ern' => 'southern'
+                            if (currentPos < tokenText.Length && tokenText[currentPos] == '|')
```
KABoissonneault, line 459: I'm not sure such a pattern would play well with localization. I'm also not keen on changing macro evaluation at this stage - what if someone wanted to make a text-based table (ex: '|%di|...|...|' -> '|south|...|...|). A bit contrived, but with a macro of fixed size, it would make more sense

Jagget: That was added specifically to enhance localization. We need postfixes in other languages. 
And I don't think this table (`|%di|...|...|`) would eve exist, honestly :D

Jagget: Or by that time, they would fiure out something

KABoissonneault:
Alright I made myself a small test suite for running this.
I tried these strings as a reference.

>        "Hello",
>        "Hello, %di!"
>        "Hello %di, mate!"
>        "Hello, %di|ern"
 >       "Hello, %di\\|ern"

Results before:
image

Results after:
image

The line with the \ is there in order to support the case where someone might want to print a special character immediately after a macro. This PR does not change how this is handled, so we can still add support for this in the future.

I think this is fine.

KABoissonneault (approved):

PR #2553, opened 2023-12-16
Fix for region macro localization

Jagget:
The `____place_ ` macro was not fully localized.

Before:
Screenshot 2023-12-16 001904

After:
Screenshot 2023-12-16 001823

The text of the letter that is on screenshot:
```txt
Message:  1011
  %pcf,
<ce>
  Картографы поручили мне сопровождать вас
  в поисках пути в ____place_.
  Давайте встретимся в ___place_, чтобы мы могли пойти
  вместе. Пунктом назначения будет __home_,
  и упражнение должно быть выполнено не позже,
  чем через =timer_ дней, чтобы поддержать
  вашу репутацию в ассоциации.
<ce>
                _qg_
                Исследователь, полевой офицер
                Ассоциация Картографов
                ___place_, ____place_
```

### DIFF
--- a/Assets/Scripts/Game/Questing/Place.cs
+++ b/Assets/Scripts/Game/Questing/Place.cs
@@ -273,7 +273,9 @@ namespace DaggerfallWorkshop.Game.Questing
                         // Workaround for older saves where regionIndex was not present and will always be 0 in save data (Alik'r Desert)
                         // This can result in improper region name being displayed when loading an older save and quest not actually set in Alik'r Desert.
                         // In these cases display name using the legacy regionName field stored in place data
-                        textOut = siteDetails.regionName;
+                        var index =
+                            DaggerfallUnity.Instance.ContentReader.MapFileReader.GetRegionIndex(siteDetails.regionName);
+                        textOut = TextManager.Instance.GetLocalizedRegionName(index);
                     }
                     else
                     {


Interkarma:
The code you changed is the fallback when region index is not set in older saves. It shouldn't be changed. There will be another reason your text isn't localized. This PR shouldn't be accepted as now.

I actually found and fixed one of these recently for 0.16.2. Is that the version you're testing on? Quest will need to be recompiled to pick up region index correctly.

Jagget:
> I actually found and fixed one of these recently for 0.16.2. Is that the version you're testing on? Quest will need to be recompiled to pick up region index correctly.

@Interkarma 

I'm running the master branch. And this quest started while I was running the Unity Player.

Interkarma:
OK, that's the core problem then. A quest that compiles on a recent version should never reach that code. It's only there as a handler for old saves, like the comment above it describes.

The proper fix will be to determine why region index isn't set by that point when quest compiled.

Jagget:
@Interkarma 

2 questions then:
1. What's wrong with translating fallback for old quests?
2. When is the quest compiled? On the game starts or on the quest starts?

Interkarma:
I'm happy to work with you, but what's the answer to my question above? Did this quest actually compile in Dragontail or Alik'r? That's important for me to determine how the code reached fallback in the first place.

To answer your questions: quests are compiled when the quest starts, and fallbacks are typically isolated from the localization system by design. That may be OK to rework in this case, but the problem you're trying to fix here actually indicates something else is happening. I want to dig into that before a fix is implemented.

Jagget:
> I'm happy to work with you, but what's the answer to my question above? Did this quest actually compile in Dragontail or Alik'r? That's important for me to determine how the code reached fallback in the first place.

OK, I'll try to explain how I got there. It is "Cartographer Guild" quest line from JayH. I started this quest line in Daggerfall. I don't really know if the game compiles consequent quests, but this exact quest has been auto started (I got the letter) in Dragontail region.

> That may be OK to rework in this case, but the problem you're trying to fix here actually indicates something else is happening.

Actually, I started a new game and tested it for longer. After some time, I got a quest that uses `____place_` macro, and in case of new game I didn't get the untranslated region.

Interkarma:
>Actually, I started a new game and tested it for longer. After some time, I got a quest that uses ____place_ macro, and in case of new game I didn't get the untranslated region.

Excellent. Yeah, I believe this should already be fixed in 0.16.2, I missed setting one index case previously that I fixed in below PR.

https://github.com/Interkarma/daggerfall-unity/commit/e6bb5cc72727cd39ac98c5f099165166797d488e

But the quest has to recompile for index to be passed down, which for long-running quests can be unclear when that happens. In the first case, quest was probably already compiled to QuestMachine and running with the missing index, so it went through to fallback.

My reasoning for isolating fallback from the localization system is that if something has already failed to localize for some reason, we shouldn't try and fallback to the system that already failed in the first place. But I will say your solution is self-contained and will probably be OK. It's just doing a region name lookup in TextManager to retry the index, then getting the localized name. After some more thought, I think this will be fine.

I'll reopen and change can be reviewed as a better fallback method.

Jagget:
Great! Thank you very much!

KABoissonneault (approved):

KABoissonneault:
The change is only fixing a data fixup for old saves, but it can't hurt to have. It should never affect new saves, and could help someone playing on an older save with a new version. Worth just getting in

Interkarma (approved):

PR #2556, opened 2023-12-31
Allow Portable Installs

RicardoLuis0:
This allows data to be stored in a `PortableAppdata` folder instead of in `persistentDataPath` if an empty `Portable.txt` file is in the same folder as the executable. If the install is portable, it also 
opens the browse window on the exe folder and stores paths as relative in the ini.


KABoissonneault:
I've seen a few people comment that this change would be useful to them, so we should probably review this. 
I don't know about asking users to create a dummy file to enable the feature, but I'm admittedly not familiar with how this is usually done - I assume through an installer wizard setting that does something similar.
I think the feature might be good for now, I just need to find the best way to test that no outside files get created. We can figure out about the user experience enabling this later

RicardoLuis0:
from what i've checked when writing it, everything that fetches/stores data goes through `PersistentDataPath`/`MyDaggerfallPath`/`MyDaggerfallUnitySavePath`/`MyDaggerfallUnityScreenshotsPath`, which i've addressed, but if i've missed something, it should be simple enough to add

Assets/Scripts/Game/UserInterface/FolderBrowser.cs
```diff
@@ -166,7 +166,22 @@ void Setup()
...
+            
+            if(DaggerfallUnity.Settings.PortableInstall)
+            { // start browsing from exe path for portable installs
+                currentPath = AppDomain.CurrentDomain.BaseDirectory;
+                if(SystemInfo.operatingSystemFamily == OperatingSystemFamily.Windows)
+                {
+                    var drive = Path.GetPathRoot(currentPath);
+                    driveList.SelectIndexSilent(drives.FindIndex( d => d == drive ));
+                }
+                UpdatePathText();
```
KABoissonneault, line 178: I'm not sure this is behaving properly. 

image

The ".." item is missing (probably because you made the "SelectIndex" silent and FolderList_OnUseSelectedItem could not add it). This probably needs to be added manually.

KABoissonneault (changes requested):
Initial selection should have a way to back out of the current folder

KABoissonneault (approved):

KABoissonneault (changes requested):
This change mostly works as expected now. Except, we still generate a Player.log in the AppData folder, Unity's idea of the Persistent Path. I'm testing some workarounds, but until we resolve that, I think this PR fails at achieving the goal of not modifying the environment outside the installation

KABoissonneault:
I've updated this feature with the new "DaggerfallUnityApplication", so that log files are properly in the portable settings folder. The feature should be good to go.

KABoissonneault (approved):

KABoissonneault:
Actually, one thing about this feature is that it saves the Daggerfall install and the save/screenshot folders as relative. However, I don't think we should do that if those folders are outside the DFU installation. Relative for inner folders is fine, since we want them to relocate if users move those folders. Gonna rework the feature to actually check for that before saving as relative.

Note that it's already possible to put your DF installation under StreamingAssets/GameFiles.

RicardoLuis0:
> Relative for inner folders is fine, since we want them to relocate if users move those folders.

ah, sorry, i didn't take into account that someone might do that 😅 (point it to an outside folder)

KABoissonneault:
In addition to the changes described above, I've reverted the changes to the FolderBrowser (and associated changes to ListBox.cs). The change was made with the idea that people would use this to embed their DF install into the DFU installation, but the proper way to do this is to put the DF install in StreamingAssets/GameFiles. There is no reason to make the FolderBrowser behave differently otherwise.

I have a test Win64 build here for reviewers, with the Portable.txt file already included. Simply run it, it shouldn't affect your global DFU settings at all.
https://mega.nz/file/f1Z0lJaQ#dpJYkImQhRX4NczAmkSshuWzvVWquAKm86eM1nJQzGs

KABoissonneault (approved):

KABoissonneault:
I'd really like to get this merged for the v1.1.1 official release soon. I have demo builds you can download, should be really easy to test

Jagget (approved):

Issue #2560, opened 2024-01-04
Advanced climbing bug - possible cause


red-moss2:
**Describe the issue**

As I understand it, climbing makes frequent checks involving Climbing skill and STR when in climbing mode, and will cause the player to fall if the check fails. So, if I have a character with say 11% Climbing and a STR of 43, my observed behaviour is trying to climb the face of a large house will see my character constantly slip down, reflecting my poor ability.

However, with Advanced Climbing System enabled, although initially it seems to honor the above system, after doing a bit of climbing your character is able to climb freely without any future fear of falling.

I think I have pinpointed the issue. If you start a new character with ACS enabled, your first attempt at climbing by going straight up a wall should behave as expected, your character will slip consistently. It seems however, that if you manage to complete an entire climb, the game stops making future slip checks thus making you a god-tier climber forever.

Looking at the code it is a feature that hanging doesn't do checks, which is fine. I think that it's completing a climb is what might be breaking future checks in ACS. Specific repro steps below.

**To Reproduce**

1. Enable ACS
2. Enable Start outside of Privateer's Hold option
3. Start a new game
4. Choose a character, I went High Rock, Male, Mage, Fast Start
5. Put all your points in anything but STR and Climbing so you end up with a character of about STR 50, Climbing 11% (i.e. quite weak)
6. Fast travel to Gothway Garden
7. Try and climb one of the bigger buildings by just going straight up, holding the move forwards key down but do not let go
8. You will observe that your character keeps slipping repeatedly, as expected from vanilla behaviour
9. Now, try a fresh climb somewhere and complete it
10. All future climbing will now be flawless

**Expected behavior**

With ACS turned on, we should still see weak climbers constantly slip and fall when they are doing anything except hanging, even after their first successful climb.

**Desktop (please complete the following information):**

 - OS: Windows
 - Version: 1.0.0

KABoissonneault:
I can confirm I can reproduce the issue. This code hasn't been touched for at least four years, and has lots of "TODOs" left. I'll try adding some debugging visualization and see where the issue is

KABoissonneault:
![image](https://github.com/Interkarma/daggerfall-unity/assets/5789925/ca2dabe5-71d5-45fd-9105-59ea0d4283c3)

Yeah as suspected, the code has an override for stopping the skill check once you're hanging at the top, but never turns it off.
The override stays on if you reload to a save before climbing (so it is not affected by save/load).
But goes back to false if you reboot the game.

[...]

PR #2561, opened 2024-01-04
Fix for Advanced Climbing bug #2560

KABoissonneault:
Issue at #2560 

Fixed "overrideSkillCheck" not being turned off after being done Climbing. 
overrideSkillCheck is used in the transition from wall to the floor above it, and should only be true during that automatic movement.

image

Moved "ClearStateOnRestore" to always run on restore, not just when climbing with Advanced.

The second fix is not strictly required, but seems cleaner. No risk of regression, since non-advanced climbing should have no expectation of state if saving/reloading during climbing (it's supposed to drop the moment you let the key go).

### DIFF
--- a/Assets/Scripts/Game/Player/ClimbingMotor.cs
+++ b/Assets/Scripts/Game/Player/ClimbingMotor.cs
@@ -1,11 +1,10 @@
 using UnityEngine;
-using System;
-using System.Collections;
 using DaggerfallConnect;
 using DaggerfallWorkshop.Game.Utility;
 using DaggerfallWorkshop.Game.Formulas;
 using DaggerfallWorkshop.Game.Serialization;
-using FullSerializer;
+using Wenzil.Console;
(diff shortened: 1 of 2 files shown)


KABoissonneault:
Things not in the PR but I'm considering:
- the debug hud (I think a setup like this could be useful for future debugging, given that a lot of things we're working with are not from the original implementors)
- a cheat code for flawless climbing (maybe some people will be disappointed that their low climb character falls off much more often now)

Interkarma:
All good if you want to include the debug HUD or a flawless climbing cheat, either in this PR or separate PRs.

Good point on low climb characters being disappointed with the fix.

numidium:
I tested this on my end. My character slips when doing long climbs as expected but if I stop mid-climb, move sideways, and continue moving up and down then I can scale a building like a spider indefinitely without slipping. Not sure if this is intentional.

numidium (approved):

Interkarma (approved):

PR #2562, opened 2024-01-08
Add shading mechanism to first-person graphics

DunnyOfPenwick:
Adds public Tint properties to first-person graphical components so that mods can manipulate the shading.

### DIFF
--- a/Assets/Scripts/Game/FPSWeapon.cs
+++ b/Assets/Scripts/Game/FPSWeapon.cs
@@ -105,6 +105,8 @@ namespace DaggerfallWorkshop.Game
 
         public WeaponStates WeaponState { get { return weaponState; } }
 
+        public Color Tint { get; set; } = Color.white;
+
         #endregion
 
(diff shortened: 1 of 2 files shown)


KABoissonneault:
Is this easily customizable to get a different tint per weapon type and per transport type? From what I understand, you currently only have one tint for the whole manager. But this would make sense if there's an event mods can react to when players select a specific weapon or transport

DunnyOfPenwick:
This change is to support a mod I made a year ago but haven't published due to lack of core code support.
The mod shades weapons/horses to match local lighting conditions, e.g. if the player is standing in darkness, their weapon will be tinted black.

KABoissonneault:
Ah I see, so those changes are always "global" to all weapons/transports. Makes sense, if you wanted to treat different weapons differently, you'd just change that texture's colors

KABoissonneault (approved):

Interkarma (approved):

PR #2563, opened 2024-01-08
Enemy component access

DunnyOfPenwick:
Replaces some method calls in enemy components with delegates so mods can inject additional logic.

Also includes a change to prevent player allies from attacking pacified creatures. (EnemySenses)
Also changes ApplyGravity in EnemyMotor to allow enemies to slowfall.
Also adds option to base enemy spell cost on enemy skills rather than player skills. (EntityEffectManager)

### DIFF
--- a/Assets/Scripts/Game/EnemyMotor.cs
+++ b/Assets/Scripts/Game/EnemyMotor.cs
@@ -325,7 +325,7 @@ namespace DaggerfallWorkshop.Game
         void ApplyGravity()
         {
             // Apply gravity
-            if (entity.IsSlowFalling && !flies && !swims && !controller.isGrounded)
+            if (entity.IsSlowFalling && !flies && !swims && !controller.isGrounded && !IsLevitating)
             {
                 Vector3 velocity = controller.velocity * 0.97f; //gradually slow x/z movement
                 velocity.y = -1; //slow downward fall
@@ -344,10 +344,7 @@ namespace DaggerfallWorkshop.Game
                     CanAct = false;
             }
 
-            if (IsLevitating || entity.IsSlowFalling)
-                LastGroundedY = transform.position.y;
-
-            if (flyerFalls && flies && !IsLevitating)
+            if (flyerFalls && flies && !IsLevitating && !entity.IsSlowFalling)
             {
                 controller.SimpleMove(Vector3.zero);
                 Falls = true;
@@ -1418,6 +1415,9 @@ namespace DaggerfallWorkshop.Game
             // For flying enemies, "lastGroundedY" is really "lastAltitudeControlY"
             else if (flies && !flyerFalls)
                 LastGroundedY = transform.position.y;
+            else if (IsLevitating || entity.IsSlowFalling)
+                LastGroundedY = transform.position.y;
+
         }
 
         /// <summary>
--- a/Assets/Scripts/Game/EnemyMotor.cs
+++ b/Assets/Scripts/Game/EnemyMotor.cs
@@ -98,7 +98,8 @@ namespace DaggerfallWorkshop.Game
         public float LastGroundedY { get; set; }    // Used for fall damage
         public bool Falls { get; private set; }
 
-        //Delegates to allow mods to extend motor behaviour.
+        //============Delegates to allow mods to extend motor behaviour.
+        //==When setting a new handler, it may be desired to store the original and call it before/after your own logic.
         public delegate void TakeActionCallback();
         public TakeActionCallback TakeActionHandler { get; set; }
 
@@ -1412,12 +1413,11 @@ namespace DaggerfallWorkshop.Game
                 LastGroundedY = transform.position.y;
             }
             // For flying enemies, "lastGroundedY" is really "lastAltitudeControlY"
-            else if (flies && !flyerFalls)
-                LastGroundedY = transform.position.y;
-            else if (IsLevitating || entity.IsSlowFalling)
+            else if ((flies && !flyerFalls) || IsLevitating || entity.IsSlowFalling)
                 LastGroundedY = transform.position.y;
 
         }
+
 
         /// <summary>
         /// Open doors that are in the way.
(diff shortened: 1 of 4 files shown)


Assets/Scripts/Game/MagicAndEffects/EntityEffectManager.cs
```diff
@@ -316,8 +318,12 @@ public bool SetReadySpell(EntityEffectBundle spell, bool noSpellPointCost = fals
             if (spell == null || spell.Settings.Version < minAcceptedSpellVersion)
                 return false;
 
+            //By default, enemy spell costs are calculated using player-character skill levels.
```
KABoissonneault, line 321: Is this confirmed to be an original DF bug? I'd rather we just fix that. For what it's worth, we've fixed it for Dodge, and this would be a lot less controversial

DunnyOfPenwick: I brought this up on dfworkshop a while back.
See:
 https://forums.dfworkshop.net/viewtopic.php?p=64057&hilit=spell+cost#p64057

Assets/Scripts/Game/EnemyMotor.cs
```diff
@@ -309,22 +325,32 @@ void KnockbackMovement()
         void ApplyGravity()
         {
             // Apply gravity
-            if (!flies && !swims && !IsLevitating && !controller.isGrounded)
+            if (entity.IsSlowFalling && !flies && !swims && !controller.isGrounded)
+            {
+                Vector3 velocity = controller.velocity * 0.97f; //gradually slow x/z movement
```
KABoissonneault, line 330: Where does this code come from? I thought this was moved from somewhere else, but I can't find it. Is this some mod code you accidentally slipped here, or an intended feature of the PR?

DunnyOfPenwick: It's an intended feature of the PR.
There wasn't any logic in ApplyGravity to handle enemy slowfalling so I added it in.

DunnyOfPenwick: I think I copied the logic from the player motor.

KABoissonneault: I feel like your implementation is a bit too specific for general DFU code. Mostly the part where it slows horizontal movement by 3% every frame. The fall speed is probably debatable too.

I get it's probably a bit hard to add through a mod otherwise... but it could be an event at this point.

I do think it's fine for the "lastGroundedY" to be updated every frame while slow falling otherwise, that should be how it works in all cases.

DunnyOfPenwick: DFU does not currently allow enemies to slowfall, only fly.
In the base game, enemies don't have slowfall in their spellbooks, so it doesn't factor in.
If a mod (like mine) applies a slowfall effect to an enemy or adds it to their spellbook, it would be completely ineffective.

Assets/Scripts/Game/EnemySenses.cs
```diff
@@ -184,8 +183,37 @@ public float TargetRateOfApproach
...
+        public BlockedByIllusionEffectCallback BlockedByIllusionEffect { get; set; }
+
+        public delegate bool CanSeeTargetCallback(DaggerfallEntityBehaviour target);
+        public CanSeeTargetCallback CanSeeTarget { get; set; }
+
+        public delegate bool CanHearTargetCallback();
+        public CanHearTargetCallback CanHearTarget { get; set; }
+
+        public delegate bool CanSeePlayerLightCallback();
+        public CanSeePlayerLightCallback CanSeePlayerLight { get; set; }
```
KABoissonneault, line 206: I like the extra feature for detecting players, but I don't see why it should have the name "light". It's probably how your mod is going to use it. However, other mods might run other kinds of checks there.

I'm fine with it if you can rename it to something like "CanSeePlayerExtra" or "CanSeePlayerOther"

Assets/Scripts/Game/EnemySenses.cs
```diff
@@ -624,7 +655,7 @@ public bool StealthCheck()
             return Dice100.FailedRoll(stealthChance);
         }
 
-        public bool BlockedByIllusionEffect()
+        public bool BlockedByIllusionEffectDefault()
```
KABoissonneault, line 658: I'm a bit bothered by the change of a public API. After 1.0, this should generally be unacceptable. However, I doubt any mod is using this specifically when none of the other functionality is exposed.

KABoissonneault:
I'm almost ready to accept this, if not for the CanSeePlayerLight comment

KABoissonneault (approved):

Assets/Scripts/Game/EnemySenses.cs
DunnyOfPenwick: Yes, that is oddly specific.  I'm changing it to CanDetectPlayerOther().

Assets/Scripts/Game/EnemySenses.cs
DunnyOfPenwick: Oops, I didn't notice that was a public API.  I could revert the method name and rename the member if need be.

Assets/Scripts/Game/EnemySenses.cs
DunnyOfPenwick: I changed it again to CanDetectOtherwise and removed the player-specific check.

Assets/Scripts/Game/EnemySenses.cs
DunnyOfPenwick: I reverted the method name back to BlockedByIllusionsEffect.

Assets/Scripts/Game/EnemyMotor.cs
```diff
@@ -1385,11 +1409,14 @@ void ApplyFallDamage()
                     }
                 }
 
-                lastGroundedY = transform.position.y;
+                LastGroundedY = transform.position.y;
             }
             // For flying enemies, "lastGroundedY" is really "lastAltitudeControlY"
             else if (flies && !flyerFalls)
```
KABoissonneault, line 1415: Now I feel I gotta nitpick and ask you to put write this as `if((flies && !flyerFalls) || IsLevitating || entity.IsSlowFalling)`

DunnyOfPenwick: Done.

KABoissonneault (approved):

Interkarma (approved):

PR #2566, opened 2024-01-26
Fix typo that breaks 3d object scaling in RMBLayout.cs

drcarademono:
Modders working with RMBs have known for a long time that rescaling 3d objects will break them. Either the scale will be wrong or objects will disappear entirely when scaled. The screenshot below demonstrates this bug with a modded RMB but otherwise unmodded game:
Screenshot from 2024-01-25 20-58-56

This bug turns out to be caused by a typo in the GetModelScaleVector method, in which obj.ZScale is incorrectly set to obj.YScale on line 76. Correcting this typo fixes the scaling bug:
Screenshot from 2024-01-25 20-57-23


drcarademono:
Grep searching for the string "Scale" in a folder with all RMB jsons extracted from DF data shows that vanilla Daggerfall objects are never scaled at all. This change should only affect modded RMBs.

KABoissonneault (approved):

numidium (approved):

PR #2570, opened 2024-02-01
RmbBlockEditor file cleanup

KABoissonneault:
This is what appears in my Git status when I checkout master and open Unity now

image

I wish I'd caught this before merging the PR.

Just adding the meta files with default values - if anything else was intended, it should be fixed by whoever knows what these files should have.

I'm adding Editor settings to the .gitignore. I figure these files should be local to each repo


Jagget (approved):

petchema:
Mmmh so far we've committed *.meta files in the repository, and personally I think it would be safer in the long run to also commit those (even if right now it will raise merge conflict for some), say if we want to tweak the metas of the PNG files. 
Not sure what those *.json files contain though (buildings-catalog.json etc)

KABoissonneault:
> Mmmh so far we've committed *.meta files in the repository, and personally I think it would be safer in the long run to also commit those (even if right now it will raise merge conflict for some), say if we want to tweak the metas of the PNG files. Not sure what those *.json files contain though (buildings-catalog.json etc)

And I am, for files which are persistent. 

Settings are trickier, they only get created after usage. These would get lots of local changes, so even if I committed an empty version of those files, they would show in the Git status after usage without any intention of actually committing them to the repo. I think .gitignore for any settings is fine (I'd even add ProjectSettings/EditorSettings.asset to be honest, it's a permanent change for me).

Podleron:
Ah, hold on, you've actually added the meta files, not removed them.
I thought that any generated files should not be included in the code, but to be honest I don't know enough about them and what they do exactly. I will go read up on it and in the meantime the change is ok by me.

Podleron (approved):

KABoissonneault:
@petchema You think you could approve the PR? You're one of the few people who count for the "2 approving reviews" requirements (sorry Podleron, you don't have write access to this repo).

I wouldn't say no to some of my other PRs too :P

petchema (approved):
LGTM

Interkarma (approved):

Issue #2569, opened 2024-01-30
Exceptions in DaggerfallMobileUnit


KABoissonneault:
Users are experiencing frequent errors with the following callstack.

```
DaggerfallWorkshop.DaggerfallMobileUnit.OrientEnemy (System.Int32 orientation) (at <9bf28c68d79b48218c80373c9e1cf18f>:0)
DaggerfallWorkshop.DaggerfallMobileUnit.UpdateOrientation () (at <9bf28c68d79b48218c80373c9e1cf18f>:0)
DaggerfallWorkshop.DaggerfallMobileUnit.Update () (at <9bf28c68d79b48218c80373c9e1cf18f>:0)
```

The specific exception is "IndexOutOfRangeException: Index was outside the bounds of the array."

The problem is entirely contained in DFU code, but I suspect the cause might be mod related. Probably custom enemies or some graphics mod.

It could be a lot of arrays that have out-of-bounds... The MobileUnit summary has `StateAnims`, `RecordSizes`, `ImportedTextures` with `Albedo` and `EmissionMaps`, `AtlasRects`

Because of the random nature of this issue, I'd like to remedy this with proper sanitization and logs first. A verbose log is probably better than throwing an exception every frame. After that, I can know what exactly is invalid and fix the real issue later

KABoissonneault:
Fixed by #2571

PR #2571, opened 2024-02-03
Fixed frame overflow in DaggerfallMobileUnit. 

KABoissonneault:
Fixes for #2569. 

I found an issue where archive 288 has 4 frames for record 0, 1, 2, and 4, but 8 frames for record 3. Record 3 is the "back diagonal" sprites (both left and right). Normally, all mobile archives have the same number of frames on each orientation, but the Ancient Lich is an exception. When changing orientation, the `currentFrame` is preserved so that the animation keeps playing from where it was. However, for the Ancient Lich, if you change orientation from the back diagonal while on frame 4, 5, 6, or 7 is playing, the `currentFrame` will now be past the allowed frames for the new orientation (back or side). On unmodded DFU, this reaches into a "Casting spell" frame in `summary.AtlasRect` (because `rectIndex` has enough room to overflow into the next records, rather than go out of bounds of the AtlasRect). When using texture replacements from mods (ex: Vanilla Enhanced), you get the exception shown in #2569 (because we don't use the atlas anymore).

The fix is on line 371 - I simply reset `currentFrame` to 0 if the frame is not valid for the new orientation. A bit jarring, but better than the current behavior.

The rest of the changes are more bounds checking and error logs for these issues. I find DaggerfallMobileUnit has lots of moving parts and lots of arrays to go out of bounds, so it helps to have better logs there.

The output looks like this:

image

Hopefully this will help diagnose future issues with mobiles, if any remain. In any case, with mods like Daggerfall Enemy Expansion, it's easy to accidentally create invalid MobileEnemy data that creates random errors in DaggerfallMobileUnit, and this will help mod development.

### DIFF
--- a/Assets/Scripts/Internal/DaggerfallMobileUnit.cs
+++ b/Assets/Scripts/Internal/DaggerfallMobileUnit.cs
@@ -23,6 +23,7 @@ using DaggerfallConnect.Arena2;
 using DaggerfallWorkshop.Utility;
 using DaggerfallWorkshop.Utility.AssetInjection;
 using DaggerfallWorkshop.Game.Utility;
+using DaggerfallWorkshop.Game;
 
 namespace DaggerfallWorkshop
 {
@@ -161,6 +162,12 @@ namespace DaggerfallWorkshop
             }
         }
 
+        protected void LogMobileError(string  message)
+        {
+            string enemyName = TextManager.Instance.GetLocalizedEnemyName(summary.Enemy.ID);
+            Debug.LogError($"Enemy '{enemyName}' (state={summary.EnemyState}): {message}");
+        }
+
         protected override void ApplyEnemy(DaggerfallUnity dfUnity)
         {
             // Load enemy content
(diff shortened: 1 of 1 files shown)


petchema:
It's hard to talk about a "best solution" without knowing the original intent (and checking what's going on in Daggerfall classic will probably be difficult), but at least for the record, since both 4 and 8 frame cycles picture a single walking cycle (left foot forward, right foot forward), another solution would be to apply cross-multiplication when orientation changes
```
--- a/Assets/Scripts/Internal/DaggerfallMobileUnit.cs
+++ b/Assets/Scripts/Internal/DaggerfallMobileUnit.cs
@@ -328,6 +328,8 @@ namespace DaggerfallWorkshop
             int orientation = - Mathf.RoundToInt(enemyFacingAngle / anglePerOrientation);
             // Wrap values to 0 .. numberOrientations-1
             orientation = (orientation + numberOrientations) % numberOrientations;
+            if (lastOrientation >= 0)
+                currentFrame = currentFrame * summary.StateAnims[orientation].NumFrames / summary.StateAnims[lastOrientation].NumFrames;
 
             // Change enemy to this orientation
             if (orientation != lastOrientation)
```

Assets/Scripts/Internal/DaggerfallMobileUnit.cs
```diff
@@ -331,7 +338,17 @@ private void UpdateOrientation()
 
             // Change enemy to this orientation
             if (orientation != lastOrientation)
+            {
+                // Different orientations may not have the same amount of frames
+                // For example, archive 288 (Ancient Lich) has 4 frames in six orientations (front, back, front diagonals, sides), but 8 frames for the two back diagonals
+                // If you change orientation from a back diagonal during frame 4, 5, 6, or 7, you overflow the other anims, where only 0 to 3 are valid
+                if (lastOrientation >= 0)
+                {
+                    currentFrame = currentFrame * summary.StateAnims[orientation].NumFrames / summary.StateAnims[lastOrientation].NumFrames;
```
Interkarma, line 347: This is really clever, I like this

Interkarma (approved):

numidium (approved):

PR #2574, opened 2024-02-05
Seek localized book files and quest files from mods

Jagget:
Seek localized book files and quest files from mods

Loose files can be packed, and it will make installing new books mods and quest localizations easier and more straightforward.


KABoissonneault (approved):
Love a PR that removes some TODOs

Interkarma (approved):

PR #2578, opened 2024-02-09
Updated DFU tutorial, minor quest fixes

JayH2971:
Warning: I don't know what "multiple PRs" look like. I am a github novice.

This contains very basic quest fixes, almost unnoticeable. It also has my updated DFU tutorial in file _TUTOR__.txt.

### DIFF
--- a/Assets/StreamingAssets/Quests/A0C01Y01.txt
+++ b/Assets/StreamingAssets/Quests/A0C01Y01.txt
@@ -119,8 +119,8 @@ _S.00_ task:
 	end quest 
 
 _S.01_ task:
-	daily from 21:00 to 23:59
-	--changed 24:00 to 23:59. 24:00 shouldn't exist.
+	daily from 21:00 to 22:59
+	--changed 24:00 to 23:59. 24:00 shouldn't exist. Goes to 11 PM due to "latehour" task added below.
 	place npc _qgiver_ at _qgiverhome_ 
 
 _S.02_ task:
--- a/Assets/StreamingAssets/Quests/A0C01Y01.txt
+++ b/Assets/StreamingAssets/Quests/A0C01Y01.txt
@@ -119,7 +119,7 @@ _S.00_ task:
 	end quest 
 
 _S.01_ task:
-	daily from 21:00 to 22:59
+	daily from 21:00 to 23:59
 	--changed 24:00 to 23:59. 24:00 shouldn't exist.
 	place npc _qgiver_ at _qgiverhome_ 
 
--- a/Assets/StreamingAssets/Quests/M0B00Y06.txt
+++ b/Assets/StreamingAssets/Quests/M0B00Y06.txt
@@ -40,27 +40,19 @@ QuestComplete:  [1004]
 <ce>                       Here's your _gold_ gold.
 
 RumorsDuringQuest:  [1005]
-<ce>                   Did you hear? An animal got into
-<ce>                    a building in town. I hope they
-<ce>                             kill it soon.
+Did you hear? An animal got into a building in town. I hope they kill it soon.
 
 RumorsPostfailure:  [1006]
-<ce>                    The Fighter's Guild killed that
-<ce>                      bear that got into _house_.
-<ce>                 It's a good thing we have them around.
+ hat creature mauled some woman before being chased out of town. What a pity.
 
 RumorsPostsuccess:  [1007]
-<ce>                That creature mauled some woman before
-<ce>                being chased out of town. What a pity.
+The Fighter's Guild killed that bear that got into _house_. It's a good thing we have them around.
 
 QuestorPostsuccess:  [1008]
-<ce>                     Good job killin' that beast.
-<ce>                        What can I do for you?
+Good job killin' that beast. What can I do for you?
 
 QuestorPostfailure:  [1009]
-<ce>                   I hear that a critter ate the guy
-<ce>                    that owned the house. Gave us a
-<ce>                 bad rep. Try to do better next time.
+I hear that a critter ate the guy that owned the house. Gave us a bad rep. Try to do better next time.
 
 QuestLogEntry:  [1010]
 %qdt:
--- a/Assets/StreamingAssets/Quests/M0B00Y06.txt
+++ b/Assets/StreamingAssets/Quests/M0B00Y06.txt
@@ -43,7 +43,7 @@ RumorsDuringQuest:  [1005]
 Did you hear? An animal got into a building in town. I hope they kill it soon.
 
 RumorsPostfailure:  [1006]
- hat creature mauled some woman before being chased out of town. What a pity.
+That creature mauled some woman before being chased out of town. What a pity.
 
 RumorsPostsuccess:  [1007]
 The Fighter's Guild killed that bear that got into _house_. It's a good thing we have them around.
--- a/Assets/StreamingAssets/Quests/_TUTOR__.txt
+++ b/Assets/StreamingAssets/Quests/_TUTOR__.txt
@@ -40,14 +40,17 @@ QuestLogEntry:  [1010]
 Message:  1012
 <ce>                           Page 1
 <ce>                                    
-<ce>           The very first thing you want to do is to check your keybinds.
-<ce>           Press Escape and click on Controls. The most important ones for now are
-<ce>           your movement (WASD) and your Swing Weapon (Mouse1 by default).
-<ce>           Attacking may involve holding the Swing Weapon button down and moving the mouse;
-<ce>           OR clicking the Swing Weapon button, OR holding it down. This can be changed
+<ce>           The very first thing you want to do is to check
+<ce>           your keybinds. Press Escape and click on Controls.
+<ce>           The most important ones for now are your movement (WASD)
+<ce>           and your Swing Weapon (Mouse1 by default). Attacking
+<ce>           may involve holding the Swing Weapon button down
+<ce>           and moving the mouse; OR clicking the Swing Weapon
+<ce>           button, OR holding it down. This can be changed
 <ce>           in the Advanced section of the Controls screen.
 <ce>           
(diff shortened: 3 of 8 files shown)


Assets/StreamingAssets/Quests/A0C01Y01.txt
```diff
@@ -119,7 +119,7 @@ _S.00_ task:
 	end quest 
 
 _S.01_ task:
-	daily from 21:00 to 22:59
+	daily from 21:00 to 23:59
```
KABoissonneault, line 122: When you change this from 23:59 to 22:59, you added the `_latehour_` task below.
https://github.com/Interkarma/daggerfall-unity/commit/21fadbd6cbf32888bb070fcbe786bd5e4f87f8db

Is the `_latehour_` still needed or not?

Assets/StreamingAssets/Quests/M0B00Y06.txt
```diff
@@ -40,27 +40,19 @@ QuestComplete:  [1004]
...
-<ce>                   Did you hear? An animal got into
-<ce>                    a building in town. I hope they
-<ce>                             kill it soon.
+Did you hear? An animal got into a building in town. I hope they kill it soon.
 
 RumorsPostfailure:  [1006]
-<ce>                    The Fighter's Guild killed that
-<ce>                      bear that got into _house_.
-<ce>                 It's a good thing we have them around.
+ hat creature mauled some woman before being chased out of town. What a pity.
```
KABoissonneault, line 46: Missing a T at the start.

KABoissonneault:
Are the <ce> tokens on rumors an issue?

From what I'm seeing so far, it seems quests generally tend to not have <ce> on Rumors, but generally tend to have it on Questor. This makes sense to me, since the rumors is in the Talk UI, while the Questor is a popup.
Your changes seems to remove them on both Rumors and Questor, though. Was it intended to remove them from Questor? Also, should we do a pass and remove <ce> from all rumors?

Assets/StreamingAssets/Quests/A0C01Y01.txt
JayH2971: Good eye. I added latehour to ensure the enemy would have to spawn somehow, and that happens at 11PM. I restored the original spawning window and kept latehour as it is.

Assets/StreamingAssets/Quests/_TUTOR__.txt
```diff
@@ -9,388 +9,355 @@ Quest: _TUTOR__
 QRC:
 
 RumorsDuringQuest:  [1005]
-<ce>                       I'll give you a hint %ra.
-<ce>                   If you want to succeed in magic,
-<ce>                         join the mages guild.
+I'll give you a hint %ra.  If you want to succeed in magic, join the mages guild.
```
KABoissonneault, line 12: You got a few double spaces in there. I'd do a pass with Find "  " and replace it with " "

JayH2971:
> This makes sense to me, since the rumors is in the Talk UI, while the Questor is a popup.
The questor only gives a popup message if it's invoked in a quest action like "say" or "givepcreward." The messages  "QuestorPostfailure" and "QuestorPostsuccess" only appear if you enter the conversation window with the questgiver after the quest ends. A lot of them are misformatted, and I haven't fixed them.

For the other questions, I need some more clarification on them.

KABoissonneault:
After discussing offline, it seems any "Talk UI" message on multiple lines does not get displayed properly. This is an issue with many quests in the game. We might wanna just fix them all

Jagget:
Are you saying that RumorsDuringQuest, RumorsPostfailure, RumorsPostsuccess, QuestorPostsuccess, and QuestorPostfailure only ever displayed in dialog interface, and that all 5 sections do not need formatting? 

Btw, there are `say QuestorPostsuccess` in S0000003, and `say QuestorPostfailure` in S0000013

JayH2971:
Yes, those five fields only are used in the Talk UI. Kab tested it earlier today, and <ce> lines and multiple-line formatting do not interfere with their appearance inside the Talk UI; the message is displayed fine.

Finding Questor messages in the main quest is odd but that's classic Daggerfall data. No doubt one of many things they changed during development.

Jagget (approved):

KABoissonneault:
Yes I want to be on record that after testing offline, the Talk UI does handle this, by simply replacing <ce> tokens with spaces. Gonna run the PR locally to see what the tutorial looks like in action now before approving

ajrb:
I've read this thread a couple of times and I am still confused. Are formatting tags like <ce> an issue for rumors / postsuccess etc message that are used in talk window really an issue? I'd like to know because I have 2 blocks like that in my R&R quest.

KABoissonneault:
@ajrb In my tests, no. When I initially reported this, this was based on my conversation with JayH, but after that conversation, I was not able to reproduce the issue. We're moving forward as if it's not an issue. However, I still think it's better if new quests avoid using <ce> on Talk text, simply before it's not used and can lead to confusion when reading the quest

KABoissonneault:
I _think_ I'd like to prioritize this for 1.1. The earlier we release this, the more feedback we can get. However, I think this could be a controversial change, if it were to be less useful than the original. Maybe some purists won't like it, but this can probably be modded back in easily.

I'll test it myself, and maybe we could playtest it on Lysandus' Tomb?

KABoissonneault:
Some feedback on the Tutorial quest

image

- The swing type can be changed anytime in Controls -> Advanced. I'd remove mention of the title screen, to avoid having people restart the game to change swing type. Just mention that after Controls, it's under  Advanced
- Is there supposed to be a Yes No?

image

- Along with the first one, I find 5 seconds for a call to action might be a bit short. I'm worried someone will start looking at their keyboard for too long, and then the next tutorial will start. It's not like the user is stuck waiting, I think it's fine to have slightly longer waits. 15 seconds on could be fine. I guess we could playtest it and adjust.

image

- Not sure about the phrasing on the saving throws. Yes, some enemies have less than 50% on their saving throws, but I don't want to make it seem like magic is only meant to be used against them. Having half your spells miss is how magic is expected to work. Maybe just mention that it's normal for half of your spells to miss.

image
- Not sure about the suggestion of Daggerfall. Gothway Garden is an iconic entry town. Pretty arbitrary I suppose

image
- Knightly Orders are also unique, might be worth mentioning

image
- The tutorial avoids mentioning hotkeys for the most part. I kind of get why you do it here, but worth considering that they might have been changed

image
- Yay

Overall, I'm kind of okay with the short timers on messages. I get it might be better to just get them over with. Just one thing, is that I find it pretty difficult to get the RumorsDuringQuest messages with your version. Do you think we could replace them with RumorsPostSuccess? You'd get helper tips for 7 days after reading all of it. You still get the basic plot hooks if you said No to the quest (assuming you make it to a city in 7 days).

JayH2971:
Okay, sounds good. Most of those are small changes, not hard to do. The reason I chose Daggerfall City is because it's the only place with a permanent location in the region of Daggerfall, so I thought I could do a "when enters" condition on it, but even that failed, so I changed it to "any city." I'll have to look up what location Gothway Garden is.

KABoissonneault:
Information I like from the original that I see missing here.

Object activation and loot piles
image
- At least the part about activating the door and telling users to check corpses for loot
- Page 9 does tell you about activating the throne later on, but I feel the first door and basic loot piles are essential to mention early

Inventory window
image
- A bit more awkward since messages go by so far, you'd get this prompt before you actually found it. Maybe it's fine

> The reason I chose Daggerfall City is because it's the only place with a permanent location in the region of Daggerfall

Yeah to be fair so did the original. My bad, it's fine like this
image

JayH2971:
Good points, I'm fine with them. I'll try to find a chance to change them.

Assets/StreamingAssets/Quests/_TUTOR__.txt
```diff
@@ -8,7 +8,8 @@ Quest: _TUTOR__
 -- Message panels
 QRC:
 
-RumorsDuringQuest:  [1005]
+
+RumorsPostfailure:  [1006]
```
KABoissonneault, line 12: Do these only occur if you say No to the tutorial, or does completing the tutorial to the end count as failure? I guess I'll test it out

KABoissonneault: Oh just saw that the last page just calls the _no_ task, got it

JayH2971: The way this is written, all ends lead to failure. The only way to get a quest to be a success is if you do "give pc ITEM" of some kind, or "give pc nothing." They invoke message 1004 as well. Anything without them is a failure.

KABoissonneault (approved):

KABoissonneault (changes requested):
I think you should do a pass with SDF Fonts turned off to make sure the maximum width is similar to the original
image

JayH2971:
It's too bad I'm only learning about this now. A lot of QP1 would probably have the same problem. But for this I think it's fixed, I tested every window in-game on non-SDF now.

KABoissonneault:
Much better. I still wish the message boxes would fit inside retro frames, but maybe we need a tool to detect and message sizes in quests instead. We'll go with this for now

image
image

KABoissonneault (approved):

KABoissonneault (approved):

Mentill:
I don't know how much additional work was going to be put into this, but as an FYI I've proofread all of the quest files and intend to do a PR for them if/when my current PR is approved.
I'll work my changes around these, however that'll be easier if fewer quest files have typos fixed here.

JayH2971:
I don't intend to add anything else to this PR, unless someone finds a game-breaking bug of some kind. It's easier on Kab to let this one settle, so it doesn't have to get approved again.

I'll defer to the other maintainers on the question of text fixes.

petchema (approved):

PR #2579, opened 2024-02-10
Support loading font file from .dfmod

Jagget:
- I changed `ModManager` load priority, so now it initialized before `DaggerfallUI`.
- Init logic of the `ModManager` was moved from `Start()` to `Awake()`

image

### DIFF
--- a/Assets/Game/Addons/ModSupport/ModManager.cs
+++ b/Assets/Game/Addons/ModSupport/ModManager.cs
@@ -134,9 +134,6 @@ namespace DaggerfallWorkshop.Game.Utility.ModSupport
             if (string.IsNullOrEmpty(ModDirectory))
                 ModDirectory = Path.Combine(Application.streamingAssetsPath, "Mods");
 
-            // Rest of the logic is moved from Start() method.
-            // There's nothing we cannot initialize right away.
-
             SetupSingleton();
 
             if (Instance == this)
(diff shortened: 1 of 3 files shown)


KABoissonneault:
I understand that shipping fonts in dfmods is one of the last steps to let a localization mod ship only a single file. I do think we can prioritize this for DFU 1.1. That font is needed any time text is shown to the user.

Can you send me a font file I can use for testing this? I have only one concern for this change. I wanna make sure the font is properly removed if the user disables the mod and restarts the game. I'm a bit hesitant to introduce more mods that affect the Launcher, after the mod "Mute Launcher Music" caused user confusion.

image

Jagget:
> shipping fonts in dfmods is one of the last steps to let a localization mod ship only a single file
BIOGs. There's always BIOGs... 

Here's the dfmod with fonts attached. For testing  [fonts.zip](https://github.com/Interkarma/daggerfall-unity/files/14325963/fonts.zip)

KABoissonneault:
I removed the comment from ModManager.Awake. No need to put comments about the "past state" of the code in the code, you can keep that in the commit message and PR review comments.

I didn't know about Script Execution Order settings. Here's an example of the DFU execution order values after this PR.
image

Some things I'm noticing while testing.

- Disabling the mod system still keeps the font active

image

That one I don't really blame on the PR. The code in ModManager that checks for `if (!DaggerfallUnity.Settings.LypyL_ModSystem)` destroys `this` but does not end the initialization of the mod manager. It seems like disabling the mod system is pretty broken to begin with.

- Starting the main menu with the "fonts" mod disabled then enabling the "fonts" mod will not apply the font on the main screen, but it will apply it on new UI screens like the Settings. Might be a bit complicated to fix, and it's not that bad

Overall, I think it's good enough for v1.1.1. It should reduce the number of files that localization mods have to ship

KABoissonneault (approved):

KABoissonneault:
I'd really like to get this merged for the v1.1.1 official release soon.

numidium (approved):

PR #2580, opened 2024-02-11
Add ability to declare QuestLists in .dfmod file.

Jagget:
Currently you can activate new Quest Pack (quest list) by adding the file to StreamingAssets\QuestPacks\ folder.

In case of packing quests in .dfmod file, you must call `QuestListsManager.RegisterQuestList("MyQuests")`.

It is not possible to ship quests in .dfmod without C# script.

I propose to register them in "Contributes" ModInfo section, the same way as BooksMappings 

Example:

```json

{
    "ModTitle": "QuestPack",
    "ModVersion": "1.0.0",
    "Files": [
        "QUEST1.txt",
        "QUEST2.txt",
        "QUEST3.txt",
        "QuestList-MyQuests.txt"
    ],
    "Contributes": {
        "QuestLists": [
            "MyQuests"
        ]
    }
}

```
image

image

Screenshot 2024-02-10 212729

### DIFF
--- a/Assets/Game/Addons/ModSupport/ModManager.cs
+++ b/Assets/Game/Addons/ModSupport/ModManager.cs
@@ -942,13 +942,14 @@ namespace DaggerfallWorkshop.Game.Utility.ModSupport
         /// Seeks asset contributes for the target mod, reading the folder name of each asset.
         /// </summary>
         /// <param name="modInfo">Manifest data for a mod, which will be filled with retrieved contributes.</param>
+        /// <param name="automaticallyRegisterQuestLists">Optional parameter that triggers Quest Lists declaration</param>
         /// <remarks>
         /// Assets are imported from loose files according to folder name,
         /// for example all textures inside `SpellIcons` are considered icon atlases.
         /// This method replicates the same behaviour for mods, doing all the hard work at build time.
         /// Results are stored to json manifest file for performant queries at runtime.
         /// </remarks>
-        public static void SeekModContributes(ModInfo modInfo, bool automaticallyRegisterQuestLists)
+        public static void SeekModContributes(ModInfo modInfo, bool automaticallyRegisterQuestLists = false)
         {
             List<string> spellIcons = null;
             List<string> booksMapping = null;
@@ -958,15 +959,15 @@ namespace DaggerfallWorkshop.Game.Utility.ModSupport
             {
                 var directory = Path.GetDirectoryName(file);
 
-                if (directory != null && directory.EndsWith("SpellIcons"))
+                if (!string.IsNullOrEmpty(directory) && directory.EndsWith("SpellIcons"))
                     AddNameToList(ref spellIcons, file);
-                else if (directory != null && directory.EndsWith("Books/Mapping"))
+                else if (!string.IsNullOrEmpty(directory) && directory.EndsWith("Books/Mapping"))
                     AddNameToList(ref booksMapping, file);
 
                 if (automaticallyRegisterQuestLists)
                 {
                     var name = Path.GetFileNameWithoutExtension(file);
-                    if (name.StartsWith("QuestList-"))
+                    if (!string.IsNullOrEmpty(name) && name.StartsWith("QuestList-"))
                         AddNameToList(ref questLists, name.Substring(10));
                 }
             }
--- a/Assets/Game/Addons/ModSupport/ModManager.cs
+++ b/Assets/Game/Addons/ModSupport/ModManager.cs
@@ -948,27 +948,36 @@ namespace DaggerfallWorkshop.Game.Utility.ModSupport
         /// This method replicates the same behaviour for mods, doing all the hard work at build time.
         /// Results are stored to json manifest file for performant queries at runtime.
         /// </remarks>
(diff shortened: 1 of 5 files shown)


ajrb:
My thinking behind this was that if you were packaging quests into a mod then you'd very likely have some c# script even if just to set things up and act on settings etc. The quest pack idea was because Jay (and others) didn't want to create a dfmod package and script files etc for purely quests. Given some of the abuse of DFU quest packs have enabled I am not sure it was a good idea, but there you go. :-)

This may have been a better way to go than questpacks, but unless we intend to deprecate and remove quest pack functionality I don't really see what this achieves. What am I missing?

KABoissonneault:
> My thinking behind this was that if you were packaging quests into a mod then you'd very likely have some c# script even if just to set things up and act on settings etc. The quest pack idea was because Jay (and others) didn't want to create a dfmod package and script files etc for purely quests. 

The culture has evolved, even quest-only modders can press Build on the Mod Builder nowadays, without any scripting knowledge. 

> Given some of the abuse of DFU quest packs have enabled I am not sure it was a good idea, but there you go. :-)

I'm with you in avoiding mods using quests to create unrelated systems, but DFU has to support them regardless.

Outside of that, my only issue with the PR is that it seems to require modifying the JSON manually to enable the feature, which is a step better than C#, but should probably be a Mod Builder feature. Granted, I think we can push that part to a later PR

Jagget:
> Outside of that, my only issue with the PR is that it seems to require modifying the JSON manually to enable the feature, which is a step better than C#, but should probably be a Mod Builder feature. Granted, I think we can push that part to a later PR

It is actually the same for Bookmappings and Spellicons. There's no editor for "Contributes" part of the `dfmod.json`

KABoissonneault:
Yeah I recognize that it's an existing issue, we'll consider it later if users want one.

I want to add that in terms of having QuestPacks in .dfmod, I have observed users confused as to why Quest Pack 1 and other quest mods were not showing up in the DFU mod list. I think moving forward, I would prefer if all mods came in .dfmod files, and keep loose files for development and very specific user configuration purposes. So this PR has my approval

Assets/Game/Addons/ModSupport/ModManager.cs
```diff
@@ -948,27 +948,36 @@ public static string MakeRelativePath(string path0, string path1)
...
+            List<string> questLists = null;
 
-            foreach (string file in modInfo.Files)
+            foreach (var file in modInfo.Files)
             {
-                string directory = Path.GetDirectoryName(file);
+                var directory = Path.GetDirectoryName(file);
 
-                if (directory.EndsWith("SpellIcons"))
+                if (directory != null && directory.EndsWith("SpellIcons"))
```
KABoissonneault, line 961: For strings, it makes more sense to check `string.IsNullOrEmpty(directory)`. Especially since Path.GetDirectoryName is specified to return an empty string sometimes: https://learn.microsoft.com/en-us/dotnet/api/system.io.path.getdirectoryname

Assets/Game/Addons/ModSupport/ModManager.cs
```diff
@@ -948,27 +948,36 @@ public static string MakeRelativePath(string path0, string path1)
         /// This method replicates the same behaviour for mods, doing all the hard work at build time.
         /// Results are stored to json manifest file for performant queries at runtime.
         /// </remarks>
-        public static void SeekModContributes(ModInfo modInfo)
+        public static void SeekModContributes(ModInfo modInfo, bool automaticallyRegisterQuestLists)
```
KABoissonneault, line 951: Could have used a defaulted =false, but worst case it breaks someone's Editor tools, it doesn't break any mods

Jagget: Fixed

ajrb:
Thanks for the answers.  Looks good to me.

ajrb (dismissed):

Assets/Game/Addons/ModSupport/ModManager.cs
```diff
@@ -942,33 +942,43 @@ public static string MakeRelativePath(string path0, string path1)
...
         /// for example all textures inside `SpellIcons` are considered icon atlases.
         /// This method replicates the same behaviour for mods, doing all the hard work at build time.
         /// Results are stored to json manifest file for performant queries at runtime.
         /// </remarks>
-        public static void SeekModContributes(ModInfo modInfo)
+        public static void SeekModContributes(ModInfo modInfo, bool automaticallyRegisterQuestLists = false)
         {
             List<string> spellIcons = null;
             List<string> booksMapping = null;
+            List<string> questLists = null;
```
KABoissonneault, line 956: I think this function should make `modInfo.Contributes.QuestLists` empty if `automaticallyRegisterQuestLists` is false. Right now, you can't disable the feature from the UI. Unchecking the box and resaving does nothing, it's gonna be checked again next time you open it, and the quests will be registered, because they are in the contributes still

KABoissonneault:
Having finally tested this, the feature is almost all good, save one last outstanding issue. Gonna approve after this

KABoissonneault (approved):

ajrb (approved):

PR #2581, opened 2024-02-11
Fix booleans returned when setting variants

ajrb:
The return values were all inverted erroneously, rather embarrassing. Not seen anyone using these return values yet, so should cause no issues.

### DIFF
--- a/Assets/Scripts/Utility/AssetInjection/WorldDataVariants.cs
+++ b/Assets/Scripts/Utility/AssetInjection/WorldDataVariants.cs
@@ -95,14 +95,14 @@ namespace DaggerfallWorkshop.Utility.AssetInjection
         /// <param name="regionIndex">Region index</param>
         /// <param name="locationName">Location name (as index is assigned at runtime)</param>
         /// <param name="variant">Variant name</param>
-        /// <returns>True if overwriting an existing variant set for this location</returns>
+        /// <returns>True unless this overwrote an existing variant set for this location</returns>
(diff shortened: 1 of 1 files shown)


KABoissonneault:
Wouldn't it be safer to just update the comments to "True if this X had no existing variants", and changing the variable name? It's not like either way is more intuitive for this function, I think as long as the docs match the behavior, it should be good

ajrb:
It was wrong and I fixed it. I don't feel it needs the API changing for safety however if that's the consensus I will change it.

ajrb:
I've done as requested. 
Added method documentation for getters.
Added new block variant getter for when not iterating locations. (i.e. most mod usage)

Assets/Scripts/Utility/AssetInjection/WorldDataVariants.cs
```diff
@@ -218,6 +235,28 @@ public static string GetBuildingVariant(ref BlockRecordKey blockRecordKey, strin
             return NoVariant;
         }
 
+        /// <summary>
+        /// Gets a variant set for the specified building, based on supplied region and location.
```
KABoissonneault, line 239: >Gets a variant set for the specified building"
Should that be "block"?

ajrb: Yep, thanks for the catch.

KABoissonneault (approved):

numidium (approved):

PR #2582, opened 2024-02-11
Fixed Temple music errors

KABoissonneault:
Temple music is handled differently from other playlists in SongManager. The selected song is based on the faction of the building (the manager checks for both the "Temple" faction id and the "God" faction id). The songs are as follows:

| God       | Song    |
|-----------|---------|
| Arkay     | Good    |
| Z'en      | Bad     |
| Mara      | Good    |
| Akatosh   | Neutral |
| Julianos  | Bad     |
| Dibella   | Good    |
| Stendarr  | Bad     |
| Kynareth  | Neutral |

I found two errors in how we handle this.

First, we use separate SongPlayer objects for Exterior and Interior. This means that when moving from an Interior to outside, the Interior SongPlayer is not updated until we enter a new interior. Therefore, if we move from one temple to outside to another temple, from the perspective of the Interior SongPlayer, the context never changed, and the music is not updated, even if the FactionID would cause a different song.

This can be tested in the Daggerfall region's Chesterwick Hamlet (not Chesterwick), which has both a Julianos and Kynareth temple. Both temples should have a different music, but if you enter one then directly enter the other, then the music is preserved from the first temple. The music is properly updated if you enter a non-temple interior then go to the temple.

So, I added "factionID" to the music context, which is only updated for temples since only that playlist actually checks it. When the playlist is the same as previously but the factionID changes, we update the song only if the playlist is Temples.

The second issue is what I initially investigated. DFU was throwing an exception if a Temple location did not have a proper Temple/God faction id, which occurs in DF data in those Fighter Trainers halls in Hammerfell. I checked in classic DF, and the music that plays is actually the "unused" Knight Order music - yes, that music is actually used. This can be checked in the Abibon-Gora capital. 

So, I fixed both the error handling in the Temple music selection, and added a new music context for FighterTrainers so its classic DF music plays.

### DIFF
--- a/Assets/Scripts/Game/SongManager.cs
+++ b/Assets/Scripts/Game/SongManager.cs
@@ -39,7 +39,9 @@ namespace DaggerfallWorkshop.Game
         public SongFiles[] OvercastSongs = _overcastSongs;
         public SongFiles[] RainSongs = _rainSongs;
         public SongFiles[] SnowSongs = _snowSongs;
-        public SongFiles[] TempleSongs = _templeSongs;
+        public SongFiles[] TempleGoodSongs = _templeGoodSongs;
+        public SongFiles[] TempleNeutralSongs = _templeNeutralSongs;
+        public SongFiles[] TempleBadSongs = _templeBadSongs;
         public SongFiles[] KnightSongs = _knightSongs;
         public SongFiles[] TavernSongs = _tavernSongs;
         public SongFiles[] NightSongs = _nightSongs;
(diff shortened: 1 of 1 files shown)


petchema:
We had a similar issue with dungeons: we're using a song player dedicated to dungeons, so it doesn't notice when dungeon changes. I solved the dungeons case by listening to entering event, but that felt like an ad hoc hack, I was not satisfied at the time... #1632
Maybe we could unify the two cases using a single approach?

numidium (approved):

numidium:
I ran into the problem of teleporting from one interior type to another not changing the music in my Dynamic Music mod. My solution was to store a PlayerGPS.CurrentLocationIndex and check if it changed since the last frame. Not sure if it's the prettiest solution but it seems to work as a catch-all for determining if the player went to a different place.

petchema (approved):
LGTM
I appreciate the extra refactos :)

PR #2583, opened 2024-02-11
Fix TextLabel metrics in 4:3 mode for MultiFormatTextLabel and ListBox.

DunnyOfPenwick:
Fix for MultiFormatTextLabel and ListBox TextLabel wrapping issues spotted in 4:3 mode.

I originally noticed an issue with wrapping TextLabels when constructing a MultiFormatTextLabel while running in 4:3 mode.
Wrapped text would overlap the following line. 
In the TextLabel.CreateNewLabelLayoutWrapped() method, the LocalScale was (1,1), causing some bad line metric calculations.

More recently I noticed the bottom line of the npc conversation window was often not shown in 4:3 mode.
This appears to be the same problem.

This fix calls GetRectangle() on the TextLabel early to establish the correct LocalScale prior to other operations.

### DIFF
--- a/Assets/Scripts/Game/UserInterface/ListBox.cs
+++ b/Assets/Scripts/Game/UserInterface/ListBox.cs
@@ -544,7 +544,6 @@ namespace DaggerfallWorkshop.Game.UserInterface
             TextLabel textLabel = new TextLabel();
 
             textLabel.Parent = this; //Establish parent early.
-            _ = textLabel.Rectangle; //Call GetRectangle() early to establish LocalScale, discard return value
 
             if (UseRestrictedRenderArea)
             {
@@ -586,7 +585,6 @@ namespace DaggerfallWorkshop.Game.UserInterface
             }
 
             textLabel.Parent = this; //Establish parent early.
-            _ = textLabel.Rectangle; //Call GetRectangle() early to establish LocalScale, discard return value
 
             if (UseRestrictedRenderArea)
             {
--- a/Assets/Scripts/Game/UserInterface/ListBox.cs
+++ b/Assets/Scripts/Game/UserInterface/ListBox.cs
@@ -542,6 +542,10 @@ namespace DaggerfallWorkshop.Game.UserInterface
                 font = DaggerfallUI.DefaultFont;
 
             TextLabel textLabel = new TextLabel();
+
+            textLabel.Parent = this; //Establish parent early.
+            _ = textLabel.Rectangle; //Call GetRectangle() early to establish LocalScale, discard return value
+
             if (UseRestrictedRenderArea)
             {
                 textLabel.RectRestrictedRenderArea = RectRestrictedRenderArea;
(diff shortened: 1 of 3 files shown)


Assets/Scripts/Game/UserInterface/ListBox.cs
```diff
@@ -542,6 +542,10 @@ public void AddItem(string text, out ListItem itemOut, int position = -1, object
                 font = DaggerfallUI.DefaultFont;
 
             TextLabel textLabel = new TextLabel();
+
+            textLabel.Parent = this; //Establish parent early.
+            _ = textLabel.Rectangle; //Call GetRectangle() early to establish LocalScale, discard return value
```
KABoissonneault, line 547: This whole thing with localScale is a bit of a mess - not your fault of course. Could you possibly assign `localScale` in `BaseScreenComponent.SetParent` instead, or does that break things?

If not, at least add a  `UpdateLocalScale` function to call here rather than do the Property with side effects thing.

By the way, I'm curious, what's the code that relies on `localScale` in here? Not really obvious with all the Property assignments

DunnyOfPenwick: The LocalScale is used when calculating label length in TextLabel.CreateNewLabelLayoutWrapped().

I tried to make as low an impact change as I could, since we are post v1.0 now.  I'm reluctant to add code to SetParent due to the increased probability of inadvertently breaking something somewhere else.
I could add an UpdateLocalScale() method to TextLabel to make it clearer.

DunnyOfPenwick:
I tried moving the GetRectangle() call to TextLabel.RefreshLayout(), but that screwed up some of the normal (non-wrapped) labels.  Some of this code is finicky.  I then moved it to the top of TextLabel.CreateNewLabelLayoutWrapped() and that seems to work fine.  I ran through the full character creation routine and tutorial a couple of times.

I decided to leave the code as **_ = Rectangle;** to emphasize the specific/finicky nature of it.

KABoissonneault:
It's been some time, you think you could provide a test that reproduces the issue so I can verify the fix?

DunnyOfPenwick:
The easy way to produce the issue is by using the SDF font while in 4:3 mode and engaging in dialog.   After the dialog reaches the end of the window you should begin seeing instances of text cut-off.  I don't normally see it when using the non-sdf font.

KABoissonneault:
I unfortunately cannot reproduce the issue, neither in Editor or in-game.
Doing 320x200 or 640x400 retro mode with 4:3 or 16:10 correction, both with and without SDF

image
image

Is it a specific kind of dialogue that overflows? Is it some part of the screen I'm not noticing? At least a screenshot of what you're seeing could help

DunnyOfPenwick:
I didn't describe the problem correctly.  The text is there, but I have to use the scroll bar to see the last line of text.

KABoissonneault (approved):
Alright, the tests seem good. I'm having a hard time doing a thorough review of the code given all the implicit side effects of the property assignments, but I don't think this PR can fix that, and it doesn't seem like it would create issues
Thank you for your patience

PR #2585, opened 2024-02-15
Pacified-Enemy-Not-Attack-Allies

DunnyOfPenwick:
In a previous PR I had added code to prevent allies from attacking pacified enemies but forgot to prevent pacified enemies from attacking player allies.

Added check to prevent pacified enemies from attacking player allies.


KABoissonneault (approved):
God this function is a mess. I feel like there's a few redundancies, but your change tracks with what's already there

numidium (approved):

ajrb (approved):

Issue #2510, opened 2023-04-21
0.14.5: "Generic Knightly Order" used as faction in quest dialogue


JayH2971:
User reported the following text in a quest:

![image](https://user-images.githubusercontent.com/37541932/233721239-be481476-c1e0-4e3a-8171-e5a806fe9786.png)

Referenced quest is Q0C20Y02. The issue is the selected faction, "Generic Knightly Order."

A similar issue was seen previously regarding Generic Temple: [https://github.com/Interkarma/daggerfall-unity/issues/2368](https://github.com/Interkarma/daggerfall-unity/issues/2368)

EDIT: User adds that this occurred in Daenia, where there is no knightly order to choose from.

Interkarma:
Interesting the quest still found a location to compile. Do you know which specific quest this is? Did user provide a save?

JayH2971:
User sends this save file: 
[SAVE31__Q0C20Y02_Generic_Knightly_Order_quest_text.zip](https://github.com/Interkarma/daggerfall-unity/files/11300102/SAVE31__Q0C20Y02_Generic_Knightly_Order_quest_text.zip)

KABoissonneault:
Reposting my findings while investigating for #2589 

I just ran the test in DOS. I made a save with 50 reputation with the Glenmoril Witches and went to the NPC in front of the Thieves Guild to ask for quests.

I first tried in Daenia, and could never get this quest. I did have a few "gave the quest to some spellsword" type messages, which usually occurs when a quest fails to resolve. Maybe I needed to be in a region with a knight order.

When I tried again in Daggerfall, I did eventually land on it.

![image](https://github.com/Interkarma/daggerfall-unity/assets/5789925/62e991ec-5ceb-4b9a-99b7-798d6a9aaa60)

![image](https://github.com/Interkarma/daggerfall-unity/assets/5789925/d1d53b1e-547c-4238-941d-19239561a6a5)

This is indeed a knightly guard... of the Totambu knight order, not Daggerfall. 

The NPC did look like a guard at the site.

![image](https://github.com/Interkarma/daggerfall-unity/assets/5789925/59656319-243c-46be-86bc-7c9d8fe44188)

Seems to me like the solution to this issue is that Knightly_Guard should resolve to a random knight order, not the local one. This would fix the problem for regions without a knight order.

John-Leitch:
[...]

PR #2589, opened 2024-02-19
Fixed Q0C20Y02 faction type bug.

John-Leitch:
This PR fixes the following bug: #2510 (0.14.5: "Generic Knightly Order" used as faction in quest dialogue). The issue appears to be caused by the incorrect `factiontype` value for `_other_` in `Q0C20Y02.txt`. `factiontype` should be set to `Daedra`, not `Knightly_Guard`.

### DIFF
--- a/Assets/StreamingAssets/Quests/Q0C20Y02.txt
+++ b/Assets/StreamingAssets/Quests/Q0C20Y02.txt
@@ -125,7 +125,7 @@ Item _reward_ gem
 Item _magicitem_ magic_item
 
 Person _questgiver_ face 112 group Questor anyInfo 1011
-Person _other_ factiontype Knightly_Guard remote
+Person _other_ factiontype Daedra remote
 
 Clock _1stparton_ 00:00 0 flag 17 range 1 4
 


KABoissonneault:
Thank you for contributing to an outstanding issue! 

While I suppose any faction other than knight order is better here, how did you determine that the faction should be Daedra? It can make sense that someone who deals with witches has such alignments, but I can't find any definitive signs that this was the intent. I assume the message says "The <item> we need is held by Daedra." in the quest accept message?

JayH2971:
I'm glad to see users make an effort to overcome this bug. However, the proposed fix will not work. Attempting the quest in this state will place an actual Daedric prince in the world, which is not a part of this quest. (Ignore the ==questgiver_ statement inside my journal, that's unrelated.)

Screenshot at 2024-02-19 06-54-43
Screenshot at 2024-02-19 06-55-40

John-Leitch:
@KABoissonneault I made this evaluation based on looking at decompilation output from the original quest. However, based on what @JayH2971 is saying, I may have misinterpreted the output.

This is the decompilation output I used as a reference: 

[stellargames.github.io/Quester/Q0C20Y02.json](https://stellargames.github.io/Quester/Q0C20Y02.json)

The naming scheme is a bit confusing, with `faction` set to `"Boethiah"`. I tried that, and it did not work. Looking around at other quests, I came across R0C11Y03:

[stellargames.github.io/Quester/R0C11Y03.json](https://stellargames.github.io/Quester/R0C11Y03.json)

In that one, I noticed `npcType` was set to `"Daedra"`, and `faction` was not set. And the relevant bit in `R0C11Y03.txt` is as follows:

`Person _dummydaedra_ face 102 factiontype Daedra female remote`

So it seemed to me that `npcType` in the decompilation output actually mapped to `factiontype` in the DSL used by the quests (not sure what it's called). Is this not the case?

JayH2971:
> The naming scheme is a bit confusing, with faction set to "Boethiah". I tried that, and it did not work.

In Daggerfall's system, named NPCs are invoked as persons rather than as factions. You can see this at work in this segment of the main quest: [https://github.com/Interkarma/daggerfall-unity/blob/e1e807e742c2efc54449773b9a165dc1efc0684d/Assets/StreamingAssets/Quests/S0000988.txt#L127](url). Most importantly, in R0C11Y03 the daedra NPC isn't meant to be seen in the game world.

As for the decompilation script, I'm not familiar with that.

KABoissonneault:
This URL might make the decompiler output easier to read:
https://stellargames.github.io/Quester/#Q0C20Y02

I do think it's possible the intent is that the quest should have a _human representative_ of Boethiah show up there. The quest offer message would look something like:
```
<ce>                        The [item] we need
<ce>                   is held by Boethiah. We require it
<ce>              before the day has dawned [6] times.
<ce>                  Boethiah can be found in [place].
<ce>                           Ask after Agrynak Moorhat.
<ce>                   Let the gods and spirits guide thy
<ce>                              feet, [Player name].
```

I think I recall at least another quest that behaves with a human representing a daedric prince faction?

It's possible DFU doesn't handle this case properly, and we'd need to fix the quest engine in consequence. It's too bad this quest would be really difficult to test in classic, unless someone knows a quest editor that can make me start with enough rep with the witches

KABoissonneault:
I just ran the test in DOS. I made a save with 50 reputation with the Glenmoril Witches and went to the NPC in front of the Thieves Guild to ask for quests.

I first tried in Daenia, and could never get this quest. I did have a few "gave the quest to some spellsword" type messages, which usually occurs when a quest fails to resolve. Maybe I needed to be in a region with a knight order.

When I tried again in Daggerfall, I did eventually land on it.

image

image

This is indeed a knightly guard... of the Totambu knight order, not Daggerfall. 

The NPC did look like a guard at the site.

image

Seems to me like the real solution to this issue is that Knightly_Guard should resolve to a random knight order, not the local one. This would fix the problem for regions without a knight order.

Closing this PR. I would welcome a new solution that keeps the knightly guard. I really appreciate people who tackle existing issues

PR #2590, opened 2024-02-20
Fixed Knightly_Guard faction resolve bug.

John-Leitch:
This PR (hopefully) fixes the following bug: #2510 (0.14.5: "Generic Knightly Order" used as faction in quest dialogue). The update makes `Knightly_Guard` resolve to a random knightly order, which should fix the faction resolution issue in `Q0C20Y02`. Since the knightly orders are already hardcoded in the `DaggerfallWorkshop.Game.Guilds.KnightlyOrder.Orders` `enum`, the `enum` values were used instead of adding another hardcoded set of knightly order ids.


KABoissonneault:
Awesome. Code looks good, just gonna have to test run a few quests to approve this

Issue #2591, opened 2024-02-20
Fast Travel deducts gold even if travel is canceled


KABoissonneault:
If you start Fast Travel on the DaggerfallTravelPopup, gold is immediately deducted. However, the fast travel does not complete immediately, and can be canceled during the live update. If it is canceled, the player stays where they were, but still has their gold deducted.

Note that classic DF does not let you cancel travel, this is a DFU issue.

**To Reproduce**

1. Open Travel map
2. Click on any location
3. Start travel
4. During the update, press Esc to cancel
5. Look at gold count

**Expected behavior**

Canceling travel should not deduct gold, since no travel actually happened.

**Suggested fix**
Should simply be a matter of moving the money deduction part of `DaggerfallTravelPopUp.CallFastTravelGoldCheck` to the end of `Update`

d-blackburn:
This should be closed as it's been fixed in #2592

PR #2592, opened 2024-02-21
Fast travel gold fix

John-Leitch:
This PR fixes the following issue: #2591 (Fast Travel deducts gold even if travel is canceled)

Pretty straight forward. Moved the code to deduct the gold into a separate method, `DaggerfallTravelPopUp.DeductFastTravelGold`. Rather than invoking directly from update, I opted to do so in `DaggerfallTravelPopUp.performFastTravel`, as that method seemed to be where most of the other related changes to `GameManager.Instance.PlayerEntity` state occur. The only concern I have is related to the `DaggerfallTravelPopUp.OnPreFastTravel` and `DaggerfallTravelPopUp.OnPostFastTravel` events. Previously, the gold was deducted prior to either event being fired. In `performFastTravel`, I put it after `OnPreFastTravel`, as that seemed to make the most sense. Looking at the codebase, I did not see any subscribers to `OnPreFastTravel`, so I don't think this will break anything.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTravelPopUp.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTravelPopUp.cs
@@ -325,6 +325,8 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
         {
             RaiseOnPreFastTravelEvent();
 
+            DeductFastTravelGold();
+
             // Cache scene first, if fast travelling while on ship.
             if (GameManager.Instance.TransportManager.IsOnShip())
                 SaveLoadManager.CacheScene(GameManager.Instance.StreamingWorld.SceneName);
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTravelPopUp.cs
```diff
@@ -325,6 +325,8 @@ private void performFastTravel()
         {
             RaiseOnPreFastTravelEvent();
 
+            DeductFastTravelGold();
```
KABoissonneault, line 328: I would rather you moved this above "RaiseOnPreFastTravelEvent". 

Before this change, the player gold was removed before any code that reacted to this event. This means code might have been assuming that the gold was already taken care of, and change gold amount.

KABoissonneault: Otherwise, all is as expected

John-Leitch: @KABoissonneault Done

KABoissonneault (approved):

petchema (approved):
Modification seems straightforward

Issue #2594, opened 2024-02-24
Deleting and creating new save can preserve old mod save data


KABoissonneault:
**Describe the issue**

When deleting a save, some mod save data may be preserved and re-used on a new save. This occurs for example when the mod with the save data is currently disabled: it won't show in the list of mods in `ModManager.Instance.GetAllModsWithSaveData()`, and `SaveLoadManager.DeleteSaveFolder` won't consider the disabled mods for deleting mod save data.

When you look at the save folder after deletion, all disabled (or removed) mods should be remaining

![image](https://github.com/Interkarma/daggerfall-unity/assets/5789925/451325f3-1ae0-40fd-963d-708f3ddf4741)

When you create a new save later on, that folder might be reused by DFU, and the old mod save data will still be there on your new save.

![image](https://github.com/Interkarma/daggerfall-unity/assets/5789925/36f21768-12ab-4116-b896-3d79d0fc4f27)

(Notice the 'Date modified' column)

This will become an issue if the user re-enables the mod, as the mod will have some leftover data from an older save that is not compatible with the current one.

**To Reproduce**

1. Locate a save containing mod save data (mod_­[mod name].txt)
2. Disable one or multiple mods with such save data
3. Delete save from the in-game DFU UI
4. Re-create new saves until that folder is reused

Should be easy to test if your SAVE0 has old mod data.

**Expected behavior**

If `SaveLoadManager.DeleteSaveFolder` fails to delete a folder because it contains leftover files, the folder should be renamed to something else. It's worth keeping files DFU doesn't know, these could be important player notes. But we don't have to keep them as primary SAVE folders, we can just toss them under a backup subfolder (ex: DELETED/SAVE0). This will prevent the folder from being reused, and old save files will not accidentally get slipped into a new playthrough.

John-Leitch:
I can take this one. Should be pretty straight forward to handle, edge cases included e.g. dupes in DELETED.

KABoissonneault:
Thanks for tackling these, it does spare me some work. However, I am stopping tests on PRs until I got my changes for v1.1 done to focus, and I won't be approving untested PRs. Others can test them if they want, but I expect you might have a few lingering PRs for some weeks. Put in the work at your own risk :)

John-Leitch:
No worries. If there are any issues, I'll gladly fix them, and if they can't be accepted, so be it. I was a huge daggerfall fan growing up, so it's worth my time either way.

[...]

PR #2596, opened 2024-02-25
Added functionality to move deleted saves to DELETED dir

John-Leitch:
This PR fixes the following issue: #2594 (Deleting and creating new save can preserve old mod save data)

The update changes the logic of `SaveGameManager.DeleteSaveFolder` to move any folder to `DELETED` should it contain unexpected files or otherwise fail to delete. If a backup folder already exists with the same, a numbered suffix is appended beginning with `_2`.

image


KABoissonneault (approved):
Works as expected. I did try the case where some of the extra files would fail to get moved to the backup, and it does leave the folder intact, same as the behavior before this change. It's all good to me

PR #2603, opened 2024-02-26
More Retro Modes

DunnyOfPenwick:
Adds half-resolution widescreen graphics modes and a high-resolution 4:3 mode. 
Also corrects for texture stretching when using the widescreen modes.

### DIFF
--- a/Assets/Prefabs/Player/PlayerAdvanced.prefab
+++ b/Assets/Prefabs/Player/PlayerAdvanced.prefab
@@ -620,6 +620,10 @@ MonoBehaviour:
   m_Script: {fileID: 11500000, guid: 36bb43919da808741adb733009f14b91, type: 3}
   m_Name: 
   m_EditorClassIdentifier: 
+  playerJumpSkillMultiplier: 0.5
+  athleticismMultiplier: 0.1
+  improvedAthleticismMultiplier: 0.1
+  jumpSpellMultiplier: 0.6
   jumpSpeed: 4.5
   gravity: 20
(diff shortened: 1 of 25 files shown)


Assets/Prefabs/Player/PlayerAdvanced.prefab
```diff
@@ -620,6 +620,10 @@ MonoBehaviour:
   m_Script: {fileID: 11500000, guid: 36bb43919da808741adb733009f14b91, type: 3}
   m_Name: 
   m_EditorClassIdentifier: 
+  playerJumpSkillMultiplier: 0.5
```
KABoissonneault, line 623: Are those leftover changes from another feature you're working on?

DunnyOfPenwick: I saw that and have no idea why it added those, or how they haven't been added before.
Those are existing public members of AcrobatMotor.

KABoissonneault: Ah I see, all good then

DunnyOfPenwick:
The implementation as it stands seems to work fine on my laptop, but after additional thought I've developed concerns about the performance impact on old systems and handheld devices.
The presentation target was changed to 4k resolution to accurately handle the various resolutions, but that can have a significant impact on less capable systems.
The presentation target is only needed to handle post-processing effects such as posterization.
I'm currently modifying the code to skip the extra step if post-processing is not being used.

PR #2608, opened 2024-03-03
Use Redguard names in Redguard lands for named residences

Jagget:
So, most of the Daggerfall is Breton lands. But not all. But in quests, we always see the named residences of Bretons. Even in the Redguard lands. We should use Redguard names in Redguard lands.

Original DOS:
image

PR effect:
Sentinel
Screenshot 2024-03-02 222339

Wayrest
Screenshot 2024-03-02 222558

### DIFF
--- a/Assets/Scripts/Internal/PlayerGPS.cs
+++ b/Assets/Scripts/Internal/PlayerGPS.cs
@@ -427,19 +427,6 @@ namespace DaggerfallWorkshop
         }
 
         /// <summary>
-        /// Gets NameHelper.BankType in given region.
-        /// In practice this will always be Redguard/Breton.
-        /// Supporting other name banks for possible diversity later.
-        /// </summary>
-        public NameHelper.BankTypes GetNameBankOfRegion(int regionIndex)
-        {
-            if (regionIndex > -1)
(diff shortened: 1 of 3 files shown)


Assets/Scripts/Internal/PlayerGPS.cs
```diff
@@ -426,6 +426,19 @@ public NameHelper.BankTypes GetNameBankOfCurrentRegion()
             return NameHelper.BankTypes.Breton;
         }
 
+        /// <summary>
+        /// Gets NameHelper.BankType in given region.
+        /// In practice this will always be Redguard/Breton.
+        /// Supporting other name banks for possible diversity later.
+        /// </summary>
+        public NameHelper.BankTypes GetNameBankOfRegion(int regionIndex)
```
KABoissonneault, line 434: Why is this in PlayerGPS if it uses nothing about the player's current location? This just reaches into MapsFile based on the input. You can always put the helper in MapsFile

Jagget: Moved

KABoissonneault (approved):

numidium (approved):

Issue #2604, opened 2024-02-28
1.0: Person fails to appear in "where is" talk UI for quest


JayH2971:
Problem description:

User reports playing Azura's quest (t0c00y00) and Meridia's quest (10c00y00) and being unable to find the reward giver. Reward giver does not show up in Talk UI when asking peasants for location. The quest journal does not say where the reward giver is (which could be an alternate fix but this is a deeper engine issue). Not known when this problem began occurring, but this quest did function properly back during quest fixing, around 0.11.

Steps to reproduce:

1. startquest t0c00y00 or 10c00y00

2. Go to dungeon and kill enemy, then travel to indicated town in travel log.

3. Ask peasants to find person who gives reward. Person does not appear in the Talk UI options. Person _does_ appear in Talk UI after you enter the reward giver's house once.

John-Leitch:
I've reproduced this locally. Not familiar with this code, but it looks (indirectly) related to this in `TalkManager`: 

```cs
// only make it available for talk if...
if (!dialogPartnerIsSamePersonAsPersonResource &&   // Dialog partner is not the same person as the person resource talked about
    questResourceInfo.Value.availableForDialog &&   // It is not "hidden" by dialog link command
    questResourceInfo.Value.hasEntryInWhereIs &&    // It is meant to have an entry in the where is section (currently this is always true, TODO: check if we can get rid of this)
    IsPlayerInSameLocationWorldCell)                // If person resource is in same world map location as player
{
    listTopicPerson.Add(item);
}
```

Every condition evaluates to `true`, except `IsPlayerInSameLocationWorldCell`, which is `false`. Here's the code where I would expect it to be set to `true`:

```cs
/* - note: found a different way to do it via mapId
try
{
    IsPlayerInSameLocationWorldCell = ((Questing.Person)questResourceInfo.Value.questResource).IsPlayerInSameLocationWorldCell();
}
catch (Exception e)
{
}*/

Symbol assignedPlaceSymbol = person.GetAssignedPlaceSymbol();

if (assignedPlaceSymbol != null)
{
    Quest quest = GameManager.Instance.QuestMachine.GetQuest(questID);
    Place assignedPlace = quest.GetPlace(assignedPlaceSymbol);  // Gets actual place resource
    if (assignedPlace.SiteDetails.mapId == GameManager.Instance.PlayerGPS.CurrentMapID)
        IsPlayerInSameLocationWorldCell = true;
}
```

[...]

PR #2610, opened 2024-03-06
Fixed TalkManager issue causing missing people

John-Leitch:
This PR covers the following issue: #2604 (1.0: Person fails to appear in "where is" talk UI for quest)

A few details related to the issue can be found in my comments on the bug. The appears to stem from `Person.GetAssignedPlaceSymbol` returning `null`. Diving a little deeper, apart from the serialization code, the assigned place symbol is only set via the `Person.SetAssignedPlaceSymbol` method. In turn, that method is called in two spots:

1. The `PlaceNpc` `ActionTemplate`, which handles `place npc` statements.
2. In `Person.PlaceAtHome`, which is itself called in two spots:
a. In `Person.Tick` when `home.IsPlayerHere` returns true. This explains the NPC begins resolving once the player enters its home.
b. The `CreateNpc` `ActionTemplate`, which handles `create npc` statements.

Both `T0C00Y00` and `10C00Y00` (and maybe others) lack `place npc` or `create npc` statements for the NPC in question, thus their assigned place is always `null`.

While this is apparently a regression, searching through the history, I was not able to easily identify the exact revision that introduced this behavior, so was not entirely sure how this case was meant to be handled. Reviewing the code, I noted that `Person.homePlaceSymbol` was set, despite the NPC not being created. As this value is not public, I added the `Person.GetAssignedOrHomePlaceSymbol` method, which `null` coalesces the values. I then updated the appropriate calls in `TalkManager`, replacing `Person.GetAssignedPlaceSymbol` with `Person.GetAssignedOrHomePlaceSymbol`. This appears to work as expected, with the NPC being resolved by map id.

### DIFF
--- a/Assets/Scripts/Game/Questing/Person.cs
+++ b/Assets/Scripts/Game/Questing/Person.cs
@@ -475,6 +475,18 @@ namespace DaggerfallWorkshop.Game.Questing
         }
 
         /// <summary>
+        /// Returns the assigned or home symbol.
+        /// </summary>
+        /// <returns>
+        /// If the most recent Place symbol assigned if not null, otherwise returns
+        /// home symbol (which might be null as well).
+        /// </returns>
+        public Symbol GetAssignedOrHomePlaceSymbol()
+        {
+            return lastAssignedPlaceSymbol ?? homePlaceSymbol;
+        }
+
+        /// <summary>
         /// Checks if player in same world cell as Place this Person was assigned to.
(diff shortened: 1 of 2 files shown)


KABoissonneault:
I have to reject this solution, as it seems intended for NPCs without an assigned place to appear under "Tell Me About" rather than "Where Is", even if they have a Home Place. I'd take a fix to prevents non-placed NPCs from appearing after its building has been loaded once, but that's pretty minor given the complexity

Issue #2342, opened 2022-05-20
Inconsistency between displayed AC and to-hit formula for enchanted armor


KABoissonneault:
"AC" is the number displayed on the paperdoll when equipping armor. "Armor value" is the actual contribution to the to-hit formula provided by armor.
As a rule, AC gives you 5 armor value - it reduces hit by 5% for each AC point.

Currently, this rule breaks for `IncreasedArmorValueModifier` and `DecreasedArmorValueModifier`. `StrengthensArmor`, `BadReactionsFrom`, and `WeakensArmor` change these values by 5. 
The inconsistency is that `IncreasedArmorValueModifier` and `DecreasedArmorValueModifier` change both the displayed AC and the hit chance by 5 (see `PaperDoll.RefreshArmourValues` and `FormulaHelper.CalculateArmorToHit`). 

Having an AC of 20 reduces hit by 100, but add a +5 enchantment, and suddenly you have an AC of 25 that reduces hit by 105.

Classic does change the displayed AC by 5 for these enchantments, but I don't know about the effect on hit.

Proposed solutions:
1. if this is consistent with classic, do nothing (mods can override the `CalculateArmorToHit` formula if they don't like this)
2. Multiply `IncreasedArmorValueModifier` and `DecreasedArmorValueModifier` by 5 in the default `CalculateArmorToHit` formula (visibly consistent with classic)
3. Divide `armorMod` by 5 in `Paperdoll` (no impact on gameplay)

My favored solution is #2, as ±5 hit from the enchantment is pretty weak.

Interkarma:
I recall some conversations and tuning around this to match behaviours, but can't dig up them up right now. Still, there's always a chance I made a mistake when implementing and nobody has caught it until now.

I'm not opposed to the change in 2 however, and I agree the enchantment is pretty weak. Something to investigate for sure.

Interkarma:
Feedback on this one. I can't find any conversation history to confirm IncreasedArmorValueModifier and DecreasedArmorValueModifier should _not_ be subject to multiplier. I'll consider this a bug in my implementation and apply *5 as per solution 2 in Kab's issue.

Issue #1740, opened 2020-03-07
Strengthen & Weaken Armor Enchantments Reversed Effect


magicono43:
**Describe the bug**
Due to how the math is done, the Strengthen & Weaken Armor Enchantments have their function reversed. That being that Strengthen armor actually increases player chances of being hit, while Weaken armor reduces player chances of being hit, which is not correct.

**To Reproduce**
1. Find the "CalculateSuccessfulHit" method in the FormulaHelper.cs script. 
2. Add "Debug.LogFormat("Final Hit-Chance = {0}", chanceToHit);" just above line 1015, just before Diceroll100 is called. (This is if you wish to track the hit value through the Unity debug log.)
3. Enchant a weapon or none-armor item, one with "Strengthen Armor", the other with "Weaken Armor"
4. Find a place that you can be hit by an enemy, take off all armor.
5. Check hit value, alternate between equipping the two enchanted items and getting hit to see hit value difference, you will see that while Strength armor is equipped, the value will increase by 5. while the opposite is true with the weaken armor item.

**Expected behavior**
The values should be reversed/effects should be opposite what they currently are.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows
 - Version: Alpha 0.10.21 

**Additional context**
This can be fixed by changing line 968 of FormulaHelper.cs: `armorValue = target.ArmorValues[struckBodyPart] + target.IncreasedArmorValueModifier + target.DecreasedArmorValueModifier;`

Simply change the adds to subtract signs like so: `armorValue = target.ArmorValues[struckBodyPart] - target.IncreasedArmorValueModifier - target.DecreasedArmorValueModifier;`

The other solution would be to change whatever gives the values for these spell effects to the opposite of what they currently are, wherever those values might be held.

XJDHDR:
@magicono43 
Would you be willing to test another thing? When you equip and unequip those two enchanted items, what happens to the armour numbers around your character's paper doll? Does the `Strengthen Armor` enchantment makes these numbers go up or down?

magicono43:
> @magicono43
> Would you be willing to test another thing? When you equip and unequip those two enchanted items, what happens to the armour numbers around your character's paper doll? Does the `Strengthen Armor` enchantment makes these numbers go up or down?

[...]

PR #2616, opened 2024-03-14
Fixed "Weakens Armor" enchantment not being applied

KABoissonneault:
This PR has 3 changes related to "Weakens Armor", last one also including "Strengthens Armor".

Two uncontroversial fixes:

# Fixed "Weakens Armor" enchantment not being applied.

The "Weakens Armor" decreased armor value was just not applied by DFU. In DF, lower armor value is better: players start at 100, and armors reduce the armor value by 5 for each armor class. Therefore, a "Decreased" armor value is a +5 malus to Amor Value.

When checking which "Decreased" armor value is the worst, we take the bigger value, not the smaller one. Otherwise, the base "DecreasedArmorValue" of 0 is always smaller than the +5 of "Weakens Armor", and fails to apply the actual debuff.

# Properly apply "Bad" color for AC penalty in Paperdoll

The AC value shown in the paperdoll runs backward from the Armor Values used in formulas. AC starts at 0, and goes up when wearing better armors. +1 AC is equal to -5 Armor Value, or -5% chance to hit. So, an increased AC is a decreased armor value and is good. A decreased AC is an increased armor value and is bad.

DFU was not always like this. It used to get them backward until this issue:
https://github.com/Interkarma/daggerfall-unity/issues/1740

and this changelist inverted them.
https://github.com/Interkarma/daggerfall-unity/commit/baa983a83419b1ea1b8db7ac5105c060234331e9

So this old logic in `Paperdoll` was still accidentally thinking DecreasedArmorValue was a negative number - no, it's 0 or +5 (Weakens Armor). IncreasedArmorValue is 0 or -5. We instead add the two modifiers together, like in `FormulaHelper.CalculateArmorToHit`.

image

The more controversial fix:

# #2342 

I initially reported that the bonus from "Strengthens Armor" and "Weakens Armor" are inconsistent between the shown AC and the effect on the formula. 
https://github.com/Interkarma/daggerfall-unity/issues/2342

In short, they both give ±5 to AC and ±5 to Armor Value (Hit). When it comes to the AC and armor value from armor, 1 AC is equal to -5 Armor Value (and therefore -5 Hit). This is inconsistent, either 5 AC should be -25 Armor Value, or -5 Armor Value should be +1 AC shown on the paperdoll.

Since I reported this issue, I tested in gameplay in DF, and this inconsistency is a classic issue. "Strengthens Armor" does not have as much of an impact on gameplay as a 25% difference would make.

When I reported this issue, we initially went with changing the effect to ±25 Armor Value, 
https://github.com/Interkarma/daggerfall-unity/commit/d0b3b2f7cdd9885b2b7ce8cf3f040407619e1beb

but IK reverted this here. And I now agree with reverting it.
https://github.com/Interkarma/daggerfall-unity/commit/7825c4effabf825cad850932bf64073451758f77

Note: There is a mistake in that changelist's explanation. Daedric's base Armor Value is -5 because 21 AC means -105 Armor Value, and `100 - 105 = -5`. The effect of "Strengthens Armor" is an extra -5, not an extra -105. So you go from `100` to `-10`, not `100` to `-110`. "Strengthens Armor" is nowhere near as impactful as wearing Daedric Armor, but it does add up.

I still agree with not changing the gameplay impact. However, I would rather we stop lying to players here, especially since it's not easy for mods to fix the Paperdoll display here.

**So instead, I show the AC change as ±1**. This does not visually match classic, but it matches it in gameplay, and has no impact on a playthrough otherwise. 
It does mean players will ask about it, and we'll need to send them the patch notes when that happens. Unlike the other DFU changes, this is purely visual.

### DIFF
--- a/Assets/Scripts/Game/Entities/DaggerfallEntity.cs
+++ b/Assets/Scripts/Game/Entities/DaggerfallEntity.cs
@@ -399,7 +399,7 @@ namespace DaggerfallWorkshop.Game.Entity
 
         public void SetIncreasedArmorValueModifier(int amount)
         {
-            // Increased armor value does not stack, only effect with the highest modifier used
+            // Increased armor value does not stack, only effect with the strongest modifier used
             // In classic effects this never goes below -5 (lower modifier -> higher armor)
             if (amount < IncreasedArmorValueModifier)
             {
@@ -409,10 +409,12 @@ namespace DaggerfallWorkshop.Game.Entity
 
         public void SetDecreasedArmorValueModifier(int amount)
         {
-            // Decreased armor value does not stack, only effect with the lowest modifier uses
+            // Decreased armor value does not stack, only effect with the strongest modifier uses
             // In classic effects this never goes above +5 (higher modifier -> lower armor)
-            if (amount < DecreasedArmorValueModifier)
+            if (amount > DecreasedArmorValueModifier)
+            {
                 DecreasedArmorValueModifier = amount;
+            }
         }
 
         public void ChangeChanceToHitModifier(int amount)
--- a/Assets/Scripts/Game/UserInterface/PaperDoll.cs
+++ b/Assets/Scripts/Game/UserInterface/PaperDoll.cs
@@ -156,20 +156,29 @@ namespace DaggerfallWorkshop.Game.UserInterface
             if (!showArmorLabels)
                 return;
 
+            // "Increased" armor value modifier is negative or 0
+            // "Decreased" armor value modifier is positive or 0
+            // They are applied directly as a hit malus/bonus when an enemy attacks
+
+            // For the purpose of UI, we invert this logic
+            // We show bonuses as bigger numbers and maluses as small numbers
+            // Also, 1 AC = 5 hit modifier
+            // Therefore, a -5 "Increase" armor value gives +1 AC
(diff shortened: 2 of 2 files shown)


KABoissonneault:
I'd like for someone to review this for the 1.2 patch. A summary of the three changes are:

- Fixed "Weakens Armor" enchantment not being applied (pure DFU bug)
- Fixed "buffed" vs "nerfed" AC colors in paperdoll (pure DFU bug)
- Changed displayed AC changes from "Strengthens Armor" and "Weakens Armor" to match the real gameplay change (new DFU enhancement/fix)

Summary for variables involved:
- ArmorValue (AV) is a direct 1% per AV. Target AV is added to attacker hit chance, so +AV is bad, and -AV is good
- Armor Class is not a real gameplay variable, but a summary used in paperdoll. AC is flipped around: +AC is good, -AC is bad. +AV gives -AC, -AV gives +AC

Key places to check: 
- DaggerfallEntity.cs (IncreasedArmorValueModifier and DecreasedArmorValueModifier)
- Paperdoll.cs (the display logic for AC)
- FormulaHelper.cs (the gameplay logic for AV)

Daneel53 (approved):
If Kab tells me that the PR is good, I cannot do anything else than to approve it! ;)

PR #2619, opened 2024-03-17
Reformat books, use the built-in book formatting.

Jagget:
Working with English version of the DFU I noticed poorly formatted books. Well, most of them have some artifacts. 

Since SDF and non-SDF books in game respect book boundaries and hence could be auto formatted, I want to clean texts.

Nothing fancy, just
* One paragraph is one line
* No multi-spaces
* No extra empty lines.

Before (Broken Diamonds):

Screenshot 2024-03-16 204922

After (Broken Diamonds):

Screenshot 2024-03-16 212352

### DIFF
--- a/Assets/StreamingAssets/Text/Books/BOK00000-LOC.txt
+++ b/Assets/StreamingAssets/Text/Books/BOK00000-LOC.txt
@@ -3,191 +3,52 @@ Author: Arkan
 IsNaughty: False
 Price: 580
 IsUnique: False
-WhenVarSet: 
+WhenVarSet:
 Content:
 
-
 [/font=2]
-
 [/center]The First Scroll of Baan Dar
 
-
-
 [/font=4]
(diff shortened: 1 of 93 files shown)


petchema:
I have to say we went for automatic text re-flowing in the french translation mod too, it feels like a more modern approach
https://github.com/Daneel53/DFU-en-francais/tree/main/French%20files%20for%20DFU/Text/Books

KABoissonneault (approved):
I admit to not having read all the diffs extensively (just too many), so I'll have to trust that there's been no loss of content. 

I've tested SDF and non-SDF fonts, with both full 16:9 ratio and "4:3 fix" mode. 

Overall, the changes are good. I do notice the loss of some of the centered paragraphs, with all paragraphs now taking the entire space. The original data was not amazing with its formatting, so it's not a big loss. I just feel the need to point out the change from a - let's assume - deliberate artistic decision.
On other points, some of the formatting here is more consistent, so it can be argued either way.

Anyway, let's merge this, and we can reiterate if people find issues with it

PR #2620, opened 2024-03-17
Relink children if any new custom factions with parents were added

ajrb:
Missed this back in 2018. If new custom factions are added by mods and they specify parent factions, this needs to be processed by the child linking code FactionFile.RelinkChildren() otherwise it does nothing.

Found this when Cliffworms added a named NPC to Betony - this should be used at the provinces ruler but it wasn't. Will confirm once this has been completely tested. 

Still doesn't help being able to modify existing faction data to, for example, set the ruler value for Betony. Tried some things for that but breakpoints stopped working and I just can't continue today.

### DIFF
--- a/Assets/Scripts/Game/Player/PersistentFactionData.cs
+++ b/Assets/Scripts/Game/Player/PersistentFactionData.cs
@@ -138,13 +138,21 @@ namespace DaggerfallWorkshop.Game.Player
         /// </summary>
         public void AddCustomFactions()
         {
+            bool relink = false;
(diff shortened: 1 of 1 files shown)


ajrb:
Tested this now using Cliff's mod and it successfully adds a ruler NPC to Betony. Be nice to get this into 1.0.1 since it's nice and self contained.

KABoissonneault (approved):

petchema (approved):

PR #2622, opened 2024-03-22
Allow faction ruler type to be set by mods

ajrb:
This allows changes of rulers to have appropriate title when %rt macro is used. Without this new rulers must be a Lord which is the default. No one is clamouring for this, but I realised that Cliff's change to add a ruler luckily is okay because it's a Lord but in future this could be useful.

Tested this with a bare bones quest script that prints the %rt macro. Happy to supply this if any reviewers want it.


petchema (approved):

KABoissonneault (approved):
The type can easily be set through the public FactionDict, but it's good to have this consistency

PR #2627, opened 2024-03-29
Fix drums in custom soundfonts

Kung-vr:
Many soundfonts will only provide percussion on patch 0 of bank 128. This fix allows attempting to fall back to the default percussion patch location when a drum patch is unavailable.

See this comment for an example of the common soundfont bank and patch setup for drums: https://github.com/LMMS/lmms/issues/323#issuecomment-40160617

To verify the problem, try loading this OPL3.SF2 soundfont [OPL3.zip](https://github.com/Interkarma/daggerfall-unity/files/14804912/OPL3.zip) with the current release version and notice the percussion during 5strong.mid is missing in the startup screen. When this fix is applied the percussion will work.

### DIFF
--- a/Assets/Scripts/AudioSynthesis/Synthesis/Synthesizer.MidiControl.cs
+++ b/Assets/Scripts/AudioSynthesis/Synthesis/Synthesizer.MidiControl.cs
@@ -26,6 +26,8 @@ namespace DaggerfallWorkshop.AudioSynthesis.Synthesis
             // Get the correct instrument depending if it is a drum or not
             SynthParameters sChan = synthChannels[channel];
             Patch inst = bank.GetPatch(sChan.bankSelect, sChan.program);
+            if (inst == null && channel == 9)
+                inst = bank.GetPatch(sChan.bankSelect, 0); // Attempt falling back to patch 0 for drums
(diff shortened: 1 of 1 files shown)


KABoissonneault:
Thank you! We're currently on hold while waiting for the v1.1 changes to be done, but I'll have a look after. 
I suppose the fact that your reference says "channel 10" but your review checks channel 9 is because of 0-indexing vs 1-indexing differences?

Kung-vr:
Thanks for taking a peek so quickly! I didn't know how active this project is, but a reply with just an hour is amazing 🙏
> your reference says "channel 10" but your review checks channel 9 is because of 0-indexing vs 1-indexing differences?

Yep that's correct! Midi channels start from 1, but in code are indexed from 0, so the 9 there represents midi channel 10 which is the standard channel for percussion.

Kung-vr:
Actually, the drum channel was defined as a constant in MidiHelper.DrumChannel (still 9), so I changed it to use that instead to avoid using magic numbers.

petchema (approved):

KABoissonneault (approved):
Thank you for providing a reference soundfont. The fix works as described. All good

PR #2628, opened 2024-03-29
Move the live change warning outside of the settings panel

ajrb:
The warning for changing live settings being inside the panel messes up the careful formatting of the mod settings. I know that for my mods I spend a lot of time getting them looking good.

I also think the message is more obvious not being lost in all of the settings text.


ajrb:
With this change Travel Options looks like this:
image

Instead of this:
image

KABoissonneault (approved):

petchema (approved):

PR #2631, opened 2024-04-01
Added pet that follow you and give you hp potion when your HP < 25%

Daynero:


KABoissonneault:
Hello. This repo is not where you create mod features. If you need new DFU events to implement your standalone mod, you can propose just that.
At this point, I want a proper tech design that can apply to many mods, not just whatever hook you need for your mod specifically.
Your changes to RDB are exactly that, very specifically made for your mod only. You will have to come back with another design.

PR #2632, opened 2024-04-06
Replace PC Gold Weight Literal With Loaded Value

Mentill:
The weight of Gold Pieces in the player's inventory is stored in DaggerfallBankManager.cs as a literal value. This change instead indirectly loads it in from the itemTemplates array set in ItemHelper.cs

The intended result is that it thus exposes to mods the weight of gold pieces in the player's inventory, which it does according to my tests. It also appropriately affects certain transfers of gold pieces.

### DIFF
--- a/Assets/Scripts/Game/DaggerfallBankManager.cs
+++ b/Assets/Scripts/Game/DaggerfallBankManager.cs
@@ -77,7 +77,7 @@ namespace DaggerfallWorkshop.Game.Banking
 
     public static class DaggerfallBankManager
     {
-        public const float goldUnitWeightInKg = 0.0025f;
+        public static float goldUnitWeightInKg = DaggerfallUnity.Instance.ItemHelper.GetItemTemplate(ItemGroups.Currency,0).baseWeight;
         private const float deedSellMult = 0.85f;
         private const float housePriceMult = 1280f;
         private const uint loanRepayMinutes = DaggerfallDateTime.DaysPerYear * DaggerfallDateTime.MinutesPerDay;
--- a/Assets/Scripts/Game/DaggerfallBankManager.cs
+++ b/Assets/Scripts/Game/DaggerfallBankManager.cs
@@ -77,7 +77,17 @@ namespace DaggerfallWorkshop.Game.Banking
 
     public static class DaggerfallBankManager
     {
-        public static float goldUnitWeightInKg = DaggerfallUnity.Instance.ItemHelper.GetItemTemplate(ItemGroups.Currency,0).baseWeight;
+        private static float cachedGoldUnitWeightInKg = -1.0f;
+        public static float goldUnitWeightInKg
+        {
+            get
+            {
+                if (cachedGoldUnitWeightInKg == -1.0f)
+                    cachedGoldUnitWeightInKg = DaggerfallUnity.Instance.ItemHelper.GetItemTemplate(ItemGroups.Currency, 0).baseWeight;
+                return cachedGoldUnitWeightInKg;
+            }
+        }
+        
         private const float deedSellMult = 0.85f;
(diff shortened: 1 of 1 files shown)


numidium (approved):

KABoissonneault (changes requested):
I have a few issues with the change as it is implemented:
1) `goldUnitWeightInKg` is now a publicly mutable variable, which means its value can change at any time
2) Being a `static` variable, I can't be certain of when its initialization from the item template is made. How do we know the modded value is being taken into account?

It could be fixed using instead a C# property with private caching.

```
private static float cachedGoldUnitWeightInKg = -1.0f;
public static float goldUnitWeightInKg 
{
    get
    {
        if(cachedGoldUnitWeightrInKg == -1.0f)
            cachedGoldUnitWeightInKg = DaggerfallUnity.Instance.ItemHelper.GetItemTemplate(ItemGroups.Currency,0).baseWeight;
        return cachedGoldUnitWeightInKg;
    }
} 
```

The property cannot be modified, and the value would not be invoked until anything in DFU actually relies on the weight of gold, which should be long past the point at which mods modify item templates and such.

numidium:
> I have a few issues with the change as it is implemented:
> 
>     1. `goldUnitWeightInKg` is now a publicly mutable variable, which means its value can change at any time
> 
>     2. Being a `static` variable, I can't be certain of when its initialization from the item template is made. How do we know the modded value is being taken into account?

Good point. Couldn't we use also just use a lambda expression to just directly reference the item template?

`public static float goldUnitWeightInKg => DaggerfallUnity.Instance.ItemHelper.GetItemTemplate(ItemGroups.Currency, 0).baseWeight;`

KABoissonneault:
> Good point. Couldn't we also just use a lambda expression to directly reference the item template?

I was considering it, but I wasn't sure if we wanted to support mods that modified the template in the middle of gameplay. Plus there's the performance consideration of doing that access each time, though caching is not free either.

Assets/Scripts/Game/DaggerfallBankManager.cs
```diff
@@ -77,7 +77,7 @@ public enum ShipType
 
     public static class DaggerfallBankManager
     {
-        public const float goldUnitWeightInKg = 0.0025f;
+        public static float goldUnitWeightInKg = DaggerfallUnity.Instance.ItemHelper.GetItemTemplate(ItemGroups.Currency,0).baseWeight;
```
ajrb, line 80: The base weight of gold is set as 0.5kg in item templates so this is quite a change, gold pieces will start weighing 200 times as much as it should if this PR is merged as-is.

Mentill: ItemGroups.Currency,0 returns item template index 276, which is Gold Pieces, which has a baseWeight of 0.0025
AFAIK, Gold Pieces is the only element in ItemGroups.Currency
Even if not, I tested this change with a mod I made, and it returns the baseWeight from index 276.

Mentill:
> The property cannot be modified, and the value would not be invoked until anything in DFU actually relies on the weight of gold, which should be long past the point at which mods modify item templates and such.

I tested your code and it seems to work fine, so I'll update it to that.

Assets/Scripts/Game/DaggerfallBankManager.cs
ajrb: You're correct, I found the 'Gold' entry which is for a bar of gold. Nevermind, thanks for checking.

KABoissonneault:
No need to request my review. I'm leaving this on hold because I'm leaving DFU stable while we wait for potential v1.1 fixes. PRs will reopen once I have a plan for v1.2

KABoissonneault (approved):

PR #2634, opened 2024-04-07
Add Contextual Macros for Capitalized Gender-Specific Pronouns

Mentill:
Some quest dialog and strings include GSP (gender-specific pronoun) macros (e.g. %g1). However, the macro handler always makes these lower-case, which is improper at the beginning of a sentence, in English at least.

Instead of making GSP macros always capitalized, and unnecessarily deifying characters, it makes more sense to capitalize only the ones at the beginning of a sentence. I found an earlier proposal which suggested code changes to detect when one is used at the beginning of a sentence. However, this detection becomes problematic when accounting for whitespace, unusual punctuation, and other languages whose sentences may end with different symbols.

Therefore, I added new macros which capitalize the first letter of the GSP, which can be placed in quest/string files as needed to replace GSPs found at the beginning of sentences. I plan on modifying - if this is merged - my upcoming quest dialog improvements, in order to utilize these new macros.
I tested all these new macros and they do work; macro names are case-sensitive.

### DIFF
--- a/Assets/Scripts/Utility/MacroHelper.cs
+++ b/Assets/Scripts/Utility/MacroHelper.cs
@@ -34,6 +34,7 @@ namespace DaggerfallWorkshop.Utility
 
         public static System.Random random = new System.Random();
 
+        
         #region macro definitions and handler mappings
 
         static Dictionary<string, MacroHandler> macroHandlers = new Dictionary<string, MacroHandler>()
@@ -101,12 +102,6 @@ namespace DaggerfallWorkshop.Utility
             { "%g2self", Pronoun2self },// Himself/Herself etc...
             { "%g3", Pronoun3 },  // His/Her
             { "%g4", Pronoun4 },  // His/Hers
-            { "%G", PronounCap }, // He/She, first letter capitalized
-            { "%G1", PronounCap }, // He/She, first letter capitalized
-            { "%G2", Pronoun2Cap }, // Him/Her, first letter capitalized
-            { "%G2self", Pronoun2selfCap }, // Himself/Herself, first letter capitalized
-            { "%G3", Pronoun3Cap }, // His/Her, first letter capitalized
-            { "%G4", Pronoun4Cap }, // His/Hers, first letter capitalized
             { "%gii", GoldCarried }, // Amount of gold in hand
             { "%gdd", GodDesc }, // God description i.e. God of Logic
             { "%god", God }, // God of current region or current temple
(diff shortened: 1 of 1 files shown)


Jagget (approved):

Assets/Scripts/Utility/MacroHelper.cs
```diff
@@ -1309,13 +1315,42 @@ public static string Pronoun3(IMacroContextProvider mcp)
...
         {   // %g4
             if (mcp == null) return null;
             return mcp.GetMacroDataSource().Pronoun4();
         }
-
+        public static string PronounCap(IMacroContextProvider mcp)
+        {   // %G & %G1
+            if (mcp == null) return null;
+            string tmpPronoun = mcp.GetMacroDataSource().Pronoun();
+            return tmpPronoun.Substring(0, 1).ToUpper() + tmpPronoun.Substring(1);
```
petchema, line 1327: create a function to capitalize strings

Assets/Scripts/Utility/MacroHelper.cs
```diff
@@ -101,6 +101,12 @@ public static class MacroHelper
             { "%g2self", Pronoun2self },// Himself/Herself etc...
             { "%g3", Pronoun3 },  // His/Her
             { "%g4", Pronoun4 },  // His/Hers
+            { "%G", PronounCap }, // He/She, first letter capitalized
+            { "%G1", PronounCap }, // He/She, first letter capitalized
+            { "%G2", Pronoun2Cap }, // Him/Her, first letter capitalized
+            { "%G2self", Pronoun2selfCap }, // Himself/Herself, first letter capitalized
+            { "%G3", Pronoun3Cap }, // His/Her, first letter capitalized
+            { "%G4", Pronoun4Cap }, // His/Hers, first letter capitalized
```
ajrb, line 109: Please move these below the comment "// DF Unity - new macros:" so they're separated from original game macros

ajrb (changes requested):

Mentill:
Ok, I created a function and moved the macro definitions.

ajrb (approved):

petchema (approved):

Assets/Scripts/Utility/MacroHelper.cs
```diff
@@ -235,6 +236,12 @@ public static class MacroHelper
             { "%pg2", PlayerPronoun2 }, // Him/Her (player)
             { "%pg2self", PlayerPronoun2self },// Himself/Herself (player)
             { "%pg3", PlayerPronoun3 },  // His/Her (player)
```
Jagget, line 238: Do we need capitalized versions of `%pgXX` macros? 
Like in
```
<ce> "He ran this way!"
<ce> You started hearing
<ce> the sound of heavy boots behind you.
```

Mentill: This PR's intended scope is to resolve issues with classic quests (I found up to 80 that have this issue). If someone wants to make new quests that use uncapitalized macros at the start of a sentence, those macros can be added as needed; otherwise, we'd be making new macro handlers for every macro that can return an uncapitalized string.

KABoissonneault: I wouldn't say every. Player pronouns to match character pronouns makes sense from a consistency point of view. I still thinks the PR makes sense as it is

Jagget (approved):

PR #2635, opened 2024-04-07
Allow other transportation items to be added by mods

ajrb:
- Allows custom transportation items to be contributed by mods and actually have a chance to appear in shops.
- Allow new horses to enable the horse transportation option. (this could be done for carts as well if desired, but I don't think people really use this much so won't want to add any new ones)
- Allow custom items to have animated icons in inventory. (bugfix)


KABoissonneault (approved):
Code looks fine. Leaving ajrb in charge of maintaining this

Issue #2636, opened 2024-04-08
Mod Builder removing book mapping contributions


ajrb:
I recently released v2.6 of the Archaeologists Guild mod built using the DFU mod builder tool and up to date source code. This release had the book mapping contribution removed from the dfmod.json file by the latest mod builder. I didn't notice this when I committed because the change was at the bottom of the file but this means that v2.6 had no book and so the code for joining the guild which gives the player that book was failing and no one could join anymore. See this link for the json the mod builder removed from my json file: https://github.com/ajrb/dfunity-mods/commit/3654dd2611a6c861e8fbb4078dbfa95448926662#diff-e91627110ad60d228a2c9c75471e495fe7fc2ea922c2a7b5eca2cab912f6c99cL84-L89

This issue was caused by the work in this PR: https://github.com/Interkarma/daggerfall-unity/pull/2580

More specifically this commit: https://github.com/Interkarma/daggerfall-unity/commit/dabcbe9d000a8e785340617ca5233fdee9ad2096

I reverted that commit on my local DFU code clone and restored the json book contribution section. Re-built v2.6.1 of the mod and the mod builder no longer removes this section and the mod works properly

KABoissonneault:
Fixed by #2637

PR #2637, opened 2024-04-08
Fixed book mapping contributes not being detected on Windows

KABoissonneault:
Fix for #2636.

Before #2580, we used to leave contributes untouched if no contributes were detected. Instead now, we always update contributes, removing files that are not detected by the function.

Detection of book mappings has always been broken on Windows. The path returned by `Path.GetDirectoryName` uses `\` to separate paths, but the check for Books/Mapping checked with a forward slash.

I instead changed this function to properly detect the path Books/Mapping on any platform. As a result, the books should now properly be detected on Windows, and updating the mapping should always give the correct result.


ajrb (approved):
Thanks.

petchema (approved):

PR #2639, opened 2024-04-09
Add limited RMB Editor / World Data Editor integration 

drcarademono:
While the RMB and World Data Editors are both quite functional, in their current implementation they are poorly integrated. For example, if a modder is editing a temple RMB and wants to edit the temple's interior, the modder would first have to export the temple to a building file, which requires a number of clicks through file paths, assigning that temple a name, etc. Then the modder would open the temple building file in the WDE by once again clicking through file paths, then after modifying the building, the modder would have to save the building file, then import it again in the RMB Editor. In sum, modifying any building in the RMB Editor requires navigating _four_ separate Import/Export File dialogue boxes.

This PR would reduce that to zero. It adds two buttons the RMB Editor. The top right "WDE" one immediately opens the selected building in the World Data Editor. The bottom "Import from WDE" button imports the building right back from the World Data Editor. No need to mess with jsons or remember which filename is which.
image

Just click the building, click WDE, and it's open:
image

Make your changes in the WDE, go back to the RMB Editor, click the building again, click "Import from WDE," and your changes to the building interior are now in the RMB:
image

This implementation is a little less direct than it perhaps could be, as I couldn't figure out how to pass the building interior data directly from the RMB Editor to the WDE Editor and back again. Instead the PR writes the building to a temporary json which the WDE opens, and the WDE likewise writes a temporary json that the RMB Editor opens. In each case the json is deleted immediately after being open so it shouldn't leave any clutter behind.


KABoissonneault (approved):
I have some user experience concerns with the new changes, but the code itself is not intrusive, and does not add new warnings/errors in the editor. We can improve on this once users actually have something to use

PR #2642, opened 2024-04-12
Enable variant custom guilds

ajrb:
Try creating guild using variant so that custom guilds can have sub-groups like Temples and Knight Orders do.

Also entended the guild groups enum so we have plenty of guild entries available. Pretty sure this has no detremental effect.

### DIFF
--- a/Assets/Scripts/API/FactionFile.cs
+++ b/Assets/Scripts/API/FactionFile.cs
@@ -593,6 +593,12 @@ namespace DaggerfallConnect.Arena2
             Witches = 22,
             Vampires = 23,
             Orsinium = 24,
+            GGroup25 = 25,
+            GGroup26 = 26,
+            GGroup27 = 27,
+            GGroup28 = 28,
+            GGroup29 = 29,
+            GGroup30 = 30,
         }
 
         /// <summary>
--- a/Assets/Scripts/Game/Guilds/GuildManager.cs
+++ b/Assets/Scripts/Game/Guilds/GuildManager.cs
@@ -178,7 +178,11 @@ namespace DaggerfallWorkshop.Game.Guilds
                         return (IGuild)Activator.CreateInstance(guildType, new object[] { KnightlyOrder.GetOrder(variant) });
 
                     default:
-                        return (IGuild)Activator.CreateInstance(guildType);
+                        try {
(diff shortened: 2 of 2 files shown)


Assets/Scripts/Game/Guilds/GuildManager.cs
```diff
@@ -178,7 +178,11 @@ private IGuild CreateGuildObj(FactionFile.GuildGroups guildGroup, int variant =
                         return (IGuild)Activator.CreateInstance(guildType, new object[] { KnightlyOrder.GetOrder(variant) });
 
                     default:
-                        return (IGuild)Activator.CreateInstance(guildType);
+                        try {
+                            return (IGuild)Activator.CreateInstance(guildType, new object[] { variant });
```
KABoissonneault, line 182: I'd rather see something with `if(guildType.GetConstructor(new Type[] { typeof(int) }) != null)` than exceptions. There is the performance concern, sure, but also I set my debugger to always break when an exception is thrown in user code, so I'd have to mute that exception type entirely
Suggestion comes from: https://stackoverflow.com/a/19611584/2356414

ajrb: Originally I was going to require all custom guild classes implement default and variant param constructors but then I figured this could be handled by a try catch so mod authors were not required to do that even though I'd still recommend it. I can change it to check rather than handle the exception like you suggest if you like.

KABoissonneault: Yes, of course it should be optional, for backward compatibility and general convenience. It's really about not relying on exceptions if we can avoid it

Assets/Scripts/API/FactionFile.cs
```diff
@@ -593,6 +593,12 @@ public enum GuildGroups
             Witches = 22,
             Vampires = 23,
             Orsinium = 24,
+            GGroup25 = 25,
```
KABoissonneault, line 596: Btw this is not really needed. Users can easily write `const GuildsGroups MyRenegadeGuild = (GuildGroups)25;` themselves, no need to extend the enum here. No need to revert it either, but also let's not keep adding every few versions.

KABoissonneault (approved):

PR #2644, opened 2024-04-14
Improve custom door support for custom models

### LINKED ISSUE #2643: Static door support for mods is lacking
**Describe the issue**

Static door (doors that transition between exteriors and interiors) support currently only works for model overrides of vanilla building models with DF door data. Each custom door references a static door on the original model for transitioning inside or back out again. The original doors can be suppressed from being active but the data must exist for a custom door to work. This use of existing static door data also restricts where the custom doors are located to one of the original doors for purposes of finding the closest start marker for positioning player inside.

This means there is currently no way to add doors to custom models, even if they're based on original models - if a custom model id is used, there is no door data available so it doesn't work.

Intention is to improve CustomDoor component. 

**Improvement todo list**

1. Allow custom building models to have transition doors defined that work.
2. Allow custom doors for custom dungeon entrance models.
3. Allow custom interior exit doors for custom interior models.

ajrb:
Improve custom door support for custom models, enabled for building exterior / interior and dungeon exterior doors. CustomDoor now creates its own StaticDoor data same as any DF model StaticDoor, allowing completely new custom models using non-classic model id's to be supported. 

Retained the existing copy mechanism that uses the boxcollider to test for a hit and the copied door data. This copy mechanism has been retained for backwards compatibility, code could be simplified if it were removed. I don't think it's that useful anymore. If any published mods use CustomDoor then they should continue to work identically as long as the copy index was set.

Player placement for both entering and exiting is now correct. The exception is when replacing a classic model with all building doors disabled in favour of custom doors all copying these classic doors.

Included 2 boxcollider presets for custom doors facing X or Z with standard DF door size 1.25, 2, 0.5. They're in Assets/Editor/ because Unity insisted they went in Assets folder and that seemed the most appropriate place. Will remove from PR if requested.

Ninelan will be testing this out to check my testing didn't miss anything. Instructions added to the wiki here: https://github.com/Interkarma/daggerfall-unity/wiki/Custom-Doors

Resolves #2643

### DIFF
--- a/Assets/Scripts/Internal/DaggerfallInterior.cs
+++ b/Assets/Scripts/Internal/DaggerfallInterior.cs
@@ -482,7 +482,7 @@ namespace DaggerfallWorkshop
 
                 // Add any DF model static doors unless suppressed
                 if (modelData.Doors != null && !disableClassicDoors)
-                    doors.AddRange(GameObjectHelper.GetStaticDoors(ref modelData, entryDoor.blockIndex, entryDoor.recordIndex, modelMatrix));
+                    doors.AddRange(staticDoors);
 
 
                 if ((modelGO != null) && (isAutomapRun))
--- a/Assets/Scripts/Internal/DaggerfallInterior.cs
+++ b/Assets/Scripts/Internal/DaggerfallInterior.cs
@@ -444,9 +444,11 @@ namespace DaggerfallWorkshop
                 Vector3 modelScale = RMBLayout.GetModelScaleVector(obj);
                 Matrix4x4 modelMatrix = Matrix4x4.TRS(modelPosition, Quaternion.Euler(modelRotation), modelScale);
 
-                // Does this model have doors?
+                // Does this Daggerfall model have any static doors?
+                bool disableClassicDoors = false;
+                StaticDoor[] staticDoors = null;
                 if (modelData.Doors != null)
-                    doors.AddRange(GameObjectHelper.GetStaticDoors(ref modelData, entryDoor.blockIndex, entryDoor.recordIndex, modelMatrix));
+                    staticDoors = GameObjectHelper.GetStaticDoors(ref modelData, entryDoor.blockIndex, entryDoor.recordIndex, modelMatrix);
 
                 // Inject custom GameObject if available
                 GameObject modelGO = MeshReplacement.ImportCustomGameobject(obj.ModelIdNum, node.transform, modelMatrix);
@@ -471,6 +473,17 @@ namespace DaggerfallWorkshop
                         dfMesh.SetClimate(climateBase, climateSeason, WindowStyle.Disabled);
                     }
                 }
+                else
+                {
+                    // For a custom model initialise any custom doors
+                    List<StaticDoor> customStaticDoors = CustomDoor.InitDoorsInterior(modelGO, staticDoors, entryDoor.blockIndex, entryDoor.recordIndex, modelMatrix, out disableClassicDoors);
+                    doors.AddRange(customStaticDoors);
+                }
+
+                // Add any DF model static doors unless suppressed
+                if (modelData.Doors != null && !disableClassicDoors)
+                    doors.AddRange(GameObjectHelper.GetStaticDoors(ref modelData, entryDoor.blockIndex, entryDoor.recordIndex, modelMatrix));
+
 
                 if ((modelGO != null) && (isAutomapRun))
                     modelGO.AddComponent<AutomapModel>();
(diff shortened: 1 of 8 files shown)


Assets/Scripts/Utility/AssetInjection/Components/CustomDoor.cs
```diff
@@ -4,89 +4,148 @@
...
+                    // Calculate the door normal
+                    Vector3 doorCenter = allCustomDoors[i].DoorTrigger.center;
+                    Vector3 doorSize = allCustomDoors[i].DoorTrigger.size;
+                    Vector3 doorNormal = new Vector3();
+                    if (doorSize.x < doorSize.z && doorSize.x < doorSize.y)
+                        doorNormal.x = Mathf.Sign(doorCenter.x);
+                    else if (doorSize.z < doorSize.x && doorSize.z < doorSize.y)
+                        doorNormal.z = Mathf.Sign(doorCenter.z);
+                    else if (doorSize.y < doorSize.x && doorSize.y < doorSize.z)
+                        doorNormal.y = Mathf.Sign(doorCenter.y);
```
ajrb, line 96: @KABoissonneault This part calculating the door normal may not be how it should be done, but it does seem to work. This was my 4th revision of this code and I am open to better methods if you want to suggest anything. My 3d graphics knowledge is rudimentary at best.

Assets/Scripts/Utility/RMBLayout.cs
```diff
@@ -840,47 +840,59 @@ public static bool IsRepairShop(DFLocation.BuildingTypes buildingType)
...
-                        staticBuilding.size = new Vector3(modelData.DFMesh.Size.X, modelData.DFMesh.Size.Y, modelData.DFMesh.Size.Z) * MeshReader.GlobalScale;
+                        if (go == null || modelData.Indices != null) {
+                            staticBuilding.centre = new Vector3(modelData.DFMesh.Centre.X, modelData.DFMesh.Centre.Y, modelData.DFMesh.Centre.Z) * MeshReader.GlobalScale;
+                            staticBuilding.size = new Vector3(modelData.DFMesh.Size.X, modelData.DFMesh.Size.Y, modelData.DFMesh.Size.Z) * MeshReader.GlobalScale;
+                        } else {
+                            Renderer goRenderer = go.GetComponent<Renderer>();
+                            //staticBuilding.centre = new Vector3(goRenderer.bounds.center.x, goRenderer.bounds.center.y, goRenderer.bounds.center.z);
+                            staticBuilding.size = new Vector3(goRenderer.bounds.size.x, goRenderer.bounds.size.y, goRenderer.bounds.size.z);
+                            staticBuilding.centre = staticBuilding.size / 2;    // Seems to replicate DFMesh.Centre whereas goRenderer.bounds.center.z gives position.z for some reason.
```
ajrb, line 879: The commented out line was what I thought I should use for the centre but the z is the same as parent transform position z which is incorrect. Unless anyone objects I will remove the commented out line.

ajrb: Removed.

ajrb: Actually it turns out calculating this works fine but only if the centre of the custom model is at the exact centre which is true for DF models but not true of the test custom model. Trying to find a solution to avoid all models having to have a dead centre 0,0,0

ajrb:
So I've found a few issues when testing with Ninelan, am addressing them but hold off on reviewing this PR for now.

Assets/Scripts/Utility/RMBLayout.cs
ajrb: Solved this now.

Assets/Scripts/Internal/DaggerfallInterior.cs
```diff
@@ -471,6 +473,17 @@ private void AddModels(PlayerGPS.DiscoveredBuilding buildingData, bool isAutomap
...
+                else
+                {
+                    // For a custom model initialise any custom doors
+                    List<StaticDoor> customStaticDoors = CustomDoor.InitDoorsInterior(modelGO, staticDoors, entryDoor.blockIndex, entryDoor.recordIndex, modelMatrix, out disableClassicDoors);
+                    doors.AddRange(customStaticDoors);
+                }
+
+                // Add any DF model static doors unless suppressed
+                if (modelData.Doors != null && !disableClassicDoors)
+                    doors.AddRange(GameObjectHelper.GetStaticDoors(ref modelData, entryDoor.blockIndex, entryDoor.recordIndex, modelMatrix));
```
KABoissonneault, line 485: Do we have to call `GameObjectHelper.GetStaticDoors(ref modelData, entryDoor.blockIndex, entryDoor.recordIndex, modelMatrix)` a second time here, or would reusing `staticDoors` work?

KABoissonneault:
For the most part, I'd like to unblock this and let ajrb figure out the feature. I just have one minor issue. Then, it should be ready to merge

Assets/Scripts/Internal/DaggerfallInterior.cs
ajrb: Yeah it would, I'll push a change when I get the chance to test it.

ajrb: Tested and committed.

KABoissonneault (approved):

PR #2645, opened 2024-04-16
Add %pg4 for missing pronoun His/Hers

Jagget:
This is part for player-realted pronoun. Other languages need that part.
Didn't address this in commit 83ce86660e39021e4682c0b8ff5ca730efa4b266
His/Her is represented by %pg3, but there was no equivalent for His/Hers.
This is now represented by %pg4.
"pronounHers" text already available in localization tables.


KABoissonneault:
Change seems good. 

I'm wondering, doesn't the fact that DFU reuses the same localization key in different situations an issue in certain languages? 
For example, %g2 is him/her, and %g3 is his/her, and it uses the key "pronounHer" for both, which means a language has to translate the feminine of both %g2 and %g3 with the same word, which works in English, but I'm not sure it's universal.
Same thing for %g3 his and %g4 his.

I'm wonder if we should just duplicate these entries into two different keys, so languages where they differ can do that. Either just put a number at the end and let the localizer figure it out, or go fancier with pronounHisPossessivePronoun vs pronounHisPossessiveDeterminer

Jagget:
Yes, that fact that DFU didn't use `pronounHers` was a drawback, that why IK add `%g4` by my request. And I just recently realized that I need `%pg4` as well. 

With `go fancier` approach we might get too fancy. For example I need 7 forms of `she` and 8 forms of `he` for Russian 😄 
```
она её неё ей ней ею нею
```
```
он его ему им него нему ним нём
```

KABoissonneault:
Well we can't invent new context that DFU doesn't provide. I'm thinking that since DFU already distinguishes this context, we should provide two different hooks for localization, that's really not too hard.
We could even have a fallback so that if a language does not implement pronounHis2, it falls back to pronounHis.
Hopefully it's better than what we have now.

Just a suggestion.

Daneel53:
There are anyway translated values that are missing into Internal_Strings to adress the pronoun issue.
For example, in MacroHelper.cs, both %g2 and %pg2 sends to pronounHim and pronounHer, when we should use two different translations in both cases.
Example with %g2: Get him! or Get her! In French: Attrapez-le! or attrapez-la!
Example with %pg2: That's him! or That's her! In French: C'est lui! or C'est elle!

We should have a different translation possible for each macro: different macro, different translation.

I suggest into Internal_Strings in English:
Him,him
Her,her
pronounHim,him
pronounHer,her

That will become in another language as mine:
Him,le
Her,la
pronounHim,lui
pronounHer,elle

And so a quest sentence as "I will pay you 100 gold  pieces if you find %g2" shall become "Je vous paierai 100 pièces d'or si vous le retrouvez" and not the erroneous "Je vous paierai 100 pièces d'or si vous lui retrouvez" as today.

Daneel53:
I agree that %pg4 should be added besides %g4, but we must go further to be able to use it properly in other languages that English.
If we consider that the proposal is accepted, we will arrive to a situation where the English pronouns "his" and "her" shall be used twice in two different macros: %pag2 and %pg3 for "her", and %g3 and %g4 for "his", with only six possible translation strings in Internal_Strings.csv.

Pronouns and macros_2024-06-19_14-31-05

I propose to introduce two new pronoun definitions, pronounHer2 and pronounHis2 following the new definition given in the snapshot, and to add the two new definition into Internal_Strings.csv.

As you can see, this will change nothing for English that will keep on using "his" and "her" where needed, but this shall permit, for all languages other than English that need it, to give different values for each macro %gx and %pgx.

Jagget:
@Daneel53 
Updated the PR. Added pronounHis2/pronounHer2

ivankuzyk:
Hi all!  I have one question - is it possible to add additional macros in the Dagerfall localization, for example: %g5 %g6 %g7 %g8 %g9?

KABoissonneault:
You can open an issue if you want to open this discussion. Adding new macros is easy enough, but you'd have to define when to use each and what the expected output should be

ivankuzyk:
I want to make a Ukrainian localization of the game and I came across this problem with macros, here's an example: 
 він/вона - %g1,%g
 йому/їй - %g2
 його/її - %g4
 ним/нею - %g2self
 нього/неї - ????
Is it at least possible to introduce the %g5 macro?

Jagget:
I noticed that you don't use `g3`. Granted, in the current state, `g3` is partially equal to `g2` and `g4`, but after this PR is merged, you'll be able to make all 4 macros unique. This will give you 4 different variants. You can also use prefixes and [postfixes](https://github.com/Interkarma/daggerfall-unity/pull/2552) with macros, which also increases versatility. 

Adding a new macro is easy, yes, but needs to be justified for widespread use, I think.

ivankuzyk:
The %g3 macro is broken for both Russian and Ukrainian localizations, you yourself once begged Interkarma to add the %g4 macro to replace %g3. I propose to add a few additional macros (%g5 %g6 %g7 %g8 %g9) with the ability to localize them in their own way, and then, I think, everyone will take the right number of macros for their localization.

Jagget:
So you're saing, that you need 4 different macros. g1, g2, g4, and **`g5`**, but you don't need g5!

I just told you that after this PR is merged, you'll be able to set **g3** in a proper way. So you'll get your 4 unique macros, it'll be g1, g2, g3, and g4.

ivankuzyk:
> То есть вы говорите, что вам нужны 4 разных макроса. g1, g2, g4 и **`g5`**, но вам не нужен g5!
> 
> Я только что сказал вам, что после объединения этого PR вы сможете правильно настроить **g3 . Таким образом, вы получите 4 уникальных макроса, это будут g1, g2, g3 и g4.**
How can this be done?

Jagget:
image

ivankuzyk:
Interesting, so you're saying we can do it like this:
він/вона - %g1,%g
йому/їй - %g2
його/її - %g4
ним/нею - %g2self
нього/неї - %g3

Then in which file can I localize %g3 to use it?

Jagget:
The same file. Add **pronounHer2** and **pronounHis2** next to pronounHer and pronounHis

ivankuzyk:
> Тот же файл. Добавьте **pronounHer2** и **pronounHis2** рядом с pronounHer и pronounHis

I tried it, it doesn't work.

ivankuzyk:
I tried it, it doesn't work. From the macro "%g3", the game displays the localization with the string "pronounHer"

KABoissonneault:
Can you discuss this somewhere else? I don't think this PR should add more

KABoissonneault (approved):

PR #2654, opened 2024-05-19
FPS Weapon Movement variable

RedRoryOTheGlen:
Adds an Offset variable that lets other scripts move the FPS Weapon sprite while preserving the original weaponPosition variable, mostly made to support mods that add weapon sway/bob.

 [SAMPLE](https://streamable.com/p2o26c) of Offset working with mod (not included)

If nothing tries to change the FPSWeapon Offset, it should behave entirely like usual. I hope. Suggestions for better implementation are always welcome.

### DIFF
--- a/Assets/Scripts/Game/FPSWeapon.cs
+++ b/Assets/Scripts/Game/FPSWeapon.cs
@@ -106,6 +106,7 @@ namespace DaggerfallWorkshop.Game
         public WeaponStates WeaponState { get { return weaponState; } }
 
         public Color Tint { get; set; } = Color.white;
+        public Vector2 Offset { get; set; } = Vector2.zero;
 
         #endregion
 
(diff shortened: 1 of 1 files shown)


RedRoryOTheGlen:
Old Offset variable was renamed to Position. Added two new variables: Scale and Offset (new).
1. Scale allows a mod to change the size of the sprite.
2. The new Offset variable allows a mod to change the position of the sprite relative to its width/height without having access to the original weaponPosition rect. For example, adding a value of 0.5 to Offset.x will move the sprite horizontally by sprite.width/2 pixels.

Added a reset to the variables in LateUpdate. This lets multiple mods simultaneously affect the Position, Scale and Offset of FPSWeapon.

It should behave entirely like usual as before. I think only the LateUpdate check would run as long as the variables are all zero'd.

KABoissonneault (approved):

PR #2666, opened 2024-06-17
Preserve enemy orientation when hit by the Wabbajack

petchema:
A new gameObject is actually created for the transmuted enemy, facing north. Keep the orientation of the replaced gameObject instead.

Test: you can give yourself the Wabbajack by typing "addArtifact 6" in the console

Forums: https://forums.dfworkshop.net/viewtopic.php?t=6709&sid=e5aade01a84f5052bc8b32e13e7ee934

### DIFF
--- a/Assets/Scripts/Game/MagicAndEffects/Effects/Special/WabbajackEffect.cs
+++ b/Assets/Scripts/Game/MagicAndEffects/Effects/Special/WabbajackEffect.cs
@@ -69,16 +69,17 @@ namespace DaggerfallWorkshop.Game.MagicAndEffects.MagicEffects
                 if (enemy == null || enemy.WabbajackActive)
                     return null;
 
-                // Get new enemy career and transform
-                MobileTypes enemyType = careerIDs[Random.Range(0, careerIDs.Length)];
-                if ((int)enemyType == enemy.CareerIndex)
-                    enemyType = (MobileTypes)(((int)enemyType + 1) % careerIDs.Length);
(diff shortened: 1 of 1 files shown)


petchema:
I _think_ it's classic behavior, I haven't tried getting the Wabbajack in classic, but it seems @numidium did years ago:
https://www.youtube.com/watch?v=exj1xrX_vtU&t=296s
most enemies are hit facing the player and keep facing the player, but once around 4:56 he hits a guard in the back and he transforms in a fire atronach, apparently facing the same direction

KABoissonneault:
Yeah no worries, it's not an area where I expect classic consistency. Let's just do what seems more natural, your change should be good

petchema:
Noticed something else funky in the code making sure the enemyType changed:
```
                if ((int)enemyType == enemy.CareerIndex)
                    enemyType = (MobileTypes)(((int)enemyType + 1) % careerIDs.Length);
```
If the goal is to pick the next existing MobileType, then modulo should be MobileTypes.Length.
But it would mean the Wabbajack can generate a monster type that's not in careerIDs, which looks suspiciously undesirable.

And if the goal was to pick the next MobileType in careerIDs array, then the code is totally incorrect, that's the array index that should be incremented, not the value taken from the array (also it would put a higher probability to generate one MobileType than others, for no good reason).

So I suggest replacing this code by a simple retry (much simpler than other correct alternatives).

KABoissonneault (approved):

Issue #2656, opened 2024-05-20
Add code to permit the use of grammar processors


Daneel53:
Hello,
I would like to ask officially to the team that is today in charge of DFU code to accept to introduce into DFU some code that will permit mods in other languages than English to use a grammar processor in order to produce better translated text in a grammatical point of view.

I wrote a document that explain everything to be clearer.

If the modification is accepted by the team and implemented in a future release of DFU, the French translation shall be the first to benefit of a grammar processor to ameliorate the French translation, it is already working on my PC. Then other translators may try, if they feel ready, to write a mod with their own grammar processor for their language, the French one being an example for them.
[Grammar processors for DFU.pdf](https://github.com/Interkarma/daggerfall-unity/files/15379255/Grammar.processors.for.DFU.pdf)

You can also contact me via my Discord account for further discussion.

Daneel53:
Hi,
Think that I've finally found all the scripts where I have to add calls to both methods of the grammar processor in order to deal with grammar tokens in all the screens of the game.

For calls to ProcessGrammar():

- Game\PlayerActivate.cs
- Game\TextManager.cs
- Game\UserInterface\MultiFormatTextlabel.cs
- Game\UserInterfaceWindows\DaggerfallExteriorAutomapWindow.cs
- Game\UserInterfaceWindows\DaggerfallPlayerHistoryWindow.cs
- Game\UserInterfaceWindows\DaggerfallTalkWindow.cs

For calls to SetGenreHero():

- API\Save\CharacterRecord.cs: get character genre when loading a classic save
- Game\Serialization\SerializablePlayer.cs: get character genre when loading a DFU save
- Game\UserInterfaceWindows\CreateCharGenderSelect.cs: genre selection during a new game creation

I've tested in Unity editor all modification with the current code contained into the Github master branch, everything's working fine.

I'm preparing the package and I will create the PR in Github as soon as I will be ready.

Daneel53:
Final state of PR 2667 was integrated into DFU on May 17, 2026.

PR #2667, opened 2024-06-21
Add the use of grammar processors

Daneel53:
Create two scripts with new classes into Game/Localization and add calls to grammar processor methods into nine existing scripts that display text to screen.
See issue #2656 for explanation.

### DIFF
--- a/Assets/Scripts/Localization/Grammar.cs
+++ b/Assets/Scripts/Localization/Grammar.cs
@@ -14,6 +14,7 @@ namespace GrammarModule
     {
         public abstract string ProcessGrammar(string text);
         public abstract void SetGenreHero(string Genre);
+        public abstract void SetGenreNPC(string Genre);
     }
 
     public static class GrammarManager
--- a/Assets/Scripts/Localization/Grammar.cs
+++ b/Assets/Scripts/Localization/Grammar.cs
@@ -8,19 +8,20 @@
 
 using DaggerfallWorkshop.Game.Entity;
 using DaggerfallWorkshop.Localization.LanguageRules;
+using System;
 
 namespace DaggerfallWorkshop.Localization
 {
     public abstract class GrammarRules
     {
         public abstract string ProcessGrammar(string text);
-        public abstract void SetHeroGender(Genders Gender);
-        public abstract void SetNPCGender(Genders Gender);
+        public abstract void SetHeroGenderGetter(Func<Genders> HeroGenderGetter);
+        public abstract void SetNPCGenderGetter(Func<Genders> NPCGenderGetter);
     }
 
     public static class GrammarManager
     {
-        public static GrammarRules grammarProcessor = (GrammarRules) new DefaultGrammarRules();
+        public static GrammarRules grammarProcessor = new DefaultGrammarRules();
     }
 }
 
--- a/Assets/Scripts/Localization/Grammar.cs
+++ b/Assets/Scripts/Localization/Grammar.cs
@@ -6,15 +6,16 @@
 // Original Author: Daneel53
 // 
 
-using GrammarModule.LanguageRules;
+using DaggerfallWorkshop.Game.Entity;
+using DaggerfallWorkshop.Localization.LanguageRules;
 
-namespace GrammarModule
+namespace DaggerfallWorkshop.Localization
 {
     public abstract class GrammarRules
     {
         public abstract string ProcessGrammar(string text);
-        public abstract void SetGenreHero(string Genre);
-        public abstract void SetGenreNPC(string Genre);
+        public abstract void SetHeroGender(Genders Gender);
+        public abstract void SetNPCGender(Genders Gender);
     }
 
     public static class GrammarManager
--- a/Assets/Scripts/Localization/Grammar.cs
+++ b/Assets/Scripts/Localization/Grammar.cs
@@ -1,0 +1,25 @@
+// Project:         Daggerfall Unity
+// Copyright:       Copyright (C) 2009-2024 Daggerfall Workshop
+// Web Site:        http://www.dfworkshop.net
+// License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
+// Source Code:     https://github.com/Interkarma/daggerfall-unity
+// Original Author: Daneel53
+// 
+
+using GrammarModule.LanguageRules;
+
+namespace GrammarModule
+{
+    public abstract class GrammarRules
+    {
+        public abstract string ProcessGrammar(string text);
+        public abstract void SetGenreHero(string Genre);
+    }
+
+    public static class GrammarManager
+    {
+        public static GrammarRules grammarProcessor = (GrammarRules) new DefaultGrammarRules();
+    }
+}
+
+
--- a/Assets/Scripts/API/Save/CharacterRecord.cs
+++ b/Assets/Scripts/API/Save/CharacterRecord.cs
@@ -18,7 +18,6 @@ using DaggerfallWorkshop.Game.Player;
 using DaggerfallWorkshop.Game.Entity;
 using DaggerfallWorkshop.Game.Formulas;
 using DaggerfallWorkshop;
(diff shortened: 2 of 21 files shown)


Assets/Scripts/Localization/Grammar.cs
```diff
@@ -0,0 +1,26 @@
...
+
+using GrammarModule.LanguageRules;
+
+namespace GrammarModule
+{
+    public abstract class GrammarRules
+    {
+        public abstract string ProcessGrammar(string text);
+        public abstract void SetGenreHero(string Genre);
+        public abstract void SetGenreNPC(string Genre);
```
Jagget, line 17: 1) Maybe something more English-like?
```suggestion
        public abstract void SetHeroGender(string Gender);
        public abstract void SetNPCGender(string Gender);
```
2) Just curious. Why string? Can we use built-in `Genders.Male` and `Genders.Female`?

Jagget (commented):
I'm not sure you covered all possible places where we need Grammar Processor, but I like the idea.

Assets/Scripts/Localization/Grammar.cs
Daneel53: 1 - You're perfectly right! I realize that I almost always use in English  the word "genre" (as in French) instead of "gender", and because the word "genre" exists also in English, the spellcheckers never says anything. Oups! 
So I agree, I will change for SetHeroGender() and SetNPCGender().

2 - Why string? That is because the grammar processor may be used with another game and I don't want to link its code to the code of DFU. There is no "using..." of whatever is DFU definition into the code of the French grammar processor. So I prefer to use the classic and neutral string values "M" and "F" as parameter. Moreover, it is always possible that into another game there may be more than two genders, a kind of "Neutral" or something like that. I saw recently, on the discord of the Bannerlord mod Banner Kings, a player asking to introduce "non-binary" characters with the pronouns "their" and "them". So we must be ready... 

3 - I'm not fully sure either to have covered all places where a call to ProcessGrammar() would be useful, but for the moment I don't see a place where a grammar token is not interpreted. What is coming very soon is the use of the new method SetNPCGender() that I've added two days ago. It shall be used into the quests where we need to know the gender of NPCs as _cousin_ or _healer_. We can handle the last NPC gender when macros as %g2 are interpreted, but it seems also possible with values as _cousin_ or __cousin_ or ___cousin_. I'm working on it, it already works with macros as %gx, should have a definitive answer today or tomorrow.

KABoissonneault: > That is because the grammar processor may be used with another game and I don't want to link its code to the code of DFU

If that's the case, then you probably shouldn't put the code in the DFU code folders. 
The simple place would be under Assets/Scripts/External, but it would be nice if you could find how to place it under Game/Addons, like the CSharpCompiler or the UnityConsole plugins
If you don't do that, I think you should fully respect the DFU namespace and DFU coding convention. But we do already have external modules, so I think it would be fine for you to add yours there

KABoissonneault: If you go that direction, you should change the namespace to DaggerfallWorkshop.Localization rather than GrammarProcessor.
I still think you can just move the file to Assets/Scripts/External, to make it clear this is being contributed as an external module

Daneel53: I already made (and tested) all the changes needed with the new names of the methods and the use of the enum Genders. I just not yet uploaded them to the PR but finally it's a good thing.
I'm ready to keep on changing things so that it is convenient for you. My goal is that this PR is inside DFU 1.2, so I will do what is needed.
If I change the namespace from GrammarModule to DaggerfallWorkshop.localisation, I will have to change the `using GrammarModule` that are in the code where the grammar methods are called. OK, I can do.
If I move grammar.cs and DefaultGrammaRules.cs to Assets/Scripts/External, I suppose that I can keep the namespace GrammarModule.
What is your preference? I think that this is the second one, but tell me, and I will do.

KABoissonneault: Latter. It's gonna be easier to update your scripts as you copy-paste them from your other games if you don't have to port them to DFU standards each time

Daneel53: Sometimes I would like to be able to talk quietly with other people in my language, I would probably be more clear...

My script is not yet included into another game, but my goal is that it can be usable for another game. But when I look carefully to the details, obviously I will have to modifiy it a little in order to be able to insert it properly in each game. The core of the script is the method ProcessGrammar() that will be into FrenchGrammarRules.cs, and this one is enough isolated from any game so that I can reuse it with very little effort in another game. Even with DFU, this script and method will be into the French mod, and not into DFU code, so that's fine.

What I'm inserting today into DFU with this PR is the backbone needed to be able to plug a grammar processor to DFU : the basic abstract class GrammarRules and the GrammarManager into Grammar.cs, the DefaultGrammarRules class into DefaultGrammarRules.cs, and all the calls to ProcessGrammar() where needed into the DFU code so that a real grammar processor can make its job. If there is no grammar processor plugged through a mod, no matter, the game will act as before due to the default grammar processor. But if you plug a specific language processor via a mod, it will do its job without modifying anything to DFU code once this PR is inserted.

In my point of view, this PR add a functionality into the heart of DFU that, I agree, was not existing in the original Daggerfall. But its the same for all function linked to localization that Interkarma added to DFU. That's why I thought that Scripts\Localization was the right place. But if you think that Scripts\External is a better place, no problem. Because no matter for me: this will change nothing to the heart code of my grammar processor that will be enclosed into the French mod, except some `using` and the enum Genders that can be changed easily for another game.

We just need to find an agreement: Grammar.cs and DefaultGrammarRules.cs into Localization or External?
If into Localization, namespace DaggerfallWorkshop.Localization (seems more logical, I agree) or GrammarModule?
And that's it. The rest is on my side. ;)

Daneel53: I've taken the following decisions:
- Grammar.cs and DefaultGrammarRules.cs stay into Localization.
- Namespace changed for DaggerfallWorkshop.Localization for coherency with other scripts into Localization.
- Names of the methods and parameters are now fully in English.
- Parameter for both methods SetxxxGender() is now coming from DFU enum Genders.

The 11 scripts impacted by this update have been pushed into the current PR that is now ready to be approved if nobody have other remarks on it.

The only thing that is still on my desk is the way to find and use properly the gender of the NPCs into the quests so that the localized text is correct ("mon cousin" or "ma cousine" for `_cousin_` for example). If I cannot find a correct way to realize that before 1.2 is out, it will be for 1.3.

KABoissonneault:
Would you be willing to post a sample script of how you would use this in your French localization for testing? Doesn't have to be the full thing if you don't want to, just an example of how you're using this.

Daneel53:
I will post a package with the current 3 "tokenized" French translation files that I'm using into Unity Editor with current state of the DFU Github.... and the scripts modified by this PR, for sure.
I will join a document to explain how it works and what is the result in game with some easy examples.
Just let me the time needed to pack all this.

Daneel53:
Found the right place to get the NPC gender in quests and send it to the grammar processor.
Added a call to ProcessGrammar() so that the NPC gender is used peoperly into the localized text of quests.

The PR is complete for 1.2.

KABoissonneault:
Tell me when you think the PR is stable for merge. At this stage, I'll have to trust you and get this in while you make sure nothing breaks in the default English build

Daneel53:
I cannot guarantee 100% that I (or a future user) won't find one day any other small place where the French grammar tokens are not managed by a call to ProcessGrammar(), but today I think that the PR is complete.
Concerning the English build, testers and moders in Lysandus' Tomb that use 1.2 Test2 and Test3 versions since the one of Carademono in June 2025 play the game with this PR included and nobody never notice it is in, so I'm pretty sure that it is totally harmless for players in English and any other language.

KABoissonneault (approved):

Issue #2672, opened 2024-07-03
PlayerGPS.GetCourtOfCurrentRegion() returns 0 courts in Mournoth


drcarademono:
PlayerGPS.GetCourtOfCurrentRegion() returns 0 courts in Mournoth. I wasn't able to determine why that is. This bug causes Advanced Dialogue to break in Mournoth, because it expects all regions to have a court.

Error msg:
GetCourtOfCurrentRegion() did not find exactly 1 match.
UnityEngine.Debug:LogError (object)
DaggerfallWorkshop.PlayerGPS:GetCourtOfCurrentRegion () (at Assets/Scripts/Internal/PlayerGPS.cs:496)
DaggerfallWorkshop.Game.UserInterface.ADTalkWindow:GetFilterData () (at Assets/Game/Mods/advanced-dialogue/ADTalkWindow.cs:2085)
DaggerfallWorkshop.Game.UserInterface.ADTalkWindow:OnPush () (at Assets/Game/Mods/advanced-dialogue/ADTalkWindow.cs:376)
DaggerfallWorkshop.Game.UserInterface.UserInterfaceManager:AddWindow (DaggerfallWorkshop.Game.UserInterface.IUserInterfaceWindow) (at Assets/Scripts/Game/UserInterface/UserInterfaceManager.cs:182)
DaggerfallWorkshop.Game.UserInterface.UserInterfaceManager:PushWindow (DaggerfallWorkshop.Game.UserInterface.IUserInterfaceWindow) (at Assets/Scripts/Game/UserInterface/UserInterfaceManager.cs:82)
DaggerfallWorkshop.Game.TalkManager:TalkToNpc () (at Assets/Scripts/Game/TalkManager.cs:2634)
DaggerfallWorkshop.Game.TalkManager:TalkToMobileNPC (DaggerfallWorkshop.Game.MobilePersonNPC) (at Assets/Scripts/Game/TalkManager.cs:722)
DaggerfallWorkshop.Game.PlayerActivate:ActivateMobileNPC (UnityEngine.RaycastHit,DaggerfallWorkshop.Game.MobilePersonNPC) (at Assets/Scripts/Game/PlayerActivate.cs:747)
DaggerfallWorkshop.Game.PlayerActivate:Update () (at Assets/Scripts/Game/PlayerActivate.cs:413)

KABoissonneault:
[...]

PR #2673, opened 2024-07-04
Fixed error during GetCourtOfCurrentRegion query in Mournoth

KABoissonneault:
In FACTION.TXT, you can find 44 factions of the "Court" guild type. Out of these, 43 have the social group Nobility - we have only one exception, the Court of Mournoth.

```
    #567
    type: 14
    name: Court of Mournoth
    rep: 0
    summon: -1
    region: 53
    power: 51
    face: -1
    race: 2
    flat: 183 3
    flat: 183 9
>>> sgroup: 2
    ggroup: 15
    flags: 0
    minf: 20
    maxf: 80
```

As documented at the top of FACTION.TXT, 
```
; The field sgroup should be set to one of the following values:
; COMMONER	0
; MERCHANT	1
; SCHOLAR	2
; NOBILITY	3
; UNDERWORLD	4
```
sgroup of 2 is Scholar.

In DFU's search for the current region's Court, it specified that the faction it's looking *has* to be Nobility. This is an unnecessary restriction, and causes the Scholarly Court of Mournoth to fail to return here, causing an exception to be thrown. 

This PR simply allows all social groups to be considered a court, as long as it's of the regional guild group and the court faction type

### DIFF
--- a/Assets/Scripts/Internal/PlayerGPS.cs
+++ b/Assets/Scripts/Internal/PlayerGPS.cs
@@ -471,7 +471,7 @@ namespace DaggerfallWorkshop
             // Find court in current region
             FactionFile.FactionData[] factions = GameManager.Instance.PlayerEntity.FactionData.FindFactions(
                 (int)FactionFile.FactionTypes.Courts,
-                (int)FactionFile.SocialGroups.Nobility,
+                -1,
                 (int)FactionFile.GuildGroups.Region,
                 CurrentRegionIndex);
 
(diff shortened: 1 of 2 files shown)


KABoissonneault:
Fixes issue #2672

df-ferital:
Hi @KABoissonneault, hope you're well. I've been away from DF Unity for a while, hopefully I'll be able to contribute again soon.

I wanted to add that while your patch is fine, I would have also fixed FACTION.TXT, as the court factions are indeed supposed to belong to the Nobility social group. This will have an impact in the game when talking to members of this faction, such as the NPCs in town palaces.

KABoissonneault:
Good point. I had assumed without looking that this file was part of DF data, but no, we do ship our own, which we have patched in the past. Gonna keep the current fix for mods, and fix FACTION.TXT as you suggest

Jagget (approved):

PR #2674, opened 2024-07-05
Reverse order of merchants selling quotes

petchema:
Buying and selling dialogs are in reverse order of "success" in the resources, fix how offset is computed.

Light testing with a character with low mercantile and personality skills, messages when selling stuff went from "I'm losing my shirt" to "take it or leave it", which felt more in line. More testing welcome. 

Forums: https://forums.dfworkshop.net/viewtopic.php?t=6723

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTradeWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTradeWindow.cs
@@ -1116,7 +1116,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
                         msgOffset = 1;
                 }
                 if (WindowMode == WindowModes.Sell || WindowMode == WindowModes.SellMagic)
-                    msgOffset += 3;
+                    msgOffset = 5 - msgOffset;
 
                 DaggerfallMessageBox messageBox = new DaggerfallMessageBox(uiManager, this);
                 TextFile.Token[] tokens = DaggerfallUnity.Instance.TextProvider.GetRandomTokens(TradeMessageBaseId + msgOffset);


ajrb:
I've been looking at this for ages from the messages POV and it looks correct as is to me. TBH I am notsure I really understand the messages and how they relate to merch skill etc.

To me both 260-262 (buy) and 263-265 (sell) are in success->failure for both ranges. I only looked at the first of the randomised variant messages however.

KABoissonneault:
I think it's best to edit a classic DF save and test there, see what happens. I'm usually fairly confident that this has been reverse engineered and matches what DF does, but mistakes can happen, so it's worth checking. I'll decide whether a DFU fix is worth it after that

petchema:
Following @KABoissonneault suggestion, I attempted selling a Cloth Amulet of Shocking in The King's Hockshop of Daggefall,
Very bad personality and mercantile (1):
"Enough of this endless dickering. You are obviously in great need. Normally I would pay far less, but for you I will offer 220 gold."
"Listen, I will offer you the absolute maximum I can without losing my shirt. Here is 220 gold. Take it and tell all your friends you cheated me."
"Not bad, not bad. I'd give most people less, but for you, 220 gold. That's my absolute best offer.
"I'll give the price I only offer to my closest friends and family. The market value of that is much less, but I think I can give you 220 gold."

Very high personality and mercantile (100):
"Maybe you can get more than 411 gold for that somewhere, but it's a fair price. And it's my only offer."
"I can give you 411 gold pieces. If someone else is offering more, I'd advise you go to him. 411 gold is all it's worth to me."
"Yeah, you're right. Good quality all around. I'll offer a little more than the standard price -- 411 gold pieces"
"There's not a big market for those, but I can sell it. I'll give you 411 gold pieces, a perfectly fair price."

So it seems the current order of messages in Daggerfall Unity is correct, good call @KABoissonneault!

PR #2675, opened 2024-07-06
Use magic items at range while silenced

petchema:
In classic you can use all magic items while silenced, but in Daggerfall Unity you couldn't use magic items at range, because while silence lets you ready magic items, it does not let you release their spells.
Most likely an overlook.
    
Also, SilenceCheck() can do side effets (display "You are silenced" and reset readied spell), so when used with short-circuited logic operators, order matters and it should be placed late/last.
    
Forums: https://forums.dfworkshop.net/viewtopic.php?t=6725

### DIFF
--- a/Assets/Scripts/Game/MagicAndEffects/EntityEffectManager.cs
+++ b/Assets/Scripts/Game/MagicAndEffects/EntityEffectManager.cs
@@ -312,7 +312,7 @@ namespace DaggerfallWorkshop.Game.MagicAndEffects
         public bool SetReadySpell(EntityEffectBundle spell, bool noSpellPointCost = false)
         {
             // Do nothing if silenced or cast already in progress
-            if ((SilenceCheck() && !noSpellPointCost) || castInProgress)
+            if ((!noSpellPointCost && SilenceCheck()) || castInProgress)
                 return false;
 
             // Spell must appear valid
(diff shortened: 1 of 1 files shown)


numidium (approved):

KABoissonneault (commented):
I'm not too sure about equating "no spell cost" with "can be used while silenced". 
"No spell cost" is used in the following situations:
1) player magical items
2) enemy attack abilities (spider/scorpion poison)
3) player lycanthropy spell
4) the "Cast Spell" RDB action for "caster target" spells

I'm not sure where or if the last one is used honestly. But it does seem to be related to player spellcasting, so it might be relevant to consider. If it's used for main quests, then it might be essential to prevent silence from breaking them. But if it's used for like, a magic artifact that gives you a buff... then it might be fine to let it work through silence too.
0
I'm fine with the fix for case 1 and 3, but I think 2 is debatable. It makes sense logically that natural poisons should still work while silenced, but maybe players like the ability to use Silence to beat these. I'd go with classic for sure if it doesn't stop spider/scorpion poison, but I'd like to hear opinions from others if it does.

petchema:
Ah, enemies bypass the "ready" stage

It seems silenced scorpions can still paralyze me in classic?
[SAVE4.zip](https://github.com/user-attachments/files/16571220/SAVE4.zip)

KABoissonneault (approved):

Issue #2660, opened 2024-05-31
Breton's magicka resist does not show in history.


fartingporkchop:
**Describe the issue**

Creating a character, Breton's magic resistance does not show in class history summary.  A Nord's frost resistance, and High Elf's immune to paralysis does show when checking history.  Bretons might not have magic resistance like they should?

**To Reproduce**
Create a Breton character fully, then check in-game.  Magic resistance does not show.
Create a Nord character fully, then check in-game.  Frost resistance shows in history.
Same with High Elf, Paralysis immunity shows in history.

Save file doesn't show magic resistance either as a flag, which it does for Nords and High Elves abilities.  Is the Breton's resistance hard coded elsewhere or not active?

**Expected behavior**

Breton's magic resistance should behave like the Nord's and High Elf's resistance.  Otherwise, it should be like in DosBox where these resistances do not show up in the history for all races.

**Screenshots and Logs**

n/a

**Desktop (please complete the following information):**

 - OS: Windows
 - Version 1.1.1

**Additional context**

No mods.

KABoissonneault:
Seems like the Racial part was added by @numidium in https://github.com/Interkarma/daggerfall-unity/commit/42eed869702b8277cade12488780873ec6ece3cd

From the message, the intent was to show the racial bonuses from vampirism and lycanthropy, but it seems the Nord and High Elf data also have these flags, and it gets picked up by this code.

Because the race flags are not used by `FormulaHelper.SavingThrow`, I'm tempted to just remove the elemental resistance lines entirely. It's not useful info with the base game data. The Racial Special Ability lines can be kept, since those are actually relevant to vampirism/lycanthropy

We could potentially make `FormulaHelper.SavingThrow` use the flags, except that won't cover the Breton, which does not have the flag in the DF data.

![image](https://github.com/Interkarma/daggerfall-unity/assets/5789925/e6064eaf-1ed5-4607-acb3-31efc197e4c2)

numidium:
> From the message, the intent was to show the racial bonuses from vampirism and lycanthropy...

Sorry, I could have worded that better. If I remember correctly, at the time I wanted to cover all the racial bonuses and I put the text there to say that it includes vampirism/lycanthropy _in addition to_ all the base races.

[...]

PR #2676, opened 2024-07-07
Refactor FormulaHelper to read racial resistances

numidium:
FormulaHelper previously applied racial resistances by reading the player's race directly. With this change, the saving throw calculation will be modified by the contents of the player's RaceTemplate.

Addresses the following issue:
https://github.com/Interkarma/daggerfall-unity/issues/2660

### DIFF
--- a/Assets/Scripts/Game/Formulas/FormulaHelper.cs
+++ b/Assets/Scripts/Game/Formulas/FormulaHelper.cs
@@ -1514,7 +1514,7 @@ namespace DaggerfallWorkshop.Game.Formulas
             var raceTemplate = playerEntity.GetLiveRaceTemplate();
             if ((elementType == DFCareer.Elements.Fire && (raceTemplate.ResistanceFlags & DFCareer.EffectFlags.Fire) != 0) ||
                 (elementType == DFCareer.Elements.Frost && (raceTemplate.ResistanceFlags & DFCareer.EffectFlags.Frost) != 0) ||
-                (elementType == DFCareer.Elements.DiseaseOrPoison && (raceTemplate.ResistanceFlags & (DFCareer.EffectFlags.Disease | DFCareer.EffectFlags.Poison)) != 0) ||
+                (elementType == DFCareer.Elements.DiseaseOrPoison && (raceTemplate.ResistanceFlags & effectFlags & (DFCareer.EffectFlags.Disease | DFCareer.EffectFlags.Poison)) != 0) ||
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Game/Formulas/FormulaHelper.cs
```diff
@@ -1509,9 +1509,14 @@ public static int SavingThrow(DFCareer.Elements elementType, DFCareer.EffectFlag
                 savingThrow += 25;
 
             savingThrow += biographyMod + modifier;
-            if (elementType == DFCareer.Elements.Frost && target == playerEntity && playerEntity.Race == Races.Nord)
-                savingThrow += 30;
-            else if (elementType == DFCareer.Elements.Magic && target == playerEntity && playerEntity.Race == Races.Breton)
+
+            // Apply racial resistances
+            var raceTemplate = playerEntity.GetLiveRaceTemplate();
```
KABoissonneault, line 1514: Seems like the check for whether the target is the player entity is missing.

numidium: Yep, fixed that now.

petchema (approved):

ajrb (dismissed):

KABoissonneault:
I'm not convinced with the changes in this PR. It feels incomplete.
1. High Elf paralysis resistance is still hardcoded rather than going through immunity flags
2. Only racial Resistance is checked, not Immunity, Low Tolerance, or Critical Weakness

I guess the main point is really missing Breton flag for Resistance, so maybe I should let this PR pass. It just feels like a wasted opportunity if we just do the data-driven formula halfway

KABoissonneault (changes requested):
Numidium told me he would rework this PR for completeness

numidium:
Added a new commit that uses RaceTemplate for all saving throw-related flags. See commit description for details.

KABoissonneault:
Seems mostly good. 

1. I think a race that has resistance for both paralysis and the element should apply the effect twice. Not relevant with any of the vanilla race, so I'll keep that as a later improvement.
2. The saving throw for High Elves never goes through. The EntityEffectManager blocks the spell immediately with `IsEntityImmuneToParalysis`. It does not make this PR wrong, just kind of pointless. Perhaps we should remove the race template check in the EntityEffectManager, and let the SavingThrow formula handle it

I'm gonna merge this in the meantime, no critical issues

KABoissonneault (approved):

PR #2678, opened 2024-07-20
Receice quests from MG and Knights based on either rank or level

petchema:
In classic Daggerfall, Mages guild and knightly orders can provide quests not only if the player's rank is sufficient, but also if his/her level is sufficient. This PR implements this behavior.

This uses a bit in FactionData's flags, so I think it could be set by mods too.

I provided stubs for mods backward compatibility, even if I don't know if any mod overrides QuestListManager.GetGuildQuest() or QuestListManager.GetGuildQuestPool()

Remark: FACTION.TXT is parsed then cached in FactionData.txt in saves, so you may have to edit or remove FactionData.txt to get updated faction data?

Forums:
https://forums.dfworkshop.net/viewtopic.php?p=68013#p68013
https://forums.dfworkshop.net/viewtopic.php?p=43591#p43591

### DIFF
--- a/Assets/Scripts/API/FactionFile.cs
+++ b/Assets/Scripts/API/FactionFile.cs
@@ -630,7 +630,7 @@ namespace DaggerfallConnect.Arena2
         {
             RulerImmune = 0x10,
             Summoned = 0x40,
-            questByRankOrLevel = 0x100, // Not in classic
+            QuestByRankOrLevel = 0x100, // Not in classic
         }
 
         #endregion
--- a/Assets/Scripts/API/FactionFile.cs
+++ b/Assets/Scripts/API/FactionFile.cs
@@ -630,6 +630,7 @@ namespace DaggerfallConnect.Arena2
         {
             RulerImmune = 0x10,
             Summoned = 0x40,
+            questByRankOrLevel = 0x100, // Not in classic
         }
 
         #endregion
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallWitchesCovenPopupWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallWitchesCovenPopupWindow.cs
@@ -131,10 +131,11 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             // Get the faction id for affecting reputation on success/failure, and current rep
             int factionId = witchNPC.Data.factionID;
             int reputation = GameManager.Instance.PlayerEntity.FactionData.GetReputation(factionId);
+            int rank = 0;
             int level = GameManager.Instance.PlayerEntity.Level;     // Not a proper guild so rank = player level
 
             // Select a quest at random from appropriate pool
-            offeredQuest = GameManager.Instance.QuestListsManager.GetGuildQuest(FactionFile.GuildGroups.Witches, MembershipStatus.Nonmember, factionId, reputation, level);
+            offeredQuest = GameManager.Instance.QuestListsManager.GetGuildQuest(FactionFile.GuildGroups.Witches, MembershipStatus.Nonmember, factionId, reputation, rank, level);
             if (offeredQuest != null)
             {
                 // Log offered quest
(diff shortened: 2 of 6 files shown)


Assets/Scripts/Game/UserInterfaceWindows/DaggerfallWitchesCovenPopupWindow.cs
```diff
@@ -131,10 +131,11 @@ protected override void GetQuest()
             // Get the faction id for affecting reputation on success/failure, and current rep
             int factionId = witchNPC.Data.factionID;
             int reputation = GameManager.Instance.PlayerEntity.FactionData.GetReputation(factionId);
+            int rank = 0;
             int level = GameManager.Instance.PlayerEntity.Level;     // Not a proper guild so rank = player level
```
petchema, line 135: Just noticed that I broke the witches coven, that so far intentionally sent the level as rank parameter.
Either we keep doing that, or add the questByRankOrLevel to them so they enter the PR framework

Assets/Scripts/API/FactionFile.cs
```diff
@@ -630,6 +630,7 @@ public enum Flags
         {
             RulerImmune = 0x10,
             Summoned = 0x40,
+            questByRankOrLevel = 0x100, // Not in classic
```
KABoissonneault, line 633: Should start with a capital letter

Assets/StreamingAssets/Factions/FACTION.TXT
```diff
@@ -649,7 +649,7 @@ maxf: 90
 		summon: -1
 		region: -1
 		power: 25
-		flags: 0
+		flags: 256
```
KABoissonneault, line 652: Not sure it was necessary to cover the religious knights, but I guess it doesn't hurt

KABoissonneault (changes requested):

KABoissonneault:
Change seems good. Was the Coven and Vampire clans from your testing? I can't find info on whether their quests accept level as a rank too, but it also seems under documented

KABoissonneault (changes requested):
I'm not comfortable with the current "existing save" issue. I feel like if we ship a fix saying quest distribution now works like classic, we're constantly going to get questions from people saying it doesn't work for them.

I think a contingency plan would be to add a new `FactionData_v3` structure (same data as v2), and if a save contains v2 data, add 256 to the flag of the concerned factions. If a save already has v3 data, don't do anything (since mods might want to remove the flags).
Maybe we could also just add a bool to `FactionData_v2`, that's always set to true on save? It would default to false if a save did not have the bool from a previous version, and we can then detect on load if a migration is needed. There's similar code in `SerializablePlayer.RestoreFactionData`

ajrb:
Hi all, I only found out about this a couple days ago when github emailed me. This was something that I recall discussions about how quests were allocated back in the day but no one ever found that mages & knightly orders behaved differently - I think the assumption was that they were consistent with other guilds who's behaviour was checked - likely the Fighters from my side as that's where I started.

Pechema, I do wish you'd contacted me about this right from the start. I'd have happily worked on a fix for this. It looks like you've used a faction data flag or something for this, but the design in DFU was to not drive the quest offer system off of the flawed Daggerfall faction data - hence the quest list stuff etc. I hate to be that guy after the last month of effort on this PR, but I would like the opportunity to submit an alternative approach that's more aligned with the re-implementation I did for DFU if that would be okay? This will also avoid the faction data change issue for existing saves.

From what I've read here, the change required is just MG and Knight Orders allow quests by rep (as already implemented) or by player level, unless I'm missing something subtle here?

KABoissonneault:
Allegedly, the fix should cover vampire clans and witch covens too. I'm not sure where this has been verified

ajrb:
That would complicate things a little since they're not actually proper 'guilds', can anyone verify thatin vanilla? I don't have any saves that would enable doing that quickly.

petchema:
@ajrb This is an idea I experimented with after seeing a streamer (Grimith I guess) being offered by classic DF two rank 7 knight quests right after joining the order, so there was an obvious discrepancy with what DFU currently does.

 I have no issue with another implementation replacing this one at all, that's something I tried for myself on a subsystem I don't know well, so I'm sure I did plenty of things wrong ;)

ajrb:
@petchema Thanks, will submit my alternate proposed solution soon. I noticed that UESP now correctly states this for MG & Knights pages. I'm fairly sure that wasn't the case back in March 2018 when I wrote the quest list code originally, but maybe I was just not dilligent enough back then. :)

Did you have some evidence for this behaviour also being used for Vampire Clan and Witch Covern quest dispensing that KAB mentioned? 
It doesn't make sense to me for Vampire quest as they are dispensed by delivered letters once the player has joined the vamps, rank or level is not relevant. 
For witches coverns, UESP states reputation is how quests are selected. It would be easy enough to change if someone does provide evidence from classic.

petchema:
Obsoleted by #2688

PR #2679, opened 2024-07-24
Control-U during character class creation resets bonus pool

petchema:
Classic Daggerfall uses plain U key, but since in Daggerfall Unity the class name input field always has focus, that key would add an U character to the class name instead.

Remark: Under Linux, keyboard handling behaves a bit differently in the Unity Editor and in the Standalone Player: standalone player filters out control characters, while the editor does not, adding a "?" in the class name when Control-U is pressed. They could be filtered explicitly:

```diff
diff --git a/Assets/Scripts/Game/UserInterface/TextBox.cs b/Assets/Scripts/Game/UserInterface/TextBox.cs
index 9da458cef..6d660339f 100644
--- a/Assets/Scripts/Game/UserInterface/TextBox.cs +++ b/Assets/Scripts/Game/UserInterface/TextBox.cs @@ -453,6 +453,14 @@ namespace DaggerfallWorkshop.Game.UserInterface
                         }
                     }
                 }
+                else // !numeric
+                {
+                    // Filter control characters
+                    if ((int)character < 0x20)
+                    {
+                        return;
+                    }
+                }

                 // For upper only, force everything to caps
                 if (upperOnly)
```
but it seems less risky to just be aware of this small problem happening in the Editor only.
(just tested that the problem doesn't occur under Windows).

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/CreateCharCustomClass.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/CreateCharCustomClass.cs
@@ -100,6 +100,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
         Rect specialAdvantageButtonRect = new Rect(249, 98, 66, 22);
         Rect specialDisadvantageButtonRect = new Rect(249, 122, 66, 22);
         Rect reputationButtonRect = new Rect(249, 146, 66, 22);
+        Rect resetButtonRect = new Rect(0, 0, 0,0);
         Rect exitButtonRect = new Rect(263, 172, 38, 21);
 
         #endregion
@@ -113,6 +114,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
         Button specialAdvantageButton;
         Button specialDisadvantageButton;
         Button reputationButton;
+        Button resetButton;
         Button exitButton;
 
         #endregion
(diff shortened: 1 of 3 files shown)


Assets/Scripts/Game/UserInterfaceWindows/CreateCharCustomClass.cs
```diff
@@ -394,6 +401,15 @@ void ReputationButton_OnMouseClick(BaseScreenComponent sender, Vector2 position)
             uiManager.PushWindow(createCharReputationWindow);
         }
 
+        protected virtual void ResetButton_OnKeyboardEvent(BaseScreenComponent sender, Event keyboardEvent)
+        {
+            if (keyboardEvent.type == EventType.KeyDown)
+            {
+                DaggerfallUI.Instance.PlayOneShot(SoundClips.ButtonClick);
+                statsRollout.BonusPool = 0;
```
KABoissonneault, line 409: It seems to be like this should also reset all stats to 50, rather than leave the current values. As it is, it's more of a cheat code to get all stats to 75 for free.

image
image

KABoissonneault (changes requested):

Assets/Scripts/Game/UserInterfaceWindows/CreateCharCustomClass.cs
petchema: No, it's not a "reset" shortcut.
It operates exactly like in classic, this allows to create unbalanced classes, for example Grimith used it (by running character creation under classic then switching to DFU), in conjunction with LevelUp Adjuster mod, to create a character with very low starting abilities.
Daggerfall is a single player game, while balancing is the standard rule for creating characters, it doesn't need to be a hard rule.

KABoissonneault (approved):

petchema:
Rereading the PR title, now I understand how confusing it was ;)
Anyway, thanks for merge!

PR #2680, opened 2024-07-30
Fix spell absorption capping

petchema:
When absorbing magicka from own spells, DFU explicitly ensures that magicka recovered is not larger than the magicka that was used. For this, it needs to track how much magicka was spent for the last spell: lastReadySpellCastingCost

A special case was not correctly handled though: when the last spell came from a magic item, lastReadySpellCastingCost should be set 0 so that next absorption of own spells is not capped.
That worked by luck when the game was just started because lastReadySpellCastingCost was still 0, but broke after first spell casting.

Forums: https://forums.dfworkshop.net/viewtopic.php?t=6755

### DIFF
--- a/Assets/Scripts/Game/MagicAndEffects/EntityEffectManager.cs
+++ b/Assets/Scripts/Game/MagicAndEffects/EntityEffectManager.cs
@@ -2131,6 +2131,10 @@ namespace DaggerfallWorkshop.Game.MagicAndEffects
                 lastSpell = readySpell;
                 lastReadySpellCastingCost = readySpellCastingCost;
             }
+            else
+            {
+                lastReadySpellCastingCost = 0;
+            }
             readySpell = null;
             readySpellCastingCost = 0;
             instantCast = false;
--- a/Assets/Scripts/Game/MagicAndEffects/EntityEffectManager.cs
+++ b/Assets/Scripts/Game/MagicAndEffects/EntityEffectManager.cs
@@ -2126,14 +2126,14 @@ namespace DaggerfallWorkshop.Game.MagicAndEffects
 
             // Clear ready spell and reset casting - do not update last spell if casting from item
             RaiseOnCastReadySpell(readySpell);
-            if (!readySpellDoesNotCostSpellPoints)
+            if (readySpellDoesNotCostSpellPoints)
             {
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Game/MagicAndEffects/EntityEffectManager.cs
```diff
@@ -2131,6 +2131,10 @@ private void PlayerSpellCasting_OnReleaseFrame()
                 lastSpell = readySpell;
                 lastReadySpellCastingCost = readySpellCastingCost;
             }
+            else
+            {
+                lastReadySpellCastingCost = 0;
+            }
```
Jagget, line 2137: Can we flip this **if/else** so it would check the positive condition? It's just a personal preference, feel free to ignore this comment

```cs
            if (readySpellDoesNotCostSpellPoints)
            {
                lastReadySpellCastingCost = 0;
            }
            else
            {
                lastSpell = readySpell;
                lastReadySpellCastingCost = readySpellCastingCost;
            }
```

petchema: Done, but that barely improves the code readability, I also considered handling lastReadySpellCastingCost separately from lastSpell because I don't really consider their lifecycles to be related...
```
if (!readySpellDoesNotCostSpellPoints)
{
    lastSpell = readySpell;
}
lastReadySpellCastingCost = readySpellDoesNotCostSpellPoints ? 0 : readySpellCastingCost;
```

I wonder if that code could have used less state, and less separate scalar variables to represent that state...
But submitted the smallest change that fixes the bug

petchema:
Test protocol:
- notice current  spellpoints
- use the magic item to cast fireball at the facing wall
- check that spellpoints increased by 62, cost of the fireball spell for the character
- cast slowfall from the spellbook
- notice current spellpoints
- use the magic item to cast fireball at the facing wall
- check that spellpoints increased by less than 62 (actually increased by the cost of slowfall)

(sorry for the traces of french translation mod in the save)
[SAVE37.zip](https://github.com/user-attachments/files/16462383/SAVE37.zip)

Jagget (approved):

KABoissonneault (approved):

PR #2681, opened 2024-08-03
Fix enemies sometimes hearing thru walls

petchema:
Previous code checked that the first obstacle between an enemy and its target was an object tagged "static" to determine if the enemy could hear the target (followed by a distance check).
In some cases that allowed enemies to hear targets when they shouldn't (other static tagged objects along the "line of hearing" not checked), leading to enemies facing walls or still chasing enemies thru walls when you encounter them.

This PR first checks for distance to target (like CanSeeTarget() does), then checks if they're no object tagged static inbetween the enemy and its target.
As an optimization, only objects on layer "Default" are considered, and Physics.RaycastNonAlloc() API is used to lower memory allocations.

Forums: https://forums.dfworkshop.net/viewtopic.php?t=6759

### DIFF
--- a/Assets/Scripts/Game/EnemySenses.cs
+++ b/Assets/Scripts/Game/EnemySenses.cs
@@ -923,24 +923,30 @@ namespace DaggerfallWorkshop.Game
             return seen;
         }
 
+        readonly int defaultLayerOnlyMask = 1 << LayerMask.NameToLayer("Default");
+
         bool CanHearTarget()
         {
             float hearingScale = 1f;
 
-            // If something is between enemy and target then return false (was reduce hearingScale by half), to minimize
-            // enemies walking against walls.
-            // Hearing is not impeded by doors or other non-static objects
-            RaycastHit hit;
-            Ray ray = new Ray(transform.position, directionToTarget);
-            if (Physics.Raycast(ray, out hit))
+            // TODO: Modify this by how much noise the target is making
+            if (distanceToTarget < (HearingRadius * hearingScale) + mobile.Enemy.HearingModifier)
             {
-                //DaggerfallEntityBehaviour entity = hit.transform.gameObject.GetComponent<DaggerfallEntityBehaviour>();
-                if (GameObjectHelper.IsStaticGeometry(hit.transform.gameObject))
-                    return false;
+                // If something is between enemy and target then return false (was reduce hearingScale by half), to minimize
+                // enemies walking against walls.
+                // Hearing is not impeded by doors or other non-static objects
+                Ray ray = new Ray(transform.position, directionToTarget);
+                RaycastHit[] hits = Physics.RaycastAll(ray, distanceToTarget); // , defaultLayerOnlyMask 
+                foreach (RaycastHit hit in hits)
+                {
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Game/EnemySenses.cs
```diff
@@ -923,24 +923,41 @@ bool CanSeeTarget(DaggerfallEntityBehaviour target)
...
-                //DaggerfallEntityBehaviour entity = hit.transform.gameObject.GetComponent<DaggerfallEntityBehaviour>();
-                if (GameObjectHelper.IsStaticGeometry(hit.transform.gameObject))
-                    return false;
+                // If something is between enemy and target then return false (was reduce hearingScale by half), to minimize
+                // enemies walking against walls.
+                // Hearing is not impeded by doors or other non-static objects
+                Ray ray = new Ray(transform.position, directionToTarget);
+                int nhits;
+                while (true) {
+                    nhits = Physics.RaycastNonAlloc(ray, hitsBuffer, distanceToTarget, defaultLayerOnlyMask);
```
KABoissonneault, line 945: I have issues with this approach. 

We use a fixed buffer for the multi-cast (cast with multiple results). And we're aware that the results may be partial due to buffer size. If we manage to fill the whole buffer, we retry with a larger buffer, in case we missed results.

A few things jump out to me
1) Initial buffer size is 4, which seems small
2) Buffer size increases by 1 every iteration

If a situation occurs where we hit 1000 objects at once (when entering a dungeon for the first time), we have to go through this loop up to 996 times.
I'd be much easier with an initial size of 8 (seems like it leaves room for 3-4 static models and a few entities), and a growth strategy of x2 every iteration. This would greatly limit the potential hitches due to a sudden large growth, while keeping the benefits of amortizing to "no allocation" eventually.

KABoissonneault: Admittedly, I could not hit a case of this many objects. Mostly because the hearing radius is so small. But I'd still request the x2 growth, just in case a mod goes crazy with hearing distances and object density

petchema: Yes, max distance is limited, but I can double array size to get O(1) amortized allocation, no problem.

By the way, there's another reason why enemies often face walls, chasing one another thru walls: stealthiness, or the lack of thereof, is implemented by creatures having some probability to periodically "leak" their current position. So it's just Daggerfall mechanics at play (exactly what [DunnyOfPenwick was saying](https://forums.dfworkshop.net/viewtopic.php?p=68078#p68078)).
So this patch will only fix the bug of non-static first obstacle allowing to see thru walls behind them.

KABoissonneault (approved):

petchema:
As it occurred to me after the fact, a more efficient way to solve this issue would be to dedicate a layer to static objects, so a simple raycast with a layer mask would be sufficient, instead of retrieving hits and checking individual tags...

KABoissonneault:
Yeah probably. I'm just not sure about the risk, when it comes to modded dungeon blocks. I'm not sure how much we can expect existing mods that have objects with the static tag but default layer, expecting them to block sound

PR #2683, opened 2024-08-06
Always display negociated prices

petchema:
When trying to buy from a shop and not having enough gold to pay the negociated price,

In classic Daggerfall:
- displays the same dialog box as usual, with merchant negociated price message, and Yes/No buttons;
- if you click on No, dialog box just closes;
- if you click on Yes, a message box "You don't have enough gold" is displayed

In Daggerfall Unity:
- you get the message box "You don't have enough gold"

So you never learn the negociated price, and how far you are from being able to afford buying what you are interested in.

This PR replaces the "You don't have enough gold" message with the merchant negociated price message.


petchema:
not enough gold, improved

petchema:
Why choose? Keeping the old message to be as obvious as possible
not enough gold, further improved

petchema:
> Can we make it also centred? I feel like it'll look better that way

Done, it's consistent with other messages mixing quotes and non-quotes (say, totem telepathic message)

KABoissonneault (approved):

Issue #2684, opened 2024-08-11
Location list lines after search on travel map are limited to 29 characters


Daneel53:
After having entered a search clue on the travel map, the resulting list has lines cut at 29 characters.
This is probably an issue only for other languages than English that are more verbose.

![DaggerfallUnity_2024-08-11_09-03-34](https://github.com/user-attachments/assets/aeb0dfb0-23a3-4105-90fc-a1cd25f52d13)

This comes from the script DaggerfallTravelMapWindow.cs where the listbox is explicitely created with MaxCharacters = 29.

![notepad++_2024-08-11_10-51-48](https://github.com/user-attachments/assets/f35af90a-6e35-4ba8-b9bd-758ee57ca66c)

I changed the 29 for 50 with the Unity editor and the list is is now correctly displayed.

![Unity_2024-08-10_22-13-49](https://github.com/user-attachments/assets/5f197539-9ca6-448b-b365-348aba21ebd7)

Please change this value in DFU code.

petchema:
I assume this is done for compatibility with SDF fonts off:
![Capture d’écran du 2024-08-11 13-21-14](https://github.com/user-attachments/assets/3aa80ab7-3798-4a46-94b2-ae401cd599c6)
So it will have to depend on current font mode (or better, if this can be implemented, restrict text drawing to scrollable rectangle, which would work for all fonts, specially variable-spaced ones)

Daneel53:
I suppose that 29 characters always were enough for the English names, whatever the font is. So, for English, no matter if the line length is pushed to 50, additional chars shall not be used.
With other languages, we just cannot play with SDF font because this font don't have any character with accent, which make it totally not usable as your example above with esoteric characters that replace the é and the â shows clearly. So, in other languages than English, SDF font is not to be taken into account.

PR #2685, opened 2024-08-12
Allow for longer names in location picker

petchema:
Answer to request #2684

- set limit to 50 characters instead of 29 when SDF fonts (that are narrower) are enabled;
- make that limit more accessible to modding (did I get that part right? I'm not familiar with that aspect)

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallBankingWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallBankingWindow.cs
@@ -88,7 +88,7 @@ namespace DaggerfallWorkshop.Game.UserInterface
             inventoryAmount.Position        = new Vector2(156, 24);
             inventoryAmount.Size            = new Vector2(64, 13);
             inventoryAmount.Name            = "inv_total_label";
-            inventoryAmount.MaxCharacters   = 11;
+            inventoryAmount.MaxCharacters   = 20;
             mainPanel.Components.Add(inventoryAmount);
 
             loanAmountDue               = new TextLabel();
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallSpellBookWindow.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallSpellBookWindow.cs
@@ -347,7 +347,8 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
             spellsListBox.Position = new Vector2(spellsListBoxRect.x, spellsListBoxRect.y);
             spellsListBox.Size = new Vector2(spellsListBoxRect.width, spellsListBoxRect.height);
             spellsListBox.RowsDisplayed = 16;
-            spellsListBox.MaxCharacters = 22;
+            spellsListBox.RectRestrictedRenderArea = new Rect(spellsListBox.Position, spellsListBox.Size);
+            spellsListBox.RestrictedRenderAreaCoordinateType = BaseScreenComponent.RestrictedRenderArea_CoordinateType.ParentCoordinates;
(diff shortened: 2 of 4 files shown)


KABoissonneault:
I'm not sure public is the right accessibility yet. Usually, we would make this protected, so that a mod providing a window override can use a custom value.
Public could be right if there was a way for mods to register to an event that would let it change the value after `Setup()` (so that its value wins over the default) but before `ShowLocationPicker` (so it actually gets used). This way, mods could customize this behavior without overriding the window. I haven't looked if we have any such event

KABoissonneault:
I'm not sure I like the explicit check for SDF vs non-SDF. Is that what we do for books and other auto-formatted text boxes? I know quest text is manually formatted to support non-SDF, but I think books are handled in code

petchema:
> I'm not sure public is the right accessibility yet. Usually, we would make this protected, so that a mod providing a window override can use a custom value. Public could be right if there was a way for mods to register to an event that would let it change the value after `Setup()` (so that its value wins over the default) but before `ShowLocationPicker` (so it actually gets used). This way, mods could customize this behavior without overriding the window. I haven't looked if we have any such event

Okay, if simplest/most convenient way is subclassing, making it a `protected virtual` property

petchema:
Also, the same issue happens with the list of spells (spell merchant, spellbook)

petchema:
> but somehow it didn't work (text not clipped inside the ListBox. I'm not sure what I'm missing

My current hypothesis is that ShowLocationPicker is calling AddItem() of the ListBox before its parent DaggerfallListPickerWindow has been SetUp(), which is a problem because it's AddItem() that propagates the RestrictedArea of the ListBox to its elements

petchema:
My analysis was correct, and simplest fix I could think of was to subclass DaggerfallListPickerWindow to delay calls to listBox.AddItem(s) to the end of Setup()
I end up fixing locations list, spell mechant/spell book lists, bookshelves list

petchema:
sdf locations list after
nosdf locations list after

petchema:
books list before
books list after

petchema:
spells list before
spells list after

petchema:
Could be factorized further if we decide DaggerfallListPickerWindow should always use restricted area, or would benefit from a constructor with a List<string> to initialize its content; I just tried to keep changes as local as possible.

petchema:
While I'm fixing form issues, fixing banking inventory amount whose length was not increased when wagon content display was added

KABoissonneault:
Can anyone point me to where I can get the French localization installed, so I can try the width changes?

petchema:
> Can anyone point me to where I can get the French localization installed, so I can try the width changes?

Last binary release (1.1.1a) is on NexusMods: https://www.nexusmods.com/daggerfallunity/mods/456

Sources are on GitHub, to get the development version you can just copy the content of this directory in StreamingAssets: https://github.com/Daneel53/DFU-en-francais/tree/main/French%20files%20for%20DFU

KABoissonneault (approved):
Not exactly a fan of the duplicated derived class. From what I understand, the ListBox of the picker does not set its own Position and Size until you push the window, so you cannot set the `.RectRestrictedRenderArea = new Rect(listBox.Position, listBox.Size)` before adding the elements... I feel like they could be a way to generalize this into a `DaggerfallFixedListPickerWindow`, that sets the listbox size and restricted render area on construction rather than on Setup

Gonna approve this anyway. I'll let this on hold for some time in case you want to make that new class. Otherwise, we'll stick with that

petchema:
That `DaggerfallFixedListPickerWindow` approach, moving some UI code to places inconsistent with other windows, makes me more nervous than a few duplicated lines...

KABoissonneault (approved):

PR #2688, opened 2024-08-16
Allow guild quests to be dispensed by player level as well as rank

ajrb:
Guild quests can be dispensed by player level as well as rank if minReq is <10 depending on the guild implementation. Proposed alternative to #2678 .

Enabled for Mages Guild and Knightly Orders.

Will add for Witch Coverns if evidence from classic Daggerfall is provided that contradicts https://en.uesp.net/wiki/Daggerfall:Witch_Covens

Only other thought is whether this change should have a way to disable this for players who prefer how DFU 1.0 currently works. Happy to add a setting or ini file property if you think worthwhile, @KABoissonneault & @petchema ?

### DIFF
--- a/Assets/Scripts/Game/Guilds/Guild.cs
+++ b/Assets/Scripts/Game/Guilds/Guild.cs
@@ -48,6 +48,11 @@ namespace DaggerfallWorkshop.Game.Guilds
 
         public abstract List<DFCareer.Skills> TrainingSkills { get; }
 
+        public virtual bool IsAllowLevelQuests()
+        {
+            return false;
+        }
+
         #endregion
 
         #region Guild Ranks
--- a/Assets/Scripts/Game/Guilds/Guild.cs
+++ b/Assets/Scripts/Game/Guilds/Guild.cs
@@ -48,7 +48,7 @@ namespace DaggerfallWorkshop.Game.Guilds
 
         public abstract List<DFCareer.Skills> TrainingSkills { get; }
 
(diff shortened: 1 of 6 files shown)


petchema:
That's not so simple to check if witches coven consider rank or level, given that all their quests are reputation-gated (when there's any precondition).
Unless you were thinking about adding a rank-gated witches quest to classic Daggerfall to check how it'd behave? That's a bit out of my league.
If that helps, a save near the Tamarilyn Witches:
[SAVE5 - Looking for witches.zip](https://github.com/user-attachments/files/16642257/SAVE5.-.Looking.for.witches.zip)

petchema (approved):
Tested to work as advertised, with or without player quest selection.
No remarks about code either

ajrb:
Switched witches quest dispensing to player level as proved in classic. It's possible that reputation also enables offers as UESP states, but I'm not implementing that unless it's proven to be the case. This would require having a witch covern rep at > 10x player level and asking for quests. I think this is unlikely to happen in game but could be tested in classic by finding a covern, doing 3 covern quests successfully without reaching level 2, then getting offered one of the last 3 quests from https://en.uesp.net/wiki/Daggerfall:Witch_Covens. If someone does show this is how classic works, then I'll implement a special case for witches coverns which was the only faction quests that remained based on reputation since they were all changed back in December 2018 when it was shown that UESP was wrong (at that time) and all switched to rank/level based dispensing.

petchema (approved):
LGTM

PR #2690, opened 2024-08-19
Removes totem delivery to foes during "Who Gets the Totem"

JayH2971:
See bug report here: https://www.reddit.com/gallery/1evrwhg . Handing over a nonexistent totem is a separate problem, and will require a more complicated fix. Not as urgent.

### DIFF
--- a/Assets/StreamingAssets/Quests/A0C01Y01.txt
+++ b/Assets/StreamingAssets/Quests/A0C01Y01.txt
@@ -1,11 +1,3 @@
--- Quest: C:\QUESTS\WORKING\A0C01Y01.Qbn.
--- StartsBy: NPC
--- Questee: anyone
--- Questor: peasant
--- Repute: 01
--- QuestId: 1
--- Edited for Daggerfall Unity by Jay_H
-Messages: 14
 Quest: A0C01Y01
 DisplayName: The Bodyguard
 -- Message panels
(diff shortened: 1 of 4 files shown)


Mentill:
The screenshot in the linked thread appears to be of a modded game. Is it confirmed this problem happens in the unmodded game?
Perhaps a better solution is to remove it from the inventory rather than using the 'give item' command. I see a line that says `take _totem_ from pc`, that should work.
 Or, `F.00` could be defined as just a single enemy. Just dummying out what's there would keep the Totem in the player's inventory. Another option is to despawn the knights after they say their message.

JayH2971:
> Is it confirmed this problem happens in the unmodded game?

There are no mods that affect the "give item to" action.

> Perhaps a better solution is to remove it from the inventory rather than using the 'give item' command.

This is what I have done.

> Another option is to despawn the knights after they say their message.

It's another option. Either one of them is fine for now, though there can always be discussion on which one would be better, and why.

petchema:
Looking at the code, GiveItem doesn't check that the player's character has the item before handing it to the recipient, because that's not part of this action's semantics (it's never said it's the character giving the item). All it does as a last step is remove the item from the inventory, if it was there.
So it's up to the quest to ensure GiveItem is attempted under the expected conditions, which it probably should anyway (ideally player should not be asked if you want to hand an object the character don't have).
(I'm not sure why/how the Totem has been duplicated)

KABoissonneault (changes requested):
With your changes, if the player _does_ have the totem, it no longer gets removed from their inventory when choosing Yes on the prompt. 
I do think this issue needs fixing, but it either needs a new flow to handle if the player does not have the totem, or we need a line to remove the item from the player. Or give the totem to a dummy NPC that does not get spawned

KABoissonneault (approved):
I can't do an extensive review here, quests are a bit complicated. But I'm hoping players will be able to test this

PR #2692, opened 2024-08-28
Auto-Dismount After OnPreTransition

ArshvirGoraya:
# Auto Dismount AFTER OnPreTransition

For a mod, I wanted to use the `OnPreTransition` event listener defined in "PlayerEnterExit.cs" to know what the `TransportMode` was prior to transitioning into an interior. However, "TransportManager.cs" automatically sets the `TransportMode` to foot on this event. So, what the `TransportMode` was prior to the transition isn't able to be known by any `OnPreTransition` event listeners after that one has been triggered, and the workaround for this isn't as performant as just checking what the `TransportMode` is on this event.

In my opinion, it makes more sense semantically for the `OnPreTransition` event to be able to access what the `TransportMode` was pre the transition. 

Therefore, this PR simply sets the `TransportMode` to foot in the "OnTransitionInterior" events instead. 
 - The "OnTransitionExterior" events aren't added as we can't change the `TransportMode` inside, so when we exit, we will already be on foot and don't need to auto dismount.

The changes in this PR shouldn't change anything for the player and there doesn't seem to be any problems in my play testing. There is nothing in the code that relies on the `TransportMode` being set to foot on the `OnPreTransition` event, so nothing should break. 

Ultimately, it would be nice to allow developers and modders to know what the `TranportMode` is prior to transitioning into an interior using the `OnPreTransition` event, and not require a less performant workaround.

### DIFF
--- a/Assets/Scripts/Game/TransportManager.cs
+++ b/Assets/Scripts/Game/TransportManager.cs
@@ -199,7 +199,8 @@ namespace DaggerfallWorkshop.Game
             neighClip = dfAudioSource.GetAudioClip((int)horseSound);
 
             // Init event listener for transitions.
-            PlayerEnterExit.OnPreTransition += new PlayerEnterExit.OnPreTransitionEventHandler(HandleTransition);
+            PlayerEnterExit.OnTransitionInterior += HandleTransition;
+            PlayerEnterExit.OnTransitionDungeonInterior += HandleTransition;
         }
 
         // Handle interior/exterior transition events by setting transport mode to Foot.
--- a/Assets/Scripts/Game/TransportManager.cs
+++ b/Assets/Scripts/Game/TransportManager.cs
@@ -50,6 +50,10 @@ namespace DaggerfallWorkshop.Game
             get { return mode; }
             set { UpdateMode(value); }
         }
+        public TransportModes PreviousTransportMode
+        {
+            get { return previousTransportMode; }
+        }
 
         /// <summary>True when player is on foot.</summary>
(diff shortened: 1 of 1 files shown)


ajrb:
What's your use case for this?

KABoissonneault:
I think the use case is pretty clear. However, I'm not too good with the side effects of changing the order of events for mods currently using these events.
Couldn't the transport manager store its last transport mode and provide a new getter?

ArshvirGoraya:
@KABoissonneault 

>Get what transport the player was using before entering the dungeon/interior

This is it. I also agree that the potential side effects for mods already using this event isn't good. 
I like your suggestion to store the previous transport mode. It's also something I mentioned doing in this PR: https://github.com/Interkarma/daggerfall-unity/pull/2694

I made the changes here.

ajrb:
I wanted to know if there was anything else required that I wasn't seeing. I think this can be done much simpler. I'll submit my proposal this weekend and you can take a look and verify it supports what you need.

ajrb:
Sorry I have had some RL things come up with my aging dad and now I am on holiday for 2 weeks. I will get to what I said I'd do but it will be a little while longer. Sorry.

ajrb:
I apologise for the 6 weeks it took me to submit a simpler approach, but I finally got enough time to get it done. Please check #2701 to see if it could replace this PR and the other.

ArshvirGoraya:
Closing PR in favor of this instead: https://github.com/Interkarma/daggerfall-unity/pull/2701

PR #2694, opened 2024-08-31
Transport Manager Events

ArshvirGoraya:
# Add Events for Transport Mode Changes

Adds events inside of "TransportManager.cs" and triggers them inside the UpdateMode method.

- OnTransportModeChanged 
  - Usage: `TransportManager.OnTransportModeChanged += TransportManager_OnTransportModeChanged;`
  - A simple event that is raised when the TransportMode changes. Has the new TransportMode value as a parameter.
- OnTransportModeShipChanged 
  - Usage: `TransportManager.OnTransportModeShipChanged += TransportManager_OnTransportModeShipChanged;`
  - An event that is raised when the transport mode is changed to or from ship. Has a ToShip or FromShip parameter to indicate if we are warping to the ship or out of it.

Why have a separate event for the ship?

- The code doesn't handle the ship mode like the others. It sets it to foot immediately after the ship is warped into or out of. This can be problematic if a developer makes some [mod](https://github.com/ArshvirGoraya/Remember-Transport-Mode) that sets the `TransportMode` to something right after warping to the ship. Whatever they set it to will be overwritten by the automatic "foot" transport mode set by the code. So an event specifically for the ship is better for those cases as those will only be raised when the ship has been warped to/from and after the foot transport mode has been automatically set. It ends up being much cleaner for any code that needs to run right after the TransportMode is set to the ship.

It may also be useful to have an event like "ChangingFromTransportMode," to let developers know when the TransportMode is changing away from something. I haven't added this however, since a nice alternative is just creating an accessible variable called PreviousTransportMode, which they can check whenever the above `OnTransportModeChanged` is fired. This isn't needed for my mod specifically, but I thought it was worth mentioning as it may be useful for others.

### DIFF
--- a/Assets/Scripts/Game/TransportManager.cs
+++ b/Assets/Scripts/Game/TransportManager.cs
@@ -325,6 +325,47 @@ namespace DaggerfallWorkshop.Game
 
         #endregion
 
+        #region Event Arguments
+
+        /// <summary>
+        /// Types of ship transition encountered by TransportManager.
+        /// </summary>
+        public enum TransportShipType
+        {
+            ToShip,
+            FropShip,
+        }
+        #endregion
+
+        #region Events
+
+        // OnTransportModeChanged
+        public delegate void OnTransportModeChangedEventHandler(TransportModes TransportMode);
+        /// <summary>
+        /// An event that is raised when the transport mode is changed.
+        /// </summary>
+        public static event OnTransportModeChangedEventHandler OnTransportModeChanged;
+        protected virtual void RaiseOnTransportModeChangedEvent(TransportModes TransportMode)
+        {
+            if (OnTransportModeChanged != null){
+                OnTransportModeChanged(TransportMode);
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


ajrb:
I think the ship event is redundant. The change mode event could specify the from and to mode easily. What exactly is the use case you're trying to achieve?

ArshvirGoraya:
@ajrb 

The use case is to optimize one of my mods in the next release: https://github.com/ArshvirGoraya/Remember-Transport-Mode. 

One of the things the mod does is saves the TransportMode whenever the player changes it in exteriors. In order to do this, I currently just have some logic in my Update() function that checks when the TransportMode changes by comparing it to what it was last time the Update() function ran. However, it would be more optimal if I could just listen to TransportMode change events.

For the Ship event: you're right. I changed the code to remove the ship event and modified the ModeChange event to include ship warp information.

ArshvirGoraya:
Closing PR in favor of this instead: https://github.com/Interkarma/daggerfall-unity/pull/2701

PR #2697, opened 2024-09-09
Removed the "protected internal" keywords from the DaggerfallTravelPopUp.EndPos setter.

RedRoryOTheGlen:
Allows mods to change the endPos of DaggerfallTravelPopUp.

[Video](https://streamable.com/o0a2f8) of proof-of-concept fast travel encounter mod I was working on. If it decides to roll an encounter, it will change the destination of the travel in the DaggerfallTravelPopUp.OnPreFastTravel event. Nothing fancy. This threw me for a loop because it pretty much worked flawlessly as a virtual mod and only caused an error when building the DFMOD.

### DIFF
--- a/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTravelPopUp.cs
+++ b/Assets/Scripts/Game/UserInterfaceWindows/DaggerfallTravelPopUp.cs
@@ -94,7 +94,7 @@ namespace DaggerfallWorkshop.Game.UserInterfaceWindows
 
         #region Properties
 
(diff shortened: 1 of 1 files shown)


KABoissonneault:
For the "works in editor, fails in dfmod" issue, you only need to remove the `internal`, which I'm fine with removing (I don't think it's a great C# feature when working with Unity). I'd rather not remove the `protected` unless you had some justification for it.

RedRoryOTheGlen:
image

I tried it with the EndPos setter as "protected" only and now even the DaggerfallTravelMapWindow set throws an error in the Editor

KABoissonneault (approved):

PR #2701, opened 2024-10-13
Add OnTransportModeChanged event to TransportManager

ajrb:
This is an simpler alternative way to add a OnTransportModeChanged event to the TransportManager which provides the new and old transport mode after all the changes have been made, providing mods with all of the info required to post-manipulate the mode as desired.

The only oddity is with ship as this isn't really a transport mode, just a toggle teleport that was added to the others. When moving to ship it will be to Foot and when returning it'll be to Ship.

This is intended as an alternative to both PR's #2692 and #2694 and I would appreciate if @ArshvirGoraya could check whether this meets their needs and report back if additional info is needed for some use case?

### DIFF
--- a/Assets/Scripts/Game/TransportManager.cs
+++ b/Assets/Scripts/Game/TransportManager.cs
@@ -330,6 +330,7 @@ namespace DaggerfallWorkshop.Game
         private void UpdateMode(TransportModes transportMode)
         {
             // Update the transport mode and stop any riding sounds playing.
+            TransportModes previousMode = mode;
             mode = transportMode;
             if (ridingAudioSource.isPlaying)
                 ridingAudioSource.Stop();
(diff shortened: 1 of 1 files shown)


ArshvirGoraya:
This works very well for me. 

@ajrb Could you just double check this:
> When moving to ship it will be to Foot and when returning it'll be to Ship.

For me, if I warp in or out of the ship, the `newMode` is always ship.

I ask because I was worried about not being able to know if we are warping to and from ship but that's very easy with the following code (which wouldn't work if, for some people, `newMode` is not always ship when warping in and out of it):

```CS
void TransportManager_OnTransportModeChanged(TransportModes newMode, TransportModes previousMode){
    if (newMode == TransportModes.Ship){
        if (transportManager.IsOnShip()){
          Debug.Log($"Warped TO Ship"); 
        }
        else{ 
          Debug.Log($"Warped FROM Ship");
        }
    }
    Debug.Log($"Transport Mode Changed: {previousMode} -> {newMode}");
}
```

Also attached a video with the above code running: https://github.com/user-attachments/assets/06c087df-7efc-4fc3-80f0-0bdbd9513fc4

ajrb:
Sorry fo the delay (again) - managed to get laid low by covid on return from holiday. Only just getting back on my feet again. If it works as you need it to that's great. My warning was redundant then I guess.

If you're good with this PR then close yours and hopefully this one can get reviewed and committed soon.

I really appreciate your patience with me. Life gets in the way sometimes.

Daneel53 (approved):
ArshvirGoraya closed their both PRs to let this one alive.
Time to integrate this new event into DFU.

Issue #2705, opened 2024-10-27
Mod searchbar suggestion


EvieBigCheese:
The game lacks any sort of a searchbar to quickly locate a mod you're looking for, making it especially tedious for people with many mods installed

**To Reproduce**

1. Go to the mod menu
2.  Voila

**Expected behavior**

A searchbar allowing user to quickly locate a mod

**Desktop (please complete the following information):**

 - OS: Windows 10

PR #2707, opened 2024-11-11
mods list search

petchema:
Mod list search (#2705) implementation

At first I tried to implement view filtering (only display mods that matched), but that made the semantic of "increase mod order" and "decrease mod order" buttons... complicated. Also, the mod manager relies heavily on several arrays using the same indices, so it was almost intractable.

So I implemented a simpler search; Within the UI components limitations, I added a text field, and "up" and "down" arrow to find the previous and next match. It's serviceable.


petchema:
modlist search

Daneel53 (approved):
Works as intended.

PR #2709, opened 2024-11-12
Allow mods to add new RMB blocks

drcarademono:
Mods currently cannot access custom RMB blocks. New RMB blocks are only added to WorldDataReplacement's new block index when a player visits a regular Daggerfall location. This is a problem for mods like LocationLoader (in World of Daggerfall), which aims to add custom RMB blocks throughout the world, not just at regular locations. This PR simply sets WorldDataReplacement.AssignNextIndex to public, allowing LocationLoader to call it and add a new RMB block to the index.

### DIFF
--- a/Assets/Scripts/Utility/AssetInjection/WorldDataReplacement.cs
+++ b/Assets/Scripts/Utility/AssetInjection/WorldDataReplacement.cs
@@ -504,7 +504,7 @@ namespace DaggerfallWorkshop.Utility.AssetInjection
             return false;
         }
 
-        private static void AssignNextIndex(string blockName)
+        public static void AssignNextIndex(string blockName)
         {
             newBlockNames[nextBlockIndex] = blockName;
             newBlockIndices[blockName] = nextBlockIndex;


ajrb:
RMB blocks by definition are only used at daggerfall locations as defined by location data. Any usage outside of this is also outside of the location & block indexes you're referring to.

drcarademono:
This PR allows RMB blocks to be used by mods, and not just at Daggerfall locations as defined by location data.

Kab added the ability in Location Loader (World of Daggerfall's utility mod) to place an RMB block anywhere on the map, including outside traditional Daggerfall locations. However, in order for a mod to use custom RMB blocks, they need to first be added to WorldDataReplacement's new block index, because Location Loader calls RMBLayout.CreateBaseGameObject and other RMBLayout methods.

I should add that this PR has been replaced by 2718 (https://github.com/Interkarma/daggerfall-unity/pull/2718) which both allows mods to add custom RMB blocks _and_ fixes the potentially game-breaking bug we discussed a couple of months ago caused by dynamically assigned block indices for custom RMBs.

drcarademono:
Not a solution for the performance issue, but I re-added the original method as a fallback to assure backwards compatibility with any mods that don't use unique Index values in their custom RMBs. I don't think this is a big issue, because (as discussed earlier) to the best of my knowledge only Cliffworms, Wayton, and I have added custom RMBs, but this commit will make sure that any others are handled as they were before.

PR #2712, opened 2024-11-29
Fix The Akatosh Chantry bug

drcarademono:
This fixes a bug in which temples of Akatosh are incorrectly named The Order of the Hour instead of The Akatosh Chantry, even though all NPCs inside refer to it as The Akatosh Chantry.

The bug was [noticed six years ago here](https://forums.dfworkshop.net/viewtopic.php?t=877) and likely replicates DOS Daggerfall behavior. The cause is a mistake in vanilla location data for Akatosh temples. Most location data sets the Temple FactionID as the relevant god (ie 33 for Stendarr) and then this method grabs the first child (106 for Temple of Stendarr). However, the location data for Akatosh temples refer directly to The Akatosh Chantry (FactionID 92) instead of the god (FactionID 26). The first child of The Akatosh Chantry is The Order of the Hour.

This PR makes the method first traverse to the top-level factionID and then takes its first child for temple names.

### DIFF
--- a/Assets/Scripts/Game/Formulas/FormulaHelper.cs
+++ b/Assets/Scripts/Game/Formulas/FormulaHelper.cs
@@ -3025,9 +3025,23 @@ namespace DaggerfallWorkshop.Game.Formulas
                     break;
 
                 case DFLocation.BuildingTypes.Temple:
-                    // Temples get name from faction data - always seem to be first child of factionID
+                    // Temples get name from faction data - always seem to be first child of the top-level factionID
                     if (GameManager.Instance.PlayerEntity.FactionData.GetFactionData(factionID, out factionData))
                     {
+                        // Traverse up to find the top-level faction
+                        while (factionData.parent != 0)
+                        {
+                            if (GameManager.Instance.PlayerEntity.FactionData.GetFactionData(factionData.parent, out FactionFile.FactionData parentFaction))
+                            {
+                                factionData = parentFaction;
+                            }
+                            else
+                            {
+                                break; // If we can't find the parent, stop the loop
+                            }
+                        }
+
+                        // Now that we have the top-level faction, get its first child
                         if (factionData.children.Count > 0)
                         {
                             FactionFile.FactionData firstChild;


KABoissonneault:
If this replicates classic behavior and does not cause issues in DFU, why change it? I understand that this might be an area hard or impossible to mod for setups that wish to use the correct names, but in that case the solution should be to make this more moddable.

drcarademono:
That's reasonable. The only issue is the mismatch between the building name and the dialogue of the NPCs inside the building. But that's hardly game-breaking, and modders can also fix it by correcting the location data via mod.

df-ferital:
This is actually a bug from classic Daggerfall, where all temples, with the exception of the Akatasoh Chantry, are marked with their respective god's faction ID instead of the temple one. And rather than patching BLOCKS.BSA to fix this bug, a somewhat ugly hack has been implemented in classic DF code itself, to ensure that when displaying the name of a temple dedicated to a deity, the temple name is show instead of the name of the said god. E.g. for temples marked with the faction ID "Kynareth", their name is always displayed as "The Temple of Kynareth". On the other hand, when you speak to individuals inside these temples, they will claim to belong to the "Kynareth" faction, which makes no sense.

Anyway, the devs at the time didn't pay attention to the fact that only Akatosh temples were marked with their correct faction ID. As a result, the hack I mentioned above means that when the name of an Akatosh Chantry is displayed, the name of its first child faction appears instead, namely "The Order of the Hour".

In my opinion, as this is a classic bug, and Daggerfall Unity is supposed to fix classic bugs when it can, it should be corrected, but not with a new hack, but rather by modifying BLOCKS.BSA with the patch mechanism I've set up years ago to correct bad models.

As a side note, BLOCKS.BSA is litteraly plagued with bad faction IDs assignment. A well-known example is that of some Fighters Guild buildings in Hammerfell which use the faction ID of "Fighters Trainers".

drcarademono:
I agree that this bug should be fixed and that patching BLOCKS.BSA with the patch mechanism would be the best way to go about it. I don't know how to do that though, so hoping for another PR on this issue.

PR #2715, opened 2024-12-24
Mergable building override automap data

ajrb:
This allows building overrides to only replace the automap data for the relevant area of the RMB for the specific building. This effectively enables merging of overridden map data. The picture below shows the town map, first is with the building override data being used (current functionality), second is without the building override (the modified RMB map data taken from Beautiful Cities), and the third is with functionality added by this PR where the building override only overwrites the part it needs to.

autodataMerge

You can see it overwrites some of the nearby buildings, but after adjusting the amount of area it looks very good. (note Beautiful Cities map data doesn't include the random objects and flats that can be displayed using the red grid icon)

autodataMerge2

The unused value of '30' is used to signify no change for that pixel in a building override automap data byte array. See the attached modified building file for an example where the automap data has been changed to only affect the building area. It can be compared to the original from my [Archaeologists mod](https://github.com/ajrb/dfunity-mods/blob/master/Archaeologists/WorldData/PAWNAL03.RMB-857-building3.json).

[PAWNAL03.RMB-857-building3.json](https://github.com/user-attachments/files/18240766/PAWNAL03.RMB-857-building3.json)

### DIFF
--- a/Assets/Scripts/API/BlocksFile.cs
+++ b/Assets/Scripts/API/BlocksFile.cs
@@ -846,7 +846,7 @@ namespace DaggerfallConnect.Arena2
             BuildingReplacementData buildingReplacementData;
             for (int i = 0; i < recordCount; i++)
             {
-                // Check for replacement building data and use it if found
+                // Check for replacement building data and use it, if found
                 if (WorldDataReplacement.GetBuildingReplacementData(blocks[block].Name, block, i, out buildingReplacementData))
                 {
                     blocks[block].DFBlock.RmbBlock.SubRecords[i] = buildingReplacementData.RmbSubRecord;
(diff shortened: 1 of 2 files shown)


ajrb:
This PR relies on #2714 being merged first and the branch branches off that, so includes those changes.

KABoissonneault (dismissed):
As with the other change, I'll trust Hazel to maintain this part

KABoissonneault (approved):

KABoissonneault (approved):
Github won't let me merge, sorry for spam

ajrb:
Yeah I have never understood how to do this correctly with git so I expected this. Will re-submit a new PR with just this change.

Issue #2703, opened 2024-10-23
Saves in buildings within custom RMBs cannot be loaded


drcarademono:
Mods like Fixed Desert Architecture and Beautiful Cities of Daggerfall add completely new RMBs using unique filenames (TEMPLEBA00.RMB.json or FARMBA04.RMB.json), and then insert these RMBs into the game by modifying the BlockNames array in location files. These custom RMBs work correctly, however, if the player saves inside a building in one of these custom blocks, the save will be unloadable. If a player restarts DFU and attempts to load the save, they'll get a black screen and the following error in the log:

`Exception: Could not load RMB block index 1439
DaggerfallWorkshop.DaggerfallInterior.AssignBlockData (DaggerfallWorkshop.StaticDoor door) (at Assets/Scripts/Internal/DaggerfallInterior.cs:384)
DaggerfallWorkshop.DaggerfallInterior.DoLayout (UnityEngine.Transform doorOwner, DaggerfallWorkshop.StaticDoor door, DaggerfallWorkshop.ClimateBases climateBase, DaggerfallWorkshop.PlayerGPS+DiscoveredBuilding buildingData) (at Assets/Scripts/Internal/DaggerfallInterior.cs:152)
DaggerfallWorkshop.Game.PlayerEnterExit.TransitionInterior (UnityEngine.Transform doorOwner, DaggerfallWorkshop.StaticDoor door, System.Boolean doFade, System.Boolean start) (at Assets/Scripts/Game/PlayerEnterExit.cs:721)
UnityEngine.Debug:LogException(Exception)
DaggerfallWorkshop.Game.PlayerEnterExit:TransitionInterior(Transform, StaticDoor, Boolean, Boolean) (at Assets/Scripts/Game/PlayerEnterExit.cs:726)
DaggerfallWorkshop.Game.PlayerEnterExit:StartBuildingInterior(DFLocation, StaticDoor, Boolean) (at Assets/Scripts/Game/PlayerEnterExit.cs:1034)
DaggerfallWorkshop.Game.<Respawner>d__105:MoveNext() (at Assets/Scripts/Game/PlayerEnterExit.cs:545)
UnityEngine.GUIUtility:ProcessEvent(Int32, IntPtr) (at /home/bokken/buildslave/unity/build/Modules/IMGUI/GUIUtility.cs:197)`

[...]

PR #2718, opened 2025-01-12
Assign constant rather than dynamic blockIndices to modded RMBs

drcarademono:
This PR is aimed at fixing this rather severe issue with modded RMB files (those with non-vanilla blockNames) first described here: https://github.com/Interkarma/daggerfall-unity/issues/2703. Because modded RMBs are assigned their blockIndices dynamically rather than deterministically, their blockIndices change when the player restarts the game or loads a save. This breaks all interiors associated with modded RMBs. This means that saves within buildings in modded RMBs are corrupted and cannot be loaded. It also means that questors assigned to buildings in modded RMBs disappear, preventing players from completed quests -- including the main quest. Unfortunately this issue affects some of the most widely used mods, including Fixed Desert Architecture, Finding My Religion, Beautiful Cities/Villages, and Cities Overhauled.

This PR fixes the issue in the simplest possible manner by allowing the modder to assign a constant blockIndex to modded RMBs. I replaced the AssignNextIndex method with the AssignNewIndex method, which looks up the Index value from the RMB.json file:
`
{
    "Position": 14551075,
    "Index": 2012,
    "Name": "TEMPASB0.RMB",
    "Type": "Rmb",
    "RmbBlock": {
`
Here AssignNewIndex will always assign this non-vanilla RMB block, TEMPASB0.RMB, the constant index 2012, which is well above the highest vanilla blockIndex (1294). 

This PR will require RMB modders to change the Index values in all of their non-vanilla RMBs and to claim unique RMB blockIndices in the [Daggerfall Workshop Resource Register](https://forums.dfworkshop.net/viewtopic.php?t=3023). This will not require much adjustment since this is already standard practice for all other assets that require a unique, constant value, like LocationIds, FactionIds, and texture archives. To the best of my knowledge, only Cliffworms, Wayton, and I have made modded new RMBs so far, and coordinating between us will be easy.

I have tested this PR on modded new blocks in Beautiful Cities/Beautiful Villages and can affirm that saves inside modded RMB buildings successfully load and questors appear correctly.

I made this PR as a minimal change. One downside of my approach is that it requires deserializing modded RMB.jsons twice, once to get the Index value in AssignNewIndex and once to get the rest of the data in GetDFBlockReplacementData. While I don't find that this noticeably increases load time at startup, a more adept coder might be able to reorganize the WorldDataReplacements class or BlocksFiles so that it deserializes modded RMB.jsons only once.

Finally, please note that I made AssignNewIndex public. This is so that mods (and here I'm particularly thinking of World of Daggerfall) can add new modded RMBs as well. Therefore, this PR replaces [2709](https://github.com/Interkarma/daggerfall-unity/pull/2709).

### DIFF
--- a/Assets/Scripts/Utility/AssetInjection/WorldDataReplacement.cs
+++ b/Assets/Scripts/Utility/AssetInjection/WorldDataReplacement.cs
@@ -486,7 +486,7 @@ namespace DaggerfallWorkshop.Utility.AssetInjection
                     // RMB blocks
                     foreach (string blockName in dfLocation.Exterior.ExteriorData.BlockNames)
                         if (blocksFile.GetBlockIndex(blockName) == -1)
-                            AssignNextIndex(blockName);
+                            AssignNewIndex(blockName);
 
                     // RDB blocks
                     if (dfLocation.Dungeon.Blocks != null)
@@ -495,7 +495,7 @@ namespace DaggerfallWorkshop.Utility.AssetInjection
                         {
                             string blockName = dungeonBlock.BlockName;
                             if (blocksFile.GetBlockIndex(blockName) == -1)
-                                AssignNextIndex(blockName);
+                                AssignNewIndex(blockName);
                         }
                     }
                     return true;
@@ -504,12 +504,40 @@ namespace DaggerfallWorkshop.Utility.AssetInjection
             return false;
         }
 
-        private static void AssignNextIndex(string blockName)
+        public static void AssignNewIndex(string blockName)
         {
-            newBlockNames[nextBlockIndex] = blockName;
-            newBlockIndices[blockName] = nextBlockIndex;
-            Debug.LogFormat("Found a new DFBlock: {0}, (assigned index: {1})", blockName, nextBlockIndex);
-            nextBlockIndex++;
+            string fileName = GetDFBlockReplacementFilename(blockName, WorldDataVariants.NoVariant);
+
+            int jsonBlockIndex;
+            DFBlock? dfBlock = null; // Make blockData nullable
+            TextAsset blockReplacementJsonAsset;
+
+            // Seek from loose files
+            if (File.Exists(Path.Combine(worldDataPath, fileName)))
+            {
+                string blockReplacementJson = File.ReadAllText(Path.Combine(worldDataPath, fileName));
+                dfBlock = (DFBlock)SaveLoadManager.Deserialize(typeof(DFBlock), blockReplacementJson);
+            }
+            // Seek from mods
+            else if (ModManager.Instance != null && ModManager.Instance.TryGetAsset(fileName, false, out blockReplacementJsonAsset))
+            {
+                dfBlock = (DFBlock)SaveLoadManager.Deserialize(typeof(DFBlock), blockReplacementJsonAsset.text);
+            }
+
+            // Ensure blockData was successfully assigned
+            if (!dfBlock.HasValue) // Check if blockData has a value
+            {
+                Debug.LogError($"Failed to load block data for blockName: {blockName}");
+                throw new System.Exception($"Block {blockName} does not have a valid Index in its JSON file.");
+            }
+
+            // Check for the "Index" field and assign its value
+            jsonBlockIndex = dfBlock.Value.Index; // Use .Value to access the DFBlock inside the nullable
+
+            // Add to cache
+            newBlockNames[jsonBlockIndex] = blockName;
+            newBlockIndices[blockName] = jsonBlockIndex;
+            Debug.LogFormat("Found a new DFBlock: {0}, (assigned index: {1})", blockName, jsonBlockIndex);
         }
 
         #endregion


drcarademono:
Testing this a bit more, I have noticed a performance hit caused by deserializing lots of large jsons twice (once to get the index and a second time to get the replacement data. I'm hoping that a better coder than I could restructure this to either a) deserialize custom RMBs once, grabbing the index and replacement data all in one go, or b) use the filename method I suggested above.

drcarademono:
Not a solution for the performance issue, but I re-added the original method as a fallback to insure backwards compatibility with any mods that don't use unique Index values in their custom RMBs. I don't think this is a big issue, because (as discussed earlier) to the best of my knowledge only Cliffworms, Wayton, and I have added custom RMBs, but this commit will make sure that any others are handled as they were before.

Assets/Scripts/Utility/AssetInjection/WorldDataReplacement.cs
```diff
@@ -504,6 +504,49 @@ private static bool AssignBlockIndices(ref DFLocation dfLocation)
...
+            {
+                Debug.LogError($"Failed to load block data for blockName: {blockName}");
+                throw new System.Exception($"Block {blockName} does not have a valid Index in its JSON file.");
+            }
+
+            // Check for the "Index" field and assign its value
+            jsonBlockIndex = dfBlock.Value.Index;
+
+            // If jsonBlockIndex is invalid (less than or equal to blocksFile.BsaFile.Count), use fallback method
+            if (jsonBlockIndex <= blocksFile.BsaFile.Count)
```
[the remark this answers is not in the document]
drcarademono: Good catch and a good solution. Thanks, Z-Machine!

drcarademono:
Solved the double-deserialization issue with the simple and in retrospect obvious fix of caching the DFBlock on the first deserialization. Now GetDFBlockReplacementData will get the DFBlock from the cache rather than deserializing the json the second time. There should no longer be any performance issues with this PR.

Assets/Scripts/Utility/AssetInjection/WorldDataReplacement.cs
```diff
@@ -504,6 +513,83 @@ private static bool AssignBlockIndices(ref DFLocation dfLocation)
...
+            jsonBlockIndex = dfBlock.Value.Index;
+
+            // If jsonBlockIndex is invalid (less than or equal to nextBlockIndex), use fallback method
+            if (jsonBlockIndex <= nextBlockIndex)
+            {
+                AssignNextIndex(blockName);
+
+                // Cache the full DFBlock
+                if (!blocks.ContainsKey(blockKey))
+                    blocks[blockKey] = dfBlock.Value;
```
TwelveEyes, line 578: The cached DFBlock for the `AssignNextIndex` fallback also needs to have it's internal `Index` value updated. What I would recommend:
```diff
diff --git a/Assets/Scripts/Utility/AssetInjection/WorldDataReplacement.cs b/Assets/Scripts/Utility/AssetInjection/WorldDataReplacement.cs
index 92ddee749..73875e6e4 100644
--- a/Assets/Scripts/Utility/AssetInjection/WorldDataReplacement.cs
+++ b/Assets/Scripts/Utility/AssetInjection/WorldDataReplacement.cs
@@ -627,11 +627,19 @@ namespace DaggerfallWorkshop.Utility.AssetInjection
             // If jsonBlockIndex is invalid (less than or equal to nextBlockIndex), use fallback method
             if (jsonBlockIndex <= nextBlockIndex)
             {
+                int newIndex = nextBlockIndex;
+
                 AssignNextIndex(blockName);
 
                 // Cache the full DFBlock
                 if (!blocks.ContainsKey(blockKey))
-                    blocks[blockKey] = dfBlock.Value;
+                {
+                    // Update DFBlock index
+                    DFBlock block = dfBlock.Value;
+                    block.Index = newIndex;
+
+                    blocks[blockKey] = block;
+                }
 
                 return;
             }
```

Otherwise `AssignBlockData()` will potentially pull the wrong data for RMBs assigned an index through the `AssignNextIndex()` fallback, causing the player to enter the wrong interior, or throwing an exception because the `door.resourceIndex` is outside the bounds of the RMB's `SubRecords`.

This bug is clearly visible [here](https://discord.com/channels/559842137374457876/559843513521537025/1382096850386944030) with Cliffworm's release version of Lively Cities.

For more insight I'll use the mod Village of Fishguard by dotWayton to demonstrate the bug step by step.
Here you can see Fishguard's RMBs are both set to index 308 within their JSON files.



And here we can see Fishguard's RMBs have been assigned new indexes by `AssignNextIndex()`:


Attempting to enter a house in `FISHER03.RMB` results in this:


We can see `AssignBlockData()` grabbed two RMBs for our location, `FISHER02.RMB` and `FISHER03.RMB` (lines 4531 and 4532). And we can immediately see three things are wrong upon it trying to load the record data (on line 4533): it is pulling from `Block Name` `FISHER02.RMB` instead of `FISHER03.RMB`, the `Block Index` and the `Door BlockIndex` are both wrong (`308` instead of `1296`), and the `Door RecordIndex` far exceeds the `Block SubRecord length` (which causes the exception on line 4534). This is because the internal index value for the block wasn't updated, so the door is still using the old block index of 308, and since `FISHER02.RMB` is the first block that matches the door's block index of 308 it gets selected.

After the change in the diff above was applied:


It's now pulling from the correct block `FISHER03.RMB` instead of `FISHER02.RMB`, with the correct `Block Index` and `Door BlockIndex` of `1297` instead of `308`, and with the `Door RecordIndex` correctly within the bounds of the `Block SubRecords length`.

Assets/Scripts/Utility/AssetInjection/WorldDataReplacement.cs
```diff
@@ -504,6 +513,83 @@ private static bool AssignBlockIndices(ref DFLocation dfLocation)
...
+
+                return;
+            }
+
+            // Add to cache
+            newBlockNames[jsonBlockIndex] = blockName;
+            newBlockIndices[blockName] = jsonBlockIndex;
```
TwelveEyes, line 585: Probably want to add a fallback to `AssignNextIndex` here too in case there are already entries for the key `jsonBlockIndex` and `blockName`. This way two different modded RMBs with the same custom index (by accident) can still get along.

PR #2719, opened 2025-01-24
Fix Former Student part III reward not triggering 

JayH2971:
Had to do full surgery on this quest to understand it. The ending depended on a "click item" trigger that DFU didn't interpret the same as classic. Changed that mechanic (see "_takemageitem_" task), tested and it works properly now.


JayH2971:
There's a problem in Former Student Part III with the quest action "Make X permanent," which is supposed to remove the quest condition from an item (remove green background, make droppable). It's not functioning here, and I don't know why. Sometimes it works, and sometimes not. These screenshots show me having the quest items in my inventory while no quests are active, which is supposed to be impossible. Since the quest is properly scripted and all events light up correctly, I can't fix it more than that.

Screenshot at 2025-01-24 09-36-04
Screenshot at 2025-01-24 09-36-08

Daneel53:
Just resolved text changes between the quests taken by JayH in 2025 and today's state after the big PR #2780 Proofreading Part II from Mentill.
Quest S0000503 as changed here by JayZ is still to be tested.

PR #2723, opened 2025-02-05
Mergable building override automap data (take 2)

ajrb:
This allows building overrides to only replace the automap data for the relevant area of the RMB for the specific building. This effectively enables merging of overridden map data. The picture below shows the town map, first is with the building override data being used (current functionality), second is without the building override (the modified RMB map data taken from Beautiful Cities), and the third is with functionality added by this PR where the building override only overwrites the part it needs to.

autodataMerge

You can see it overwrites some of the nearby buildings, but after adjusting the amount of area it looks very good. (note Beautiful Cities map data doesn't include the random objects and flats that can be displayed using the red grid icon)

autodataMerge2

The unused value of '30' is used to signify no change for that pixel in a building override automap data byte array. See the attached modified building file for an example where the automap data has been changed to only affect the building area. It can be compared to the original from my [Archaeologists mod](https://github.com/ajrb/dfunity-mods/blob/master/Archaeologists/WorldData/PAWNAL03.RMB-857-building3.json).

[PAWNAL03.RMB-857-building3.json](https://github.com/user-attachments/files/18240766/PAWNAL03.RMB-857-building3.json)

### DIFF
--- a/Assets/Scripts/Utility/AssetInjection/WorldDataReplacement.cs
+++ b/Assets/Scripts/Utility/AssetInjection/WorldDataReplacement.cs
@@ -42,6 +42,7 @@ namespace DaggerfallWorkshop.Utility.AssetInjection
     {
         #region Fields & Constants
 
+        public const int AutoMapDataSize = 64 * 64;
         const int noReplacementIndicator = -1;
         const string worldData = "WorldData";
         static readonly string worldDataPath = Path.Combine(Application.streamingAssetsPath, worldData);
@@ -417,8 +418,8 @@ namespace DaggerfallWorkshop.Utility.AssetInjection
                         dfBlock.RmbBlock.FldHeader.BuildingDataList[i].Quality = buildingReplacementData.Quality;
                     if (buildingReplacementData.NameSeed > 0)
                         dfBlock.RmbBlock.FldHeader.BuildingDataList[i].NameSeed = buildingReplacementData.NameSeed;
-                    if (buildingReplacementData.AutoMapData != null && buildingReplacementData.AutoMapData.Length == 64 * 64)
-                        dfBlock.RmbBlock.FldHeader.AutoMapData = buildingReplacementData.AutoMapData;
+
+                    ApplyBuildingReplacementAutoMapData(buildingReplacementData, ref dfBlock.RmbBlock.FldHeader.AutoMapData);
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


ajrb:
I will update all the building overrides in my mods if this PR is merged into a released DFU. WIll also update the tutorial on the forums.

KABoissonneault:
Am I understanding that with this change, mods _have_ to be updated? Or is that just going to be a new best practice? I'd rather avoid any sort of breaking change - granted, world overrides are probably the kind of mod that can always be updated (text files are easy to extract even from packaged mods).

drcarademono:
I'm actually not aware of any mod except for Hazel's that uses the AutoMapData feature in building overrides -- in fact, I'd wager that no one but Hazel knew it exists -- so there's probably little danger of breaking other mods with this.

Assets/Scripts/Utility/AssetInjection/WorldDataReplacement.cs
```diff
@@ -485,6 +486,23 @@ public static int MakeLocationKey(int regionIndex, int locationIndex)
...
+        /// </summary>
+        /// <param name="buildingData">BuildingReplacementData input</param>
+        /// <param name="blockAutoMapData">A reference to the block automap data byte array to merge into</param>
+        public static void ApplyBuildingReplacementAutoMapData(BuildingReplacementData buildingData, ref byte[] blockAutoMapData)
+        {
+            if (buildingData.AutoMapData != null && buildingData.AutoMapData.Length == AutoMapDataSize)
+            {
+                for (int i = 0; i < AutoMapDataSize; i++)
+                {
+                    if (buildingData.AutoMapData[i] != 30)
```
KABoissonneault, line 500: I wish we had an enum for these values, but I think I can see from AutomapEditor.cs `GetColor` what values are used for what. 30 does seem out of supported range

KABoissonneault (approved):

ajrb:
> Am I understanding that with this change, mods _have_ to be updated? Or is that just going to be a new best practice? I'd rather avoid any sort of breaking change - granted, world overrides are probably the kind of mod that can always be updated (text files are easy to extract even from packaged mods).

@KABoissonneault Sorry for the delay replying, been dealing with my dad. This change won't affect any mods at all. All it does is allow for building overrides to only affect the parts of the RMB automap that the building is in and use this previously unused value to mark the rest as unaffected. Before only 1 automap definition would win, now it's possible to merge parts together. Hope that's clear now.

PR #2725, opened 2025-02-13
Missed 2 UI classes

Jagget:
Adding access to UI controls that I missed in https://github.com/Interkarma/daggerfall-unity/pull/2618


KABoissonneault (approved):
Seems good

PR #2726, opened 2025-02-18
Fix UI/popups background color to use the one set in settings.ini

themrleon:
Fix most of the UI/popups backdrop/dim that was supposed to be based on the `DimAlphaStrength` property from the `settings.ini` file, also fix the inconsistency with the "Rest" screen that has black background. Here some examples of **before and after based on the default 50% dim**:

image1912

This PR would also allow mods to expose this setting so players can choose between 0% and 100% dim via mod settings. 

This is specially useful when there are lots of text and buttons on UI stacked at same time like in the map travel scenario (2nd example).

### DIFF
--- a/Assets/Scripts/Game/EnemyAttack.cs
+++ b/Assets/Scripts/Game/EnemyAttack.cs
@@ -278,7 +278,7 @@ namespace DaggerfallWorkshop.Game
 
                         DaggerfallMessageBox messageBox = new DaggerfallMessageBox(DaggerfallUI.UIManager);
                         messageBox.SetTextTokens(DaggerfallUnity.Instance.TextProvider.GetRSCTokens(doYouSurrenderToGuardsTextID));
-                        messageBox.ParentPanel.BackgroundColor = Color.clear;
+                        messageBox.ParentPanel.BackgroundColor = new Color(0, 0, 0, DaggerfallUnity.Settings.DimAlphaStrength);
                         messageBox.AddButton(DaggerfallMessageBox.MessageBoxButtons.Yes);
                         messageBox.AddButton(DaggerfallMessageBox.MessageBoxButtons.No);
                         messageBox.OnButtonClick += SurrenderToGuardsDialogue_OnButtonClick;
--- a/Assets/Scripts/Game/Questing/QuestMachine.cs
+++ b/Assets/Scripts/Game/Questing/QuestMachine.cs
@@ -904,7 +904,7 @@ namespace DaggerfallWorkshop.Game.Questing
             DaggerfallMessageBox messageBox = new DaggerfallMessageBox(DaggerfallUI.UIManager, DaggerfallMessageBox.CommonMessageBoxButtons.YesNo, tokens);
             messageBox.ClickAnywhereToClose = false;
             messageBox.AllowCancel = false;
-            messageBox.ParentPanel.BackgroundColor = Color.clear;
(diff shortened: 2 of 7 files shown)


KABoissonneault:
# Context

I'm trying to piece the history of this feature.

The setting was added in here: https://github.com/Interkarma/daggerfall-unity/commit/2d57b9f942a7f3e59e7440a368d55464448642c6#diff-8f948b53959f0fb4d0b4b9db7dac6e54689d035b1134df89cddeb4e9dc86d9adR95
This was intended as the screen dim alpha for DaggerfallPopupWindow. 

After, some changelist changed it so it's only the default value, allowing derived classes to not depend on the setting at all: https://github.com/Interkarma/daggerfall-unity/commit/3a2f6ffd18ab783fa8173ca0671b63a5cbd36401

This was finally fully removed in this changelist by Interkarma: https://github.com/Interkarma/daggerfall-unity/commit/15eddf3416e6041b03c4615b14b5aee09d9f9a77

The justification in the commit is: "Using a RenderTexture with GUI.DrawTexture() means that for now any alpha blends are not additive."

Not saying anything as to whether this feature should be restored, I'm just trying to get the context. 

We're still relying on GUI.DrawTexture, so the issue might still apply, even though I'm not certain what the concern even is. Maybe @Interkarma remembers any of this.

# My analysis

1. I generally appreciate the dim on popups, and historically DFU has not implemented all of the ones the original game had (the spellmaker was missing it until this PR https://github.com/Interkarma/daggerfall-unity/pull/2532 )
2. I don't think the Rest screen should use the same value as the rest of the popups. The base DF expectation is that the Rest screen is fully opaque while the others mostly transparent. 
3. In part because of the above issue, I'm not sure having this feature as a single global INI config value. I feel like this is the responsibility of a UI mod to add general "polish" 

I'd be more satisfied with a solution that:
1) does not touch the Rest window (let a mod choose whether this should allow transparency at all) 
2) uses a solution that applies universally for all existing dims (ex: DaggerfallEffectSettingsEditorWindow)

Interkarma:
>This was finally fully removed in this changelist by Interkarma: https://github.com/Interkarma/daggerfall-unity/commit/15eddf3416e6041b03c4615b14b5aee09d9f9a77

Key Kab! For some context around this change at the time in 2018, I was supporting InconsolableCellist working on a VR mod for DFU (start [here](https://forums.dfworkshop.net/viewtopic.php?p=13070#p13070) in forum topic).

In VR mode, the UI needs to be projected into the game world, e.g. onto a quad rect in front of player's face, for the user to point at with motion tracking controllers. The UI required a little refactoring to make this possible, as at the time it simply drew the UI directly on top game viewport after everything else. Search other changes made around May-Jul 2018 for more.

I can't recall the exact problem, but it reads like a compatibility problem after changing from straight GUI.DrawTexture to DrawTexture wrappers in DaggerfallUI that can change UI rendering pathway based on internal requirements. From memory, I also had to create a pathway due to differences in Mac UI rendering output - so be careful tinkering in there and make sure it's tested on all platforms.

Not sure why I didn't remove ScreenDimColor entirely, and I agree it's best to let the individual UI window or mod decide on this rather than a global setting. I likely meant to loop back to it later and forgot.

themrleon:
Thanks for the context @KABoissonneault and @Interkarma, reason I tried a PR instead of a mod is because I am not sure this can be achieved with a mod. But guess I will only stick with mods for now.

PR #2732, opened 2025-03-13
Fix: ItemData_v1.GetSaveData() also saves className now.

AncientGrief:
I added a custom DaggerfallUnityItem to the game. Works fine until I load a savegame with the item in my inventory (or in store shelves)

It's because the className is not saved and upon deserialization it will become a regular DaggerfallUnityItem with the default `UseItem`-Method and no custom icon (it will have no icon at all)

A workaround would be this (on my custom item class)
```csharp
public override ItemData_v1 GetSaveData()
{
    ItemData_v1 saveData = base.GetSaveData();
    saveData.className = GetType().ToString();
    
    return saveData;
}
```
But maybe it doesn't hurt to let DFU take care of this, because while deserializing it's already checking for `className`.

### DIFF
--- a/Assets/Scripts/Game/Items/DaggerfallUnityItem.cs
+++ b/Assets/Scripts/Game/Items/DaggerfallUnityItem.cs
@@ -813,6 +813,7 @@ namespace DaggerfallWorkshop.Game.Items
             data.timeForItemToDisappear = timeForItemToDisappear;
             data.timeHealthLeechLastUsed = timeHealthLeechLastUsed;
             data.artifactIndexBitfield = artifactIndexBitfield;
+            data.className = GetType().ToString();
 
             return data;
         }


KABoissonneault:
This is a virtual function. Derived classes are meant to override `GetSaveData()` and add the class name themselves.

Here are some mods adding items:
https://github.com/ajrb/dfunity-mods/blob/master/Archaeologists/Scripts/LocatorItem.cs#L121
https://github.com/magicono43/DFU-Mod_Repair-Tools/blob/c3ee81777e9458c14a46d6cd99471a85a4a07dc5/Scripts/ItemsRepairTools.cs#L208

A mod can already work around this issue that way, without a change to DFU.

I think there could be a way to make it easier on mods, to avoid having to do this override in every derived class. However, I am not certain about changing this behavior for existing mods, which might want to avoid setting the class name. 

I admit I am not certain why that hasn't been the solution from the start, as I cannot conceive of why someone would want the className null on their derived item. I'd rather not merge this anyway unless we can be certain no one would mind the change.

AncientGrief:
Totally fair. I just thought it was odd and cost me some time to figure out what was going on (I started modding DFU about 1.5 weeks ago, still not very familiar with the code).

PR #2735, opened 2025-04-09
Extends functionality for modders to modify horse riding speed.

tuncturel:
Hi, I've found out that it was not possible for modders to modify the horse riding speed in DFU. I took some liberty in refactoring this script and adding this exact functionality. I followed the same format with which walking and running speeds were modified to keep my addition predictable and in line with the existing material. I also took a bit of liberty in renaming a few things, removing some possibly redundant comments, rewriting missing/incorrect comments as well. I made sure none of my changes impacted existing mods by keeping everything backwards compatible. I did some light testing to see if things are working as intended and they seemed okay but would love it if someone took the time to test it thoroughly. I'm by no means a stranger to Unity or coding with C#. I got many years under my belt but this is my first contribution in a public repo ever so please let me know if something doesn't make sense or if you have any input. I'm just a guy who wants to mod his horse to ride faster. 🐴


n-ubo (commented):
LGTM

tuncturel:
Below is the download link for the Unity package of my sample mod. You can click the link, unzip it and unpack the unity package to the Assets>Game>Mods folder of your DFU Unity project. I'm hoping this can help with the testing process for those who are interested in giving it a try.

[Click this link to download FasterHorseRidingMod.zip](https://github.com/user-attachments/files/19673668/FasterHorseRidingMod.zip)

I have also provided a snippet from the unity package below so that you can see how the mod would utilize the functionality I've written in the merge request.


```
using DaggerfallWorkshop.Game;
using DaggerfallWorkshop.Game.Utility.ModSupport;
using DaggerfallWorkshop.Game.Utility.ModSupport.ModSettings;
using UnityEngine;


namespace FasterHorseRiding
{
    public class FasterHorseRidingModManager : MonoBehaviour
    {
        static Mod s_Mod;
        float m_HorseSpeedMultiplier = 1.0f;
        PlayerSpeedChanger m_PlayerSpeedChanger;
        PlayerMotor m_PlayerMotor = null;
        string m_SpeedModUID = null;

        [Invoke(StateManager.StateTypes.Game, 0)]
        public static void Init(InitParams initParams)
        {
            s_Mod = initParams.Mod;
            GameObject modGameObject = new GameObject(s_Mod.Title);
            modGameObject.AddComponent<FasterHorseRidingModManager>();
        }

        void Awake()
        {
            s_Mod = ModManager.Instance.GetMod("FasterHorseRiding");
            ModSettings modSettings = s_Mod.GetSettings();
            s_Mod.IsReady = true;
        }

        void Start()
        {
            Initialize();
        }

        void Initialize()
        {
            m_PlayerMotor = GameManager.Instance.PlayerMotor;
            m_HorseSpeedMultiplier = s_Mod.GetSettings().GetValue<float>("Main Settings", "Horse Speed Multiplier");
            m_PlayerSpeedChanger = GameManager.Instance.PlayerObject.GetComponent<PlayerSpeedChanger>();
            m_PlayerSpeedChanger.AddRideSpeedMod(out m_SpeedModUID, m_HorseSpeedMultiplier, true);
            Debug.Log("FasterHorseRidingModManager: Added ride speed of x" + m_HorseSpeedMultiplier.ToString() + " mod with UID: " + m_SpeedModUID);
        }

        void OnDestroy()
        {
            m_PlayerSpeedChanger.RemoveSpeedMod(m_SpeedModUID);
        }
    }
}
```

KABoissonneault:
@ajrb Was that really not possible before? I know you've played with something similar, like the sprint you have in R&R. Maybe this is different, I'm not sure

tuncturel:
Hi @KABoissonneault , if it's possible to do it another way then I'm happy to consider that.👍 I did talk to other modders on the Lysandus' Tomb Discord working on a mod that increases the variety of rides among other things. They told me the following: 

> "We've only been able to make the horse speed slower. We still can't make it faster than its vanilla speed. The other aspect we've been able to modify is the "run" or "gallop" speed."

I wonder if @ajrb can weigh on it when he has the time. I'm always happy to delete or modify the pull request, of course.

ajrb:
Hi, thanks for being patient and waiting for me to respond. The PlayerSpeedChanger class was not one I was veyr involved in, I only added changes to support the horse speed. This was based on the player speeds and thus the player speed stat - hopefully matching classic DF for horse and cart as best we could. I always considered that a mod of movement speed would operate on the base speeds. Anyway I have no problems with these changes, however I will add some review comments.

Assets/Scripts/Game/Player/PlayerSpeedChanger.cs
```diff
@@ -30,16 +30,20 @@ public class PlayerSpeedChanger : MonoBehaviour
         public const float classicToUnitySpeedUnitRatio = 39.5f; // was estimated from comparing a walk over the same distance in classic and DF Unity
         public const float dfWalkBase = 150f;
         private const float dfCrouchBase = 50f;
-        private const float dfRideBase = dfWalkBase + 225f;
-        private const float dfCartBase = dfWalkBase + 100f;
+        private const float dfRideBase = 375f;
+        private const float dfCartBase = 250f;
```
ajrb, line 34: Why change these definitions? Making this change would lose the context that the base speeds are based on the difference to the walk speed base.

Assets/Scripts/Game/Player/PlayerSpeedChanger.cs
```diff
@@ -148,15 +158,20 @@ public float GetBaseSpeed()
...
             {
-                float rideSpeed = (GameManager.Instance.TransportManager.TransportMode == TransportModes.Cart) ? dfCartBase : dfRideBase;
-                baseSpeed = (playerSpeed + rideSpeed) / classicToUnitySpeedUnitRatio;
+                float currentRideBase = 0;
+                bool playerIsRidingHorseWithNoCart = GameManager.Instance.TransportManager.TransportMode == TransportModes.Horse;
+                if (playerIsRidingHorseWithNoCart)
+                    currentRideBase = dfRideBase;
+                else
+                    currentRideBase = dfCartBase;
+                baseSpeed = (playerSpeed + currentRideBase) / classicToUnitySpeedUnitRatio;
```
ajrb, line 174: I am not sure why this code is being changed.. I can't see any difference in what it does, it's just more verbose and in my eyes less easy to see what is being done. I might be biased since these lines were added by me, happy to take a consensus if many feel it's better written the new way, but otherwise I'd rather this change is removed.

ajrb:
> @ajrb Was that really not possible before? I know you've played with something similar, like the sprint you have in R&R. Maybe this is different, I'm not sure

The speeds are based on normal char speeds with modifiers for horse/cart, normally sprinting is disabled when riding but R&R enables it when riding a horse.

Assets/Scripts/Game/Player/PlayerSpeedChanger.cs
tuncturel: Hi, this was purely a personal taste change and **I think it can be discarded as it has no functional impact**. These values are constants meanings they will never change during gameplay (if I'm not mistaken). Since they were not multiplied by the walk speed and were simply used in an addition I preferred to see them as their actual value instead of having to go and look at the walk speed.

Assets/Scripts/Game/Player/PlayerSpeedChanger.cs
tuncturel: Just making it simple stupid and more 'verbose' as you put it. I like reading code like prose and don't mind extra steps sometimes. I think it can make it easier to grasp and consume for others as well. **I believe no functional change in here so please feel free to pick the version you like.** 👍

Assets/Scripts/Game/Player/PlayerSpeedChanger.cs
Daneel53: I think it's better to keep the lines than explain the context.

Assets/Scripts/Game/Player/PlayerSpeedChanger.cs
Daneel53: ajrb line uses a syntax that is more in the way people code in C# today.
We have to get used of it.

Daneel53 (dismissed):
Apart from the two conversations where finally the result is the same, this PR was reviewed and should work as expected.

Ah, being the last editor, I can't do an approval review...

Assets/Scripts/Game/Player/PlayerSpeedChanger.cs
```diff
@@ -216,14 +231,39 @@ public bool AddRunSpeedMod(out string UID, float speedModifier = 0, bool refresh
...
         /// </summary>
-        /// <param name="speedModifier">the amount to change players base walk speed by percentages. AKA, .75 will lower player movement by 25%. Using 0 or negatives will do nothing but return null.</param>
         /// <param name="UID">The Unique Universal ID created and provided when original value was added to dictionary. Store this value to reference your speed modifier later.</param>
-        /// <param name="refreshWalkSpeed">will cause routine to also update the player speed using the list to sequentially multiply the current base value by the list modifier values.</param>
-        /// <returns></returns>
-        public bool RemoveSpeedMod(string UID, bool removeRunSpeed = false, bool refreshSpeed = true)
+        /// <param name="refreshSpeed">will cause routine to also update the player speed using the list to sequentially multiply the current base value by the list modifier values.</param>
+        /// <returns></returns>   
+        public bool RemoveSpeedMod(string UID, bool refreshSpeed = true)
```
KABoissonneault: I don't think I can accept removing a parameter from a public interface. This will break mods that relied on this function parameter.
I think the parameter should, unfortunately, be kept and do nothing instead.

@Daneel53 You think you could push that change? Just add the bool here with the same name, nothing else needs to change

KABoissonneault (changes requested):
Review is mostly good other than my one request

Assets/Scripts/Game/Player/PlayerSpeedChanger.cs
Daneel53: @KABoissonneault  Done

PR #2741, opened 2025-05-22
Fix fatigue bar vital

TwelveEyes:
Fixes `fatigueBar.Amount` erroneously being set instead of `fatigueBarGain.Amount`.

### DIFF
--- a/Assets/Scripts/Game/UserInterface/HUDVitals.cs
+++ b/Assets/Scripts/Game/UserInterface/HUDVitals.cs
@@ -4,8 +4,8 @@
 // License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
 // Source Code:     https://github.com/Interkarma/daggerfall-unity
 // Original Author: Gavin Clayton (interkarma@dfworkshop.net)
-// Contributors:    
-// 
+// Contributors:
+//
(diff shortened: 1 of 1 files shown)


KABoissonneault:
Excuse me, can you explain what issue this is fixing exactly? What difference will the player observe? Is this only relevant for modders? 
It does seem like it might be fixing something, but I'd like to know what to test before I can review this

TwelveEyes:
It simply fixes the way the fatigue bar updates so it matches the health and magicka bars. Without this fix the fatigue bar change upon gain/loss is messed up.

https://github.com/Interkarma/daggerfall-unity/blob/master/Assets/Scripts/Game/UserInterface/HUDVitals.cs#L297
https://github.com/Interkarma/daggerfall-unity/blob/master/Assets/Scripts/Game/UserInterface/HUDVitals.cs#L344

KABoissonneault:
Okay, so if I cast Heal Fatigue and compare it to other vitals, it should look "off". I would have liked more details but this will have to do I suppose.

Daneel53 (approved):
A simple line of code, already discussed and approved.

MeteoricDragon:
Thanks for fixing this!

PR #2742, opened 2025-06-04
Fix PlaceNpc action not serializing marker

### LINKED ISSUE #2700: 1.1.1: Quest S0000501, Lord K'avar Part 3 can't spawn enemy in correct marker
**Describe the issue**

Quest S0000501 tries to place an enemy and an NPC at the same marker in a dungeon, at two different times. The NPC spawns in that marker, but the enemy gets sent to a completely different marker in the dungeon. The quest file places both entities at quest marker 3 in the dungeon, so the quest script has no error.

[https://github.com/Interkarma/daggerfall-unity/blob/21f0a0f31170cb49de9810f8ba4c24d5207dc285/Assets/StreamingAssets/Quests/S0000501.txt](https://github.com/Interkarma/daggerfall-unity/blob/21f0a0f31170cb49de9810f8ba4c24d5207dc285/Assets/StreamingAssets/Quests/S0000501.txt)

**To Reproduce**

1. Start quest S0000501.
2. Find Lord K'avar in the dungeon and click on him.
3. His NPC sprite will disappear, but his enemy form will not be spawned there. It will be spawned in a different quest marker far away.

**Expected behavior**

The enemy should spawn in the same marker as the NPC sprite (effectively replacing it).

 - Version 1.1.1

**Additional context**

The spawn point used by DFU is different from DOS. I don't know why this is. This is less important, of course.

Other quests have used the same marker twice, such as the vanilla switcheroo quest.

TwelveEyes:
Marker indexes for the `PlaceNpc` quest action are not properly serialized, causing quest npcs to be placed at a random marker when the action is loaded. This is quite obvious in the quest `Lord K'avar Part III`, where the Lord K'avar flat could be located at a incorrect marker, and the foe at the correct one, causing K'avar to seemingly teleport across the dungeon.

Also edited `PlaceFoe`'s `AssignQuestResource` pattern to match that of `PlaceNpc`'s `AssignQuestResource` for consistency.

Fixes #2700

### DIFF
--- a/Assets/Scripts/Game/Questing/Actions/PlaceFoe.cs
+++ b/Assets/Scripts/Game/Questing/Actions/PlaceFoe.cs
@@ -4,8 +4,8 @@
 // License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
 // Source Code:     https://github.com/Interkarma/daggerfall-unity
 // Original Author: Gavin Clayton (interkarma@dfworkshop.net)
-// Contributors:    
-// 
+// Contributors:
+//
 // Notes:
 //
 
@@ -79,7 +79,7 @@ namespace DaggerfallWorkshop.Game.Questing.Actions
             }
 
             // Assign Foe to Place
-            place.AssignQuestResource(foeSymbol, marker);
+            place.AssignQuestResource(foe.Symbol, marker);
 
             SetComplete();
         }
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Game/Questing/Actions/PlaceFoe.cs
```diff
@@ -79,7 +79,7 @@ public override void Update(Task caller)
             }
 
             // Assign Foe to Place
-            place.AssignQuestResource(foeSymbol, marker);
+            place.AssignQuestResource(foe.Symbol, marker);
```
KABoissonneault, line 82: Just curious, is there any way `foeSymbol` could differ from `foe.Symbol`?

Daneel53 (approved):
Already discussed.
Was into Test2 and into Test3 since the beginning.

Issue #2262, opened 2021-10-27
0.13.1: Quest item does not become "permanent" upon enemy death


JayH2971:
Custom quests that attempt to convert a quest item into "permanent" state will not achieve it if "permanent" is linked to a "killed foe" condition. The following quest script is used:

> _kill2_ task:
>    killed _frenemy_
>    make _reward_ permanent
    
When the enemy dies, the item retains its green background.

This zip file includes Save888, which requires you to kill the enemy at the end of the hallway who holds the item, as well as the quest file.
[QuestItemStaysGreen.zip](https://github.com/Interkarma/daggerfall-unity/files/7423511/QuestItemStaysGreen.zip)

BadLuckBurt:
The UID of the quest item has changed by the time that EnemyDeath.CompleteDeath() starts to transfer the enemy's inventory to the lootpile. The quest UID is still intact so I'm trying to track down where the item's UID changes atm. 

Depending on what I find, the solution will either be to update the item's UID in the quest action or make sure the item's UID doesn't change.

```
EnemyDeath.CompleteDeath() - found quest item with item.UID: 33569042

ItemCollection.TransferAll() - Transferring quest item with UID: 33569042 
ItemCollection.TransferAll() - Quest UID: 33568744 

QuestAction MakePermanent - Item UID = 33568745 
QuestAction MakePermanent - Quest UID = 33568744
```

Interkarma:
The problem here is how the virtual quest Item lifecycle is handled relative to the DaggerfallUnityItem. In some cases the DaggerfallUnityItem tracked by Item is transferred and in some cases it's reinstantiated/cloned with a new UID. Depending on process this can lead to permanent status becoming desynced.
As this could happen in a few cases, and desynced permanent status can already be in save data (like in Jay's save above), I've opted to fix this in a low-impact way by fixing permanent status when player picks up item.
The item on slain foe will still have green background in corpse loot container, but it will be patched back to permanent when player picks up item. Tested in save attached to issue.
This should workaround most (if not all) cases of this problem and not cause any regressions.

PR #2745, opened 2025-06-05
Fix for possible zombie items from MakePermanent action

### LINKED ISSUE #2720: Quest action "make X permanent" refuses to trigger in S0000503
There's a problem in Former Student Part III with the quest action "Make X permanent," which is supposed to remove the quest condition from an item (remove green background, make droppable). It's not functioning here, and I don't know why. Sometimes it works, and sometimes not. These screenshots show me having the quest items in my inventory while no quests are active, which is supposed to be impossible. Since the quest is properly scripted and all events light up correctly, I can't fix it more than that.

Screenshot at 2025-01-24 09-36-04
Screenshot at 2025-01-24 09-36-08

_Originally posted by @JayH2971 in https://github.com/Interkarma/daggerfall-unity/issues/2719#issuecomment-2612948620_

TwelveEyes:
This was somewhat [addressed](https://github.com/Interkarma/daggerfall-unity/issues/2262#issuecomment-1261602732) in the past, but it only resyncs permanency between the virtual quest item and the inventory item with item transfer (e.g. foe has quest items; those quest items are made permanent; foe is killed and player loots those quest items). It doesn't cover items already in the player inventory when a `MakePermanent` action happens, potentially leading to a desynced state — the dreaded "zombie" item.

~~This workaround should fix the issue in a low impact way by resyncing permanency when the inventory window is setup and when it refreshes.~~

Changed to be much cleaner, though it won't fix bugged items already made permanent in saves retroactively.

Fixes #2720

### DIFF
--- a/Assets/Scripts/Game/Questing/Item.cs
+++ b/Assets/Scripts/Game/Questing/Item.cs
@@ -5,7 +5,7 @@
 // Source Code:     https://github.com/Interkarma/daggerfall-unity
 // Original Author: Lypyl (lypyldf@gmail.com), Gavin Clayton (interkarma@dfworkshop.net)
 // Contributors:    Hazelnut
-// 
+//
 // Notes:
 //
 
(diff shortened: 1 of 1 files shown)


Daneel53 (approved):
Was into Test2 and into Test3 since the beginning. No remark since then.

PR #2746, opened 2025-06-06
Fix run speed not being reduced while crouched

### LINKED ISSUE #2670: You can run at full speed while crouched
**Describe the issue**

If you press the run key  while crouched, you'll move at full running speed, same as if you were standing.

**To Reproduce**

1. Crouch
2. Walk forward. Notice the speed
3. Press your run toggle key
4. Notice the huge speed boost

**Expected behavior**

In classic, crouching + running seems faster than crouching alone, but much slower than running alone.
I suspect running applies the same speed multiplier whether you're standing or crouched, but I haven't measured to be sure. 

 - OS: Linux
 - Version 1.1.1

TwelveEyes:
Addresses running while crouched being the same speed as running while standing.

Fixes #2670


KABoissonneault:
Thank you, turns out we never turned petchema's commit into a PR.

Do you mind matching the formatting we used for https://github.com/petchema/daggerfall-unity/commit/26ac2a7f728bb16d63c70b7aaecd670ca848997f ?
I find using this many embedded conditional expressions harder to read

TwelveEyes:
> Do you mind matching the formatting we used for https://github.com/petchema/daggerfall-unity/commit/26ac2a7f728bb16d63c70b7aaecd670ca848997f ?

I don't mind changing it. Done.

petchema (approved):

PR #2748, opened 2025-06-07
Fix static NPCs in houses not being hidden at night

### LINKED ISSUE #2662: Non-quest Billboard NPCs in locked buildings don't disable at night
**Describe the issue**

when breaking into buildings, the normal npcs are still active.  In DOS, building NPCs are disabled like the shop npcs when breaking and entering at night, to simulate npcs asleep or away.

**To Reproduce**

break into any residence when the door is locked at night.  the house npc is still active.

**Expected behavior**

non-quest npcs should be disabled when the residence is locked at night so house appears empty.  This behavior is already in the shops and guild halls.

**Screenshots and Logs**

n/a

 - OS: Windows
 - Version 1.1.1

**Additional context**

Unsure if non-quest npcs should be disabled during quests at night at quest location since the residence is usually not locked.  it may make sense for thief quests.
not a high priority, but immersion breaking when thieving that the house npc is active and staring at you.

TwelveEyes:
Houses were being omitted from the conditional to disable NPCs at night. This adds building types `House1` to `House4` to the NPC disable/enable functions. Also includes new entries to the open/closed hours arrays for `House2` to `Ship`.

Fixes #2662

### DIFF
--- a/Assets/Scripts/Game/PlayerActivate.cs
+++ b/Assets/Scripts/Game/PlayerActivate.cs
@@ -5,7 +5,7 @@
 // Source Code:     https://github.com/Interkarma/daggerfall-unity
 // Original Author: Gavin Clayton (interkarma@dfworkshop.net)
 // Contributors:    Allofich, Numidium, TheLacus
-// 
+//
 // Notes:
 //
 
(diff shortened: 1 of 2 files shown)


Daneel53 (approved):

KABoissonneault (approved):
I admit the code was difficult to review, but I think this is indeed free of possible regressions. I do appreciate the `openHours`/`closeHours` arrays covering more building types given we never check if the buildingType fits into the array.

It's not a full solution for the concerns listed in the TODO at line 1286 of PlayerActivate.cs, but it's any improvement.

PR #2749, opened 2025-06-08
Allow looting from building types House5, House6, and Town23

### LINKED ISSUE #2648: Town 23 building type - Stealable containers do not work
The Town 23 building type (City walls, towers gates) cannot have containers that can be stolen from. I'm aware that in DOS, those buildings are empty and devoid of any containers to steal from and that, as such, this concerns mostly the possibility for players to steal from containers I've added in Detailed City Walls. 

According to Magicono, it could be because Town 23 is excluded from being lootable in DFLocation.cs and DaggerfallLoot as pictured below.

TwelveEyes:
Adds `House5`, `House6`, and `Town23` to the private property loot byte arrays and expands the conditional check to include those building types.

Fixes #2648


Daneel53 (approved):
Seems fine to me.

PR #2750, opened 2025-06-08
Add city wall automap display name

### LINKED ISSUE #2649: Town 23 building type - "City Wall" not shown on automap
**Describe the issue**

In DOS, the city walls are marked on the automap when the PC enters one of their buildings. In DFU, they're not identified on the automap. 
A DOS save file and a screenshot are provided.

[SAVE3.zip](https://github.com/Interkarma/daggerfall-unity/files/15206352/SAVE3.zip)

DOS City Wall

TwelveEyes:
Adds `Town23` to the building type switch within `GenerateBuildingName()`. After using info on or entering the building it will be marked as "City Wall".

This was a bit of a headache to PR, as the `Internal_Strings.csv` file is a horrible mix of `CRLF` and `LF` line endings. Had to fight git and my editor so the commit diff wasn't 500+ line changes.

Fixes #2649


Daneel53 (approved):
Seems fine to me.

KABoissonneault:
I'm not sure if the City Wall should be a singleton entry on the Automap, but better not spam for now

PR #2179, opened 2021-07-16
Moved absorb logic from EntityEffectManager to FormulaHelper

dakoller123:
So it can be easily overriden by mods

PR #2752, opened 2025-06-09
Moved absorb logic from EntityEffectManager to FormulaHelper

dakoller123:
Hey Guys!

I'd like to attempt another PR regarding spell absorb. 4 years ago it was closed due to feature freeze. Today I re-forked unity and re-did the whole change from scratch, I've also made an experimental dfumod that overrid the absorb effect.

What has changed: 
- Moved TryAbsorbtion method from EntityEffectManager to FormulaHelper, alongside it's helper methods
- TryAbsorption now returns the absorbed spell point count instead of a bool and an out variable
- TryAbsorption has the target caster entity and the absorb incumbent effects as function parameters now so its side effect-free
- Tiny refactor around the usage of TryAbsorb, the same 'if bundletype is spell' was reused 3 times

What was out of scope:
- Moving Reflection and Resistance to the FormulaHelper in a similar fashion. Once this PR is approved I might do that as well

How it was tested:
- I built my own binaries from my fork and used an experimental version of Mighty Magick mod where I overrid the formula. It all worked as expected.

Link to the old PR for reference:
https://github.com/Interkarma/daggerfall-unity/pull/2179


Daneel53 (approved):
Following discussion in old PR 2179, it is a good change.
Present into Test3, no remark since then.

PR #2753, opened 2025-06-10
Restore letter signoff macro expansion for names

TwelveEyes:
Restores macro expansions (except location name macros) in letter signoff inside the hover tooltip.

Interkarma [disabled](https://github.com/Interkarma/daggerfall-unity/commit/319e5c0ef0f286e243d7fef11e14e7541f335c69) macro expansion for letter signoffs because location data lookup could potentially cause serious performance degradation from world reloads ([dfworkshop thread source](https://forums.dfworkshop.net/viewtopic.php?p=66826#p66826)).

This restores letter signoff macro expansion for all macros _except_ location based ones, so it shouldn't impact performance.


KABoissonneault:
Interesting. I was interested in a solution for these, since I do think it was a common issue. Hopefully we can get this tested before 1.2

Daneel53 (approved):

Daneel53 (approved):
Seems fine to me.
Into Test3 since the beginning, no remark since then.

PR #2755, opened 2025-06-12
Fix loss of mage light dlight while spell is active

TwelveEyes:
While mage light is active, going through a terrain change (exterior <---> interior or using the teleport spell/mage guild service) would cause loss of the dynamic light. The only way to get it back is to wait until the spell expires (or dispel it manually) and recast — a recast while the spell is active won't do it. This simply checks if the spell lost the light somehow and recreates it.


Daneel53 (approved):
Into Test3 since a long time. No remark after use.

PR #2756, opened 2025-06-13
Push inventory window directly instead of through message queue

TwelveEyes:
Occasionally activating a loot container (e.g. a loot pile, a corpse, a store shelf, a quest reward, etc) would open the inventory and the `lootTarget` would somehow be null — the inventory would open but the loot column on the right would just be the ground instead of the loot container. It seems to always happen upon the very first activation of a loot container after a load of a save game.

The cause may be a possible race condition between the `lootTarget` being set and the UI message queue being processed. Pushing the UI window directly instead of going through the UI message queue seems to solve this.

Dungeon wagon access prompts and knightly order armor reward prompts have been changed as well, just in case this bug affects them too.


Daneel53 (approved):
Already discussed.
Put into Test3, no remark since then.

Issue #2740, opened 2025-05-17
[Modding Support] Custom Item support for Cloak slots


RedRoryOTheGlen:
**ISSUE**
In the current build of DFU, it is impossible to set a custom item to be equipped in the cloak slot.

If any item is equipped as a cloak in either slot, the PaperDollRenderer will try to render the cloak interior.
![Image](https://github.com/user-attachments/assets/83b54934-b614-49e8-98e2-0749d29d683b)

To get the cloak interior ImageData it will first check if the item is a cloak. If not, it will return a blank ImageData.
![Image](https://github.com/user-attachments/assets/98c3cc0c-a034-4c4c-ac25-a7a5468ffe23)

First problem is that it will only recognize the vanilla Casual and Formal Cloak items as Cloaks. So any other item (even if its a cloak, like a scarf or such) will not be able to have any kind of Cloak Interior.
![Image](https://github.com/user-attachments/assets/0b4e8b9f-b87e-40a0-9e8e-69f505279f40)

Continuing on. With ImageData in hand, it will attempt to call DrawTexture with the blank ImageData and immediately throw an exception due to ImageData's contents being null. This results in the paperdoll being unable to refresh.
![Image](https://github.com/user-attachments/assets/57ed158e-4cb9-4ae8-bc38-c13eba7c3df0)
![Image](https://github.com/user-attachments/assets/b84201ea-b178-43a8-84e2-ce5f273435d7)

**EXPECTED BEHAVIOR**
First:
I am not quite sure how, but some way of being able to dictate whether an item has a Cloak Interior texture without having to replace the vanilla cloaks would be nice.

Second:
Adding a null check for the ImageData to either the DrawTexture or BlitCloakInterior methods completely avoids this issue. 

![Image](https://github.com/user-attachments/assets/bc62c005-3bfc-4a6c-bc21-c223d7cc5804)
![Image](https://github.com/user-attachments/assets/b67b0bc7-6dc8-4f84-a465-84ee5b807068)

No exception is thrown and the paperdoll updates as normal. Attached is my testing of equipping clothes to the cloak slot to enable layering chest clothing:
![Image](https://github.com/user-attachments/assets/f76b2bb0-fdcb-41f9-9523-57ceca4dd76c)
![Image](https://github.com/user-attachments/assets/3fe50fc1-3ee7-4d8a-a3f1-18ca8beebc8a)

**OTHERS**
[...]

PR #2757, opened 2025-06-14
Null check ImageData for BlitCloakInterior

TwelveEyes:
Fixes the paperdoll throwing a null exception trying to blit a cloak interior for a custom item that goes into the cloak slot and doesn't count as one of the hard coded cloaks (`Formal_cloak` or `Casual_cloak`).

Addresses part of #2740.


petchema (dismissed):

TwelveEyes:
> Maybe we could have put those null checks inside `DrawTexture` itself, to make sure all paths are covered

Makes sense to me. Done.

KABoissonneault (approved):

RedRoryOTheGlen:
Hello, I've been trying this out in the test build and when equipping a custom item in the Cloak slot, an exception is still thrown in the overload here:

image

I suppose it is because since **srcImage** is null, then **srcImage.offset** is also null.

Adding the null check to the overload prevents this exception from occurring.

image

image

Thanks for your work!

PR #2759, opened 2025-06-15
Fix custom item type being lost upon splitting item stack

### LINKED ISSUE #2758: Manual Item Stack Splitting Behavior Does Not Properly Account For Custom/Modded Items
**Describe the issue**

If using the "manual stack split" behavior in the inventory, necessary attributes for custom/modded items outside the vanilla items won't be carried over properly or at all.

This causes functions for the child stack to no longer work, such as not being able to be restacked with the original parent stack (or any other stack for that matter.) of otherwise identical items. As well as breaking any additional functionality that custom item may have had, such as any "OnUse" functions attached to it normally.

**To Reproduce**

Have any mod that adds custom items that also stack. When you manually split said stack, the child will no longer function properly, as a result of the "className" attribute now being "null" when looking at the save-data for said child item stack.

**Proposed Solution**

I think the best way to potentially correct this behavior would be to simply update the "SplitStack" method's current way it recreates the child stack to use the ItemBuilder class' method "CreateItem" rather than how it currently does it.

Specifically changing this:

DaggerfallUnityItem pickedItems = new DaggerfallUnityItem(stack);

To This:

DaggerfallUnityItem pickedItems = ItemBuilder.CreateItem(stack.ItemGroup, stack.TemplateIndex);

Image

TwelveEyes:
Upon splitting a stack `new DaggerfallUnityItem()` is called instead of `ItemBuilder.CreateItem()`, so any custom items lose their custom class type as `(DaggerfallUnityItem)Activator.CreateInstance(itemClassType)` is not called. This fixes that so custom items properly retain their type upon split.

Fixes #2758

### DIFF
--- a/Assets/Scripts/Game/Items/ItemCollection.cs
+++ b/Assets/Scripts/Game/Items/ItemCollection.cs
@@ -264,7 +264,7 @@ namespace DaggerfallWorkshop.Game.Items
                 return null;
             if (numberToPick == stack.stackCount)
                 return stack;
-            DaggerfallUnityItem pickedItems = new DaggerfallUnityItem(stack);
+            DaggerfallUnityItem pickedItems = ItemBuilder.CreateItem(stack.ItemGroup, stack.TemplateIndex);
             pickedItems.stackCount = numberToPick;
             AddItem(pickedItems, noStack: true);
             stack.stackCount -= numberToPick;


Daneel53 (approved):
Put into Test3, no remark since then.

PR #2762, opened 2025-07-14
Use a snow emitter around the camera

petchema:
Specially when going down hills, it's obvious that the snow emitter is above the player and snow particles can't keep up with player's movement.

Simplest fix is to use an emitter around the camera instead of above; It's not perfect but greatly improves the visual effect worse cases.

Test:

    set_weather 6
    location 3

(maybe use location 3 multiple times until you spawn on the sea side of Sentinel)
Then descend in the direction of the sea, preferably on horse

Source:
https://gamedev.stackexchange.com/questions/198424/in-a-top-down-3d-game-how-can-i-simulate-snow-without-actually-covering-entire


Daneel53 (approved):
Kab was OK, present into Test3, no remark.

PR #2770, opened 2025-10-25
#2768 and #2767 (fixed bugs) and some optimization

### LINKED ISSUE #2767: WeaponManager attack raycast layer mask collides with objects on the "Ignore Raycasts" layer
As per the title, the Layer Mask variable of the WeaponManager class only excludes the "Player" layer. This is how it is initialized in [WeaponManager](https://github.com/Interkarma/daggerfall-unity/blob/master/Assets/Scripts/Game/WeaponManager.cs#L199).
Image

Copying the layer mask to a public variable to view in the Inspector: 
Image

This results in objects on the "Ignore Raycast" layer (like corpses and lootpiles) to "block" the attack spherecast. The corpses aren't considered doors or static colliders, so the backup raycast never fires. This makes hitting enemies while fighting around corpses unreliable as you get no feedback on whether the attack just missed or it hit a corpse and vanished. Especially noticeable with small enemies like rats, but can also be problematic with the camera level due to the size of the spherecast.

Image

Theoretically, this can also mean that the automap marker models can block attacks, though I have not tried this yet.

**SOLUTION:**
Simply removing "Ignore Raycast" (and maybe "Automap" as well) from the included layers in the Layer Mask would fix this behavior.

### LINKED ISSUE #2768: Touch spell hit detection targets the player when aiming downwards.
Firstly, with a touch spell readied, the expected behavior is that you are unable to cast it unless you have a target entity in range. But in practice, if you look slightly downwards (around 15-30 degrees below level) the touch spell can be cast. 

Secondly, when casting touch spells on small enemies like rats where you have to look down to target them, you will commonly see the spell be cast, but the rat will not proc a spell save or any observable effect occur.

Using some debug gizmos with the touch spell targeting, I tried to figure out why this problem was occurring.

Casting a touch spell:
Image

Aiming downwards a bit more:
Image

Based on the **DaggerfallMissile.GetEntityTargetInTouchRange** [method](https://github.com/Interkarma/daggerfall-unity/blob/master/Assets/Scripts/Game/DaggerfallMissile.cs#L391), it seems like the Touch spell spherecast originates a little bit behind the AimPosition. If the player looks downwards by a certain margin, the Touch spell actually collides with the player's own collider and sets them as the target. The spell is then cast and fizzles because the caster cannot be the target of the Touch spell.

This leads to instances where the player is trying to cast a touch spell on a rat or similar enemy and the casting animation is firing but no discernable effect (like a spell save) occurs.

**SOLUTION:**
This only seems to be a major problem with the player experience so I think adding a Layer Mask to the spherecast excluding the Player layer (as well as Ignore Raycast and Automap like in this other [issue](https://github.com/Interkarma/daggerfall-unity/issues/2767)) would work if the player is the caster.

DanyilYedelkin:
Fix: #2768  — Touch spell 
Fix: #2767 — WeaponManager hit detection issues
Perf: Optimized DaggerfallMissile runtime (GC + CPU improvements)

### Bug Fixes
- **#2768** — Touch spell hit detection no longer targets the player when aiming downwards.
  - Added proper layer mask excluding Player, Ignore Raycast, and Automap layers.
  - Slightly nudged spherecast origin forward to prevent self-collision.
- **#2767** — WeaponManager attack layer mask no longer collides with objects on the "Ignore Raycast" or "Automap" layers.

### Optimization
- Replaced `Physics.OverlapSphere()` + `new List<>` with `Physics.OverlapSphereNonAlloc()` and a reusable buffer to eliminate heap allocations during AoE checks.
- Cached references to `GameManager`, `WeaponManager`, and frequently used components (`Collider`, `EnemySenses`, `EnemyAttack`, `CharacterController`) to reduce expensive `GetComponent<T>()` and singleton lookups.

### Impact
- Fixes inconsistent spell and melee hit detection.
- Eliminates GC spikes and microstutters during combat.
- Reduces CPU cost per missile instance, improving performance in large fights.

### Verification
- Tested touch spells, melee attacks, and AoE effects on multiple enemies.
- Verified consistent hit registration and unchanged gameplay behavior.

### DIFF
--- a/Assets/Scripts/Game/DaggerfallMissile.cs
+++ b/Assets/Scripts/Game/DaggerfallMissile.cs
@@ -55,6 +55,8 @@ namespace DaggerfallWorkshop.Game
 
         #region Fields
 
+        private static int TouchMask = -1;
+
         const int coldMissileArchive = 376;
         const int fireMissileArchive = 375;
         const int magicMissileArchive = 379;
@@ -390,15 +392,19 @@ namespace DaggerfallWorkshop.Game
 
         public static DaggerfallEntityBehaviour GetEntityTargetInTouchRange(Vector3 aimPosition, Vector3 aimDirection)
         {
-            // Fire ray along caster facing
-            // Origin point of ray is set back slightly to fix issue where strikes against target capsules touching caster capsule do not connect
+            // Nudge origin slightly forward to avoid intersecting the caster capsule when looking down
+            const float originNudge = 0.01f;
+            Vector3 origin = aimPosition + aimDirection * originNudge;
+
             RaycastHit hit;
-            aimPosition -= aimDirection * 0.1f;
-            Ray ray = new Ray(aimPosition, aimDirection);
-            if (Physics.SphereCast(ray, SphereCastRadius, out hit, TouchRange))
+            Ray ray = new Ray(origin, aimDirection);
+
+            if (Physics.SphereCast(ray, SphereCastRadius, out hit, TouchRange, GetTouchMask(), QueryTriggerInteraction.Ignore))
+            {
                 return hit.transform.GetComponent<DaggerfallEntityBehaviour>();
-            else
-                return null;
+            }
+
+            return null;
         }
 
         #endregion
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Game/DaggerfallMissile.cs
```diff
@@ -14,6 +14,7 @@
 using DaggerfallWorkshop.Utility;
 using DaggerfallWorkshop.Game.MagicAndEffects;
 using DaggerfallWorkshop.Game.Entity;
+using System.Runtime.InteropServices;
```
KABoissonneault: Curious, what is this used for?

KABoissonneault:
This PR was merged into #2781 and can now be found in f9382b6b1f86ba4cecec69f6dce6516e62401db2

PR #2777, opened 2026-02-04
Update startquest console command error message

Daneel53:
Change the error message of the console command startquest when the creation of a quest is not possible.

When launching a quest with console command startquest, it happens that the quest cannot be created because an instruction in then quest file cannot be done. The error message just tell that the quest file was not found, when in reality it was found but a parsing error occured.
The new error message is not anymore limited to "Quest file not found" and send to the file quest_log.txt for further explanation.


KABoissonneault (approved):
Simple enough

PR #2781, opened 2026-03-31
Raycast look-ahead for projectiles and Bounding Box-based weapon collision + fixes and optimizations.

RedRoryOTheGlen:
Implemented raycast-based collision for DaggerfallMissile from PR #2472 by [numidium](https://github.com/numidium) . Revised the layer mask switching and fixed the now-off-center arrow model resulting from the change. Also applies the layer mask to the Touch spell detection and tweaks the GetAimPosition method to prevent the player from targeting themselves with their own Touch spells addressing Issue #2768.

Additionally, incorporated the optimizations to DaggerfallMissile from PR #2770 by [DanyilYedelkin](https://github.com/DanyilYedelkin), taking into account the new raycast-based collision when needed. Original PR had an issue which applied the Player-excluding layer mask to all projectiles and area-of-effect regardless of caster which has been corrected. Also included the changes to WeaponManager that fixes the player melee attacks being blocked by colliders on the IgnoreRaycast layer (Issue #2767), as well as an optimization.

UPDATE:
Implemented the bounding box-based weapon collision for players. At the default settings, this does the following things when the player performs a melee attack:
1. Uses Physics.OverlapBox() to grab all colliders (using the WeaponManager layermask) in front of the player in an area based on the default weapon reach + spherecast radius.
2. For each collider, check if it is an entity
3. Check if collider's center is inside the camera's field-of-view
4. Check if line-of-sight to collider's center is unobstructed by environment and other entities (as well as within the weapon reach value)
5. Add collider to list (hitColliders)
6. After all overlapped colliders have been finished, perform WeaponDamage() on all colliders in hitColliders list
7. If hitColliders is empty, perform the old spherecast check to handle bashing doors and other actionable objects.

Two settings are included: **Hit Detection** and **Protect Friendlies and Neutrals**.

By default, **Hit Detection** is set to **Performance** and behaves as described above. If **Hit Detection** is set to **Quality**, then there are additional checks that happen in Steps 3 and 4 if their initial checks fail. There can be up to 4 tests checking if some other part of the target is in view or has clear LOS. If a test succeeds, the following tests are skipped. This lets the player hit entities that may be partially covered by the environment or other entities.

**Protect Friendlies and Neutrals** is enabled by default. While enabled, the check in Step 1 will skip wandering commoners and entities that are pacified, and these targets can only be damaged/killed by the old spherecast check in Step 7. This prevents attacking them by accident if the player is fighting an enemy while they're around.

Both settings can be accessed in the DFU Advanced Settings menu under Gameplay. They can also be changed mid-game from the Advanced Controls window.

For illustration:
image
image

The green capsule is the default DFU attack spherecast. The red box is the OverlapBox check. After the field-of-view, line-of-sight and weapon reach checks, the player's attack area is reduced to the semi-spherical shape inside the blue lines. In effect, it is somewhat like sweeping the spherecast across the whole of the player's view.

If the player increases the camera FOV through the settings, this will also increase the attack area but not the total weapon reach, for consistency:
image

### DIFF
--- a/Assets/Localization/StringTables/Internal_Settings Shared Data.asset
+++ b/Assets/Localization/StringTables/Internal_Settings Shared Data.asset
@@ -143,6 +143,18 @@ MonoBehaviour:
     m_Key: textureArrayInfo
     m_Metadata:
       m_Items: []
+  - m_Id: 392878225761951744
+    m_Key: meleeAttackDetectionModes
+    m_Metadata:
+      m_Items: []
+  - m_Id: 392920400377864192
+    m_Key: meleeAttackDetection
+    m_Metadata:
+      m_Items: []
+  - m_Id: 392920453876211712
+    m_Key: meleeAttackFriendlyProtection
+    m_Metadata:
+      m_Items: []
   m_Metadata:
     m_Items: []
   m_KeyGenerator:
--- a/Assets/Localization/StringTables/Internal_Settings_en.asset
+++ b/Assets/Localization/StringTables/Internal_Settings_en.asset
@@ -236,5 +236,19 @@ MonoBehaviour:
       and modding support
     m_Metadata:
       m_Items: []
+  - m_Id: 392878225761951744
+    m_Localized: 'Performance
+
+      Quality'
+    m_Metadata:
+      m_Items: []
+  - m_Id: 392920400377864192
+    m_Localized: Hit Detection
+    m_Metadata:
+      m_Items: []
+  - m_Id: 392920453876211712
+    m_Localized: Protect Friendlies and Neutrals
+    m_Metadata:
+      m_Items: []
   references:
     version: 1
--- a/Assets/Prefabs/Missiles/ArrowMissile.prefab
+++ b/Assets/Prefabs/Missiles/ArrowMissile.prefab
@@ -11,11 +11,8 @@ GameObject:
   - component: {fileID: 4928002457579466}
   - component: {fileID: 82343321333735398}
   - component: {fileID: 114835290752812540}
-  - component: {fileID: 135236093773449130}
   - component: {fileID: 108732554255048532}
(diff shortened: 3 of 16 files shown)


numidium (dismissed):

KABoissonneault:
Did you get to try it out numidium, or is that from reading the code? I do like that this PR replaces two others, I could probably knock down both, but it would probably help to have people - other than the author - test it

RedRoryOTheGlen:
I have been focusing on the Touch usage a bit more and last commit is supposed to fix the self-targeting issues still happening while moving and sprinting due to enabling Headbob and the player's SmoothFollower lagging behind the player collider.
Would definitely appreciate more people trying it out to see if things can be handled better.

RedRoryOTheGlen:
Additionally, since this already changes WeaponManager, I would like to try implementing bounding box based player melee attacks with multi-hit in this PR as well, if that would be acceptable?
If not, I can make a new PR for it later, but I would definitely use this PR as a base.

[Test footage of bounding box-based weapon collision](https://streamable.com/f4u470) Even with lots of playtime on DF, I still don't really know how Daggerfall did it, but I attempted to achieve the same feel with just an OverlapBox and Raycasts for a quick LOS check. Then I added some (optional) enhancements like friendly fire protection, radial reach and a complex-LOS check as QOL.

RedRoryOTheGlen:
After further testing on 1.2-Test3, I've implemented TOB-style radial reach check to prevent the player from attacking beyond the default DFU weapon reach value.

Updated the OP with some more info including the effect of the new check.

Daneel53 (commented):
This PR was deeply tested into 1.2 Test3 in the same time as Vwing's PR 2790.
Both authors confirm that all is now working as intended.
Both PR are ready to be integrated into the game.

Daneel53 (approved):

PR #2782, opened 2026-04-04
Replace WWW with UnityWebRequest for sound replacement

AncientGrief:
Updates the audio loading logic to use UnityWebRequestMultimedia instead of the deprecated WWW class. This implementation uses synchronous waiting to maintain compatibility with the existing architecture.

This implementation is based around how the whole sound caching in DFU works.

It would need a little bit more effort if going the full and correct asynchronous workflow but works for now and gets rid of the warning + might also fixes #2760 

Tested with .wav and .ogg replacing music and in-game sounds.

### DIFF
--- a/Assets/Scripts/Utility/AssetInjection/SoundReplacement.cs
+++ b/Assets/Scripts/Utility/AssetInjection/SoundReplacement.cs
@@ -5,7 +5,7 @@
 // Source Code:     https://github.com/Interkarma/daggerfall-unity
 // Original Author: TheLacus
 // Contributors:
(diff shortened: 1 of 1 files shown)


Daneel53:
I've tested the new code with both a song replacement with a .ogg file and a sound replacement with a .wav file, everything works as expected. Replace the proposed PR 2760 that can be closed.
Can be approved.

Daneel53 (approved):
Works with both Unity engines 2019 and 2022.
Will remove calls to obsolete API.

Issue #2737, opened 2025-05-03
City Watch enemies causing significant performance issues


RedRoryOTheGlen:
When there are many City Watch enemies present and hunting the player, performance will significantly decrease.

![Image](https://github.com/user-attachments/assets/4ceefbfb-a9c5-41b3-8eb2-22d0603cb563)

![Image](https://github.com/user-attachments/assets/ea9ea4f5-1c82-4784-92c6-d1f12667c6de)

This seems to be caused by each individual City Watch enemy's **EnemySounds.IsReady()** repeatedly calling **dfAudioSource.GetAudioClip()** twice every Fixed Update due to the City Watch having their **Move Sound** and **Attack Sound** variables set to **SoundClips.None**. Changing these values to SoundClips.EnemyMove or EnemyAttack either in EnemyBasics or through a mod accessing the static EnemyBasics.Enemies array results in a massive performance boost even when there are numerous City Watch engaging the player.

![Image](https://github.com/user-attachments/assets/5d718dab-5732-49f5-a96c-f992c093ed4e)

KABoissonneault:
Have you measured this issue with a game build? If you enable Development Build when creating a game build from the Editor, you can then play a normal game with profiling enabled (which you run from the editor but remotely connected to your game client).

I'm saying this because the editor has much, much worse performance on file access than the game itself. The editor loads from disk each time, while the game loads those files ahead of time.

I've observed a few times in the past people trying to rewrite everything for something that only affects the editor.

I'm not against a simple fix if it can make the editor better, but it's not as much of a priority.

RedRoryOTheGlen:
True enough. It definitely seems to be negligible in the standalone build. I've only noticed it happening in the editor because I usually avoid aggro'ing guards on my main install. I guess it's all good since I've already mitigated it on my end.

Thanks for your time!

RedRoryOTheGlen:
Just chiming back in here with a follow-up. The performance hit from the Guards seems to be extremely significant on lower-end and mobile devices, especially with the Android port. I sent reporters (there were at least 3, with [the most recent one](https://discord.com/channels/559842137374457876/1468424698974044193/1468424698974044193) being on the DFU modding Discord server) a small mod that simply sets the Move and Attack sounds to the same SoundClip as other human enemies on game start and it visibly improved their performance quite handily.

All that the mod does:
[...]

PR #2783, opened 2026-04-05
Replaced references to SoundClips.None in EnemyBasics class with appropriate SoundClips

RedRoryOTheGlen:
As title. Made the PR to address Issue #2737 myself since it's a really small change that only really affects low-end machines and mobile devices.


KABoissonneault (approved):
This is indeed a behavior I've observed, where trying to fetch None will prevent the cache from functioning and attempt to fetch every frame, so I agree with this change.

I admit I would have preferred that we fix EnemySounds to not try to play sounds if the value is None, but at least this fixes this particular issue

PR #2625, opened 2024-03-25
Daggerfall Proofreading Project, Part 1: String Files (Take Two)

Mentill:
This contains ~200 corrections to spelling & grammar errors/typos to the string files (excluding internal_locations.csv), all done by hand.

The Internal_Locations.csv file was copied to the Text folder so it can override the misspelled names normally loaded directly from MAPS.BSA. I didn't go through all ~15k entries, but rather a representative sample since they're procedurally-generated.

Some changes were made to improve consistency. In particular:

- the one message that refers to the Cart was changed to Wagon, to be consistent with every other message; the item name was left alone

- the Pickpocket skill was renamed to Pickpocketing to match the strings, and for consistency with Backstabbing and Lockpicking

- references to Critical Striking were changed to Critical Strike, as that's the skill's current name, which I feel sounds better despite being less-consistent with the other skill names

I wasn't sure how to edit the images in INVE00I0.IMG & INVE01I0.IMG, which contain the rendered 'Wagon' buttons for the inventory screen, or if it were allowable to include modified versions in DFU's loose texture folder. Otherwise, I would've changed every 'Wagon' reference to 'Cart', including that button. I left the Transport menu's 'Cart' button graphic unmodified for the same reason.

It was recommended that I split my project into parts submitted one at a time, so this first part is just the string files. Next will be the quests, of which I corrected typos in ~170 of them; if that's too many, I can split that, too.
I had improved the formatting of the books, but it seems another PR did a better job on that; I didn't proofread many of those, though.
Hopefully I did this all correctly this time, the file diffs all look good to me...

[strings notes.txt](https://github.com/Interkarma/daggerfall-unity/files/14745936/strings.notes.txt)

PR #2784, opened 2026-04-08
Update Editor to the latest LTS 2019.4.41f2

KABoissonneault:
While the DFU release has been updated to use secure binaries, the project is still setup on the vulnerable 2019.4.40f1 version. This simple commit changes the project to the latest LTS. The impact should be essentially none, beyond removing the warning signs on Unity Hub.

I've also removed a duplicate file introduced by PR #2625, it was adding noise since its .meta file was missing


Daneel53 (approved):
Internal_Locations.csv should not be there anyway.
And yes, it's time to update Unity engine 2019.

PR #2785, opened 2026-04-10
Added null srcImage check to the other DrawTexture overload.

RedRoryOTheGlen:
Extends #2757 to prevent the exception in #2740 from happening by calling the overload.

### DIFF
--- a/Assets/Scripts/Game/Utility/PaperDollRenderer.cs
+++ b/Assets/Scripts/Game/Utility/PaperDollRenderer.cs
@@ -237,7 +237,7 @@ namespace DaggerfallWorkshop.Game.Utility
 
         void DrawTexture(ImageData srcImage, DaggerfallUnityItem item = null)
         {
-            if (srcImage.texture == null)
+            if (srcImage.texture == null || srcImage.offset == null)
                 return;
 
             DrawTexture(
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Game/Utility/PaperDollRenderer.cs
```diff
@@ -237,6 +237,9 @@ public byte GetEquipIndex(int x, int y)
 
         void DrawTexture(ImageData srcImage, DaggerfallUnityItem item = null)
         {
+            if (srcImage.texture == null)
```
KABoissonneault: Question, why does this not check `srcImage` and `srcImage.offset` for null instead?

RedRoryOTheGlen: - **srcImage** can be not null since GetCloakInteriorImage() returns [a new uninitialized ImageData](https://github.com/Interkarma/daggerfall-unity/blob/master/Assets/Scripts/Game/Items/ItemHelper.cs#L709) if the item is not a cloak
- I wanted to keep it consistent with #2757 
- I suppose that there's no use-case for calling DrawTexture() if the **srcImage** has no texture so it's better to abort it here than later?

KABoissonneault: Fair enough about the first point, but the second point is irrelevant. The convention that matters here is checking for null before access. The third point is fine, you can check the texture for null even if it's not directly accessed, but for safety I'd wanna make sure any field that can be null is checked before access.
In short, you can have `if(srcImage.texture == null || srcImage.offset == null)`

RedRoryOTheGlen: Understood. I'll revise it as directed.

KABoissonneault (approved):

PR #2788, opened 2026-04-15
Suppress compilation warnings for CS0108 and CS0168

Daneel53:
Compilation warnings linked to the name of an inherited method (automap.cs) and data declared but not used (5 other files).
Code was updated so that these warnings are not anymore displayed.

### DIFF
--- a/Assets/Game/Addons/RmbBlockEditor/Persisted/BuildingTemplates.cs
+++ b/Assets/Game/Addons/RmbBlockEditor/Persisted/BuildingTemplates.cs
@@ -1,4 +1,4 @@
-﻿// Project:         Daggerfall Tools For Unity
+// Project:         Daggerfall Tools For Unity
 // Copyright:       Copyright (C) 2009-2022 Daggerfall Workshop
 // Web Site:        http://www.dfworkshop.net
 // License:         MIT License (http://www.opensource.org/licenses/mit-license.php)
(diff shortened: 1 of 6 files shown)


KABoissonneault (commented):
For the Exception ones, I would have rather add a `Debug.LogException(error);`

Daneel53:
See what you mean. Wiil try that.

Assets/Game/Addons/RmbBlockEditor/Persisted/BuildingTemplates.cs
```diff
@@ -1,15 +1,16 @@
...
 // Original Author: Podleron (podleron@gmail.com)
 
+using DaggerfallWorkshop.Utility.AssetInjection;
+using Newtonsoft.Json;
 using System;
 using System.Collections.Generic;
 using System.IO;
-using DaggerfallWorkshop.Utility.AssetInjection;
-using Newtonsoft.Json;
+using UnityEditor.PackageManager;
```
KABoissonneault: Is this necessary for the log function, or was that added by the IDE?

Daneel53: Don't really know from where this line comes.
Anyway, following Visual Studio this using is useless (the line is grayed), so we can remove it.

KABoissonneault (approved):

PR #2789, opened 2026-05-24
Performance Enhancement: GameObject Culling

Vwing:
### Performance Enhancement: GameObject Culling

This feature has been part of the Android fork for the past couple years. It greatly improves performance on mobile, and the performance gains are pretty great on desktop as well.

On my machine in the Wayrest dungeon with culling disabled, Profiler shows constant stuttering to 30fps and an occasional 15fps lag spike. With culling _enabled_, it stays closer to 60fps, occasionally spiking to 30fps. The remaining lag is due to AutoMapper, which is optimized in Android and I'd like to open a separate PR for that.

#### How it works

The CulledGameObjectManager disables Billboards, Foe Spawners, Enemies, Action Doors, Static NPCs, Loot, and Dungeon Blocks that are further than 2060 unscaled units away from the player. It does so by moving them under a disabled game object parent, so that their local enabled/disabled state is unchanged.

It culls these object types on separate frames, which helps smooth out any lag from enabling and disabling a ton of GameObjects all at once.

#### Implementation Notes

* added relevant 'Object Culling' setting to Video settings (defaulted to enabled)
* ActiveGameObjectDatabase now supports inactive object lookups (required for culling disabled objects, otherwise you can have e.g. enemies walking around in disabled dungeon blocks)
* billboards are now registered with the ActiveGameObjectDatabase (so they may also be culled; we only bother doing so when the player is outside)
* added CulledGameObjectManager to the game scene

### DIFF
--- a/Assets/Resources/defaults.ini.txt
+++ b/Assets/Resources/defaults.ini.txt
@@ -31,6 +31,7 @@ DungeonShadowDistance=20.0
 InteriorShadowDistance=30.0
 ExteriorShadowDistance=90.0
 EnableTextureArrays=True
+EnableObjectCulling=True
 RandomDungeonTextures=0
 
 [Effects]
--- a/Assets/Scenes/DaggerfallUnityGame.unity
+++ b/Assets/Scenes/DaggerfallUnityGame.unity
@@ -2894,3 +2894,78 @@ PrefabInstance:
       objectReference: {fileID: 1769263313}
     m_RemovedComponents: []
   m_SourcePrefab: {fileID: 100100000, guid: bbe5e15c9f6b3dc47bb9485a437750a0, type: 3}
+--- !u!114 &69466200359737372
+MonoBehaviour:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  m_GameObject: {fileID: 1834928754549345153}
+  m_Enabled: 1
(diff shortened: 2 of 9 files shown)


KABoissonneault:
Thank you for sharing your improvements back to the main repo. 

I admit at this stage, I'm a bit hesitant to add a new bunch of unknowns into the build, but this is also more in Daneel's hands, so that's his call whether he wants to add a new round of testing with your latest PRs.

I'm aware these have been played on Android for a long time, but it was also on another Unity version, and I'm not sure I can trust Unity to not have weird inconsistencies in runtime behavior between even minor versions.

Again, leaving this to Daneel.

Daneel53:
Will push this one into next Test3 so that we can test it.

Daneel53 (dismissed):
Pushed into Test3, no remark.

Vwing:
I changed the public method definitions for ActiveGameObjectDatabase.cs in the original implementation of this, which ended up breaking already-built mod compatibility, requiring a rebuild for affected mods.

Now I have changed them back to how they were, but with an overloaded method containing the 'includeInactive' bool this feature required. This should restore mod compatibility.

Vwing:
Squeezed in another fix.

Users on the Android version of DFU were reporting that platforms were invisible in the Mantellan Crux. Looking at the game data, it seems that those platforms extended below the assumed bounds of RDBs, which meant that they were culled even when you were standing on them.

So the fix, here, is to ignore vertical distance when culling dungeon blocks. The platforms should now be visible.

Daneel53:
New code was pushed into Test3-2029 and Test3-2022 since July 17.
No remark since then.

PR #2791, opened 2026-05-24
Performance Enhancement: Font rendering in a single drawcall

Vwing:
Both SDF and classic fonts were being rendered a single character at a time, causing potentially tens of thousands of draw calls in a single frame. I've reduced it down to a single draw call per text item.

Before (18k draw calls and 30 fps):
image

After (500 draw calls and 60 fps):
image

The old multi-call path is still present and called if and when an error ever occurs in the new path.

### DIFF
--- a/Assets/Scripts/Game/UserInterface/DaggerfallFont.cs
+++ b/Assets/Scripts/Game/UserInterface/DaggerfallFont.cs
@@ -135,6 +135,16 @@ namespace DaggerfallWorkshop.Game.UserInterface
             get { return sdfFontInfo.Value; }
         }
 
+        public Texture2D AtlasTexture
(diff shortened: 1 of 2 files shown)


Vwing:
Note the drawcalls could be reduced further in books by batching lines together, and I believe I'm doing that in the Android fork, but it might cause some appearance differences. Will examine later, but it's a lower priority; I think this PR is fine without that addition.

Daneel53:
Seems good to have a try, will push it into the next 1.2 Test3.

Daneel53 (approved):
Tested into Test3 and books are displayed without issue.

KABoissonneault (approved):

PR #2792, opened 2026-05-27
Fix for overlapping text in some books

Vwing:
#2619 introduced an issue with books sometimes having overlapping text.

Example: The Real Barenziah, Part V

image

The fix was a single line. Just needed to set the TextLabel's parent to the pagePanel *before* calculating its size.

After the fix:
image


KABoissonneault:
Does this only happen with SDF fonts? What aspect ratio have you tested? I figured it's worth having if it fixes one case, I'm just curious if you've tested the other scenarios (no SDF font, faked aspect ratios), both before and after.

Vwing:
Different aspect ratios don't seem to change anything; the books end up always drawing at the same size, drawing off the screen if the screen is too skinny.

The bug only affects SDF fonts 👍

Daneel53:
Added into Test3 and tested with 640:480 display: this simple line of code solve the issue.

Daneel53 (approved):
This simple line of code fixes the issue.

PR #2798, opened 2026-06-10
Implement error logging and exception catching in `FolderBrowser`

### LINKED ISSUE #2796: DFU Fails to show the contents of `steamapps/common` directory
## Tested on:
- Windows 11
- Linux (Fedora KDE)

## The exact problem
I tried to set DFU up on two machines, one running Bazzite (based off of Fedora with KDE) and the other running Windows 11. On both machines, DFU was not able to show the contents of the `steamapps/common` directory, so I was not able to select the DOS daggerfall files.

## Workaround
In both cases, I was able to get around this by moving the DOS daggerfall installation to another folder (either through Steam or just by cutting and pasting the folder myself).

If we're able to find the source, I'll gladly contribute a PR for this myself. Thanks for what you do!

ChristosMaragkos:
`FolderBrowser` uses `Directory.GetDirectories` to scan the filesystem, which can throw for numerous reasons (path too long, any subdirectory failing, etc.). This is fine but, currently, in the event of an error:
- Nothing is communicated to the user
- The parent directory is not added to the list so the user must click on their filesystem root to get the game back to an interactable state

This PR fixes those issues by:
- Explicitly adding the `..` directory to the list if an exception is caught
- Logging all directory errors to the console with Unity's APIs

Closes #2796 
*Be aware that this does not fix any underlying potential issue. It just clearly communicates any errors to the player.*

### DIFF
--- a/Assets/Scripts/Game/UserInterface/FolderBrowser.cs
+++ b/Assets/Scripts/Game/UserInterface/FolderBrowser.cs
@@ -240,7 +240,8 @@ namespace DaggerfallWorkshop.Game.UserInterface
             {
                 folders.Clear();
                 folderList.ClearItems();
-                return;
+                if (currentPath != drives[driveList.SelectedIndex])
+                    folderList.AddItem(parentDirectory);
             }
         }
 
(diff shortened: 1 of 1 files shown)


ChristosMaragkos:
I also took the liberty of adding a fake directory when an exception is caught to communicate to the user that an error occured in-game. Will revert if too much.

KABoissonneault:
Did you manage to get any info about your issue from this? Running it should give you the log with the exception now, no?

ChristosMaragkos:
I wasn't able to because for some reason the game did *not* fail to scan the directory this time around. Which is a net positive either way.

PR #2801, opened 2026-06-14
Update WeatherTable.json

Daneel53:
Total values for Woodlands/Winter was only 90 and not 100 as per all other places/seasons.

### DIFF
--- a/Assets/Resources/WeatherTable.json
+++ b/Assets/Resources/WeatherTable.json
@@ -159,6 +159,7 @@
       "Cloudy": 15,
       "Overcast": 20,
       "Fog": 5,
+	  "Rain": 10,
       "Snow": 25
     },
     "Spring": {


Daneel53:
Tested in game, the following warning when loading the save in Daggerfall City disappear.
Unity_2026-06-13_18-55-48

petchema:
Bug originates from the Daggerfall Chronicles, where those tables came from https://en.uesp.net/wiki/Daggerfall:Climate

DFU "fixes" the issue by normalizing total to 100% at runtime. It's hard to tell what is more "right", short of investigating classic's assets

Daneel53:
We will never know if the Chronicles are wrong for the Woodlands/Winter values or if these values were already present into DF code. But I checked all the numbers and this entry is the only one where the total is not 100, so there is somewhere a bug.
I decided to add rain in Winter because it seemed not logical with 20% in Autumn and 25% in Spring to have no rain at all in Winter into woods.
Anyway, what is bothering me is the yellow warning that indicates an issue into the code when finally the issue is in the numbers of the table. If we don't want to modifiy one number in the table to respect what is written in the Chronicles, I think that at least we should downgrade the yellow Debug.LogWarning() to a simple grey Debug.Log() into Weather.cs so that we don't think there is a code error to fix.

KABoissonneault (approved):
If the data comes from Chronicles and not reverse-engineering, then I really don't mind what we put. Chronicles has not been a reliable source for anything except design intent

PR #2802, opened 2026-06-19
Bug fix in Dialogs window

Daneel53:
Replace the JustifyCenter token by a space so that successive lines of a record into Internal_RSC separated by a [/center] are not sticked together  when displayed into the Dialogs window.

Example:
8570,"Ah, at last! I was beginning to think you[/center]
wouldn't get this far.[/center]
[/end]"

With current code into TextManager.cs:
Daggerfall_Unity_2026-06-16_14-34-31

After bug fix:
Unity_2026-06-17_15-19-19

### DIFF
--- a/Assets/Scripts/Game/TalkManager.cs
+++ b/Assets/Scripts/Game/TalkManager.cs
@@ -3570,6 +3570,9 @@ namespace DaggerfallWorkshop.Game
                 string textFragment = tokens[i].text;
                 if (textFragment != null && textFragment != string.Empty)
                     builder.Append(textFragment);
+                else if (tokens[i].formatting == TextFile.Formatting.JustifyCenter)
+					// To avoid consecutive lines with [/center] in dialog screen to stick together.
+                    builder.Append(" ");
                 else
                     builder.Append(separatorString);
             }


KABoissonneault:
I can see how the center token can generate this issue, given that neither `"Ah, at last! I was beginning to think you` or `wouldn't get this far.` contain a space to separate them. I'm having a hard time understanding why this only happens if we have multiple lines with this center token.

The current solution seems to work for this issue, I'm just not sure if this could have other side effects, and if there's not a wider issue I'm missing

Daneel53:
For the Internal_RSC entries that are displayed into the Dialogs window, there are only 3 possible tags : [/record], [/end] and [/center]. The two first tags indicates an end of line, so they can be replaced by "", but [/center] is just not anymore taken into account **into this specific window**, as all text is left aligned. To add a space when you join one line that finish with [/center] with the following one was forgotten.
In other text as rumors, from entry 1400 to 1456, it was choosen to write only one line for each record, all the original [/center] were suppressed. I suppose that the one that changed these entries was not sure of the other entries that are into this long file that are displayed into the Dialogs window, so they stayed as into the classic DF.
Anyway, as the changed code is into TalkManager.cs, that should not chage anything to other windows. For example, the [/center] into quests windows are still taken into account.

PR #2807, opened 2026-08-16
Fix #2750 incomplete

petchema:
When looking at city walls, DFU now displays the LocaleText-NotFound message.

The text was added to the locales template, but not to the English locale. If current town is corrupted, you need to run the "refresh_buildingnames" command in the console to fix it.

(Also, should translation be "City Wall" or "City Walls"? I've seen the latter mentioned in other texts)


Daneel53 (approved):
Tested into 1.2 Test3, now "City Wall" is displayed in game when clicking on a city wall in Info mode.

